PR #26, opened 2016-12-09
Feat/5/https oauth

embmk:


codders:
Looks like this needs a rebase and squash - I can see a lot of older / unassociated commits there.

Also, but I guess this is too late, it seems that you've run the linting for the first time which has made a bunch of edits unrelated to the code you are introducing. It would have been nice to have those whitespace changes in a separate pull request.

embmk:
I installed an addon in Eclipse that applies the code style on each save when I saw that clang-format was introduced. Due to the messed up commit history I created a new branch locally and cherry-picked. This would allow reverting the code format changes, but I thought it would be recommended to run the qa target to see if it finds unpleasant code.

codders:
Sure. I think there's no point getting a separate pull request out now that this is already done. Please just squash the extra commits (we don't need 22 commits here), and assign to me for review.

embmk:
I'd like to squash the intermediate merge with the master commit but when I start a rebase it shows me tons of commits as seen before...

embmk:
I assigned you as assignee before instead of choosing you as reviewer, I think you meant assigning you as reviewer...

embmk:
I was a bit afraid that linting would become part of the integration builds, but I should have checked the changes you introduced on the master more in detail, sorry...

PR #33, opened 2016-12-12
Feat/6/api request

embmk:


embmk:
I thought about using the example config file for a part of the coverage. This would increase coverage partly. I'll try this. In addition we can try creating "incomplete" or empty config files in CMake and use them to get the "fail" branches. Or I can use some arbitrary files from the repository, maybe the README as it will always be available...

embmk:
This may become tricky. It is hard to emulate curl and we don't want the tests do any traffic of course. Executing the new code depends on getting a token from the authorization server.
Using an invalid, say non-yml, file leads to an exception thrown by the yml-library but this is not the problem when it comes to increasing the coverage.
A minimal inversive

embmk:
Pardon me, I had a nervous mouse and did an accidental close. Hope this increases coverage.

embmk:
Ok this doesn't really work, I'm afraid the only way would be to exclude the lines in question from the coverage analysis.

codders:
This actually looks good and seems to increase coverage now. The only problem is that it shows as 28 commits. Can you rebase against master so that those extra commits disappear?

embmk:
I intermediately found out how to use the markers that exclude lines from coverage which could be used on all BOOST_FAIL macros as these shall never be reached if everything works as expected. But maybe we can introduce this later...

embmk:
How do I see that a rebase succeeds? I did it three times now but tig always shows me an awful long commit history...

codders:
I really don't know why this seems to cause so much trouble for you in particular. I think it might be that you are sometimes doing a 'git pull' or 'git merge' - I would recommend avoiding both of those, and configuring your tooling not to use them (if you are using the git eclipse integration).

But since you seem to have ended up with a messy branch again, you need to create a fresh branch from the latest master, cherry-pick the changes that you want to the new branch, and force push to this branch.

embmk:
I quit using the gui two on thursday, but these are unfortunately branches that have been creating with the gui doin kind of unpredictable stuff (not really but one have to get into it...).
I did to additional rebases and now git tells me I am up to date with both, the remote branch and the local master which has to be up to date as I just cloned...

codders:
You need to create a fresh branch from the latest master, cherry-pick the changes that you want to the new branch, and force push to this branch.

embmk:
I am doing cherry picking now. I started with cherry picking when you purposed it.

embmk:
The commits seem to be fine or at least less now. Do you want me to exclude the testcode from the coverage analysis in order to get a decent increase of coverage?

Issue #42, opened 2016-12-15
Tidy up CMakeLists.txt


cajun-rat:
The CMakeLists should be improved to:

- use find_package rather than explicitly naming link libraries like libboost_system.so
- Split it into separate top-level and project CMakeLists.txt. The top-level CMakeLists should find the packages, and the 'project' one should contain definitions of the executables.

Please feel free to change the existing directory structure.  I would be happy with all the source code in a single directory for now. We can split it out when the project is big enough. Tests should be called oauthtoken_test.cc, and in the same directory as oauthtoken.cc. https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

I'm no cmake expert, but the example at https://github.com/advancedtelematic/sota-tools/blob/master/CMakeLists.txt might be a good starting point. Feel free to copy/paste from that code.

PR #44, opened 2016-12-15
Cmake lib fix

patriotyk:
Added lbrary dpendencies in proper way
Changed layout of source code, and cmake files
Renamed and moved tests

#42


cajun-rat:
This looks great, but could you take a look at the Travis CI build that is failing, please?

PR #46, opened 2016-12-16
Re-format code to Google Style Guide

patriotyk:
Changed code to be more compatible with "Google Style Guide" style


cajun-rat:
Nice job, thank you

PR #56, opened 2016-12-22
Check for updates

patriotyk:
There is a big test coverage decrease, but it is because of previous tests do only the execution but don't check code and functionality. I have cleaned and moved tests to separate folder. Another side is also that to have a good coverage and good written tests we should use some mock framework, without mocks it is not possible to cover all code and also it is much difficult to write fake objects. I suggest to use a  google test framework instead of boost.
Also I have removed some cmake tests. That tests test executable output and return code, but now we run in forever loop, and executable is never exits.


src/httpclient.cc
```diff
@@ -0,0 +1,158 @@
...
+
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include "logger.h"
+
+/*****************************************************************************/
+/**
+ * \par Description:
+ *    A writeback handler for the curl library. I handles writing response
```
jerrytrieu: s/I/It/

src/httpclient.cc
```diff
@@ -0,0 +1,158 @@
...
+static size_t writeString(void* contents, size_t size, size_t nmemb,
+                          void* userp) {
+  // clear the provided string
+  ((std::string*)userp)->clear();
+
+  // append the writeback data to the provided string
+  ((std::string*)userp)->append((char*)contents, size * nmemb);
+
+  // return size of written data
+  return size * nmemb;
```
jerrytrieu: how about `return str.assign(buf,size).length();`

src/httpclient.cc
```diff
@@ -0,0 +1,158 @@
...
+ * \par Description:
+ *    A writeback handler for the curl library. I handles writing response
+ *    data from curl into a string.
+ *    https://curl.haxx.se/libcurl/c/CURLOPT_WRITEFUNCTION.html
+ *
+ */
+static size_t writeString(void* contents, size_t size, size_t nmemb,
+                          void* userp) {
+  // clear the provided string
+  ((std::string*)userp)->clear();
```
jerrytrieu: Better to use c++ static_cast.

src/httpclient.cc
```diff
@@ -0,0 +1,158 @@
...
+  return size * nmemb;
+}
+
+static size_t writeFile(void* contents, size_t size, size_t nmemb, FILE* fp) {
+  size_t written = fwrite(contents, size, nmemb, fp);
+  return written;
+}
+
+HttpClient::HttpClient() {
+  curl_global_init(CURL_GLOBAL_ALL);
```
jerrytrieu: Is this safe to do here in class ctor?

src/httpclient.h
```diff
@@ -0,0 +1,27 @@
...
+  ~HttpClient();
+  bool authenticate(const AuthConfig &conf);
+  std::string getResponse() { return response; }
+  boost::property_tree::ptree get(const std::string &url);
+  boost::property_tree::ptree post(const std::string &url,
+                                   const std::string &data);
+  bool download(const std::string &url, const std::string &filename);
+
+ private:
+  CURL *curl;
```
jerrytrieu: Need copy-ctor/assign-op for this, or declare w/o implement.

src/config.h
```diff
@@ -44,5 +47,15 @@ class Config {
   CoreConfig core;
   AuthConfig auth;
   DeviceConfig device;
+
+ private:
+  boost::property_tree::ptree pt;
+
+  std::string getStringValue(const std::string &name) {
```
patriotyk: I don't  think so, this is private member function and should be accessible only in this class

src/config.h
```diff
@@ -44,5 +47,15 @@ class Config {
...
+    std::string ret_val = pt.get<std::string>(name);
+    ret_val.erase(std::remove(ret_val.begin(), ret_val.end(), '\"'),
```
patriotyk: I think it is the best ))) Do you know how to do it better?

src/httpclient.cc
patriotyk: Sorry?

src/httpclient.cc
patriotyk: No, because callback may be called twice for one request, but there was an error in prev line , I will remove string clearing in callback and will do it before request perform.

src/httpclient.cc
patriotyk: Yes, according to the documentation multiple calls of curl_global_init have the same effect as one call.

patriotyk:
I have made changes to things I didn't comment only. This not so quick task to split to two commits, this will require manual splitting, so I noticed your recommendation for the future commits

src/config.h
```diff
@@ -44,5 +47,14 @@ class Config {
   CoreConfig core;
   AuthConfig auth;
   DeviceConfig device;
+
+ private:
+  std::string getStringValue(boost::property_tree::ptree pt,
```
jerrytrieu: This is static helper function that doesn't need to be in this class. Let's put in anonymous namespace in the cpp. Need to keep header files clean as possible.

src/httpclient.h
jerrytrieu: Is this class with ptr safe to copy? Will open bug for this.

src/httpclient.cc
jerrytrieu: Typo. Nevermind.

src/sotahttpclient.cc
```diff
@@ -0,0 +1,45 @@
...
+    const boost::property_tree::ptree &update) {
+  std::string url = core_url + "/device_updates/" + config.device.uuid + "/" +
+                    update.get<std::string>("requestId") + "/download";
+  std::string filename =
+      config.device.packages_dir + update.get<std::string>("packageId.name");
+  http.download(url, filename);
+  boost::property_tree::ptree status;
+  status.put("updateId", update.get<std::string>("requestId"));
+  status.put("resultCode", 0);
+  status.put("resultText", "Downloaded");
```
jerrytrieu: The downloaded result isn't used. These fields are part of a download. Just test code?

src/httpclient.h
patriotyk: I made this class noncopyable please look for the base class

src/httpclient.cc
```diff
@@ -0,0 +1,156 @@
...
+  curl_easy_setopt(curl_post, CURLOPT_URL, url.c_str());
+  curl_easy_setopt(curl_post, CURLOPT_POSTFIELDS, data.c_str());
+  boost::property_tree::ptree result = perform(curl_post);
+  curl_easy_cleanup(curl_post);
+  return result;
+}
+
+boost::property_tree::ptree HttpClient::perform(CURL* curl) {
+  boost::property_tree::ptree pt;
+  response.empty();
```
jerrytrieu: Can use temporary string buffer instead of member `response`?

src/sotahttpclient.cc
patriotyk: Yes, temporary it is stubs

patriotyk:
@jerrytrieu could You merge my pull request? Or are You waiting for some updates from me?

Issue #52, opened 2016-12-20
JSON Event parsing


taheris:
For each [`Event`](https://github.com/advancedtelematic/rvi_sota_client/blob/master/src/datatype/event.rs), implement serialization and deserialization to and from JSON.

The parsing format should be: `{"variant":{EventName},"fields":[{arg1},{arg2},...,{argN}]}`

For example, if the `Event` was `UpdatesReceived` the JSON representation would be:
```
{
  "variant": "UpdatesReceived",
  "fields": [
    [{
      "requestId": "04179cd3-a73b-4b87-82f7-c0c6616d5f0f",
      "status": "Pending",
      "packageId": {
        "name": "libats",
        "version": "1.0.0"
      },
      "installPos": 0,
      "createdAt": "2016-12-19T16:03:09.171Z"
    }]
  ]
}
```

In this case `{arg1}` is another list containing each `UpdateRequest` (which contains a `requestId`, `status`, `packageId`, `installPos` and `createdAt`). Whitespace was added in this example for clarity but should be excluded from the actual implementation.

codders:
Can you enumerate the JSON serialisations and deserialisations in, for example, and RVI to DBus message?

taheris:
For the RVI -> DBus case, there would be a notification at the RVI Edge server of a new package, which will trigger a `DownloadComplete` message when all chunks are transferred.

The DBus gateway will listen for these events internally, then write a `downloadComplete` message to DBus which will be picked up by the Software Loading Manager.

Because the events are internal, no serialisation or deserialisation to and from JSON is required in this instance, however a `DownloadComplete` event that was serialised would look like:

```
{
  "variant": "DownloadComplete",
  "fields": [{
    update_id: "03edd73f-4cb0-4eb2-b0c3-ec7d08c170d4",
    update_image: "/tmp/03edd73f-4cb0-4eb2-b0c3-ec7d08c170d4",
    signature: ""
  }]
}
```

PR #57, opened 2016-12-28
 #52 JSON Event parsing

patriotyk:
#52

### DIFF
--- a/src/events.cc
+++ b/src/events.cc
@@ -1,0 +1,300 @@
+#include "events.h"
+
+namespace event {
+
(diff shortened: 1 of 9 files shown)


src/events.cc
```diff
@@ -0,0 +1,300 @@
+#include "events.h"
+
+namespace Event {
```
jerrytrieu: Let's use all lowercase for namespace.

patriotyk: Done

patriotyk:
I put data and event to different namespaces because some classes have equal names.

Issue #51, opened 2016-12-20
JSON Command Parsing


taheris:
For each [`Command`](https://github.com/advancedtelematic/rvi_sota_client/blob/master/src/datatype/command.rs), implement serialization and deserialization to and from JSON.

The parsing format should be: `{"variant":{CommandName},"fields":[{arg1},{arg2},...,{argN}]}`

For example, if the `Command` was `SendInstalledPackages` the JSON representation would be:
```
{
  "variant": "SendInstalledPackages",
  "fields": [
    [{
      "name": "libats",
      "version": "1.0.0"
    }]
  ]
}
```

In this case `{arg1}` is another list containing each `Package` (which contains a `name` and `version`). Whitespace was added in this example for clarity but should be excluded from the actual implementation.

Note that this does not follow the syntax in the Rust client of `{CommandName} {arg1} {arg2} ... {argN}`.

PR #58, opened 2016-12-29
Implemented Comand events parsing #51

patriotyk:


src/comands.cc
```diff
@@ -0,0 +1,167 @@
+#include "comands.h"
```
jerrytrieu: Typo, should be "commands.h"

src/comands.cc
```diff
@@ -0,0 +1,167 @@
...
+
+namespace Comand {
```
jerrytrieu: Should be lowercase namespace `command`

src/comands.h
```diff
@@ -0,0 +1,93 @@
+#ifndef COMANDS_H_
+#define COMANDS_H_
+#include <json/json.h>
+#include <string>
+#include "types.h"
+
+namespace Comand {
+
+class BaseComand {
```
jerrytrieu: Should be `BaseCommand`

src/comands.h
```diff
@@ -0,0 +1,93 @@
...
+ public:
```
jerrytrieu: Can be struct if only public?

src/comands.h
```diff
@@ -0,0 +1,93 @@
...
+  std::string variant;
+  Json::Value toJson();
+};
+
+class Authenticate : public BaseComand {
+ public:
+  Authenticate(const data::ClientCredentials& client_credentials_in);
+  data::ClientCredentials client_credentials;
+  std::string toJson();
```
jerrytrieu: Confusing to have same name as non-virtual function in base with different return type?

src/comands.h
```diff
@@ -0,0 +1,93 @@
...
+  static Authenticate fromJson(const std::string& json_str);
```
jerrytrieu: Can be a templated function to reuse across types?

patriotyk: Which function?

jerrytrieu: It's ok, we can refactor later.

src/comands.h
patriotyk: why do You think it is confusing? In any way I call it through BaseCommand

jerrytrieu: It's better to avoid this. Though the different return type will be caught at compile...? Refactor later.

PR #61, opened 2016-12-29
Fix a shadowing issue

OYTIS:
gcc-5.3.0 returns an error with current warning flags


src/httpclient.h
```diff
@@ -19,8 +19,8 @@ class HttpClient : private boost::noncopyable {
 
  private:
   CURL *curl;
```
jerrytrieu: We should name all members with `_` suffix. Do later.

Issue #59, opened 2016-12-29
Numerous issues with jsoncpp-1.7.5


OYTIS:
jsoncpp-1.7.5 is default on Arch Linux. Some issues are due to immaturity of jsoncpp-1.7.5 itself (the fail to include <cstdint> to their config.h), some are because of incompatible type definitions and very strict compilation rules. Maybe worth specifying in the docs that we require jsoncpp-1.7.2 for now, and keep these issues in mind for the future.

OYTIS:
It looks like jsoncpp-1.7.2 (default on Ubuntu 16.04) is still a bit too new, it makes test_events fail with

`unknown location(0): fatal error: in "UpdateAvailable_event_to_json": Json::LogicError: Value is not convertible to bool.
/home/oytis/playground/sota_client_cpp/tests/events_test.cc(95): last checkpoint`

Docker image uses jsoncpp-0.6.0-rc2

cajun-rat:
Making sota_client_cpp easy to build everywhere is important, so we should work out a plan to fix this.  I think the options are:

1. Bundle jsoncpp in our source tree
1. Make the CMakeLists detect the working versions (although if v0.x and v1.0 are incompatible but share the same file names it will be unbuildable on Arch)
1. Switch to a more stable json library (which may be painful now, but will only get more painful in the future)

cajun-rat:
Cool, thank you

PR #64, opened 2017-01-03
include jsoncpp to source tree

patriotyk:
We need to exclude jsoncpp from codecov to exclude code coverage calculation for this lib. 
#59


cajun-rat:
@patriotyk What do you think of patch I just pushed, which moves jsoncpp to a 'third_party' directory and excludes that from the coverage?

patriotyk:
Cool, I did not know how to exclude it. I thought it can be done somewhere in settings. Thanks.

cajun-rat:
(Note the drop in coverage is because we are excluding the tests/ directory as well as third_party)

Issue #55, opened 2016-12-21
Create Gateway abstraction


taheris:
A Gateway has two purposes:

1) Forward `Command` messages to the main interpreter and receive the final `Event` response.
2) Subscribe to different kinds of `Event` messages (e.g. all `DownloadComplete` events).

PR #65, opened 2017-01-03
prepare code for easy integrate with gateways

patriotyk:
#55


cajun-rat:
Looks good. Could you add some basic tests for the Channel class please?  It looks correct, but some tests would stop it getting broken in the future, and also help our codecov score :)

PR #67, opened 2017-01-04
Generate compile_commands.json

taheris:
`compile_commands.json` is used by [rtags](https://github.com/Andersbakken/rtags) for code indexing.

Also removes superfluous comments for readability but if there is a style guide that requires them they can be kept.


cajun-rat:
Review Score: 11/11

Issue #48, opened 2016-12-19
Add D-Bus gateway


taheris:
This should support the same operations that exist in the Rust implementation. Namely:

- publish the `UpdateAvailable`, `DownloadComplete` and `InstalledSoftwareNeeded` events to the `software_manager{,_path}` paths defined in the config.
- handle the `initiateDownload`, `abortDownload` and `updateReport` methods

A Franca IDL interface description of the methods can be found here: https://github.com/advancedtelematic/genivi_swm/blob/master/franca_idl/SotaClient.fidl.

The sd-bus library from systemd should be used. In addition, conditional compilation should be added such that the D-Bus gateway may be excluded (including the libraries) at compile time.

patriotyk:
During the work on this issue, I have found one problem. sd-bus library uses some C extensions that is not supported by c++ compiler, as result it is not possible to compile our dbusgateway code. https://github.com/systemd/systemd/issues/4858  Even if we rewrite existent macroses to c++, I am not sure there is no another C-only code. Another solution might be to use GENIVIs c++ dbus api https://github.com/GENIVI/capicxx-dbus-tools/wiki/CommonAPI-C---D-Bus-in-10-minutes Also it includes tools that can generate nice c++ stub code from Franca IDL file.

patriotyk:
GENIVI Common API requires c++11 ((

cajun-rat:
I'm fine with requiring C++11 for the GENIVI integration.  'Old' (pre-C++11) systems will use a different method to install packages, so requiring C++11 in this part only will be fine

PR #74, opened 2017-01-10
Dbusgateway implementation

patriotyk:
I have not covered events, because without mocks it is not possible, and I didn't have time to integrate gtests #48


taheris:
@patriotyk: would it be possible to remove the generated code from source control?
(@cajun-rat: unless you have a different opinion?)

cajun-rat:
Avoiding checking in generated code would be nice, but I'd like to understand the trade-off first:  Is generating these files at build time easy?  What tools are required?

If it turns out to be a pain (e.g. requires Java), could you put a note in the commit message that details how these files were generated (command line and the versions of the tools used), please?

patriotyk:
I don't think it will be easy to autogenerate code at build time because codegenerators requires java and maven to build it. Also GENIVI provides builded codegenerators, but they are architecture dependant. This librarry is designed to work with codegenerators only, and there is no any documentation how to use it without codegenerators

cajun-rat:
Agreed. 'maven' is a pretty good reason to check in the generated code.

Could you update the commit message with the commands used to generate the files, please?  Then it is good to merge.

patriotyk:
Ok, and I will add also fixed .fidl file.

cajun-rat:
Sorry, I've just spotted an issue that I missed the first time. When building with -DBUILD_WITH_DBUS_GATEWAY=ON, there are several checked in files that are modified as part of the build. These appear to be generated as part of the build process, and so shouldn't be checked in.

To Reproduce:
```
git clone into /tmp/foo
cd /tmp/foo
mkdir build
cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_WITH_DBUS_GATEWAY=ON ..
make
git status # Produces lots of output
```

The files that appear as modified should probably be removed from source control with 'git rm', and the 'Untracked' files should be .gitignore'd

It is a bit annoying that the autotools build process generates files in the source directory: it is possible to build those third_party libraries with cmake?

patriotyk:
I have changed libdbus to build with cmake but I still need to run ./autogen.sh to generate package-config file, because commonapi-dbus use it.

cajun-rat:
I'm having trouble building this on my desktop machine, but since it works on CI I will merge it anyway to unblock future work and I'll investigate a fix.

Issue #83, opened 2017-01-23
Abort on startup


taheris:
With a new release build, the following crash report is returned on startup:

```
root@b6b8f70db894:~/build/target# ./sota_client --config sota.toml
[2017-Jan-23 10:09:57.140174]: config read from sota.toml :
	Authtentification Server: https://auth-plus.atsgarage.com
	Sota Server: https://sota-core.atsgarage.com
	Pooling enabled : 0
	Pooling interval : 10
	Client: 432aece6-a001-4c9a-9941-ddb971c265ed
	Secret: YvbGH9pMRQ
	Device UUID: ba0dac05-da73-4add-a808-4db18e80a6ba
[2017-Jan-23 10:09:57.141997]: no commandline option 'loglevel' provided
[2017-Jan-23 10:09:57.142332]: http-gateway turned on, use it
process 7893: arguments to dbus_bus_request_name() were incorrect, assertion "connection != NULL" failed in file /source/third_party/dbus-1.9.0/dbus/dbus-bus.c line 1120.
This is normally a bug in some application using the D-Bus library.
  /source/libdbus_build/lib/libdbus-1.so(_dbus_print_backtrace+0x2e) [0x7f895951b45d]
  /source/libdbus_build/lib/libdbus-1.so(_dbus_abort+0xd) [0x7f8959516013]
  /source/libdbus_build/lib/libdbus-1.so(_dbus_warn_check_failed+0x111) [0x7f89595013f3]
  /source/libdbus_build/lib/libdbus-1.so(dbus_bus_request_name+0x8f) [0x7f89594bacad]
  /source/dbusapi_install/lib/libCommonAPI-DBus.so.3(_ZNK9CommonAPI4DBus14DBusConnection26requestServiceNameAndBlockERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0xfa) [0x7f895c4534f6]
  /source/dbusapi_install/lib/libCommonAPI-DBus.so.3(_ZN9CommonAPI4DBus7Factory19registerStubAdapterESt10shared_ptrINS0_15DBusStubAdapterEE+0x3c8) [0x7f895c48d0fa]
[...]

PR #84, opened 2017-01-23
Use patches, to see our changes on common-api-dbus library

patriotyk:
This pull request fixses #83, also all changes to dbus librarry moved to patch.


cajun-rat:
This looks good: clearly separating out the changes we are making to the libraries is a good thing.

We/I need to figure out how this will work when being built in GDP. The ideal case would be to get these changes upstreamed to GDP, but if we can't do that then we'll have to rethink how the Yocto build will work. The current plan for the Yocto recipe is to build against the GDP-provided platform builds of these libraries.

patriotyk:
Sorry, what is GDP?

taheris:
[GENIVI Development Platform](https://at.projects.genivi.org/wiki/pages/viewpage.action?pageId=11567210).

cajun-rat:
GDP stands for Genivi Development Platform, it is a Yocto/OpenEmbedded/Poky based embedded Linux distribution for cars. It operates differently to desktop Linux distributions like Debian because the entire system is built from source. This means that it is possible to apply patches to 'global' system libraries like dbus when building software in GDP.

When we get sota_client_cpp building in GDP we will want to use these patched 'global' libraries rather than the locally built versions we have to use for Debian-like desktop Linux. This is because patching dbus is quite simple in GDP compared to patching the Debian (or Red Hat) dbus library.

Issue #87, opened 2017-02-01
Segfault on startup


taheris:
Testing out a release build of the latest master segfaults on startup:

```
# ./sota_client --loglevel 0 --config sota.toml
[2017-Feb-01 16:39:52.639032]: no 'network.socket_events' option have been found in config file: temp, Falling back to default value
[2017-Feb-01 16:39:52.639242]: config read from temp :
	Authtentification Server: https://auth-plus-staging.atsgarage.com
	Sota Server: https://sota-core-staging.atsgarage.com
	Pooling enabled : 1
	Pooling interval : 10
	Client: e3a63586-37fa-4a98-8f80-011306bfffd1
	Secret: I5oNTaAtkI
	Device UUID: e4a36666-a434-4588-adf2-659e428fdcef
[2017-Feb-01 16:39:52.639380]: boost command line option: loglevel detected.
[2017-Feb-01 16:39:52.639451]: loglevel set to trace
[2017-Feb-01 16:39:52.639576]: http-gateway turned on, use it
Segmentation fault
```

The config is as follows:

```
[auth]
server = "https://auth-plus-staging.atsgarage.com"
client_id = [removed]
client_secret = [removed]

[core]
server = "https://sota-core-staging.atsgarage.com"
polling = true
polling_sec = 10

[device]
uuid = "e4a36666-a434-4588-adf2-659e428fdcef"
packages_dir = "/tmp/"
package_manager = "off"
certificates_path = "/sota_certificates"

[gateway]
socket = true

[network]
socket_commands_path = "/tmp/sota-commands.socket"
socket_events_path = "/tmp/sota-events.socket"
```

PR #88, opened 2017-02-01
fixes crash on startup

patriotyk:
#87


cajun-rat:
LGTM, @taheris can you check this resolves the bug you saw, please?

PR #92, opened 2017-02-06
Feat/53/rvi client

patriotyk:


cajun-rat:
Looks like http://git.projects.genivi.org is down

patriotyk:
I have added commit of moving from commonapi to this pull request because it is much easier and don't requires mach merging. You can review by each commit, because there is a lot of changes.

PR #96, opened 2017-02-09
Change client name to 'aktualizr'

tkfu:


tkfu:
Note: this will break CI; that should be updated before this is merged.

cajun-rat:
Hi Jon, I'd like to sit on this for a couple of days.  @patriotyk has a couple of branches open, and the build is broken by git.genivi.org going offline, so it would be good to get those merged first since this change will be easier to rebase.

PR #99, opened 2017-02-14
Refactor CMakeLists to support Yocto build

cajun-rat:
This includes several changes to support building in Yocto
 - Both RVI and SWM support are now behind a single BUILD_GENIVI flag
 - All third_party dependencies are build with cmake
 - A new flag disables building tests, to avoid a dependency on gtest/gmock

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,50 +1,31 @@
 
 
-IF( NOT GTEST_ROOT )
+if(NOT GTEST_ROOT )
     set(GTEST_ROOT /usr/src/gtest)
-ENDIF()
-
-IF( NOT GMOCK_ROOT )
-    set(GMOCK_ROOT /usr/src/gmock)
-ENDIF()
-
-# build google test as static library
-set(BUILD_SHARED_LIBS OFF)
-set(BUILD_SHARED_LIBS_AUTOMATIC_OFF 1)
-
-add_subdirectory(${GMOCK_ROOT} ${CMAKE_CURRENT_BINARY_DIR}/gmock EXCLUDE_FROM_ALL)
-#add_subdirectory(${GTEST_ROOT} ${CMAKE_CURRENT_BINARY_DIR}/gtest EXCLUDE_FROM_ALL)
-
-set(BUILD_SHARED_LIBS ON)
-set(BUILD_SHARED_LIBS_AUTOMATIC_OFF 0)
-
-set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake-modules)
-
-# Setup warnings
-
-if (CMAKE_COMPILER_IS_GNUCXX)
-        add_definitions(-fstack-protector-all)
-        # Enable maximum of Warnings :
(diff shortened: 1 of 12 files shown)


cajun-rat:
@taheris Can you re-check this please?

patriotyk:
Looks good to me, but do we really need session.conf file?

cajun-rat:
@patriotyk On my Ubuntu desktop, I had trouble with dbus permission errors without it.  I don't really know enough dbus, but I _think_ that dbus-run-session uses the config file that happens to be installed in the system by default.  Isolating the tests from that seemed to be a good idea, but I'm no expert.

tests/CMakeLists.txt
```diff
@@ -53,36 +34,33 @@ if(BUILD_WITH_CODE_COVERAGE)
...
 
 if(CLANG_FORMAT)
-add_custom_target(format-tools
-    COMMAND ${CLANG_FORMAT} -i -style file ${CMAKE_CURRENT_SOURCE_DIR}/src/*.cc ${CMAKE_CURRENT_SOURCE_DIR}/src/*.h ${CMAKE_CURRENT_SOURCE_DIR}/tests/*.cc
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-    COMMENT "Running clang-format"
-    VERBATIM)
-add_dependencies(qa format-tools)
+    add_custom_target(format-tests
+        COMMAND ${CLANG_FORMAT} -i -style=file ${ALL_TEST_SRCS}
```
patriotyk: With this change, You have not provided target for formatting sources, only tests.

taheris:
lgtm

tests/CMakeLists.txt
cajun-rat: I've put a separate clang-format invocation in src/CMakeLists.txt, too.  This means we call it twice, but it doesn't have to reference files between the two directories.  Does that make sense?

patriotyk:
Found one bug. all executables placed in root folder instead target

tests/CMakeLists.txt
patriotyk: I think it is ok. But why did You split it?

cajun-rat:
@patriotyk What is the advantage of putting them in the 'target' subfolder?

tests/CMakeLists.txt
cajun-rat: The problem is that clang-format needs a list of files to operate on. Globbing for these seems a little ugly, and so the two options seem to be:
1) Pass the list of files between the src/ and test/ CMakeLists.txt 
2) Run clang-format twice

My thought is that keeping the CMakeLists.txt self-contained such that they only operate on files in their folders is a bit nicer?

cajun-rat:
Oh, do you normally make 'in-tree' builds?  I normally create a separate build output directory (like 'build') and run "cd build; cmake .."

patriotyk:
Yes I do.

cajun-rat:
Ah OK.  Would switching to out-of-tree builds (or having the build output in the project root directory) be a problem?

patriotyk:
No, all tests pass, I just thought it was a mistake, so if for You it is better, for me it is also ok.

cajun-rat:
Cool. Thank you!

PR #106, opened 2017-02-15
fixed problems with bad network, and implemented download resume

patriotyk:
Also code is formatted according 160 columns


cajun-rat:
In the future could you try and use separate commits for big, automated changes (such as the line length change) and functional changes, please?  That makes reviewing the code easier.

PR #107, opened 2017-02-15
Sources formtatted with 160 columns

patriotyk:
There is no logical changes, all sources automatically formatted with clang-format


taheris:
Some of these lines are a little too wide to read comfortably. Can we set a 120 max-width instead?

patriotyk:
Yes, sure

PR #109, opened 2017-02-16
Added verbose logging to rvi_lib

patriotyk:
All logs goes to stderror

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -80,6 +80,7 @@ find_package(Boost COMPONENTS system thread program_options log log_setup regex 
 enable_testing()
 
 include_directories(${PROJECT_SOURCE_DIR}/src)
+include_directories(${PROJECT_SOURCE_DIR}/third_party/rvi_lib/include)
 
 # use the same libiraries as the normal target but add boost test
 set (TEST_LIBS ${Boost_LIBRARIES} ${CURL_LIBRARIES}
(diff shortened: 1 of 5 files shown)


tests/CMakeLists.txt
```diff
@@ -178,7 +178,7 @@ if(BUILD_GENIVI)
                      src/logger.cc
                      tests/rvisotaclient_test.cc)
     add_executable(aktualizr_test_rvi_client ${TEST_RVI_SRC})
-    target_link_libraries(aktualizr_test_rvi_client rvi ${TEST_LIBS} gtest)
+    target_link_libraries(aktualizr_test_rvi_client ${TEST_LIBS} gtest)
```
cajun-rat: I think this is required: The CI build fails with "/source/src/sotarviclient.h:3:17: fatal error: rvi.h: No such file or directory"

patriotyk: No it is not, it is linker flag, in tests I don't use rvi_lib, instead I defined all symbols in test. But rvi.h is required so the problem is with compilation in finding include files.

cajun-rat: Ah OK. The CMakeLists.txt in rvi_lib includes the line "target_include_directories(rvi PUBLIC include)" which automatically includes the RVI public interface headers in any target that links to it.  I can see that this isn't what you want in this case though :)

PR #110, opened 2017-02-16
Much faster build, and simplified CmakeFiles.txt for tests

patriotyk:
Optimized build process because it is to boring, also simplified CmakeLists.txt

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -25,15 +25,6 @@ if (CMAKE_COMPILER_IS_GNUCXX)
     endif()
 endif()
 
-# Setup coverage
-if(BUILD_WITH_CODE_COVERAGE)
-    include(CodeCoverage)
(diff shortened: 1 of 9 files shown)


tests/CMakeLists.txt
```diff
@@ -79,161 +70,87 @@ find_package(Boost COMPONENTS system thread program_options log log_setup regex
 #enable test features of cmake
 enable_testing()
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -std=c++0x")
```
cajun-rat: Is this needed?

patriotyk: Yes, You are right, now we don't need it

patriotyk:
There was a c++11 specific code, I have fixed it

cajun-rat:
Nice job!  This doubles the speed of the build on my machine!

PR #112, opened 2017-02-16
Gracefull thread exits

patriotyk:


cajun-rat:
Weirdly, only one of the CI builds is failing. The error looks real though?

src/sotarviclient.h
```diff
@@ -40,6 +41,8 @@ class SotaRVIClient: public SotaClient
     int connection;
     event::Channel * events_channel;
     Json::Value services;
+    bool stop;
```
cajun-rat: Should this be boost::atomic<bool>?

PR #122, opened 2017-03-02
Initial draft of automatic provisioning documentation

cajun-rat:


taheris:
We already have a [gateway] section. In the rvi_sota_client it is implemented under the [auth] section as follows:

```
[auth]
server = "http://localhost:9001" # existing server field used for both oauth2 and tls authentication
client_id = "client-id" # existing client-id field used for both
client_secret = "client-secret" # existing client-secret field only used for oauth2
tls_p12_path = "/foo/tls.p12" # new provisioning field only for tls
tls_p12_password = "" # new provisioning field only for tls
tls_expiry_days = 365 # new provisioning field only for tls
```

The validation is as follows:

```
match (id, secret, auth_p12, device_p12) {
  (_,  None,    None,    _)    => Err("Need one of auth.client_secret or auth.tls_p12_path"),
  (_,  Some(_), Some(_), _)    => Err("Got both auth-client-secret and auth.tls_p12_path"),
  (_,  _,       Some(_), None) => Err("Certificate registration needs device.p12_path"),
  (id, Some(s), _,       _)    => Ok(Auth::Credentials(ClientCredentials { client_id: id, client_secret: s })),
  (id, _,       Some(_), _)    => Ok(Auth::Registration(RegistrationCredentials { client_id: id })),
}
```

The `auth.server` field is expanded to `<auth.server>/core/<endpoint>` with certificate authentication, and `<auth.server>/api/v1/mydevice/<device-id>/<path>` otherwise.

When using certificate authentication, check the `device.p12_path` for an exisiting bundle, or download a new PKCS#12 bundle to this location otherwise. There is also a `device.p12_password` field which is used to decrypt this bundle.

On another note, there should be a new ticket added to bring the Aktualizr sota.toml config back in-line with what is currently supported.

cajun-rat:
Thank you for the pointer.  I'll update this PR in a sec

PR #131, opened 2017-03-09
Implemented auto provision and ecu registration

patriotyk:


src/httpclient.cc
```diff
@@ -46,6 +47,11 @@ HttpClient::HttpClient() {
   if (loggerGetSeverity() <= LVL_debug) {
     curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);
   }
+
+  struct curl_slist* h = curl_slist_append(headers, "Accept: */*");
+  h = curl_slist_append(headers, "Content-Type: application/json");
+  curl_easy_setopt(curl, CURLOPT_HTTPHEADER, h);
+  curl_slist_free_all(headers);
```
cajun-rat: I think will free `headers` too soon. From https://curl.haxx.se/libcurl/c/CURLOPT_HTTPHEADER.html:

"When this option is passed to curl_easy_setopt, libcurl will not copy the entire list so you must keep it around until you no longer use this handle for a transfer before you call curl_slist_free_all on the list. "

We call curl_slist_free_all in the destructor, so I think this isn't needed.

src/httpclient.cc
```diff
@@ -111,30 +124,50 @@ bool HttpClient::authenticate(const AuthConfig& conf) {
...
+  Json::Value json;
+  Json::Reader reader;
+
+  std::string response = post(url, Json::FastWriter().write(data));
+  reader.parse(response, json);
+  return json;
+}
+
+void HttpClient::setCerts(const std::string& ca, const std::string& cert) {
+  curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, false);
```
cajun-rat: Is this needed?  Disabling CURLOPT_SSL_VERIFYPEER will allow man-in-the-middle attacks against the SSL connection

src/sotahttpclient.cc
```diff
@@ -24,7 +40,171 @@ SotaHttpClient::SotaHttpClient(const Config &config_in, HttpClient *http_in, eve
...
+}
+
+bool SotaHttpClient::authenticate() {
+  if (config.core.auth_type == CERTIFICATE) {
+    return http->authenticate(config.device.certificates_path + config.tls.client_certificate,
+                              config.device.certificates_path + config.tls.ca_file);
+  } else {
+    return http->authenticate(config.auth);
+  }
+}
```
cajun-rat: Nice!

cajun-rat:
Other than CURLOPT_SSL_VERIFYPEER and free-ing header, this looks great.

PR #133, opened 2017-03-10
Removed new line at the and of generated string.

patriotyk:
This pull request removes single new line from generated string with FastWriter writer of JsonCpp parser.
By default, FastWriter writes json without any whitespaces(except end of the string), keys are sorted,  list items preserved with original order. 

Also I have tried Ukrainian symbols, and it sorts regarding to the Ukrainian alphabet.
And I have not  tested Double values. Do we will use them?


cajun-rat:
Nice!  I can't see any double values in the sample TUF metadata, so don't worry about encoding double values for now.

PR #139, opened 2017-03-30
Uptane api

patriotyk:
Implemented uptane signing
Added OStree support
Covered almost all functionality with tests


cajun-rat:
Merged in #146

PR #147, opened 2017-03-31
Remove deleted files from CMakeLists

cajun-rat:
Also fix the tests when OSTree is disabled

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,5 +2,3 @@ ignore:
   - "third_party"
   - "tests"
   - "src/dbusgateway/src-gen"
-fixes:
-  - "src/::"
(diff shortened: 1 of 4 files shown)


codecov.yml
```diff
@@ -1,6 +1,4 @@
 ignore:
   - "third_party"
   - "tests"
-  - "src/dbusgateway/src-gen"
-fixes:
-  - "src/::"
\ No newline at end of file
+  - "src/dbusgateway/src-gen"
```
patriotyk: Why You added this? We don't such folder anymore.

src/CMakeLists.txt
```diff
@@ -25,15 +25,13 @@ set(HEADERS
             src/logger.h
             src/httpclient.h
             src/sotahttpclient.h
-            src/uptanehttpclient.h
```
patriotyk: We need change this line to sotauptaneclient.h

PR #149, opened 2017-04-03
Towards a Debian stable build

cajun-rat:


Dockerfile.deb-stable
```diff
@@ -0,0 +1,14 @@
+FROM debian:stable
```
taheris: jessie uses libc 2.19, while the prior stable wheezy uses 2.13 which may be a better target

cajun-rat: Was there something special about libc 2.13?  Jessie was released in 2015, so is pretty old :)  Maybe it would be worth another Dockerfile that tests a very basic build against oldstable?

taheris: Nothing special about 2.13, other than it is a reasonably old libc to target.

PR #151, opened 2017-04-04
Use new rviJsonInit to avoid second config file

patriotyk:
This PR also includes latest librvi from genivi


cajun-rat:
Replaced by #156

PR #157, opened 2017-04-06
Debian stable builds

cajun-rat:
Set up the CI builds to build (in parallel) against Ubuntu and Debian stable, then add some fixes to make the build pass


cajun-rat:
Looks like we are hitting some build failures. Please wait until they are resolved before merging

PR #158, opened 2017-04-07
Do not overwrite cmake configuration when run tests

patriotyk:
Now it is not possible to run tests in the second time when build in the source directory. Tests overwrite cmake configuration. This PR fixes this issue.

### DIFF
--- a/tests/build_with_options
+++ b/tests/build_with_options
@@ -1,5 +1,11 @@
 echo $PWD
+mv $1/CMakeFiles $1/CMakeFiles_back
+mv $1/CMakeCache.txt $1/CMakeCache.txt_back
+
 mkdir "$2"
 cd "$2"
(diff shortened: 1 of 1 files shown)


cajun-rat:
The intention of these tests was that they should perform the build in a subdirectory away from the main build directory.  Does that not happen for you?

cajun-rat:
Is this caused by an in-tree build?  I think those tests are incompatible with in-source builds.  We could either disable those tests when an in-source build is detected, or just refuse to support in-source builds at all.

I'm happy to drop support for in-source builds completely: What do you think about PR #159?  If there is a reason to continue to support in-tree builds, then we could try another option

cajun-rat:
We switched to denying in-tree builds altogether in PR #159

PR #177, opened 2017-05-03
Fixes #176 issue

### LINKED ISSUE #176: void* userdata in rviRegisterService() points to stack
When debugging an issue with the RVI mode of aktualizr, I ran into a bug where the rvi callbacks were not working correctly.  I think the problem is with the following piece of code that registers the service callbacks:

https://github.com/advancedtelematic/aktualizr/blob/master/src/sotarviclient.cc#L60

rviRegisterService takes a void* pointer for 'user data', but in this case this pointer is shared among every iteration of the loop, and also on the stack.

One solution might be to split callbackWrapper into one function for each of "notify", "start", "chunk", "finish", "getpackages", "abort", and make userdata a pointer to the SotaRVIClient object.  I'm sure there are nicer solutions too :)

patriotyk:
Please check this fix, because I was not able to setup rvi server.

### DIFF
--- a/src/sotarviclient.cc
+++ b/src/sotarviclient.cc
@@ -9,29 +9,43 @@
 #include "logger.h"
 #include "types.h"
 
-char service_names[6][12] = {"notify", "start", "chunk", "finish", "getpackages", "abort"};
 
-void callbackWrapper(int fd, void *service_data, const char *parameters) {
-  (void)fd;
-  SotaRVIClient *client = (SotaRVIClient *)((void **)service_data)[0];
-  std::string service((char *)((void **)service_data)[1]);
+Json::Value parseParameters(const char *parameters) {
   Json::Reader reader;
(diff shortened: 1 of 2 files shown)


leon-anavi:
Thank you for the provided fix. It seems that this patch fixes the issue with the registering the services. 

Best regards,
Leon

src/sotarviclient.cc
```diff
@@ -10,11 +10,11 @@
 #include "types.h"
 
 char service_names[6][12] = {"notify", "start", "chunk", "finish", "getpackages", "abort"};
+SotaRVIClient *client;
```
cajun-rat: This is OK as a quick-fix, but I'm really not keen on global variables.

cajun-rat:
Please remember to run 'make qa' to reformat to normalised style :)

PR #181, opened 2017-05-09
Add option to connect to D-Bus's System Bus

cajun-rat:
This is useful to avoid a line like:

export DBUS_SESSION_BUS_ADDRESS="unix:path=/var/run/dbus/system_bus_socket"

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -2,6 +2,7 @@
 
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
+#include <sstream>
 
 #include "logger.h"
 
@@ -54,7 +55,7 @@ T StripQuotesFromStrings(const T& value) {
 
 template <typename T>
 void CopyFromConfig(T& dest, const std::string& option_name, LoggerLevels warning_level,
-                    boost::property_tree::ptree& pt) {
+                    const boost::property_tree::ptree& pt) {
   boost::optional<T> value = pt.get_optional<T>(option_name);
   if (value.is_initialized()) {
     dest = StripQuotesFromStrings(value.get());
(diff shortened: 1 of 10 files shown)


cajun-rat:
cc @leon-anavi

leon-anavi:
Thank you for adding "dbus-system-bus". It will be very useful for the Aktualizr's systemd service in GDP.

patriotyk:
Maybe, it will be better just to add to config file the parameter dbus-bus  which can be DBUS_BUS_SESSION or DBUS_BUS_SYSTEM. This will avoid of adding a dbusutils files. We will need just to parse dbus_bus="session or system" in config file and set the right values DBUS_BUS_SESSION or DBUS_BUS_SYSTEM for config structure. 
Do You understand what I mean?

cajun-rat:
Yes, I think that makes a lot of sense.  I'll move the option to the config file.

cajun-rat:
@leon-anavi @patriotyk I've updated it to take the config via sota.toml.  Please re-review

src/dbusgateway/dbusgateway.cc
```diff
@@ -1,14 +1,15 @@
...
+#include "dbusutilities.h"
 #include "logger.h"
 #include "types.h"
 
 DbusGateway::DbusGateway(const Config& config_in, command::Channel* command_channel_in)
     : stop(false), config(config_in), command_channel(command_channel_in), swlm(config) {
   dbus_threads_init_default();
   dbus_error_init(&err);
-  conn = dbus_bus_get(DBUS_BUS_SESSION, &err);
+  conn = connectToDbus(config_in.dbus, &err);
```
patriotyk: We don't need connectToDbus function anymore. Just use dbus_bus_get and pass the buss type from our config

src/dbusgateway/dbusutilities.cc
```diff
@@ -0,0 +1,24 @@
+//
```
patriotyk: We don't need this file

src/dbusgateway/dbusutilities.h
```diff
@@ -0,0 +1,15 @@
+//
```
patriotyk: And this file too ))

src/dbusgateway/swlm.cc
```diff
@@ -1,12 +1,16 @@
...
 #include <stdlib.h>
 #include <iostream>
+
+#include "dbusutilities.h"
 #include "logger.h"
 
 SoftwareLoadingManagerProxy::SoftwareLoadingManagerProxy(const Config &config_in) : config(config_in) {
   dbus_error_init(&err);
-  conn = dbus_bus_get(DBUS_BUS_SESSION, &err);
+  conn = connectToDbus(config_in.dbus, &err);
```
patriotyk: don't use connectToDbus

src/config.cc
```diff
@@ -2,9 +2,12 @@
 
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
+#include <sstream>
 
 #include "logger.h"
 
+using std::string;
+
```
patriotyk: Using "using namespace" considered as bad practice. I prefer to remove it and add std to all std variables.

src/dbusgateway/dbusgateway.cc
cajun-rat: We'll have to do some conditional includes because dbus is only needed when BUILD_GENIVI is defined, otherwise should be fine

PR #186, opened 2017-05-18
base implementation of uptane-vector-tests

patriotyk:


heartsucker:
@patriotyk I just made a couple of changes to http://github.com/advancedtelematic/uptane-test-vectors about an hour ago, so it would be best if you regenerated the test data before you committed this. It should only affect a couple of the vectors.

Dockerfile
```diff
@@ -9,6 +9,6 @@ RUN echo "deb http://httpredir.debian.org/debian jessie-backports main contrib n
 
 RUN apt-get update
 RUN apt-get -y install liblzma-dev bison e2fslibs-dev libgpgme11-dev libglib2.0-dev gcc g++ make cmake git psmisc dbus python-dbus python-gobject-2 libdbus-1-dev libjansson-dev libgtest-dev libssl-dev autoconf automake pkg-config libtool libexpat1-dev libboost-program-options-dev libboost-test-dev libboost-regex-dev libboost-dev libboost-system-dev libboost-thread-dev libboost-log-dev libjsoncpp-dev curl libcurl4-openssl-dev lcov clang clang-format-3.8
-RUN apt-get -y install ostree libostree-dev libsodium-dev
```
heartsucker: I'm not sure you `python3-flask` here. I could rewrite the `requirements.in` to have flask be optional since that's just a convenience thing for testing. If you're generating all the files and stuffing them into the repo, it shouldn't matter.

patriotyk: It is needed for CI tests, to run server.py

PR #188, opened 2017-05-22
Add A3 overview diagram

cajun-rat:


patriotyk:
This svg file displays incorrectly on my laptop and on github preview page
screenshot 2017-05-22 at 16 19 03

cajun-rat:
Yeah, it needs inkscape to view it. Checking in the PDF might help?

patriotyk:
SVG is good to, but You just need to convert all text to curves and save it as plain svg(not inkscape)

cajun-rat:
Updated, I now build the PDF automatically, if inkscape is detected.

I'd rather not convert all the text to curves, because that will make it hard to edit

cajun-rat:
Nudge :)

PR #192, opened 2017-05-26
Let OSTree sysroot be changed in config

OYTIS:
Normally one will need to set both ostree.sysroot and ostree.os when updating a directory, because in this case there will be no "currently booted" deployment.


cajun-rat:
Rebased to master

cajun-rat:
Review Score: 11/11

PR #193, opened 2017-05-29
Compatibility with current server implementation,

patriotyk:
Added possibility to downgrade.
Some code refactoring.


cajun-rat:
@OYTIS Could you or @tkfu try this in meta-updater and see if we can do UPTANE + OSTree with this, please?

OYTIS:
@cajun-rat Sure, but I still can't make device gateway routing the message indicating presense of my device. As soon as this is fixed, I can go on.

Issue #191, opened 2017-05-26
Configuration option to skip some UPTANE checks


cajun-rat:
The current server implementation doesn't support the full UPTANE security model. In order to test Aktualizr against the server that we currently have deployed, we need a configuration option to disable/relax some of the security checks that we perform. There are 3 levels for this configuration flag:

* Full UPTANE checks
* All the UPTANE checks except those relating to rolling-back to an older version (this is to allow demos to switch between 2 versions)
* Compatibility with https://qa-app.atsgarage.com

We will have to do some tests to discover exactly which checks need disabling for the last option, and it should be easy to audit which checks are being skipped.

OYTIS:
@cajun-rat @patriotyk @Danten 
Do we really need the roll-back? As far as I understand, if we issue a downgrade to a previous version, the server should issue a new snapshot with increased version number and the respective timestamp. From point of view of the client it will be an upgrade, not a downgrade anyway.
Correct me if I'm wrong.

cajun-rat:
Yes, the version number from the director can always increase, but the version number in the targets.json from the images repository will go backwards during a rollback (because that number is signed offline)

OYTIS:
@cajun-rat OK, I see, thank you (although I don't quite understand what kind of protection does version check on targets.json give us when we already have snapshot.json).

@patriotyk Then the implementation in https://github.com/advancedtelematic/aktualizr/pull/193 is not quite correct. uptane.allow_downgrade shouldn't let us use outdated timestamp.json, it just should disable version check for targets.json (which is now missing anyway as far as I can see).

@heartsucker Could we add test for targets.json downgrade attempt to uptane-test-vectors?

patriotyk:
@OYTIS  Yeah I see. I will prepare the PR that fixes it.

PR #194, opened 2017-05-29
Remove unnecessary check

OYTIS:
keyid MAY be a hash of the key, it's not a requirement. This check is actually incompatible with how our server generates keyids, i.e. based on binary data, not PEM representation.

### DIFF
--- a/src/uptane/tufrepository.cc
+++ b/src/uptane/tufrepository.cc
@@ -183,15 +183,6 @@ void TufRepository::initRoot(const Json::Value& new_content) {
                                                                << requiredThreshold);
       throw IllegalThreshold(name_, "root.json contains a role that requires too many signatures");
     }
-    for (Json::ValueIterator it_keyid = (*it)["keyids"].begin(); it_keyid != (*it)["keyids"].end(); ++it_keyid) {
-      if (keys_.count((*it_keyid).asString())) {
-        std::string key_can = Json::FastWriter().write(Json::Value(keys_[(*it_keyid).asString()].value));
-        std::string key_id = boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest(key_can)));
-        if ((*it_keyid).asString() != key_id) {
-          throw UnmetThreshold(name_, role);
-        }
-      }
-    }
     thresholds_[role] += requiredThreshold;
   }
   verifyRole(content);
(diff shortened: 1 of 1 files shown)


src/uptane/tufrepository.cc
```diff
@@ -183,15 +183,6 @@ void TufRepository::initRoot(const Json::Value& new_content) {
                                                                << requiredThreshold);
       throw IllegalThreshold(name_, "root.json contains a role that requires too many signatures");
     }
-    for (Json::ValueIterator it_keyid = (*it)["keyids"].begin(); it_keyid != (*it)["keyids"].end(); ++it_keyid) {
```
heartsucker: This breaks the spec. The TUF spec specifically says that key IDs *MUST* be calculated locally which means that they also *MUST* match the key IDs calculated by the server.

heartsucker: https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L549-L552

OYTIS: OK, then we will need to fix that somehow. @simao?

cajun-rat:
Can we add this to the compatibility exceptions in https://github.com/advancedtelematic/aktualizr/pull/193 and https://github.com/advancedtelematic/aktualizr/issues/191

src/uptane/tufrepository.cc
patriotyk: I have already prepared fix, and it waits for review -  #193

heartsucker: Also related: PRO-3051

patriotyk:
#193 adds option to skip keyid verification

src/uptane/tufrepository.cc
simao: `Every public key has a unique identifier (PublicKey.keyid). This identifier MAY be, for example, the SHA-256 hash of the public key." - from the UPTANE impementation spec.`

So?

heartsucker: The TUF/ Uptane guys are trying to merge the two specs to be closer to each other, and this is one place where they diverge slightly. Following TUF definitely means we're following Uptane (in this case).

simao: Ok, then this will significant some changes in the repo because uptane says a keyid is just the hash of the public key, tuf says it's the hash of the canonical json of the keys, this would be a new ticket.

PR #196, opened 2017-05-30
Add uptane.ostree_server option

OYTIS:
"custom.uri" field in targets.json is outdated.

Patch to cope with URI proliferation in config is coming.


cajun-rat:
Just pushed a commit that reformats with 'make qa'.  I should figure out why clang-format doesn't work on your machine :)

OYTIS:
@cajun-rat Sorry, I'll try to fix that ASAP.

PR #197, opened 2017-05-30
Fixed timeout because polling_sec was not used

patriotyk:


cajun-rat:
+1 for a commit that reduces the code size :)

PR #202, opened 2017-05-31
Remove downgrade option

patriotyk:
This option is wrong and unuseful


patriotyk:
Oh, there is failed tests

PR #204, opened 2017-06-01
Overwrite server options and primary_ecu_serial from cmdline

patriotyk:
Add primary_ecu_hardware_id option instead of device.uuid
Construct default urls if repo,direcor, or ostree urls are empty

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -66,6 +66,44 @@ void CopyFromConfig(T& dest, const std::string& option_name, LoggerLevels warnin
 
 // End template tricks
 
+Config::Config() { postUpdateValues(); }
+
+Config::Config(const std::string& filename) {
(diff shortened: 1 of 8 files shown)


OYTIS:
One issue there: config.core.auth_type is set based on what's in sota.toml, ignoring tls_server value provided from command line. Practically that means that if it can't find tls_server in sota.toml, it will use SotaHttpClient instead of SotaUptaneClient.

OYTIS:
@patriotyk fixed a problem with using plain variable instead of getter in sotauptaneclient.bb. The issue with auth_type is still there.

OYTIS:
Great job, thank you. /var/sota/metadata/{director, repo}/targets is still empty for whatever reason, but let's solve this problem (if there is really a problem) later.

PR #205, opened 2017-06-01
Report last deployed image as currently installed

OYTIS:


src/ostree.h
```diff
@@ -12,6 +12,7 @@ static const std::string BOOT_BRANCH = "/usr/share/sota/branchname";
 
 struct Ostree {
   static OstreeDeployment *getBootedDeployment(const std::string &path);
+  static OstreeDeployment *getStagedDeployment(const std::string &path, const std::string &ostree_os);
```
cajun-rat: Sometime we might want to think about making some kind of 'ostree' object that encapsulates the ostree root path

OYTIS: @cajun-rat A static member in Ostree maybe? Or we can imagine managing multiple OSTree repos?

OYTIS:
The issue is solved by https://github.com/advancedtelematic/aktualizr/pull/207

OYTIS:
@patriotyk @cajun-rat Sorry, maybe it was not such a bad idea actually. We just need to change the installation process sligtly, so that OSTree tracks both branchname and refhash.

OYTIS:
Finally pulling from OSTree properly. Maybe we don't need both getEcu and getCurrent now.

src/ostree.cc
```diff
@@ -23,6 +23,32 @@ OstreeSysroot *Ostree::LoadSysroot(const std::string &path) {
...
+  if (ostree_os.empty()) {
+    OstreeDeployment *merge_deployment = ostree_sysroot_get_merge_deployment(sysroot, NULL);
+    osname = ostree_deployment_get_osname(merge_deployment);
+  } else {
+    osname = ostree_os.c_str();
+  }
+
+  deployments = ostree_sysroot_get_deployments(sysroot);
+
+  for (unsigned int i = 0; i < deployments->len; i++) {
```
patriotyk: I noticed that ostree admin status shows list of deployments where the topmost deployment is the deployment that will boot on reboot. Maybe we just can take the deployment with 0 index instead?

src/ostree.cc
```diff
@@ -23,6 +23,32 @@ OstreeSysroot *Ostree::LoadSysroot(const std::string &path) {
...
 }
 
+OstreeDeployment *Ostree::getStagedDeployment(const std::string &path, const std::string &ostree_os) {
+  OstreeSysroot *sysroot = Ostree::LoadSysroot(path);
+  GPtrArray *deployments = NULL;
+  OstreeDeployment *res = NULL;
+
+  const char *osname;
+  if (ostree_os.empty()) {
+    OstreeDeployment *merge_deployment = ostree_sysroot_get_merge_deployment(sysroot, NULL);
```
patriotyk: Are You sure that merge_deployment is deployment that will boot after reboot? What will happen if we install two packages one by one than reboot, will merge_deployment point to latest one? Will it point after reboot to current deployment?

src/ostree.cc
```diff
@@ -215,17 +242,5 @@ Json::Value OstreePackage::toEcuVersion(const Json::Value &custom) {
 
 OstreePackage OstreePackage::getEcu(const std::string &ecu_serial, const std::string &ostree_sysroot,
```
patriotyk: I think we don't need this method anymore

src/ostree.h
```diff
@@ -12,18 +12,20 @@ static const std::string BOOT_BRANCH = "/usr/share/sota/branchname";
 
 struct Ostree {
   static OstreeDeployment *getBootedDeployment(const std::string &path);
```
patriotyk: Do we use getBootedDeployment now?

src/ostree.cc
OYTIS: merge_deployment gives you either currently running or just the latest deloyment. Here we're using it merely to get the OS name.

src/ostree.cc
OYTIS: Yes, it may make sense. Here we filter out all deployments that don't belong to our OS (taking "OS" as an OSTree term), but since we're only using only one "OS" it only makes things more complicated.

src/ostree.h
OYTIS: Good point. We still do (ostree.cc:158), but we probably shouldn't.

src/ostree.cc
OYTIS: Which one would you prefer to keep: getEcu or getCurrent?

patriotyk: maybe getCurrentEcu?  ))))

OYTIS: Let's keep getEcu for now. We will need to separate Uptane notion of a package (or target in Uptane terms) from OstreePackage soon anyway.

OYTIS: Ouch, they are in different classes. I'll keep both for now and later we'll refactor everything to enable multi-ecu updates.

OYTIS:
@patriotyk I've made some changes. Please review again after CI builds.

Issue #206, opened 2017-06-01
Attempt to switch a branch results in installation error


OYTIS:
I'm not yet sure who's to blame here, additional investigation needed

PR #207, opened 2017-06-01
Report correct ostree package after installation

patriotyk:


OYTIS:
@patriotyk Please take a look at the alternative implementation in https://github.com/advancedtelematic/aktualizr/pull/205

OYTIS:
@patriotyk @cajun-rat This solution should be preferred to https://github.com/advancedtelematic/aktualizr/pull/205 due to https://github.com/advancedtelematic/aktualizr/issues/206. Of course we still get refhash from OSTree while writing branchname to a file, but if it's probably better to treat both uniformly.
Now waiting for a new commit that will make NEW_PACKAGE configurable.

OYTIS:
@patriotyk I will still want to elaborate my solution a bit. You could better focus on refatoring of Config.

PR #211, opened 2017-06-06
uptane-vector-tests now is submodule

patriotyk:
There is one problem, on my local machine the 24 and 25 tests fails. The problem is in new cannonical json library. On my laptop it interprets all \n, instead of escaping it.


heartsucker:
@patriotyk I'm not sure what you're problem is with the canonical JSON. Can you run one of the `root.json` files through it and paste the results into a gist? It should look something like

```
json
{"field":"foo\nbar"}

cjson
{"field":"foo
bar"}
```

patriotyk:
https://gist.github.com/patriotyk/eec3c7829046cb28c904e7eba89a90ee

heartsucker:
That appears to be correct to me. CJSON is **not** a subset of JSON. The incoming data should be parsed as JSON, then the `signed` field of the JSON is run through a CJSON converter that outputs bytes. Signatures are calculated over these bytes.

patriotyk:
But old cannonical json parser produces following output https://gist.github.com/patriotyk/681a7e388b8259d2cae94b36210e219d
So it was wrong yes?

cajun-rat:
@heartsucker Do you know where the canonical canonical json definition is?

heartsucker:
@cajun-rat http://wiki.laptop.org/go/Canonical_JSON

@patriotyk If that link you just gave was the output of `generator.py`, then that is correct. The data interchange format is normal JSON. If that link is the output of the CJSON lib, then it is wrong. Aktulalizr receives plain JSON and parses it. It then sends the `signed` portion of the object into the CJSON function and calculates the signatures over the bytes that are output.

cajun-rat:
OK, this doesn't make any sense to me.  "Canonical JSON is parsable with any full JSON parser," but the grammar seems to allow newlines in the middle of a string, which is new invalid according to json_pp. 

The json spec says that strings may not contain a 'control-character', which I assume is referring to Unicode category [Cc].  http://www.fileformat.info/info/unicode/char/000d/index.htm says that newline is a unicode control character.

I think I believe your assertion that CJSON isn't JSON at all (i.e. neither is a subset of the other).

Great.  Now we need a new parser :)

heartsucker:
Unfortunately, there's no RFC on CJSON, and that's the best resource. If you ignore what it says about CJSON being a subset of JSON, everything is holds. I think the best method is to parse the JSON into its C++ representation, then have a function `std::Vec<char> cjson::encode(json::Value input)`.

cajun-rat:
cjson makes more sense if you consider it a way to convert a JSON parse tree into a sequence of bytes in order to checksum it. 'Parsing' CJSON is possible, but not actually ever needed.

OYTIS:
Current implementation in the backend: https://github.com/advancedtelematic/ota-tuf/blob/master/libtuf/src/main/scala/com/advancedtelematic/libtuf/crypt/CanonicalJson.scala
Doesn't seem to do anything about linebreaks.

heartsucker:
There is a PR in `libtuf` that address that. Or at least I wrote a test case that says "oh hey your encoder is badwrong."

JIRA for issue to fix encoding: https://advancedtelematic.atlassian.net/browse/PRO-3255

cajun-rat:
Current plan is to change the TUF test vectors to match the current behaviour of the server and client. See https://github.com/advancedtelematic/tuf-test-vectors/issues/42 and https://github.com/theupdateframework/tuf/issues/457

heartsucker:
You should remove the `killall python3` before you really upset some random contributor who runs the test suite.

cajun-rat:
@heartsucker Good catch, I'd forgotten to delete that line.  Killing the server process is now handled in run_vector_tests.sh

tests/run_vector_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+virtualenv -p python3 venv
+. venv/bin/activate
+pip install -r "$1/requirements.txt"
+$1/generator.py -t uptane --signature-encoding base64 -o vectors --cjson json-subset
+$1/server.py --path "$(pwd)/vectors" &
+sleep 3
+trap 'kill %1' EXIT
+./aktualizr_uptane_vector_tests vectors/vector-meta.json
+RES=$?
+kill %1
```
heartsucker: I totally made you do extra work. You can actually remove all these lines and get the same effect

```
RES=$?
kill %1
trap - EXIT
trap
exit ${RES}
```

and change the trap to

```
trap 'EC=$?; kill %1 || true; exit $EC` EXIT
```

However, that is pedantic and the current thing works fine.

PR #214, opened 2017-06-07
Added option --poll-once,

patriotyk:
Siplified code by removing retry logic, which was added to http class.


cajun-rat:
Nice, thank you

PR #220, opened 2017-06-09
run uptane-vector-tests with valgrind

patriotyk:


cajun-rat:
...rebased on master (because of a conflicting Dockerfile change)

PR #221, opened 2017-06-12
Fix ordering of warning and info log levels

cajun-rat:
Most people would expect that the 'warning' log level is more serious than the
'info' log level.

Fixes: PRO-3290


cajun-rat:
cc @koshelev

PR #224, opened 2017-06-13
wrap pointers with shared_ptr

patriotyk:


cajun-rat:
what about using boost::intrusive_ptr?

http://www.boost.org/doc/libs/1_64_0/libs/smart_ptr/intrusive_ptr.html

patriotyk:
intrusive_ptr requires implementation of manual reference counting that in our case is no needed, so shared_ptr works well here.

PR #222, opened 2017-06-12
Openssl 1.1.0 compatibility

heartsucker:

PR #225, opened 2017-06-13
Reformat

cajun-rat:


src/uptane/uptanerepository.cc
```diff
@@ -157,9 +156,9 @@ bool Repository::ecuRegister() {
   ret = PEM_write_bio_PUBKEY(bp_public, pkey);
   if (ret != 1) {
     RSA_free(r);
-    #if openssl_version_number >= 0x01010000
+#if openssl_version_number >= 0x01010000
```
OYTIS: Should be uppercase probably.

cajun-rat: good catch

cajun-rat:
#222 doesn't work at all.  Closing and I'll create a new PR with a better name

PR #228, opened 2017-06-15
Docs to help with debugging

cajun-rat:


OYTIS:
@cajun-rat Sorry, I don't quite understand what's sota_autoprov.sh. Was it just accidentially added to the commit? There is a reference to a particular GW account probably not usable by anyone outside ATS.

PR #229, opened 2017-06-21
implementation of secondary update

patriotyk:


patriotyk:
Looks like the debian repository has some problems, we need to wait, then restart build

src/crypto.cc
```diff
@@ -192,3 +209,66 @@ bool Crypto::parseP12(FILE *p12_fp, const std::string &p12_password, const std::
 
   return true;
 }
+
+bool Crypto::generateRSAKeyPair(const std::string &public_key, const std::string &private_key) {
+  int bits = 1024;
```
heartsucker: This needs to be 2048 at a minimum according to the spec.

heartsucker:
Also, I can't fully review this as I'm unfamiliar with the internal architecture, so I'm going to need to sit down with one of you in the office and look it over.

src/crypto.cc
cajun-rat: Final version will use ED25519, but the server doesn't support that yet.  I agree that we should follow the UPTANE minimum key size though.

patriotyk: Updated with 2048 size

PR #230, opened 2017-06-22
Report hardware info

patriotyk:


src/utils.cc
```diff
@@ -167,3 +168,15 @@ void Utils::writeFile(const std::string &filename, const std::string &content) {
   file << content;
   file.close();
 }
+
+std::string Utils::getHardwareInfo() {
+  char buffer[128];
+  std::string result = "";
+  boost::shared_ptr<FILE> pipe(popen("lshw -json", "r"), pclose);
+  if (!pipe) throw std::runtime_error("popen() failed!");
```
cajun-rat: If this fails, can you catch the report the error (at warning level) and continue, please? That way if someone doesn't have lshw installed then we will continue to mostly function

patriotyk: Yes, You are right

cajun-rat:
Tested on QEMU. Waiting for the CI build

PR #231, opened 2017-06-22
Report installed packages at startup

patriotyk:


cajun-rat:
I get a HTTP/1.1 400 Bad Request when trying to run this

patriotyk:
Updated, now it works

PR #232, opened 2017-06-23
Perform sync() after provisioning and update install

cajun-rat:


patriotyk:
Build have been failed with "503  Service Temporarily Unavailable" error, and it happens very often now

PR #234, opened 2017-06-27
Http now returns HttpResponse class that contains all needed data

patriotyk:
Catch all Uptane exceptions

### DIFF
--- a/src/httpclient.h
+++ b/src/httpclient.h
@@ -3,8 +3,24 @@
 
 #include <curl/curl.h>
 #include "json/json.h"
+#include "utils.h"
 
 #include "config.h"
+
+struct HttpResponse {
+  HttpResponse(const std::string &body_in, unsigned int http_status_code_in, CURLcode curl_code_in,
+               const std::string &error_message_in)
+      : body(body_in),
+        http_status_code(http_status_code_in),
+        curl_code(curl_code_in),
+        error_message(error_message_in) {}
+  std::string body;
+  unsigned int http_status_code;
+  CURLcode curl_code;
+  std::string error_message;
+  bool isOk() { return (curl_code == CURLE_OK && http_status_code >= 200 && http_status_code < 203); }
+  Json::Value getJson() { return Utils::parseJSON(body); }
+};
 
 class HttpClient {
  public:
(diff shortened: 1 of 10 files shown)


src/httpclient.h
```diff
@@ -6,18 +6,31 @@
...
+      : body(body_in),
+        http_status_code(http_status_code_in),
+        curl_code(curl_code_in),
+        error_message(error_message_in) {}
+  std::string body;
+  unsigned int http_status_code;
+  CURLcode curl_code;
+  std::string error_message;
+  operator bool() const { return curl_code == CURLE_OK; }
+  bool operator!() const { return curl_code != CURLE_OK; }
```
cajun-rat: What about merging 3 checks together here:

1) That the http request completed OK
2) That the status code was 200 (or 20x if a response isn't needed)
3) That the response is valid JSON?

The JSON parsing could then be moved into HttpResponse, and "Utils::parseJSON(http.get(server + path).body);" would become:

~~~
HttpReponse http_response = http.get(path)
if (http_response.Ok()) {
  Json::Value json = http_response.json();
} else {
  // Handle error
}
~~~

src/sotahttpclient.cc
```diff
@@ -85,7 +84,7 @@ void SotaHttpClient::startInstall(const data::UpdateRequestId &InstallingUpdate)
 void SotaHttpClient::sendUpdateReport(data::UpdateReport update_report) {
   std::string url = core_url + "/mydevice/" + config.device.uuid;
   url += "/updates/" + update_report.update_id;
-  http->post(url, update_report.toJson()["operation_results"]);
+  http->post(url, Json::FastWriter().write(update_report.toJson()["operation_results"]));
```
cajun-rat: Looking at our code base, the only thing we send via http->post and http->put is JSON.  What about putting the JSON serialization inside the HttpClient?

patriotyk:
@cajun-rat I've updated PR with all Your suggestions except of validation of json, because not always we get json. So the solution might be to check the content-type  and parse response in case it is json.

cajun-rat:
(rebased to master)

PR #241, opened 2017-06-29
Fix path to targets

OYTIS:
Thanks @koshelev


cajun-rat:
Looks like this needs a matching change in the UPTANE test vectors.  @heartsucker: Which behaviour is correct to resolve the path to the target file?

PR #244, opened 2017-06-29
Refactor TUF implementation

cajun-rat:
The goal is to create a set of classes modeled around the concepts in TUF

### DIFF
--- a/src/uptane/tuf.cc
+++ b/src/uptane/tuf.cc
@@ -1,0 +1,230 @@
+#include <logger.h>
+#include <boost/algorithm/hex.hpp>
+#include <boost/algorithm/string/case_conv.hpp>
+#include <ctime>
+#include <ostream>
+#include <sstream>
+
+#include "crypto.h"
+#include "exceptions.h"
+#include "uptane/tuf.h"
+
+using Uptane::Hash;
+using Uptane::Target;
+using Uptane::Root;
+using Uptane::Role;
+using Uptane::TimeStamp;
+using Uptane::Version;
+
+std::string Version::RoleFileName(Role role) const {
+  std::stringstream ss;
+  if (version_ != Version::ANY_VERSION) {
+    ss << version_ << ".";
+  }
+  ss << StringFromRole(role) << ".json";
+  return ss.str();
+}
+
+std::ostream &Uptane::operator<<(std::ostream &os, const Version &v) {
+  if (v.version_ == Version::ANY_VERSION) {
+    os << "vANY";
+  } else {
+    os << "v" << v.version_;
+  }
+  return os;
+}
+
+TimeStamp TimeStamp::Now() {
+  time_t raw_time;
+  struct tm time_struct;
+  time(&raw_time);
+  gmtime_r(&raw_time, &time_struct);
+  char formatted[22];
+  strftime(formatted, 22, "%Y-%m-%dT%H:%M:%SZ", &time_struct);
+  return TimeStamp(formatted);
+}
+
+TimeStamp::TimeStamp(std::string rfc3339) {
+  if (rfc3339.length() != 20 || rfc3339[19] != 'Z') {
+    time_ = "";
+  } else {
+    time_ = rfc3339;
+  }
+}
+
+bool TimeStamp::valid() const { return time_.length() != 0; }
+
+bool TimeStamp::operator<(const TimeStamp &other) const { return valid() && other.valid() && time_ < other.time_; }
+
+bool TimeStamp::operator>(const TimeStamp &other) const { return (other < *this); }
+
+std::ostream &Uptane::operator<<(std::ostream &os, const TimeStamp &t) {
+  os << t.time_;
+  return os;
+}
+
+Role Uptane::RoleFromString(const std::string &rolename) {
+  if (rolename == "root" || rolename == "Root") {
+    return Uptane::kRoot;
+  } else if (rolename == "snapshot" || rolename == "Snapshot") {
+    return Uptane::kSnapshot;
+  } else if (rolename == "targets" || rolename == "Targets") {
+    return Uptane::kTargets;
+  } else if (rolename == "timestamp" || rolename == "Timestamp") {
+    return Uptane::kTimestamp;
+  } else {
+    return Uptane::kInvalidRole;
+  }
+}
+
+std::string Uptane::StringFromRole(Role role) {
+  switch (role) {
+    case kRoot:
+      return "root";
+    case kSnapshot:
+      return "snapshot";
+    case kTargets:
+      return "targets";
+    case kTimestamp:
+      return "timestamp";
+    default:
+      return "invalidrole";
+  }
+}
+Hash::Hash(Type type, const std::string &hash) : type_(type), hash_(boost::algorithm::to_upper_copy(hash)) {}
+
+bool Hash::matchWith(const std::string &content) const {
+  switch (type_) {
+    case sha256:
+      return hash_ == boost::algorithm::hex(Crypto::sha256digest(content));
+    case sha512:
+      return hash_ == boost::algorithm::hex(Crypto::sha512digest(content));
+    default:
+      throw std::invalid_argument("type_");  // Unreachable
+  }
+}
+
+bool Hash::operator==(const Hash &other) const { return type_ == other.type_ && hash_ == other.hash_; }
+
+std::ostream &Uptane::operator<<(std::ostream &os, const Target &t) {
+  os << "Target(" << t.filename_ << " ecu_identifier:" << t.ecu_identifier() << " length:" << t.length() << ")";
+  return os;
+}
+
+Root::Root(const std::string &repository, const Json::Value &json) : policy_(kCheck) {
+  if (!json.isObject()) {
+    LOGGER_LOG(LVL_error, "Trying to construct Tuf Root from invalid json");
+    LOGGER_LOG(LVL_trace, json);
+    policy_ = kRejectAll;
+    return;
+  }
+  Json::Value keys = json["keys"];
+  for (Json::ValueIterator it = keys.begin(); it != keys.end(); ++it) {
+    std::string key_type = boost::algorithm::to_lower_copy((*it)["keytype"].asString());
+    if (key_type != "rsa" && key_type != "ed25519") {
+      throw SecurityException(repository, "Unsupported key type: " + (*it)["keytype"].asString());
+    }
+    KeyId keyid = it.key().asString();
+    PublicKey key((*it)["keyval"]["public"].asString(), (*it)["keytype"].asString());
+    keys_[keyid] = key;
+  }
+
+  Json::Value roles = json["roles"];
+  for (Json::ValueIterator it = roles.begin(); it != roles.end(); it++) {
+    std::string role_name = it.key().asString();
+    Role role = RoleFromString(role_name);
+    if (role == kInvalidRole) {
+      LOGGER_LOG(LVL_warning, "Invalid role in root.json");
+      LOGGER_LOG(LVL_trace, "Role name:" << role_name);
+      LOGGER_LOG(LVL_trace, "root.json is:" << json);
+      continue;
+    }
+    // Threshold
+    int requiredThreshold = (*it)["threshold"].asInt();
+    if (requiredThreshold < kMinSignatures) {
+      LOGGER_LOG(LVL_debug, "Failing with threshold for role " << role << " too small: " << requiredThreshold << " < "
+                                                               << kMinSignatures);
+      throw IllegalThreshold(repository, "The role " + StringFromRole(role) + " had an illegal signature threshold.");
+    }
+    if (kMaxSignatures < requiredThreshold) {
+      LOGGER_LOG(LVL_debug, "Failing with threshold for role " << role << " too large: " << kMaxSignatures << " < "
+                                                               << requiredThreshold);
+      throw IllegalThreshold(repository, "root.json contains a role that requires too many signatures");
+    }
+    thresholds_for_role_[role] = requiredThreshold;
+
+    // KeyIds
+    Json::Value keyids = (*it)["keyids"];
+    for (Json::ValueIterator itk = keyids.begin(); itk != keyids.end(); ++itk) {
+      keys_for_role_.insert(std::make_pair(role, (*itk).asString()));
+    }
+  }
+}
+
+Json::Value Root::UnpackSignedObject(TimeStamp now, std::string repository, Role role,
+                                     const Json::Value &signed_object) {
+  if (policy_ == kAcceptAll) {
+    return signed_object["signed"];
+  } else if (policy_ == kRejectAll) {
+    throw SecurityException(repository, "Root policy is kRejectAll");
+  }
+  assert(policy_ == kCheck);
+
+  std::string canonical = Json::FastWriter().write(signed_object["signed"]);
+  Json::Value signatures = signed_object["signatures"];
+  int valid_signatures = 0;
+
+  for (Json::ValueIterator sig = signatures.begin(); sig != signatures.end(); ++sig) {
+    std::string method((*sig)["method"].asString());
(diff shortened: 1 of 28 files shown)


src/uptane/tuf.h
```diff
@@ -0,0 +1,72 @@
...
+ public:
+  enum Type { invalid, sha256, sha512 };
+
+  Hash() : type_(invalid) {}
+  Hash(Type type, const std::string &hash, int64_t length);
+
+  /**
+   * Use to avoid downloading indefinitely long (corrupted) files
+   */
+  int64_t length() const { return length_; }
```
patriotyk: this is bad design to add size of target to Hash. Only target should contain its size

src/uptane/tuf.h
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * The hash of a file or TUF metadata.  File hashes/checksums in TUF include the length of the object, in order to
+ * defeat infinite download attacks.
+ */
+class Hash {
+ public:
+  enum Type { invalid, sha256, sha512 };
+
+  Hash() : type_(invalid) {}
```
patriotyk: why do we need this?

src/uptane/tuf.h
```diff
@@ -0,0 +1,72 @@
...
+   * @return
+   */
+  bool matchWith(const std::string &content) const;
+
+  bool operator==(const Hash &other) const;
+
+ private:
+  Type type_;
+  std::string hash_;
+  int64_t length_;
```
patriotyk: It is bad to add size of target to hash object

src/uptane/tuf.h
```diff
@@ -0,0 +1,72 @@
...
+#ifndef AKTUALIZR_UPTANE_TUF_H_
+#define AKTUALIZR_UPTANE_TUF_H_
+
+#include <ostream>
+
+namespace Uptane {
+
+/**
+ * The hash of a file or TUF metadata.  File hashes/checksums in TUF include the length of the object, in order to
+ * defeat infinite download attacks.
```
patriotyk: I don't understand how it could help to avoid  infinite download attacks.

src/uptane/tuf.h
heartsucker: There is also the possibility of multiple hashes per target/metadata, so you'd have redundant data.

src/uptane/tuf.h
heartsucker: You only download up to `N` bytes and then abort if it exceeds it.

src/uptane/tuf.h
cajun-rat: I put it in to make Hash default constructable.  It might make more sense to have target take a list of Hashes

src/uptane/tuf.h
cajun-rat: The plan was to have an object that holds the 'validity' of a target--before this we were passing the hash and length around together a lot, and so it makes sense to keep them together.

I might try a design that uses Target for this instead

patriotyk: Yes I think it would be better to hold it in the target

cajun-rat:
I'm working on this now, should have a v2 PR ready today

cajun-rat:
Tested on AGL/qemu

cajun-rat:
@patriotyk : Can you re-review, please?

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+    std::string keyid = (*sig)["keyid"].asString();
+    if (!keys_.count(keyid)) {
+      LOGGER_LOG(LVL_info, "Signed by unknown keyid, skipping");
+      LOGGER_LOG(LVL_trace, "Key ID: " << keyid);
+      continue;
+    }
+
+    if (!keys_for_role_.count(std::make_pair(role, keyid))) {
+      LOGGER_LOG(LVL_warning, "KeyId is not valid to sign for this role");
+      LOGGER_LOG(LVL_trace, "KeyId: " << keyid);
```
patriotyk: Is it ok, that we verify signature even it is signed by wrong key?

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+    }
+  }
+  int threshold = thresholds_for_role_[role];
+  if (threshold < kMinSignatures || kMaxSignatures < threshold) {
+    throw IllegalThreshold(repository, "Invalid signature threshold");
+  }
+  if (valid_signatures < threshold) {
+    throw UnmetThreshold(repository, StringFromRole(role));
+  }
+  Json::Value res = Utils::parseJSON(canonical);
```
patriotyk: Why not just use signed_object["signed"] instead of parsing it again?

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+
+  // TODO check _type matches role
+  // TODO check timestamp
+  Uptane::TimeStamp expiry(Uptane::TimeStamp(res["expires"].asString()));
+  if (!expiry.valid() || expiry < TimeStamp::Now()) {
```
patriotyk: maybe  'valid()' should check `expiry < TimeStamp::Now()`

src/uptane/tuf.h
```diff
@@ -0,0 +1,151 @@
+/**
+ * Base data types that are used in The Update Framework (TUF), part of UPTANE.
+ */
+
+#ifndef AKTUALIZR_UPTANE_TUF_H_
```
patriotyk: Those defines should be the first lines of the header

src/uptane/tufrepository.cc
```diff
@@ -38,207 +49,95 @@ Json::Value TufRepository::getJSON(const std::string& role) {
...
+  Json::Value content = getJSON(version.RoleFileName(role));
+  TimeStamp now(TimeStamp::Now());
+  Json::Value result = root_.UnpackSignedObject(now, name_, role, content);
+  if (role == kRoot) {
+    // Also check that the new root is suitably self-signed
+    Root new_root = Root(name_, result);
+    new_root.UnpackSignedObject(TimeStamp::Now(), name_, kRoot, content);
+  }
+  // UnpackSignedObject throws on error--write the (now known to be valid) content to disk
+  std::ofstream file((path_ / (StringFromRole(role) + ".json")).string().c_str());
```
patriotyk: we can use Utils::writeFile

src/uptane/tufrepository.cc
```diff
@@ -38,207 +49,95 @@ Json::Value TufRepository::getJSON(const std::string& role) {
...
   file << content;
   file.close();
+  return result;
+}
+
+void TufRepository::verifyRole(Uptane::Role role, const Json::Value& tuf_signed) {
+  Uptane::TimeStamp now("2017-01-01T01:00:00Z");
```
patriotyk: "2017-01-01T01:00:00Z" is it ok?

src/uptane/tufrepository.cc
```diff
@@ -38,207 +49,95 @@ Json::Value TufRepository::getJSON(const std::string& role) {
...
+      Hash::Type type;
+      std::string hash_string;
+      if (hashes.isMember("sha512")) {
+        type = Hash::sha512;
+        hash_string = hashes["sha512"].asString();
+      } else if (hashes.isMember("sha256")) {
+        type = Hash::sha256;
+        hash_string = hashes["sha256"].asString();
       }
+      Hash hash(type, hash_string);
```
patriotyk: We could reimplement Hash to accept `type` in string format, then we could iterate through `hashes` and append them to the list, then pass this list to the Target.

tests/uptane_vector_tests.cc
```diff
@@ -83,6 +84,7 @@ bool run_test(Json::Value vector) {
 }
 
 int main(int argc, char* argv[]) {
+  ::testing::InitGoogleTest(&argc, argv);
```
patriotyk: We don't use google tests in this test why do we need this?

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+using Uptane::Role;
+using Uptane::TimeStamp;
+using Uptane::Version;
+
+std::string Version::RoleFileName(Role role) const {
+  std::stringstream ss;
+  if (version_ != Version::ANY_VERSION) {
+    ss << version_ << ".";
+  }
+  ss << StringFromRole(role) << ".json";
```
cajun-rat: ...except:
https://google.github.io/styleguide/cppguide.html#Function_Names

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+
+TimeStamp::TimeStamp(std::string rfc3339) {
+  if (rfc3339.length() != 20 || rfc3339[19] != 'Z') {
+    time_ = "";
+  } else {
+    time_ = rfc3339;
+  }
+}
+
+bool TimeStamp::valid() const { return time_.length() != 0; }
```
cajun-rat: Agreed

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+
+bool TimeStamp::operator>(const TimeStamp &other) const { return (other < *this); }
+
+std::ostream &Uptane::operator<<(std::ostream &os, const TimeStamp &t) {
+  os << t.time_;
+  return os;
+}
+
+Role Uptane::RoleFromString(const std::string &rolename) {
+  if (rolename == "root" || rolename == "Root") {
```
cajun-rat: Yep, agreed

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+
+bool Hash::operator==(const Hash &other) const { return type_ == other.type_ && hash_ == other.hash_; }
+
+std::ostream &Uptane::operator<<(std::ostream &os, const Target &t) {
+  os << "Target(" << t.filename_ << " ecu_identifier:" << t.ecu_identifier() << " length:" << t.length() << ")";
+  return os;
+}
+
+Root::Root(const std::string &repository, const Json::Value &json) : policy_(kCheck) {
+  if (!json.isObject()) {
```
cajun-rat: Agreed

src/uptane/tuf.cc
cajun-rat: Good catch, this was missing a 'continue' line

src/uptane/tuf.cc
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    std::string signature = (*sig)["sig"].asString();
+    if (Crypto::VerifySignature(keys_[keyid], signature, canonical)) {
+      valid_signatures++;
+    } else {
+      LOGGER_LOG(LVL_warning, "Signature was present but invalid");
+      LOGGER_LOG(LVL_trace, "  keyid:" << keyid);
+      LOGGER_LOG(LVL_trace, "  signature:" << signature);
+      // throw SecurityException(repository, "Invalid signature, verification failed");
```
cajun-rat: I'm unsure what the right thing to do here is.  The security check is based on getting a 'threshold' number of valid signatures, so there isn't a problem if there are signatures that are present but that we don't trust.

Currently we just ignore invalid signatures (i.e. we treat them as if they weren't present, or we didn't have the matching public key).

src/uptane/tuf.cc
cajun-rat: This is is to improve the security of the system: the idea is to avoid attacks where the attacker creates two inputs that behave differently but which canonicalise to the same value. For example:

~~~
Original Signed value: {"a":"real_image"}
Hacker's modified value: {"a":"real_image", "a":"Bad_Hacker_image"}
Canonicalised string: '{"a":"real_image"}'
~~~

(In reality the structure would be much more complex, maybe abusing Unicode normalization). Both of these json blobs canonicalise to the same string, but are handled differently when they are read. 

The solution to these attacks is to re-parse exactly the same set of bytes that was used to check the signature, then this whole class of attacks is defeated.

src/uptane/tuf.cc
cajun-rat: I'd like to inject the current time in from the outside (it makes testing easier, and makes the dependency explicit).  The operator< actually checks validity (invalid behaves like NaN in floating point)

Maybe bool TimeStamp::HasExpiredAt(TimeStamp current_time)?

src/uptane/tuf.h
cajun-rat: Ack

src/uptane/tuf.h
```diff
@@ -0,0 +1,151 @@
...
+
+#include "crypto.h"
+
+namespace Uptane {
+
+typedef std::string KeyId;
+/**
+ * TUF Roles
+ */
+enum Role { kRoot, kSnapshot, kTargets, kTimestamp, kInvalidRole };
```
cajun-rat: Good idea, done

src/uptane/tuf.h
```diff
@@ -0,0 +1,151 @@
...
+
+  bool MatchWith(const std::string &content) const {
+    // TODO: We should have some priority/ordering here
+    return content.length() == length_ && hash_.matchWith(content);
+  }
+
+  int64_t length() const { return length_; }
+
+  bool IsForSecondary(const std::string &ecuIdentifier) const {
+    return length() > 0 && ecu_identifier() == ecuIdentifier;
```
cajun-rat: Yep

src/uptane/tuf.h
```diff
@@ -0,0 +1,151 @@
...
+  std::string filename_;
+  Hash hash_;
+  int64_t length_;
+};
+
+std::ostream &operator<<(std::ostream &os, const Target &t);
+
+class Root {
+ public:
+  enum Policy { kRejectAll, kAcceptAll, kCheck };
```
cajun-rat: Yes: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

src/uptane/tufrepository.cc
cajun-rat: Yeah.  Content is JSON: I'll add a "Utils::writeFile(const std::string &filename, const Json::Value &content)"

src/uptane/tufrepository.cc
cajun-rat: This is only used in testing at the moment--I'd prefer test code didn't depend on the current time.  verifyRole is maybe not a great name though.  I'll try and think of a better one

tests/uptane_vector_tests.cc
cajun-rat: clion was passing some random gtest parameters to the test class, so adding this caused them to be passed to GoogleTest (and removed from argv) rather than causing the test to fail

src/uptane/tuf.h
```diff
@@ -0,0 +1,151 @@
...
+
+class Target {
+ public:
+  Target(const Json::Value &custom, const std::string &filename, Hash hash, int64_t length)
+      : custom_(custom), filename_(filename), hash_(hash), length_(length) {}
+  bool operator==(const Target &t2) const { return (filename_ == t2.filename_ && hash_ == t2.hash_); }
+
+  std::string ecu_identifier() const { return custom_["ecuIdentifier"].asString(); }
+
+  std::string filename() const { return filename_; }
```
cajun-rat: I would like to stick to https://google.github.io/styleguide/cppguide.html#Function_Names - I know there are quite a few deviations from that in the codebase, but we should uphold it as the standard :)

src/uptane/tuf.h
```diff
@@ -0,0 +1,151 @@
...
+  enum Policy { kRejectAll, kAcceptAll, kCheck };
+  /**
+   * An empty Root, that either accepts or rejects everything
+   */
+  Root(Policy policy) : policy_(policy) {}
+  /**
+   * json should be the contents of the 'signed' portion
+   * @param json
+   */
+  Root(const std::string &repository, const Json::Value &json);
```
cajun-rat: Do you mean passing in the whole JSON blob, not just the "signed" part?

patriotyk: No, I mean that signed json object could be the class member. And all operations with it could be encapsulated in this class

cajun-rat:
OK, should be good to go now

PR #250, opened 2017-07-04
Move hardcoded firmware path to config file

patriotyk:


OYTIS:
@patriotyk Looks like all the emulated secondaries are going to write to a single file. Wouldn't it be better to make firmware_path a directory, and store blobs under <firmware_path>/<hardware_id>/<ecu_id>?

cajun-rat:
Rebased

PR #251, opened 2017-07-04
Remove unsupported sha512 digest

taheris:
Resolves [PRO-3472](https://advancedtelematic.atlassian.net/browse/PRO-3472).

@patriotyk: the director doesn't support sha512 digests yet


OYTIS:
@taheris I'd really want it fixed rather on the server-side. It will probably be the only hashing algorithm supported by physical secondaries, since it comes as a part of ed25519 and it additional algorithm will cost us 1Kb flash more.

taheris:
@oytis: fair enough. I've updated the ticket to amend the Director service so I'll close this.

cajun-rat:
Long term we need to change the director, but this is breaking prod at the moment.
(Rebased)

PR #258, opened 2017-07-06
wait more time for server start

patriotyk:


cajun-rat:
Is there any way we can wait until it has actually started, rather than adding more sleep()?

patriotyk:
We could try to connect to 80 port in loop

cajun-rat:
What about making the script pick a random port number, and write it to a temporary named pipe that is  passed in on the command line?

That would also let the tests run in parallel, which would save a huge amount of time?

cajun-rat:
I'll merge this as a temporary fix.  A proper solution (that doesn't require sleep() calls) can come as part of the work to be able to run the tests in parallel (i.e. removing CTEST_PARALLEL_LEVEL=1)

PR #267, opened 2017-07-12
Change update logic to same as in python implementation

patriotyk:
Allow redirects for image files

### DIFF
--- a/src/uptane/tuf.h
+++ b/src/uptane/tuf.h
@@ -107,7 +107,6 @@ class Target {
   Target(const std::string &name, const Json::Value &content);
 
   std::string ecu_identifier() const { return ecu_identifier_; }
-
   std::string filename() const { return filename_; }
 
   bool MatchWith(const std::string &content) const;
@@ -119,7 +118,7 @@ class Target {
   };
 
   bool operator==(const Target &t2) {
-    return (filename_ == t2.filename_ && length_ == t2.length_ && ecu_identifier_ == t2.ecu_identifier_ &&
+    return (filename_ == t2.filename_ && length_ == t2.length_ &&
             std::equal(hashes_.begin(), hashes_.end(), t2.hashes_.begin()));
   }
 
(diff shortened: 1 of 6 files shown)


src/uptane/tuf.cc
```diff
@@ -91,12 +91,9 @@ bool Hash::MatchWith(const std::string &content) const {
...
 bool Hash::operator==(const Hash &other) const { return type_ == other.type_ && hash_ == other.hash_; }
 
-Target::Target(const std::string &filename, const Json::Value &content) : filename_(filename), ecu_identifier_("") {
+Target::Target(const std::string &filename, const Json::Value &content) : filename_(filename) {
   if (content.isMember("custom")) {
-    Json::Value custom = content["custom"];
-    if (custom.isMember("ecuIdentifier")) {
-      ecu_identifier_ = content["custom"]["ecuIdentifier"].asString();
-    }
+    custom_ = content["custom"];
```
cajun-rat: Why is this needed?  I'm not keen on exposing implementation details (i.e. the format of the 'custom' JSON field, or the fact that the serialization of Targets happens to be JSON at the moment) via the API of Uptane::Target.

What didn't work with the old method?

patriotyk: It is because image and director has different formats of custom field, for example director contains "ecuIdentifier" but image not. But image contains hardware identifiers which might be needed in the future.

cajun-rat: My concern is that returning a Json::Value means that the users of Target are locked-in to the specific serialization format for targets.  If we decide to support ASN.1 DER encoded targets files, then the Json::Value Target::custom() member function will need changing everywhere.

Instead of an explicit property, why not put the logic for matching a target against an EcuIdentifier inside the Target object itself?  Done that was it doesn't matter where the Target object came from.  You just ask 'are you for this EcuIdentifier?'.  Targets from image repository will always say 'yes', but targets from director will check the custom field.

In the future we may want a 'Ecu' object that includes both the ecuIdentifier and hardwareIdentifier.  In that case the Image repository would perform a check against hardwareIdentifier, but the users of a Target object wouldn't need to worry about the details.

An std::string Targets::ecu_identifier() member function would be fine too, but it shouldn't return Json :)

PR #275, opened 2017-07-18
Update dependencies to reflect actual state of affairs.

patrickvacek:


cajun-rat:
Nice, thank you for making this better!

PR #278, opened 2017-07-19
main.cc: Handle missing configuration file

leon-anavi:
Gracefully terminate the application if the
configuration file is missing or if there is another
runtime unhandled exception.

Although the Aktualizr recipe in Yocto/OE layer
meta-updater deploys sota.toml as a read-only file,
it will be missing if the user forgets to specify
variable SOTA_PACKED_CREDENTIALS. This fix ensures
that in this case Aktualizr will not crash.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -119,10 +119,13 @@ int main(int argc, char *argv[]) {
   }
 
   // Initialize config with default values, the update with config, then with cmd
-  std::string filename = commandline_map["config"].as<std::string>();
-  Config config(filename, commandline_map);
-
-  Aktualizr aktualizr(config);
-
-  return aktualizr.run();
+  try {
+    std::string filename = commandline_map["config"].as<std::string>();
+    Config config(filename, commandline_map);
+    Aktualizr aktualizr(config);
+    return aktualizr.run();
+  } catch(std::runtime_error e) {
+    LOGGER_LOG(LVL_error, "ERROR: " << e.what());
+    exit(EXIT_FAILURE);
+  }
 }


src/main.cc
```diff
@@ -119,10 +119,13 @@ int main(int argc, char *argv[]) {
...
-
-  Aktualizr aktualizr(config);
-
-  return aktualizr.run();
+  try {
+    std::string filename = commandline_map["config"].as<std::string>();
+    Config config(filename, commandline_map);
+    Aktualizr aktualizr(config);
+    return aktualizr.run();
+  } catch(std::runtime_error e) {
```
cajun-rat: I'm not keen on using exceptions to handle this case.  In general exceptions are best kept for 'exceptional' cases, and user configuration error might be better handled with an explicit test e.g.:

~~~
if (!boost::filesystem::exists(filename)) {
  ...
~~~

(While you are at it, perhaps you could change `filename` to a more descriptive name, such as `config_filename` :)

Finally, a test would be nice--there are some examples in tests/CMakeLists.txt for how to do this.  A good example to copy is `test_log_invalid`, which uses sets the test property  `WILL_FAIL` to true to tell the test driver that the test should return 1.

There is also `PASS_REGULAR_EXPRESSION` (see https://cmake.org/cmake/help/v3.0/command/set_tests_properties.html) which can check that the right error message is given to the user.

patriotyk: This PR non needed. #281 already fixes this.

cajun-rat:
I'd still be happy with an explicit test for this, and a test case to check that it keeps working :)

PR #282, opened 2017-07-19
Fetch pre-built Docker image

cajun-rat:
These images are build using an internal CI system at ATS, and pushed to
Dockerhub.  They are only used as a cache to speed up the build process, and
the correctness of the build does not depend on them being up-to-date.


cajun-rat:
This speeds up the Docker build by about 90 seconds.

PR #284, opened 2017-07-20
Allow to create config from an existing property tree

koshelev:


src/config.cc
```diff
@@ -84,6 +84,10 @@ Config::Config(const std::string& filename, const boost::program_options::variab
   postUpdateValues();
 }
 
+Config::Config(const boost::property_tree::ptree& pt) {
+  updateFromPropertyTree(pt);
+}
```
patriotyk: To properly work, we need to add postUpdateValues()  here.

koshelev: Fixed, thanks for pointing at it.

cajun-rat:
@patriotyk's changes are now in, merging.

PR #285, opened 2017-07-20
Use absolute paths when signing the manifest

OYTIS:
Not sure how it could work before, but relative paths definitely break @koshelev's [load tests](https://github.com/advancedtelematic/ota-load-tests-aktualizr)


koshelev:
Seems like addressing @patriotyk 's comments in https://github.com/advancedtelematic/aktualizr/pull/284 fixes it for load tests. The path seems to be absolute.

cajun-rat:
Is this still needed?

OYTIS:
@cajun-rat @koshelev Probably not, these variables are already set to absolute paths in postUpdateValues(). Closing.

PR #286, opened 2017-07-20
Make device registration and ECU registration independent for more robustness

OYTIS:
A partial solution until [PRO-3630](https://advancedtelematic.atlassian.net/browse/PRO-3630) is implemented.

### DIFF
--- a/src/uptane/uptanerepository.cc
+++ b/src/uptane/uptanerepository.cc
@@ -150,6 +150,7 @@ bool Repository::deviceRegister() {
   }
   fclose(device_p12);
   sync();
+  LOGGER_LOG(LVL_info, "Provisioned successfully on Device Gateway");
   return true;
 }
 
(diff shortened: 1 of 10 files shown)


src/uptane/uptanerepository.cc
```diff
@@ -177,8 +193,14 @@ bool Repository::ecuRegister() {
   authenticate();
   HttpResponse response = http.post(config.tls.server + "/director/ecus", all_ecus);
   if (!response.isOk()) {
-    LOGGER_LOG(LVL_error, "Error registering device on Uptane, response: " << response.body);
-    return false;
+    Json::Value resp_code = Utils::parseJSON(response.body)["code"];
```
patriotyk: Response has method getJson() that returns parsed body

cajun-rat: Good point, fixed

PR #289, opened 2017-07-21
Drop HTTP-SOTA support

OYTIS:
A better observable codebase should make refactoring easier.


patriotyk:
Not all code has been removed. For example authenticate method of http. Also we have some specific events and commands. Also download method of http class is also specific. And config options.

OYTIS:
@patriotyk
Thank you for the review, I've tried to track all the unused events and commands and delete them. I've also dropped support for socketgateway after a discussion with @cajun-rat . Please take a look if I've forgotten something or maybe dropped something which is really important to us.

OYTIS:
@cajun-rat @patriotyk Returned socketgateway until we can discuss its deletion with @christianhuggerATS

OYTIS:
@patrickvacek 
Can't pass check-format. I've tried clang-format v3.9.1-2 and v3.8.1-1.

cajun-rat:
Tried reformatting using the version on my machine.  I had assumed that its behaviour would be more stable...

patrickvacek:
Rather disappointing that clang-format would yield such differences between seemingly similar versions. Could this be a configuration issue? I would hope we could rely on some amount of consistency.

src/eventsinterpreter.cc
```diff
@@ -29,17 +29,6 @@ void EventsInterpreter::run() {
     gateway_manager.processEvents(event);
 
     if (event->variant == "DownloadComplete") {
```
patriotyk: I think we don't need such empty blocks

OYTIS:
@patriotyk Done, thanks.

PR #291, opened 2017-07-25
Fix memory leaks,

patriotyk:
Instead of openssl use libsodium for hash functions

### DIFF
--- a/src/aktualizr.cc
+++ b/src/aktualizr.cc
@@ -39,18 +39,21 @@ Aktualizr::Aktualizr(const Config &config) : config_(config) {
 #endif
 }
 
-int Aktualizr::run() {
-  command::Channel *commands_channel = new command::Channel();
-  event::Channel *events_channel = new event::Channel();
+Aktualizr::~Aktualizr() { CRYPTO_cleanup_all_ex_data(); }
 
-  EventsInterpreter events_interpreter(config_, events_channel, commands_channel);
+int Aktualizr::run() {
+  command::Channel commands_channel;
+  event::Channel events_channel;
+
+  EventsInterpreter events_interpreter(config_, &events_channel, &commands_channel);
+
   // run events interpreter in background
   events_interpreter.interpret();
 
   if (config_.gateway.rvi) {
 #ifdef WITH_GENIVI
     try {
-      SotaRVIClient(config_, events_channel).runForever(commands_channel);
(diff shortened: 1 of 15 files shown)


src/aktualizr.cc
```diff
@@ -40,18 +40,23 @@ Aktualizr::Aktualizr(const Config &config) : config_(config) {
 #endif
 }
 
+Aktualizr::~Aktualizr() {
+  CRYPTO_cleanup_all_ex_data();
```
cajun-rat: I'm struggling to find good documentation on this function, but what happens if someone creates two Aktualizr objects in the same application?

Would curl_global_cleanup (to match curl_global_init) fix the problem?

src/aktualizr.cc
```diff
@@ -40,18 +40,23 @@ Aktualizr::Aktualizr(const Config &config) : config_(config) {
...
+}
+
 int Aktualizr::run() {
-  command::Channel *commands_channel = new command::Channel();
-  event::Channel *events_channel = new event::Channel();
+  command::Channel commands_channel;
+  event::Channel events_channel;
+
+  EventsInterpreter events_interpreter(config_, &events_channel, &commands_channel);
```
cajun-rat: +1 to nuking 'new'

patriotyk:
I have updated, now it is ready for review.

cajun-rat:
I get the following error from the production server with this change:

{code}
[2017-07-31 11:43:49.718849]: response: {"code":"signature_not_valid","description":"The given signature is not valid"}
{code}

Does it work for you?

cajun-rat:
I'm seeing the following test failures after rebasing against master:

~~~
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] SotaUptaneClientTest.device_registered
[  FAILED  ] SotaUptaneClientTest.RunForeverNoUpdates
[  FAILED  ] SotaUptaneClientTest.RunForeverHasUpdates
[  FAILED  ] SotaUptaneClientTest.RunForeverInstall

 4 FAILED TESTS
~~~

src/aktualizr.cc
```diff
@@ -39,18 +39,21 @@ Aktualizr::Aktualizr(const Config &config) : config_(config) {
 #endif
 }
 
+Aktualizr::~Aktualizr() { CRYPTO_cleanup_all_ex_data(); }
```
cajun-rat: Still not happy about this, but we can fix this another time

PR #294, opened 2017-07-26
Improve module isolation

OYTIS:
Moves part of application logic from Uptane::Repository.
It is the beginning of a relatively big refactoring process. Final goals:
* Make Uptane::Repository a thin layer abstracting procedures related to downloading/verifying/pushing UPTANE metadata. All the installation should occur in application-related modules (mostly SotaUptaneClient).
* Partition Config so that each module uses only its own chunk.
* Make initialization procedure traceable, gather scattered initialization-related code from config.cc and uptane/uptanerepository.cc to a single class/function.
* Replace (almost) all interaction with filesystem with transactional writes to a DB (most likely sqlite3).

This PR just moves OSTree installation to SotaUptaneClient and makes Config slightly better partitioned.

### DIFF
--- a/src/uptane/uptanerepository.cc
+++ b/src/uptane/uptanerepository.cc
@@ -62,24 +62,18 @@ void Repository::refresh() {
   image.fetchAndCheckRole(Role::Timestamp());
 }
 
-std::vector<Uptane::Target> Repository::getNewTargets() {
+std::pair<uint32_t, std::vector<Uptane::Target>> Repository::getTargets() {
   refresh();
 
-  std::vector<Uptane::Target> director_targets = director.fetchTargets();
-  std::vector<Uptane::Target> image_targets = image.fetchTargets();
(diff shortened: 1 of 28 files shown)


src/uptane/uptanerepository.cc
```diff
@@ -62,24 +62,18 @@ void Repository::refresh() {
   image.fetchAndCheckRole(Role::Timestamp());
 }
 
-std::vector<Uptane::Target> Repository::getNewTargets() {
+std::pair<uint32_t, std::vector<Uptane::Target> > Repository::getTargets() {
```
patriotyk: Why not add version to target class? This  looks too complicated

OYTIS: It's a version of targets.json, not an individual target.

cajun-rat:
Just a note that we need to duplicate certificates_directory to the [tls] section of the config file in meta-updater.

PR #296, opened 2017-07-27
Bugfix/pro 3582/reenable werror

patrickvacek:
The most controversial parts are probably getting rid of the one test (because setting an enum to an invalid value is disallowed with -Wall -Werror) and moving a few RSA arrays to the heap (because ISO C++ doesn't support VLAs).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,12 +67,27 @@ include_directories(${PROJECT_SOURCE_DIR}/third_party/rvi_lib/include)
 include_directories(${LIBOSTREE_INCLUDE_DIRS})
 include_directories(${LIBDBUS_INCLUDE_DIRS})
 
-set_source_files_properties(${PROJECT_SOURCE_DIR}/third_party/jsoncpp/jsoncpp.cpp PROPERTIES COMPILE_FLAGS -w)
-
 if(BUILD_GENIVI)
     set_property(GLOBAL PROPERTY CXX_STANDARD 11)
     add_subdirectory("third_party/rvi_lib")
 endif(BUILD_GENIVI)
+
+# Setup warnings. Do this after rvi_lib is added so that it isn't affected.
+if (CMAKE_COMPILER_IS_GNUCXX)
+    add_definitions(-fstack-protector-all)
+    # Enable maximum set of warnings. -Wno-sign-compare is required because of
+    # problems in gtest. -Wswitch-default would be nice as well, but also causes
+    # problems in gtest.
+    add_definitions(-Wall -Wextra -Wformat-security -Wfloat-equal -Wcast-qual -Wconversion -Wlogical-op -Wno-sign-compare)
+
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "4.9" OR MAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.9")
+        add_definitions(-Wshadow)
+    endif ()
+
+    if(WARNING_AS_ERROR)
+        add_definitions(-Werror)
+    endif()
+endif()
 
 add_subdirectory("src")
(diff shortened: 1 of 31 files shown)


OYTIS:
@patrickvacek 
Maybe we could use [boost::scoped_array](http://www.boost.org/doc/libs/1_55_0/libs/smart_ptr/scoped_array.htm) for these. Or just reconsider our policy about C++ standard resvision.

patrickvacek:
@OYTIS The scoped_array looks handy and would clean things up a bit, but it's still dynamically allocated memory on the heap, so it wouldn't change what this patch does. But what is our policy about C++ standard revisions?

OYTIS:
@patrickvacek Yes, sure, that's just to get proper RAII. I don't think that for Linux environment moving data from stack to heap makes huge difference otherwise.
Now our policy is just using C++03. Maybe it's worth to either make sure that aktualizr really compiles with old versions of gcc that don't support c++1x or allow some restricted subset of newer standards. We'll have to discuss that with @cajun-rat anyway.

CMakeLists.txt
```diff
@@ -67,13 +68,26 @@ include_directories(${PROJECT_SOURCE_DIR}/third_party/rvi_lib/include)
 include_directories(${LIBOSTREE_INCLUDE_DIRS})
 include_directories(${LIBDBUS_INCLUDE_DIRS})
 
-set_source_files_properties(${PROJECT_SOURCE_DIR}/third_party/jsoncpp/jsoncpp.cpp PROPERTIES COMPILE_FLAGS -w)
-
 if(BUILD_GENIVI)
-    set_property(GLOBAL PROPERTY CXX_STANDARD 11)
```
patrickvacek: Cmake doesn't have a specific c++03 option. The code still compiles if c++98 is used without -Wpedantic. That checks for several things that cause problems.

I didn't realize we might be restricted to C++03. In fact, there already was a specific reference to building with C++11, and it was set as a global property.

And by the way, I've now incorporated scoped_arrays.

patrickvacek:
Oddly, cmake doesn't have a specific c++03 option. However, the code still compiles if c++98 is used without -Wpedantic. With it, there are several things that cause problems.

I didn't realize we might be restricted to C++03. In fact, there already was a specific reference to building with C++11 in the root CMakeLists.txt, and it was set as a global property.

And by the way, I've now incorporated scoped_arrays.

OYTIS:
Great, thank you. I wonder what has happened with build time. Could it be -fstack-protector-all? I'll try to reproduce locally.

patrickvacek:
I can't figure out what the deal is with the CI builds. I restarted again just in case. I can try removing -fstack-protector-all, but that was one of the original flags, so I assumed it was desirable and that it meant CI worked with it in the past.

OYTIS:
By the way, gmock (build within `make build_tests`) doesn't compile with -Werror with gcc 7.1.1. It's an Arch-specific problem of course, but we're likely to run into it on other platforms as well. Maybe it can make sense to restrict the scope of -Werror with our own code. Adding '-Wno-error=conversion' solves the problem as well, but it's kind of a hack.

patrickvacek:
@OYTIS  That's unfortunate, but you'll note I already had to add -Wno-switch-default -Wno-sign-compare because of gtest.

You can always disable WARNING_AS_ERROR, though, if you don't want the strict error checking.

patrickvacek:
@OYTIS I removed -Wpedantic from the default flags and rebased to clean things up. CI now seems fine although I'm not sure what really changed. I did try removing fstack-protector-all, but it only slows down the build by a few seconds. If -Wconversion still gives you trouble, we can just remove it for now (but add it to my new PEDANTIC_WARNINGS section if you really want it).

PR #297, opened 2017-07-28
add 'ecu_version_manifests' to manifest

patriotyk:


cajun-rat:
I get the following error from the production server with this change:

{code}
[2017-07-31 11:43:49.718849]: response: {"code":"signature_not_valid","description":"The given signature is not valid"}
{code}

Does it work for you?

src/uptane/uptanerepository.cc
```diff
@@ -42,13 +42,22 @@ void Repository::putManifest() { putManifest(Json::nullValue); }
...
+  }
 
   Json::Value unsigned_ecu_version =
       OstreePackage::getEcu(config.uptane.primary_ecu_serial, config.ostree.sysroot).toEcuVersion(custom);
   Json::Value ecu_version_signed =
       Crypto::signTuf(config.uptane.private_key_path, config.uptane.public_key_path, unsigned_ecu_version);
   version_manifest["ecu_version_manifest"].append(ecu_version_signed);
+  version_manifest["ecu_version_manifests"][config.uptane.primary_ecu_serial] = ecu_version_signed;
+
   Json::Value tuf_signed =
```
cajun-rat: I get the following error from the production server with this change:

[2017-07-31 11:43:49.718849]: response: {"code":"signature_not_valid","description":"The given signature is not valid"}

cajun-rat:
Problem is that we can't merge this until the ecu_manifests change gets into production. @Danten: Do you have a ETA for that?

Danten:
@cajun-rat the code is in QA, so whenever the next push to production happens this will be in, I don't know when the next one of those will happen. @EpicNuts ?

EpicNuts:
@Danten It'll be Monday earliest... Tuesday if we _need_ Jerry to do it

cajun-rat:
The pull request for this change has just been merged.  It will at least a few days before it is in production.

cajun-rat:
I don't think these server-side changes are going to happen, and this PR is now rather stale. I'll keep the branch in case we need this in the future

PR #299, opened 2017-07-31
main.cc: Exit if configuration file doesn't exist

leon-anavi:
Show user friendly error message and terminate the
application if the specified configuration file
does not exist.


cajun-rat:
Nice, thank you!

PR #300, opened 2017-08-01
More logging

cajun-rat:


src/timer.h
```diff
@@ -0,0 +1,22 @@
+#ifndef AKTUALIZR_TIMER_H
```
cajun-rat: OK, now fixed (I used TIMER_H_, to match the other header files)

cajun-rat:
@patriotyk when you are happy, could you mark this review as 'OK', please?

PR #301, opened 2017-08-02
Ci provision

patriotyk:
To run tests  You need to run cmake with `SOTA_PACKED_CREDENTIALS="path_to_file.zip"` then run tests as usual  

Big coverage decrease is because we don't run real ostree class now. May be we need to add new tests to ostree tests, but this will require download ostree repo.

### DIFF
--- a/src/uptane/uptanerepository.cc
+++ b/src/uptane/uptanerepository.cc
@@ -37,9 +37,9 @@ void Repository::updateRoot(Version version) {
   image.updateRoot(version);
 }
 
-void Repository::putManifest() { putManifest(Json::nullValue); }
+bool Repository::putManifest() { return putManifest(Json::nullValue); }
 
-void Repository::putManifest(const Json::Value &custom) {
+bool Repository::putManifest(const Json::Value &custom) {
   Json::Value version_manifest;
   version_manifest["primary_ecu_serial"] = config.uptane.primary_ecu_serial;
   version_manifest["ecu_version_manifest"] = transport.getManifests();
@@ -52,9 +52,7 @@ void Repository::putManifest(const Json::Value &custom) {
   Json::Value tuf_signed =
       Crypto::signTuf(config.uptane.private_key_path, config.uptane.public_key_path, version_manifest);
   HttpResponse reponse = http.put(config.uptane.director_server + "/manifest", tuf_signed);
-  if (!reponse.isOk()) {
-    throw std::runtime_error("Could not put manifest");
-  }
+  return reponse.isOk();
 }
 
 void Repository::refresh() {
--- a/src/uptane/uptanerepository.cc
+++ b/src/uptane/uptanerepository.cc
@@ -51,7 +51,10 @@ void Repository::putManifest(const Json::Value &custom) {
   version_manifest["ecu_version_manifest"].append(ecu_version_signed);
   Json::Value tuf_signed =
       Crypto::signTuf(config.uptane.private_key_path, config.uptane.public_key_path, version_manifest);
-  http.put(config.uptane.director_server + "/manifest", tuf_signed);
+  HttpResponse reponse = http.put(config.uptane.director_server + "/manifest", tuf_signed);
+  if (!reponse.isOk()) {
+    throw std::runtime_error("Could not put manifest");
+  }
 }
 
 void Repository::refresh() {
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -27,7 +27,6 @@ add_custom_target(build_tests DEPENDS aktualizr)
 add_custom_target(check COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} -E test_valgrind_uptane_vectors\\|test_build DEPENDS build_tests)
 add_custom_target(check-full COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} -E test_uptane_vectors DEPENDS build_tests)
 
-
 # Setup coverage
 if(BUILD_WITH_CODE_COVERAGE)
     include(CodeCoverage)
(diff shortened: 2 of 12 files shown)


CMakeLists.txt
```diff
@@ -19,6 +19,10 @@ else(AKTUALIZR_VERSION)
     add_definitions(-DAKTUALIZR_VERSION="dev")
 endif(AKTUALIZR_VERSION)
 
+if(PROVISION_FILE)
+    add_definitions(-DPROVISION_FILE="${PROVISION_FILE}")
+endif(PROVISION_FILE)
+
```
cajun-rat: I'm not keen on compiling test code into the build output.  Could we instead pass this in to the test cases via the command line?

tests/CMakeLists.txt
```diff
@@ -23,9 +23,14 @@ if(BUILD_GENIVI)
     list(APPEND TEST_LIBS rvi)
 endif(BUILD_GENIVI)
 
+add_custom_target(prepare-ostree-sysroot
+    COMMAND if [ ! -d "sysroot" ]; then curl https://s3.eu-central-1.amazonaws.com/pw-dropbox/ostree-sysroot.tar.gz -o ostree-sysroot.tar.gz && tar -xf ostree-sysroot.tar.gz \; fi
+    DEPENDS aktualizr)
```
cajun-rat: It feels odd that we need a OSTree sysroot in order to test provisioning.  I know that the config validity checks expect to be able to find a ostree sysroot, but wouldn't a better solution be to move those checks to later in the process?  This would decouple the OSTree parts from the provisioning parts, rather than try meet the dependency?

tests/CMakeLists.txt
```diff
@@ -201,13 +214,20 @@ add_test(NAME test_build_all_off
 add_test(NAME test_build_all_on
         COMMAND ${PROJECT_SOURCE_DIR}/tests/build_with_options ${PROJECT_SOURCE_DIR} test_build_all_on "-DBUILD_WITH_CODE_COVERAGE=ON -DBUILD_GENIVI=ON -DBUILD_OSTREE=ON")
 
+add_test(NAME test_build_ci_tests_on
+        COMMAND ${PROJECT_SOURCE_DIR}/tests/build_with_options ${PROJECT_SOURCE_DIR} test_build_ci_tests_on "-DBUILD_WITH_CODE_COVERAGE=OFF -DBUILD_GENIVI=OFF -DBUILD_OSTREE=ON -DPROVISION_FILE=${PROJECT_BINARY_DIR}/test_data/credentials.zip")
+
```
cajun-rat: I wouldn't worry about this test case--if there _is_ a provisioning file, then we can use it, and if the _isn't_ a provisioning file we can't do this test any way :)

tests/config_test.cc
```diff
@@ -105,10 +105,12 @@ TEST(config, config_cmdl_parsing) {
 }
 
 TEST(config, config_extract_credentials) {
-  system("rm -rf tests/test_data/prov");
```
cajun-rat: system("rm -rf .. makes me nervous.  I'm happy to see this gone!

tests/uptane_ci_test.cc
```diff
@@ -0,0 +1,29 @@
...
+#include <fstream>
+#include <iostream>
+
+#include <logger.h>
+#include <boost/filesystem.hpp>
+
+#include "uptane/uptanerepository.h"
+
+TEST(SotaUptaneClientTest, UptaneSecondaryAdd) {
+  boost::filesystem::remove_all("/tmp/aktualizr");
```
cajun-rat: We really shouldn't be writing to /tmp. Can we setup the config file to store all temporary files underneath the current build directory?

CMakeLists.txt
cajun-rat: A small issue, but can we change this name to `SOTA_PACKED_CREDENTIALS`, please?  That is the name which we use elsewhere (e.g. https://docs.atsgarage.com/quickstarts/qemuvirtualbox.html), and reusing the same name will make things a little bit easier for our users.

src/uptane/uptanerepository.cc
```diff
@@ -51,7 +51,8 @@ void Repository::putManifest(const Json::Value &custom) {
   version_manifest["ecu_version_manifest"].append(ecu_version_signed);
   Json::Value tuf_signed =
       Crypto::signTuf(config.uptane.private_key_path, config.uptane.public_key_path, version_manifest);
-  http.put(config.uptane.director_server + "/manifest", tuf_signed);
+  HttpResponse reponse = http.put(config.uptane.director_server + "/manifest", tuf_signed);
+  return reponse.isOk();
```
cajun-rat: I like this!

tests/CMakeLists.txt
```diff
@@ -23,9 +23,14 @@ if(BUILD_GENIVI)
...
+add_custom_target(prepare-ostree-sysroot
+    COMMAND if [ ! -d "sysroot" ]; then curl https://s3.eu-central-1.amazonaws.com/pw-dropbox/ostree-sysroot.tar.gz -o ostree-sysroot.tar.gz && tar -xf ostree-sysroot.tar.gz \; fi
+    DEPENDS aktualizr)
 add_custom_target(build_tests DEPENDS aktualizr)
-add_custom_target(check COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} -E test_valgrind_uptane_vectors\\|test_build DEPENDS build_tests)
-add_custom_target(check-full COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} -E test_uptane_vectors DEPENDS build_tests)
+add_custom_target(build_tests_ci DEPENDS prepare-ostree-sysroot)
+add_custom_target(check COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} -E test_valgrind_uptane_vectors\\|test_build\\|test_uptane_ci DEPENDS build_tests)
+add_custom_target(check-full COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} -E test_uptane_vectors\\|test_uptane_ci DEPENDS build_tests)
+add_custom_target(check-ci COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} -R test_uptane_ci DEPENDS build_tests_ci)
```
cajun-rat: Would it be simpler to enable this test only when SOTA_PACKED_CREDENTIALS is set, and skip it otherwise?

cajun-rat:
I'm struggling to get this new test to pass, but because it is disabled by default (and there have been recent breaking changes on CI) I'll merge this now.

PR #303, opened 2017-08-04
Isolate filesystem writes to a separate class

OYTIS:
Secondary firmware and generated random device ids are still not covered.

Funny fact: transactional storage is not yet implemented, but it should be a bit easier now.


OYTIS:
Formatting check fails even with clang-format v.3.8.0

OYTIS:
Or maybe it's just me running 'make format-src' instead of 'make qa'

patrickvacek:
There's also 'make format-tests' to handle the tests. I can make a single target that does both without requiring all of 'make qa' if that'd be handy.

OYTIS:
* Share storage with boost::shared_ptr
* Use temporary files in httpclient for compatibility with gnuTLS
* Clean commented out code

src/fsstorage.cc
```diff
@@ -0,0 +1,179 @@
...
+  Utils::writeFile((director_path / "root.json").native(), metadata.director_root.toJson());
+  Utils::writeFile((director_path / "targets.json").native(), metadata.director_targets.toJson());
+  Utils::writeFile((image_path / "root.json").native(), metadata.image_root.toJson());
+  Utils::writeFile((image_path / "targets.json").native(), metadata.image_targets.toJson());
+  Utils::writeFile((image_path / "timestamp.json").native(), metadata.image_timestamp.toJson());
+  Utils::writeFile((image_path / "snapshot.json").native(), metadata.image_snapshot.toJson());
+  sync();
+}
+
+bool FSStorage::loadMetadata(Uptane::MetaPack* metadata) {
```
patriotyk: I think loadMetadata and storeMetadata should be memebres of Uptane class. 
"repo"  and "director" might have another names and only Uptane knows them

src/httpclient.cc
```diff
@@ -85,12 +94,97 @@ HttpResponse HttpClient::get(const std::string& url) {
...
+
+  tls_ca = ca;
+  tls_cert = cert;
+  tls_pkey = pkey;
+  curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, true);
+  curl_easy_setopt(curl, CURLOPT_SSL_CTX_FUNCTION, sslCtxFunction);
+  curl_easy_setopt(curl, CURLOPT_SSL_CTX_DATA, this);
+}
+
+/*void HttpClient::setCerts(const std::string& ca, const std::string& cert, const std::string& pkey) {
```
patriotyk: maybe remove this commented method?

src/crypto.cc
```diff
@@ -136,8 +147,8 @@ bool Crypto::VerifySignature(const PublicKey &public_key, const std::string &sig
   }
 }
 
-bool Crypto::parseP12(FILE *p12_fp, const std::string &p12_password, const std::string &pkey_pem,
-                      const std::string &client_pem, const std::string ca_pem) {
+bool Crypto::parseP12(FILE *p12_fp, const std::string &p12_password, std::string *out_pkey, std::string *out_cert,
```
cajun-rat: This is a Google Style Guide thing https://google.github.io/styleguide/cppguide.html#Reference_Arguments

CMakeLists.txt
```diff
@@ -79,7 +79,7 @@ if (CMAKE_COMPILER_IS_GNUCXX)
     # Enable maximum set of warnings. -Wno-sign-compare is required because of
     # problems in gtest. -Wswitch-default would be nice as well, but also causes
     # problems in gtest.
-    add_definitions(-Wall -Wextra -Wformat-security -Wfloat-equal -Wcast-qual -Wconversion -Wlogical-op -Wno-sign-compare)
+    add_definitions(-Wall -Wextra -Wformat-security -Wfloat-equal -Wcast-qual -Wconversion -Wlogical-op -Wno-sign-compare -Wno-error=conversion)
```
patrickvacek: I would actually just remove this and the -Wconversion in the same line, and maybe add to the comment right above it that -Wconversion would be nice if it didn't cause problems with gtest. I would maybe also add -Wconversion to the PEDANTIC_WARNINGS section just a few lines below.

The combination "-Werror -Wconversion -Wno-error=conversion" is perhaps undesirable since it will still show warning messages for conversion problems but won't break.

For the record, there doesn't appear to be much we can do about this type of thing at present, unless there is an easy way in cmake to manually remove flags just for a specific sub-project/subdirectory.

src/utils.cc
```diff
@@ -236,6 +236,20 @@ std::string Utils::randomUuid() {
   return boost::uuids::to_string(uuid_gen());
 }
 
+void Utils::copyDir(const boost::filesystem::path &from, const boost::filesystem::path &to) {
```
patriotyk: maybe boost::filesystem::copy do the same work?

src/fsstorage.cc
OYTIS: Unless UPTANE spec changes dramatically it shouldn't happen, and we'll have to rewrite a lot in this case anyway. The idea of the "storage" is that it lies between the app and actual storage mechanism, like file system or DB or whatever. So it is OK for it to know the application details, IMO.

src/httpclient.cc
OYTIS: Yes, there is some commented code left, I'll make a cleanup.

src/utils.cc
OYTIS: No, boost::filesystem::copy can't recursively copy directories, as far as I know.
http://www.boost.org/doc/libs/1_53_0/libs/filesystem/doc/reference.html#copy

OYTIS:
* delete storeBootstrapTlsCredentials/loadBootstrapTlsCredentials from INvStorage and FSStorage

src/bootstrap.cc
```diff
@@ -48,65 +46,34 @@ Bootstrap::Bootstrap(const std::string& provision_path_in)
...
-  }
-}
-
-std::string Bootstrap::getP12Str() const { return p12_str; }
-
-std::string Bootstrap::getPkeyPath() const { return pkey_path; }
-
-std::string Bootstrap::getCertPath() const { return cert_path; }
-
-std::string Bootstrap::getCaPath() const { return ca_path; }
```
patrickvacek: Is there a renewed need to keep the bootstrap provisioning files around? This basically reverts my work to keep all of that internal to the archive file and only create the files temporarily for the sole sake of the curl SSL call.

OYTIS: @patrickvacek Not sure if I understand you correctly.
We are still using temporary files for bootstrap credentials, their creation and deletion just moved from Bootstrap to HttpClient.

OYTIS:
Tested on RPi. Updates occur, device information is sent. The bug with ever-growing kernel command line needs to be fixed, but that is for another PR.

PR #304, opened 2017-08-04
Bugfix/pro 3484/avoid build unzip

patrickvacek:


patrickvacek:
Still dealing with a provisioning bug here, but getting much closer.

tests/config_tests_prov_bad.toml
```diff
@@ -1,35 +0,0 @@
-[gateway]
```
patriotyk: Why You removed this file?

patrickvacek: It doesn't appear to be used anywhere. Am I wrong? If we are using it, I of course have no problem keeping it!

patrickvacek:
The provisioning bug is just an issue of not having 'certificates_directory' in the [tls] section, which is not part of aktualizr. So this pull request is finally ready for review.

cajun-rat:
Looks great.  Note that this requires matching changes to meta-updater

PR #305, opened 2017-08-04
Improve secondaries, put right manifests after install

patriotyk:


cajun-rat:
@patriotyk This is failing during the CI build, could you take a look please?

cajun-rat:
(Rebased)

src/uptane/uptanerepository.cc
```diff
@@ -90,9 +92,13 @@ std::pair<uint32_t, std::vector<Uptane::Target> > Repository::getTargets() {
         throw MissMatchTarget("director and image");
```
cajun-rat: I'm struggling to understand why it is necessary to split the targets between the primary and secondary(s) here?

The split is only based on ecu_identifier, so would it make more sense to split them into two groups in SotaUptaneClient::runForever instead?

patriotyk: We cannot download target file from director target, only from image.

patriotyk:
I have updated

OYTIS:
Tested on Acer OTA stick.

PR #308, opened 2017-08-09
Use kernel arguments from the bootconfig, not the whole /proc/cmdline

OYTIS:
This resulted in unbootable device when building AGL for RPi. And indefinitely growth of kernel command line is not what we'd like to see anyway.


guyshapiro:
If I understand correctly, OSTree will take the arguments from the bootconfig if null is passed as the kernel args parameter.
https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-sysroot-deploy.c#L2007

PR #311, opened 2017-08-10
Check if aktualizr generates random ecu_serials

patriotyk:
Is the number 155 in  "\verify{\tst{155}}" comment  correct?

### DIFF
--- a/tests/config_test.cc
+++ b/tests/config_test.cc
@@ -126,6 +126,29 @@ TEST(config, config_extract_credentials) {
             "D27E3E56BEF02AAA6D6FFEFDA5357458C477A8E891C5EADF4F04CE67BB5866A4");
 }
 
+/**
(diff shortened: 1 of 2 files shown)


src/config.h
```diff
@@ -83,7 +83,7 @@ struct RviConfig {
 
 struct TlsConfig {
   TlsConfig()
-      : certificates_directory("/tmp/aktualizr"),
+      : certificates_directory("./sota_prov"),
```
patrickvacek: Is this a change we want to make? I think there is some sense to keeping the default as an absolute path.

patrickvacek:
@patriotyk Yes, 155 looks right to me. I think the format of that string isn't quite formalized yet, but that's what I'm going with for now.

src/config.h
patriotyk: Currently we don't need this change, reverting.

PR #312, opened 2017-08-10
Report installed packages through uptane_repo.http

OYTIS:


OYTIS:
And this one is really reproducable only in Travis environment. Happy debugging.

PR #315, opened 2017-08-11
Specify ports to allow parallel uptane vector test runs.

patrickvacek:
This was only a problem for me when running locally with
CTEST_PARALLEL_LEVEL > 1, but it was annoying and made me concerned that
the problem could crop up in ugly ways in other places.

### DIFF
--- a/tests/run_vector_tests.sh
+++ b/tests/run_vector_tests.sh
@@ -9,14 +9,18 @@ fi
 
 pip install -r "$1/requirements.txt"
 $1/generator.py -t uptane --signature-encoding base64 -o vectors --cjson json-subset
-$1/server.py -t uptane --signature-encoding base64 &
+if [ "$2" == "valgrind" ]; then
+  $1/server.py -t uptane --signature-encoding base64 -P 8081 &
+else
+  $1/server.py -t uptane --signature-encoding base64 -P 8080 &
+fi
 sleep 3
 trap 'kill %1' EXIT
 
 if [ "$2" == "valgrind" ]; then
-    valgrind --track-origins=yes --show-possibly-lost=no --error-exitcode=1 --suppressions=$1/../aktualizr.supp ./aktualizr_uptane_vector_tests vectors/vector-meta.json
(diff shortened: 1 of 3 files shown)


tests/run_vector_tests.sh
```diff
@@ -9,18 +9,22 @@ fi
 
 pip install -r "$1/requirements.txt"
 $1/generator.py -t uptane --signature-encoding base64 -o vectors --cjson json-subset
-$1/server.py -t uptane --signature-encoding base64 &
+if [ "$2" == "valgrind" ]; then
+  $1/server.py -t uptane --signature-encoding base64 -P 8081 &
```
patriotyk: httpclient_test.cc use bind(0) approach to get free port on the system. I suggest to create test_utils.h and move this implementation to this file. Then we could use this implementation in both tests.

patrickvacek: @patriotyk Not a bad idea, but that gets a bit challenging in this case. We need to generate the port in this script so that we can pass it to both the python server and the C++ test case.

OYTIS: @patrickvacek 
There is some probablity that `-P 0` will do the trick.

patrickvacek: @OYTIS There'd still be the problem that only the server script would know about the port. I'd have to communicate to the C++ program as well. I'll look into a way to find an open port natively in python so I can call it directly in run_vector_tests and then pass it to both programs.

patrickvacek:
@patriotyk @OYTIS Now the bash script gets an OS-assigned open port via a python script and passes that to the individual programs. Look good to you?

patriotyk:
@patrickvacek  Good, but I forget about TIME_WAIT state issue. So this could not work everywhere because when You close the socket it switches to the state TIME_WAIT, and during this state, nobody can bind this socket if don't specify special option SO_REUSEADDR. Does it works on Your local machine? Could You check the content of /proc/sys/net/ipv4/tcp_fin_timeout  file?

patrickvacek:
@patriotyk It's 60. But since I don't actually listen on the socket, do I have to worry about that? It worked for me locally, with Docker, and in CI.

patriotyk:
@patrickvacek  You are right, just have checked, get_open_port.py didn't switch port to the TIME_WAIT state

PR #316, opened 2017-08-14
Check if aktualizr can provision even of previous network issues

patriotyk:


patrickvacek:
This looks great so far. However, are there still more cases to be tested? Quoting from the test requirements:

    Repeat those steps with different network issues(timeout, 503 and 408 errors, connect to non binded port, close connection after request get, close connection after partially response sent )

All drops should be done for the following connections:

    Connection drops in between of downloading device.p12 (device registration), download interrupted.
    Connection drops after device registration, can't access director to register ECUs.
    Connection drops in between of ECU registration, POST interrupted.

tests/uptane_fake_http_test.cc
```diff
@@ -0,0 +1,99 @@
...
+#include <string>
+
+#include "fsstorage.h"
+#include "logger.h"
+#include "test_utils.h"
+#include "uptane/uptanerepository.h"
+#include "utils.h"
+
+void doDeviceRegister(const std::string &device_id, Config &conf, bool expect) {
+  system((std::string("rm -rf ") + conf.tls.certificates_directory.string() + "/*").c_str());
```
patrickvacek: Any particular reason not to use boost::filesystem::remove_all()? The directory in question is specified in the toml file as tests/test_uptane/, which is intended to be a purely temporary directory for the duration of the test.

tests/uptane_fake_http_test.cc
```diff
@@ -0,0 +1,99 @@
...
+  conf.tls.server = "http://127.0.0.1:" + port;
+  doDeviceRegister("noerrors", conf, true);
+
+  doEcuRegister("drop_request", conf, false);
+  doEcuRegister("noerrors", conf, true);
+
+  doEcuRegister("status_503", conf, false);
+  doEcuRegister("noerrors", conf, true);
+
+  doEcuRegister("status_503", conf, false);
```
patrickvacek: This is presumably a typo and should be status_408, right?

tests/fake_http_server/fake_uptane_server.py
```diff
@@ -0,0 +1,61 @@
...
+		elif data["deviceId"] == "noerrors":
+			self.send_response(200)
+			self.end_headers()
+			f = open('tests/test_data/cred.p12', 'r')
+			self.wfile.write(f.read())
+
+	def do_ecuRegister(self, data):
+		if data["primary_ecu_serial"] == "drop_request":
+			return
+		elif data["primary_ecu_serial"].startswith("status_"):
```
patrickvacek: I know this is getting tedious, but could we also have a drop_body case for the ECU registration?

OYTIS:
@patriotyk 
I can see two major problems here:
1) We create new UPTANE repositories and clear tls.certificates directory on every call to deviceRegister(). That basically means that we are not testing if we're able to restart the process. Creating new Uptane::Repository is probably fine, that roughly corresponds to restarting the aktualizr, but all the temporary data should be in place. The easiest way to ensure it is probably to use the same FSStorage instance throuhout the whole test.

2) We're using deviceId to change server's behaviour. One of the problems with the registration process is that we can't register the same deviceId twice, so the client should take care of it. So the fake server should keep track of all the registered device IDs and ECU serials and give error (as far as I remember it's 50x) on duplicates. It can also implement a dedicated end-point (e.g. '/injection') to change it's behaviour.

It the test is implemented correctly it should fail, at least until https://github.com/advancedtelematic/aktualizr/pull/318 is merged.

tests/uptane_fake_http_test.cc
patriotyk: Yes You are right

tests/fake_http_server/fake_uptane_server.py
patriotyk: We did not check the body response, so ecuRegistration will pass

cajun-rat:
@patriotyk is working on this at the moment

PR #318, opened 2017-08-15
A linear restartable initialization process

OYTIS:
Secondaries have become even a bit more messy, but they need a major refactoring. That's probably the next thing we should do.
Changes:
- deviceRegister() and ecuRegister() are merged into a single method called initialize()
- serial number and device ID generation (for both primaries and secondaries) is now a part of initialization
- Added clearXXX() functions to the storage to support repeated initialization.
- The only Crypto function still working with files is parseP12. It can be changed later for consistency (or when we decide to store bootstrap credentials in some database as well), but no urgent need yet.

### DIFF
--- a/tests/uptane_test.cc
+++ b/tests/uptane_test.cc
@@ -398,7 +398,6 @@ TEST(uptane, pet_name_provided) {
   Utils::copyDir("tests/test_data", uptane_test_dir);
   std::string test_name = "test-name-123";
   std::string device_path = uptane_test_dir + "/device_id";
-  boost::filesystem::remove(device_path);
 
   /* Make sure provided device ID is read as expected. */
   Config conf("tests/config_tests_device_id.toml");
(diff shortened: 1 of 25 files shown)


tests/uptane_test.cc
```diff
@@ -312,35 +335,213 @@ TEST(SotaUptaneClientTest, device_register_twice) {
...
+  conf.uptane.primary_ecu_serial = "testecuserial";
+  conf.uptane.private_key_path = "private.key";
+  conf.uptane.public_key_path = "public.key";
 
   boost::filesystem::remove(conf.tls.certificates_directory / conf.tls.client_certificate);
   boost::filesystem::remove(conf.tls.certificates_directory / conf.tls.ca_file);
+  boost::filesystem::remove(conf.tls.certificates_directory / conf.tls.pkey_file);
+  boost::filesystem::remove(conf.tls.certificates_directory / "bootstrap_ca.pem");
+  boost::filesystem::remove(conf.tls.certificates_directory / "bootstrap_cert.pem");
+  boost::filesystem::remove(conf.tls.certificates_directory / "bootstrap_pkey.pem");
```
patrickvacek: No big deal, but these bootstrap files aren't actually created anymore, so they don't need to be removed, right?

PR #320, opened 2017-08-17
Test/tst 149/check basic data

patriotyk:


tests/uptane_test.cc
```diff
@@ -153,9 +153,9 @@ TEST(uptane, sign) {
 }
 
 /*
- * \verify{\tst{153}} Check that aktualizr creates provisioning files if they
```
patrickvacek: Not a big deal, but could you get undo these whitespace changes? If they are caused by clang-format, let me know, but they seem undesirable.

patriotyk: Yes they are caused by clang-format

patrickvacek: Grrr. That's annoying (but presumably not your fault). What version of clang-format do you have? Our recommended version at present is 3.8.

patriotyk: clang-format version 3.8.1-12ubuntu1 (tags/RELEASE_381/final)

patrickvacek: I have clang-format version 3.8.1-18ubuntu1 (tags/RELEASE_381/final) and either way I format block comments (with or without the space after the first line), my clang-format-3.8 doesn't mess with it. I'm confused why yours does, but it's not worth losing sleep over.

tests/uptane_test.cc
```diff
@@ -667,6 +667,34 @@ TEST(SotaUptaneClientTest, UptaneSecondaryAdd) {
   boost::filesystem::remove_all(uptane_test_dir);
 }
 
+/**
+  * \verify{\tst{149}} Check that basic device info sent by aktualizr on provisioning are on server
+*/
+TEST(SotaUptaneClientTest, test149) {
```
OYTIS: A minor issue, but maybe give some meaningful name for consistency.

src/fsstorage.cc
```diff
@@ -13,35 +13,32 @@ FSStorage::FSStorage(const Config& config) : config_(config) {
 }
 FSStorage::~FSStorage() {
   // TODO: clear director_files, image_files
+  clearPrimaryKeys();
 }
 
 void FSStorage::storePrimaryKeys(const std::string& public_key, const std::string& private_key) {
-  boost::filesystem::path public_key_path = config_.tls.certificates_directory / config_.uptane.public_key_path;
-  boost::filesystem::path private_key_path = config_.tls.certificates_directory / config_.uptane.private_key_path;
+  boost::filesystem::remove(config_.uptane.public_key_path);
```
OYTIS: Regarding the names see the discussion in https://github.com/advancedtelematic/aktualizr/pull/324

patriotyk: Yes, I will rebase and remove all path changes after #324 merge

tests/uptane_test.cc
```diff
@@ -755,6 +755,34 @@ TEST(SotaUptaneClientTest, UptaneSecondaryAdd) {
   boost::filesystem::remove_all(uptane_test_dir);
 }
 
+/**
+  * \verify{\tst{149}} Check that basic device info sent by aktualizr on provisioning are on server
+*/
+TEST(SotaUptaneClientTest, provision_on_server) {
+  Config config("tests/config_tests_prov.toml");
+  std::string server = "tst149";
+  Utils::copyDir("tests/test_data", uptane_test_dir);
```
patrickvacek: This copyDir is unnecessary.

tests/uptane_test.cc
```diff
@@ -755,6 +755,34 @@ TEST(SotaUptaneClientTest, UptaneSecondaryAdd) {
...
+  event::Channel events_channel;
+  command::Channel commands_channel;
+  FSStorage storage(config);
+  HttpFake http(uptane_test_dir);
+  commands_channel << boost::make_shared<command::GetUpdateRequests>();
+  commands_channel << boost::make_shared<command::Shutdown>();
+  Uptane::Repository repo(config, storage, http);
+  SotaUptaneClient up(config, &events_channel, repo);
+  up.runForever(&commands_channel);
+}
```
patrickvacek: It would be best to have this at the end of the test:
boost::filesystem::remove_all(uptane_test_dir);

patrickvacek:
Really close! I'm surprised github isn't requiring you to rebase again, but apparently you don't have to at the moment. I still haven't figured out why your clang-format is changing the block comment spacing. I'd prefer if those whitespace changes weren't checked in, but since we haven't finalized the format for that anyway, it doesn't really matter for now.

tests/uptane_test.cc
patriotyk: I added it because I got an error that boost cant copy cred.zip file because it is already exists. I think some previous tests did not  clean that directory

PR #322, opened 2017-08-17
Test metadata on bad threshold

patriotyk:


patrickvacek:
@OYTIS isn't @heartsucker traveling right now? Do you have strong feelings of your own? Otherwise this looks good to me. I'm also okay with merging this now and moving it later if we decide it would be better to do so after all.

cajun-rat:
Let's talk to @heartsucker when he is back about how to link TST tickets to tests that are generated by uptane-vector-tests.  In the mean time I think it is easier to merge this, even if we have to move it out later.

PR #323, opened 2017-08-18
Test expired metadata

patriotyk:


tests/uptane_test.cc
```diff
@@ -422,6 +422,59 @@ TEST(uptane, pet_name_creation) {
...
+  config.uptane.repo_server = tls_server + "/repo";
+
+  FSStorage storage(config);
+  HttpFake http(uptane_test_dir);
+  Uptane::TufRepository repo("director", tls_server + "/director", config, storage, http);
+  try {
+    // Check that we don't fail on good metadata.
+    repo.verifyRole(Uptane::Role::Targets(), now, Utils::parseJSONFile("tests/test_data/targets_noupdates.json"));
+  } catch (...) {
+    FAIL();
```
cajun-rat: Same comment as for #322 - could you use EXPECT_NO_THROW and EXPECT_THROW here?  It is a bit more explicit, and also will check that the type of the exception is correct

PR #324, opened 2017-08-18
Fix bugs in initialization process

OYTIS:


src/config.cc
```diff
@@ -151,9 +151,6 @@ void Config::postUpdateValues() {
   if (uptane.director_server.empty()) uptane.director_server = tls.server + "/director";
 
   if (uptane.ostree_server.empty()) uptane.ostree_server = tls.server + "/treehub";
-
-  uptane.public_key_path = (tls.certificates_directory / uptane.public_key_path).string();
-  uptane.private_key_path = (tls.certificates_directory / uptane.private_key_path).string();
```
patriotyk: Why You removed this? I fixed this issue in https://github.com/advancedtelematic/aktualizr/pull/320 in another way. Instead of removing that lines I replaced all places with concatenation.

patrickvacek: My preference is this method, because it's inconsistent with how all other paths are calculated. The keys were the only ones with the certificates directory prepended as part of initialization.

src/uptane/initialize.cc
```diff
@@ -218,8 +236,6 @@ InitRetCode Repository::initEcuRegister() {
 
 // Postcondition: "ECUs registered" flag set in the storage
 bool Repository::initialize() {
-  if (storage.loadEcuRegistered()) return true;
-
   // if device has aready been registered, just create new device and reprovision
   //   it should be changed/fixed when the backend supports registration acknowledgement
```
patrickvacek: What do you mean in this comment by "create new device and reprovision"? Is that accurate? It looks to me like if everything is already initialized, this function will just reload everything, but it won't change the device ID or keys.

patriotyk: Do You ask me? I don't know, this is not my comments

src/config.cc
patriotyk: Maybe it would be better to do the same concatenation for others? The code will be nicer.

src/uptane/initialize.cc
patrickvacek: Sorry, I meant the question for @OYTIS.

OYTIS: @patrickvacek 
This comment really belongs to line 239 and propably needs to be rephased. I meant that if device with the same ID has been already registered to the server, aktualizr should generate a new ID and repeat registration. That is what this code hopefully does.

patrickvacek: Ah, that makes more sense. I agree, that is what the logic should do.

src/config.cc
OYTIS: In my opinion modification of config variables makes it a bit harder to trace the state of the system while reading the code. And all these pathes/file names are only used (or at least they should only be used) in fsstorage.cc and the code there is quite compact, so I don't think longer expressions is a big issue here.

OYTIS: I also find it pleasant that postUpdateValues which was previously doing a lot of things is now just performs a simple task of reading the server URL from an archive and setting a few variables accordingly.

patrickvacek:
@OYTIS After the rebase this should be good to go.

OYTIS:
I'll make an E2E test again and remove 'failing CI' once done.

OYTIS:
Great, a real heisenbug.

tests/build_with_options.sh
```diff
@@ -0,0 +1,11 @@
...
+
+echo $PWD
+# Protect against mistakes by not just blindly deleting the second input
+# argument.
+rm -Rf "${2}_prev"
+mv $2 "${2}_prev"
+mkdir -p "$2"
+cd "$2"
+cmake $3 $1
+make -j5 aktualizr
```
cajun-rat: Why is this needed?  There shouldn't be any case where a clean build gives different results to an incremental build, and if there is, we should fix _that_ bug.  I'm worried that starting from scratch here will slow down the development build process.

OYTIS:
OK, so this bug was partly my hallucination and partly bug on the director side (@danten) is aware about it. Should be OK to merge.

src/sotauptaneclient.cc
```diff
@@ -21,8 +21,8 @@ void SotaUptaneClient::run(command::Channel *commands_channel) {
 }
 
 bool SotaUptaneClient::isInstalled(const Uptane::Target &target) {
-  if (target.ecu_identifier() == config.uptane.primary_ecu_serial) {
-    return target.filename() == OstreePackage::getEcu(config.uptane.primary_ecu_serial, config.ostree.sysroot).ref_name;
+  if (target.ecu_identifier() == uptane_repo.primary_ecu_serial) {
+    return target.filename() == OstreePackage::getEcu(uptane_repo.primary_ecu_serial, config.ostree.sysroot).ref_name;
```
cajun-rat: primary_ecu_serial is private in Uptane::Repository, and we can only do this because SotaUpdateClient is a friend of Uptane::Respoitory.  Can we stop them getting too entwined?  We are quite close now to being able to stop them being 'friends', so perhaps this change belongs in a separate commit, where we can figure out how to signing can work

patrickvacek:
Works for me.

src/uptane/initialize.cc
```diff
@@ -114,6 +128,8 @@ InitRetCode Repository::initTlsCreds(const ProvisionConfig& provision_config) {
     LOGGER_LOG(LVL_trace, "Device already registered, proceeding");
     // set provisioned credentials
     http.setCerts(ca, cert, pkey);
+    director.setTlsCreds(ca, cert, pkey);
+    image.setTlsCreds(ca, cert, pkey);
```
cajun-rat: Is this right?  I think we now only construct a single implementation of HttpInterface and pass down from Aktualizr::run().  The TufRepository::setTlsCreds() member function should probably go too, because TufRepository has no business messing with something that it has been given :)

cajun-rat:
Modulo a couple of nits, this looks great

src/sotauptaneclient.cc
OYTIS: There is a straightforward solution of having a getter for primary_ecu_serial, but this will not make it any less tangled. The problem is that we need primary_ecu_serial in both SotaUptaneClient (for finding targets that should be installed on primary) and in Uptane::Repository (because primary ecu is a part of UPTANE specification and should be managed accordingly). It can also be generated dynamically which is done in the initialization procedure, which in turn I had to place in Uptane::Repository, because it uses http connection and secondaries. So a "true" solution could be to move the main initialization sequence to SotaUptaneClient, while keeping only certain subfunctions in Uptane::Repository. We could then generate primary_ecu_serial in the client and just pass it as a constructor parameter to Uptane::Repository.

cajun-rat: Sounds like the real fix is harder than I'd guessed/hoped for :)  Just adding a getter isn't really going to help us get toward the goal, so I agree that fiddling with friend's privates here is OK.

PR #326, opened 2017-08-21
Return error to director on malformed refname

OYTIS:
This PR _might_ fix [the problem with garbage package names](https://advancedtelematic.atlassian.net/browse/PRO-3825). I can't say if it really solves it because login to CI garage is currently broken, but at least it doesn't break existing functionality and makes code slightly cleaner: crashing with an exception on wrong Uptane metadata was not a good idea (yes, it was my code).

### DIFF
--- a/src/ostree.cc
+++ b/src/ostree.cc
@@ -89,13 +89,14 @@ bool Ostree::addRemote(OstreeRepo *repo, const std::string &remote, const std::s
 #include "ostree-1/ostree.h"
 
 OstreePackage::OstreePackage(const std::string &ecu_serial_in, const std::string &ref_name_in,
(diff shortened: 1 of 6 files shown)


cajun-rat:
I tried a manual test of this on meta-updater morty + qemu, and it didn't update.  I'm not currently sure where the bug is

OYTIS:
@cajun-rat 
Works on Acer stick. I'll take a look into qemu.

OYTIS:
@cajun-rat 
Works for me with qemux86-64 as well. Might be some other transient client or backend bug.

cajun-rat:
Rebased

cajun-rat:
Tested with a manual update against QA

PR #328, opened 2017-08-22
Test/159/provision keys

patriotyk:


patriotyk:
1) You must provide credentals.zip in SOTA_PACKED_CREDENTIALS variable to run it successfully
2) Done, now it pass every time.
3) Yes 
4) Done
5) Done

patrickvacek:
I'm still having a few problems here.
1. I'm still not getting this to run successfully locally. Both tests fail with "C++ exception with description "tests/config_tests.toml: cannot open file" thrown in the test body."
2/3. Should we perhaps instead automate providing credentials to this test so that it can be run in Docker/CI, and to avoid having to re-run CMake just for this test? If it doesn't matter what credentials are used, as long as they are valid, we can just use one of the zip files in tests/test_data.
4. Looks good.
5. The first test is still using tests/test_config. Maybe it would be better if both tests used something like tests/test_uptane_ci to match the convention of the other tests with their own temporary test directories?

patriotyk:
1) How do You run this test?
3) Fixed

patrickvacek:
Sorry, numbering got messed up in my last comment. Anyway, the test directory syntax looks good. Thanks!

As far as running the test, I've been running the following commands from my build directory:
```
cmake -DCMAKE_BUILD_TYPE=Debug -DSOTA_PACKED_CREDENTIALS=/home/patrick/Code/pro/client/aktualizr/tests/test_data/credentials.zip ../
make -j8 format all build_tests test CTEST_PARALLEL_LEVEL=8
```
I've also tried running it directly with ctest from inside the build directory:
```ctest -V -R test_uptane_ci```
And with gdb in the root directory:
```gdb --args /home/patrick/Code/pro/client/aktualizr/build/tests/t_uptane_ci /home/patrick/Code/pro/client/aktualizr/tests/test_data/credentials.zip```

patriotyk:
Yes You are right, ctest ran this test from wrong directory. Now it fixed

patrickvacek:
Looking better, and the second test works for me now, but the first is failing with a different error. The repo.putManifest() call is failing with "curl error 0: Peer certificate cannot be authenticated with given CA certificates". I can't quite figure out what the problem is. (Side note, though: any reason not to test the return value of that call?)

After some discussion with Phil, I now understand why we don't need/want this running in Docker/CI, so you can safely ignore my previous comments about that.

OYTIS:
@patrickvacek @patriotyk 
I also get (somewhat different) SSL errors on OneCycleUpdate.

patriotyk:
Fixed. Now it works )))

PR #330, opened 2017-08-22
Create getPrimaryEcuVersion to avoid accessing private members.

patrickvacek:


src/uptane/uptanerepository.cc
```diff
@@ -15,6 +15,7 @@
 #include "invstorage.h"
 #include "logger.h"
 #include "openssl_compat.h"
+#include "ostree.h"
```
patriotyk: One of my previous tasks, was to remove ostree dependency from uptane implementation, but this change moves it back

cajun-rat:
That is a good point.  Let's talk with @koshelev tomorrow about what a good solution would look like.

cajun-rat:
For now we will leave this as a branch for @koshelev to test against, closing this PR

PR #336, opened 2017-08-29
Feat/pro 3891/cert provider

patrickvacek:


src/cert_provider/main.cc
```diff
@@ -0,0 +1,154 @@
...
+
+/*****************************************************************************/
+int main(int argc, char *argv[]) {
+  loggerInit();
+  loggerSetSeverity(static_cast<LoggerLevels>(2));
+
+  boost::filesystem::remove("device_id");
+  boost::filesystem::remove("root.crt");
+  boost::filesystem::remove("client.pem");
+  boost::filesystem::remove("pkey.pem");
```
cajun-rat: What about using `TemporaryFile` from utils.h here?

src/cert_provider/main.cc
```diff
@@ -0,0 +1,154 @@
...
+  if (!Crypto::parseP12(device_p12, "", &pkey, &cert, &ca)) {
+    return -1;
+  }
+  fclose(device_p12);
+
+  Utils::writeFile("device_id", device_id);
+  Utils::writeFile("root.crt", ca);
+  Utils::writeFile("client.pem", cert);
+  Utils::writeFile("pkey.pem", pkey);
+  sync();
```
cajun-rat: Is this necessary?

src/cert_provider/main.cc
patrickvacek: @cajun-rat I thought about that. I was trying to allow for a use case in which the user wanted to manually do something with the certificate files instead of having them automatically `scp`ed. If that use case is unlikely or undesirable, then yes, `TemporaryFile` would be wise. Is that your preference?

src/config.cc
```diff
@@ -85,52 +85,60 @@ Config::Config(const boost::property_tree::ptree& pt) {
...
-          if (r != ARCHIVE_OK) {
-            LOGGER_LOG(LVL_error, "Error closing provision archive: " << provision.provision_path);
-          }
-          if (found) {
-            tls.server = url_stream.str();
-          } else {
-            LOGGER_LOG(LVL_error, "autoprov.url not found in provision archive: " << provision.provision_path);
-          }
-        }
+        tls.server = readServerUrl(provision.provision_path);
```
cajun-rat: Much better :)

src/cert_provider/main.cc
patrickvacek: Unlikely, but was trying to be careful, and that's what our code is doing currently in `fsstorage`, which is where I stole that logic from.

src/config.h
```diff
@@ -147,6 +147,7 @@ class Config {
   Config(const std::string& filename);
 
   void updateFromTomlString(const std::string& contents);
+  static std::string readServerUrl(const std::string& provision_path);
```
cajun-rat: Would this code be more suited to being in the `Bootstrap` class?  To my mind `Config` is mostly about holding a set of configuration information for the system, and `Bootstrap` already reads the zip file to extract the x509 certificates.

src/uptane/uptanerepository.h
```diff
@@ -33,7 +33,7 @@ class Repository {
   // void addSecondary(const std::string &ecu_serial, const std::string &hardware_identifier);
   Json::Value updateSecondaries(const std::vector<Uptane::Target> &secondary_targets);
   std::pair<int, std::vector<Uptane::Target> > getTargets();
-  std::string getPrimaryEcuSerial() { return primary_ecu_serial; };
+  std::string getPrimaryEcuSerial() const { return primary_ecu_serial; };
```
cajun-rat: +const happy

tests/CMakeLists.txt
```diff
@@ -259,7 +259,7 @@ set(ALL_TEST_SRCS
     uptane_fake_http_test.cc)
 
 add_custom_target(check-format-tests
-    COMMAND ../scripts/check-formatting.sh ${CLANG_FORMAT} ${ALL_TEST_SRCS}
+    COMMAND ${PROJECT_SOURCE_DIR}/scripts/check-formatting.sh ${CLANG_FORMAT} ${ALL_TEST_SRCS}
```
cajun-rat: yep :)

src/config.h
patrickvacek: Yeah, probably. At this point, that function isn't dependent on anything in `Config` at all (hence why I made it `static`).

src/uptane/uptanerepository.h
patrickvacek: I mean, there's no reason not to, right?

src/cert_provider/main.cc
cajun-rat: Manually copying the files to somewhere is a reasonable use case.  I'm not sure about creating (or deleting) files with magic names in the current working directory though.

Maybe add a second option to dump the files in a directory instead/as well as copying them via scp?

src/uptane/uptanerepository.h
cajun-rat: Absolutely. Adding missing const-ness is always a good thing

patrickvacek:
Now with TemporaryFiles, optional directory parameter, and readServerUrl moved to Bootstrap.

PR #337, opened 2017-08-29
Get rid of some /tmp files in tests.

patrickvacek:
This doesn't remove all /tmp usage, just what I could find and fix in a few minutes' time.


cajun-rat:
Nice!

PR #338, opened 2017-08-31
Install cert-provider to ease integration with Yocto

OYTIS:


patrickvacek:
Is this actually necessary if cert_provider is only intended to be ran on the host machine? If you just want to convenience of having it properly installed, that's just fine too.

OYTIS:
What gets where can be controlled with FILES variable in Yocto, but before it needs to be installed anyway.

patrickvacek:
I see. I think this sounds fine, then. Just to be clear, though, once we have the Yocto work done as far as getting the root CA and server URL onto the device, cert_provider will only need to copy the certificate and key to the device, and the Yocto build user will need to provide the root CA and server URL directly to Yocto.

PR #340, opened 2017-09-04
Add implicit provisioning flag (via enum).

patrickvacek:
Doesn't actually do much yet! (Doesn't really need to, either.) The files in the implicit subdirectory are just unpacked from tests/test_data/cred.p12 at the moment.


OYTIS:
Great. I was wondering if we want provisioning method to be set explicitly in the sota.toml, but that is the change we can introduce later anyway.

patrickvacek:
@OYTIS Agreed. For now, it is dependent on provision_path being set, but it would be trivial to make it something set independently.

PR #341, opened 2017-09-04
Extract device_id from the certificate

OYTIS:
As @christianhuggerATS pointed out, it's a pretty useless piece of code, we don't use device_id in the aktualizr. It might be convenient to have device_id accessible during the debugging though. If it needs to be merged is questionable.

### DIFF
--- a/src/uptane/initialize.cc
+++ b/src/uptane/initialize.cc
@@ -1,3 +1,5 @@
+#include <openssl/x509.h>
+#include <boost/scoped_array.hpp>
 #include <string>
 
 #include "bootstrap.h"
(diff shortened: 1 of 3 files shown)


patrickvacek:
I'm confused with what CI is doing. Looks like this branch doesn't have my implicit provisioning tests, and yet CI is failing on them. How is that even possible!?

OYTIS:
I think it merges the PR into target branch (master in this case) and runs tests for the resulting tree. I'll rebase and debug that.

patrickvacek:
Oh, of course. That makes more sense now. It should work after a rebase and clearing up the provision mode vs method duplication.

PR #349, opened 2017-09-13
Add 1 to device_id length to accommodate trailing null.

patrickvacek:
Fixing a bug caught by @rdanitz

### DIFF
--- a/src/uptane/initialize.cc
+++ b/src/uptane/initialize.cc
@@ -38,8 +38,8 @@ bool Repository::initDeviceId(const ProvisionConfig& provision_config, const Upt
         X509_free(x);
         return false;
       }
-      boost::scoped_array<char> buf(new char[len]);
-      X509_NAME_get_text_by_NID(X509_get_subject_name(x), NID_commonName, buf.get(), len);
+      boost::scoped_array<char> buf(new char[len + 1]);
+      X509_NAME_get_text_by_NID(X509_get_subject_name(x), NID_commonName, buf.get(), len + 1);
       device_id = std::string(buf.get());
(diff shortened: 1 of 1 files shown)


cajun-rat:
Ouch.  Is this bug anywhere else in the codebase?

patrickvacek:
I hope not. This was from the new code that reads the device_id from the common name of the certificate. To be fair, the documentation is rather vague about whether space for the trailing null is necessary.

src/uptane/initialize.cc
```diff
@@ -38,8 +38,8 @@ bool Repository::initDeviceId(const ProvisionConfig& provision_config, const Upt
         X509_free(x);
         return false;
       }
-      boost::scoped_array<char> buf(new char[len]);
-      X509_NAME_get_text_by_NID(X509_get_subject_name(x), NID_commonName, buf.get(), len);
+      boost::scoped_array<char> buf(new char[len + 1]);
+      X509_NAME_get_text_by_NID(X509_get_subject_name(x), NID_commonName, buf.get(), len + 1);
```
OYTIS: I'm not sure if we need 'len+1' here though. The documentation is a bit ambiguous on this account:

> At most len bytes will be written and the text written to buf will be null terminated

https://www.openssl.org/docs/man1.1.0/crypto/X509_NAME_get_text_by_NID.html

OYTIS: I mean the last argument to the second call to X509_NAME_get_text_by_NID

patrickvacek: I had the same concern, so I tested both ways. It is required.

PR #350, opened 2017-09-14
Feat/pro 4007/p11

patriotyk:


OYTIS:
In my opinion, it's important to have both private and public keys working before merging it. Otherwise it may happen that too much has to be rewritten.

OYTIS:
Is a subset of https://github.com/advancedtelematic/aktualizr/pull/351

PR #351, opened 2017-09-15
Feat/pro 4027/p11 tests

patriotyk:


patriotyk:
@OYTIS  I resolved the problem with softhsm ))). But I still get an error about wrong key

patriotyk:
I resolved the problem with key parsing. It is bug in OpenSC, after update to newer version I am able to put public key.  But the problem with different keys still present

OYTIS:
Merged into https://github.com/advancedtelematic/aktualizr/pull/353

PR #352, opened 2017-09-19
Utility to list Aktualizr's transitive dependencies

cajun-rat:


cajun-rat:
It looks like we need something more complex. Closing this PR for now

PR #353, opened 2017-09-19
Add support for PKCS#11 in TLS connection

OYTIS:
tls config and P11 config from @patriotyk's PR should probably be merged into a unified crypto config.
No unit tests for P11 case are present yet.
This PR depends on [another one](https://github.com/ostreedev/ostree/pull/1183) to OSTree.


OYTIS:
Added the dependencies for testing new functionality plus a dummy test, which like other provisioning tests doesn't really test much. We should include @cajun-rat's work with django server to really test that TLS communication happens.

OYTIS:
By the way, the old code extracting device_id from the client certificate is currently broken with pkcs11. It should either be deleted (we don't really need to know our device_id, although it is useful for debugging) or make it extract certificate from via pkcs11.

scripts/coverage.sh
```diff
@@ -1,8 +1,15 @@
...
+mkdir -p /var/lib/softhsm/tokens
+softhsm2-util --init-token --slot 0 --label "Virtual token" --pin 1234 --so-pin 1234
+softhsm2-util --import ./src/tests/test_data/implicit/pkey.pem --label "pkey" --id 02 --slot 0 --pin 1234
+openssl x509 -outform der -in ./src/tests/test_data/implicit/client.pem -out ./src/tests/test_data/implicit/client.der
+pkcs11-tool --module=/usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so --id 1 --write-object ./src/tests/test_data/implicit/client.der --type cert --login --pin 1234
+
 mkdir -p build-coverage
 cd build-coverage
-cmake -DBUILD_GENIVI=ON -DBUILD_OSTREE=ON -DBUILD_WITH_CODE_COVERAGE=ON ../src
+cmake -DBUILD_GENIVI=ON -DBUILD_OSTREE=ON -DBUILD_WITH_CODE_COVERAGE=ON -DTEST_PKCS11_ENGINE_PATH="/usr/lib/engines/engine_pkcs11.so" -DTEST_PKCS11_MODULE_PATH="/usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so" ../src
```
patriotyk: in newer versions(from yakkety) path "/usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so" has been changed to "/usr/lib/softhsm/libsofthsm2.so"

scripts/test.sh
```diff
@@ -1,8 +1,18 @@
 #! /bin/bash
 set -e
+
+# Virtual token is only needed for tests. As utilities' interface may vary with version, don't create token in different environments
```
patriotyk: Duplicated code(with coverage.sh) and finctionality with my PR #351

src/httpclient.cc
```diff
@@ -16,6 +18,12 @@
 #include "logger.h"
 #include "openssl_compat.h"
 
+#ifdef TEST_PKCS11_ENGINE_PATH
```
patriotyk: I would like to encapsulate all pkcs11 functionality to separate class, currently I created p11 class, but maybe it is slightly  wrong name.

scripts/coverage.sh
OYTIS: Yes, we should change it when we move to yakkety. Should work with current dockerfile though.

scripts/test.sh
OYTIS: Almost. It turned out that in order for TLS to work properly, keys should be imported with 'softhsm2-util --import'.

src/httpclient.cc
OYTIS: Maybe "crypto" is more appropriate place for everything related to OpenSSL?

OYTIS:
Added UPTANE signing with PKCS#11 (currently only for primaries and only for RSA keys) and respective tests (PRO-3803). Any suggestions about how to make the code better structured are most welcome.

src/cert_provider/main.cc
```diff
@@ -10,6 +10,7 @@
 #include "bootstrap.h"
 #include "crypto.h"
 #include "httpclient.h"
+#include "httpclient.h"
```
patrickvacek: Duplicated line here.

patrickvacek:
Should we allow compiling without the p11 libraries if the user doesn't want to use it?

OYTIS:
1. On one hand it's a reasonable thing to do, since it's a relatively rare scenario. On the other it will probably add even more complexity with conditional compilation. I'm not sure yet.
2. Storing all the TLS credentials on a HSM looks to be a quite common case. It's even supported in Firefox.

OYTIS:
BTW, here's the configuration I've used for manual end-to-end testing.

```
[tls]
certificates_directory = "/var/sota/"
ca_path = "root.crt"
cert_path = "01"
cert_source = "pkcs11"
pkey_path = "02"
pkey_source = "pkcs11"

[p11]
module = "/usr/lib/softhsm/libsofthsm2.so"
pass = "1234"

[uptane]
metadata_path = "/var/sota/metadata"
key_source = "pkcs11"
private_key_path = "03"
public_key_path = "03"

[provision]
provision_path = "/var/sota/sota_provisioning_credentials.zip"
```

OYTIS:
@patrickvacek @patriotyk @cajun-rat 
Does "sign_file" test from uptane_test.cc test anything that is not tested already in tests for crypto? I'm confused a bit with that we're creating an uptane repo instance and then don't use it just calling static functions from Crypto. I've created a similar test called sign_pkcs11, but I feel like it's redundant not to say ugly.

OYTIS:
Performed a manual e2e test with and without virtual PKCS#11 device.

patrickvacek:
@OYTIS Re: signing tests, should those tests live in the crypto and p11 test files instead? It does seem like they aren't really testing Uptane directly. No point in building the repo in that case.

src/httpclient.cc
```diff
@@ -1,6 +1,8 @@
 #include "httpclient.h"
 
 #include <assert.h>
+#include <libp11.h>
```
cajun-rat: IWe need to put this behind the #ifdef BUILD_P11

I think that we will also need to make one of the test builds run without libp11 installed, to check that the build without libp11 works

OYTIS: How about running test.sh without BUILD_P11 then?

cajun-rat: My guess is that libp11.h gets installed in a location on the default include search path, so the build will succeed on any machine with the libp11 debian package installed.

OYTIS: So do we need a separate Dockerfile to test that?

src/config.cc
```diff
@@ -176,11 +176,36 @@ void Config::updateFromPropertyTree(const boost::property_tree::ptree& pt) {
...
+  tls_source = "file";
+  CopyFromConfig(tls_source, "tls.pkey_source", LVL_warning, pt);
+  if (tls_source == "pkcs11")
+    tls.pkey_source = kPkcs11;
+  else
+    tls.pkey_source = kFile;
+
+  CopyFromConfig(tls.ca_path, "tls.ca_path", LVL_warning, pt);
+  CopyFromConfig(tls.pkey_path, "tls.pkey_path", LVL_warning, pt);
+  CopyFromConfig(tls.cert_path, "tls.cert_path", LVL_warning, pt);
```
cajun-rat: Wouldn't it be better to avoid changing these configuration file names?  The Aktualizr recipe uses them today, and otherwise we will need to offer some backwards compatibility.

OYTIS: But our sota.toml is in /usr now, we can safely update it together with the aktualizr binary.

patrickvacek:
So I'm getting errors trying to test the crypto functions. Nice and generic/unhelpful:

```
Couldn't load PKCS11 module /usr/lib/softhsm/libsofthsm2.so
10: unknown file: Failure
10: C++ exception with description "PKCS11 error" thrown in the test body.
```

I set TEST_PKCS11_MODULE_PATH to /usr/lib/softhsm/libsofthsm2.so since that's where my OS installed it; that should be fine, right? Any idea what might be going wrong?

src/httpclient.cc
```diff
@@ -54,6 +59,8 @@ HttpClient::HttpClient() : user_agent(std::string("Aktualizr/") + AKTUALIZR_VERS
   headers = curl_slist_append(headers, "Accept: */*");
   curl_easy_setopt(curl, CURLOPT_HTTPHEADER, headers);
   curl_easy_setopt(curl, CURLOPT_USERAGENT, user_agent.c_str());
+
+  pkcs11_key = pkcs11_cert = false;
```
cajun-rat: Would it be nicer to put this at the top of the ctor?

~~~
HttpClient::HttpClient() : 
    user_agent(std::string("Aktualizr/") + AKTUALIZR_VERSION) 
    pkcs11_key(false),
    pkcs11_cert(false) {
~~~

OYTIS: Yes, sure makes sense.

src/config.cc
cajun-rat: Yes, but we'll need to add both to meta-updater now, to support people who build latest master of Aktualizr.  What was the reason for changing the names?

OYTIS: Ah, we have this bleeding edge build. OK, we can revert the names. I just thought that "path" sounds more generic than "file", which is good once PKCS#11 URI are supported. And this "client_certificate" field name was an annoying inconsistency as well.

cajun-rat:
Modulo reverting the minor changes to the config file, approved by me.

One thing that bothers me is that there is lots of `#ifdef BUILD_P11` in here.  I suspect it would be possible to refactor this and put the storage of key material, signing json objects and injecting TLS settings into curl behind a simple interface. We'd then have a PKCS#11 and `INvStorage`-backed implementation of that interface. Hopefully that would reduce the number of places we have either `#ifdef BUILD_P11...` or `if ( bla == kPkcs11) { ... } else { ... }`

For now it think it is more important to get this branch finally merged, and we can make these refactorings against in the mainline codebase.

src/config.cc
cajun-rat: Yes, the name changes are nice, but last time with `certificates_path` I seemed to spend ages running into new places where the config file needed changing :)

OYTIS:
@cajun-rat 
BTW, what would you think about storing 'sota_*.toml' in aktualizr repo? Yocto recipe will then just have to choose the proper one from the provided set. It's what they're doing in rvi-sota-client and it can give use more flexibility wrt. configuration field names.

cajun-rat:
Manually tested in meta-updater vs qemu, both with and without `hsm` in `SOTA_CLIENT_FEATURES`

cajun-rat:
@OYTIS re sota_*.toml in the Aktualizr repo: that makes sense.  I've created PRO-4098 to track it.

PR #360, opened 2017-10-02
Bugfix/pro 4038/migration

patriotyk:
It fixes https://advancedtelematic.atlassian.net/browse/PRO-4038 
Instead of fake http server, I've found simpler solution for test.


cajun-rat:
@OYTIS Does this look good to you?  I think you have the best understanding of this part of the codebase

PR #362, opened 2017-10-04
Feat/pro 3799/implicit writer

patrickvacek:
This is the connection between aktualizr and Yocto for implicit provisioning. To test it properly, you will also need the corresponding branch of meta-updater.


patrickvacek:
BTW, one important matter that is still open: I did not write any code to test the functionality of implicit_writer. Since it is presumably a necessary tool for proper functionality of implicit provisioning (unlike cert_provider, which is just for testing), it should probably be tested. I think gtest is perhaps not the best method here, so should I write some ctests to verify input/output?

OYTIS:
Sorry, why do we need setters in TlsConfig?

patrickvacek:
We don't really, but I made getters so that I could control whether or not to prepend the certificates_directory. Since I did that, I figured I'd make the underlying variables private and add setters. I suppose we could skip that and keep the variables public if that's preferable.

OYTIS:
Let's keep *Config classes as dumb as possible. We can switch to absolute pathes in a separate PR which will probably eliminate the need for getters.

patrickvacek:
Unfortunately, I need to be able to have the root CA as an absolute path for implicit provisioning, unless we also want to install the client cert and key in /usr/lib/sota. For now, I was going to leave those in /var/sota but use the option to have the root CA specified as an absolute path to handle that specially.

patrickvacek:
Rebased and removed TlsConfig setters. Getters remain for now as a convenient interim solution.

PR #364, opened 2017-10-05
Integrate sota-tools into aktualizr build and CI processes

OYTIS:
Formatting is not yet done/enforced for sota-tools.


OYTIS:
"no-whitespace" test is commented out. I don't quite understand what it does and it doesn't terminate in reasonable time. Can't it be replaced with formatting checks?

OYTIS:
OK, it's just because ALL_SRCS and ALL_HEADERS are empty in tests/CMakeLists.txt. We can either move this test to where other formatting tests live or delete it.

OYTIS:
OK, I'll delete the no-whitespace and merge it. We can always add it back if trailing whitespaces get too annoying.

PR #365, opened 2017-10-05
Show correct message when installing non ostree image on ostree system

patriotyk:
This PR adds understandable error message when installing non ostree package on ostree systems. But
I think we don't need allow to install such packages on ostree systems from web panel. We currently support only ostree packages for primaries so we could easily  implement it on the server.


cajun-rat:
(rebased)

cajun-rat:
Tested on qemu vs current production (with a small patch to meta-updater to fix installed vs packaged warnings)

PR #370, opened 2017-10-10
Feat/pro 4082/garage deploy

patriotyk:


src/garage_deploy/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+endif(CLANG_CHECK)
+
+
+INSTALL(TARGETS garage-deploy DESTINATION bin COMPONENT garage_deploy)
+
+
+SET(CPACK_GENERATOR "DEB")
+SET(CPACK_DEBIAN_PACKAGE_MAINTAINER "Serhiy Stetskovych<serhiy.stetskovych@innoteka.com>") #required
+SET(CPACK_PACKAGE_DESCRIPTION_SUMMARY "garage-deploy utility")
+set(CPACK_DEB_COMPONENT_INSTALL ON)
```
cajun-rat: Please can you use lower-case names here ('set' not 'SET') to match the rest of the CMakeLists.

tests/CMakeLists.txt
```diff
@@ -314,6 +314,17 @@ if(BUILD_SOTA_TOOLS)
...
 
+
+add_test(NAME garage-deploy-option-help
+    COMMAND garage-deploy --help)
+
+add_test(NAME garage-deploy-nonexistent-option
+    COMMAND garage-deploy --test )
+set_tests_properties(garage-deploy-nonexistent-option PROPERTIES WILL_FAIL TRUE)
+
+add_test(NAME garage-deploy-build-deb COMMAND cpack WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
```
cajun-rat: I get a test failure when running this test, which I suspect is because the executables have not been built. I think that it isn't necessary to run this as a test: just making the 'qa' target depend in 'package' should be enough

cajun-rat:
I'm rebasing this to merge in my clang-format/clang-check changes, hold on a second

patrickvacek:
Actually there is a (minor) problem. `make qa` does not succeed for me because the `package` target is not found. `make package` does work, however.

cajun-rat:
Ah, it appears that `add_dependencies(qa package)` isn't supported in cmake:

https://public.kitware.com/Bug/view.php?id=8438

It does actually work with ninja, though

patrickvacek:
FYI, I found the new version of the bug. Completely untouched.

https://gitlab.kitware.com/cmake/cmake/issues/8438

patriotyk:
Sorry guys, there was not latest changes, I committed but forget push them.

cajun-rat:
I've had to make some changes and rebase to incorporate commit b47e4dbce9cce03e69434d802575de37610d2212

cajun-rat:
Replaced with #378

PR #371, opened 2017-10-11
Feat/pro 3799/implicit writer test

patrickvacek:
Pretty basic, so I'm open to suggestions if there's anything else I should check. The python script just checks the output files of implicit_writer and makes sure they match the expected values based on the inputs.


cajun-rat:
`(+3.21%) compared to 8aadd26`  makes me happy :)

PR #373, opened 2017-10-12
Add option for implicit-writer to skip writing root certificate

OYTIS:
Extracting root.crt from bootstrap.p12 doesn't work in some cases:
1. openssl (which we use for PKCS#11 case) doesn't work with root.crt stored in bootstrap credentials, it needs the issuer of client certicate there as well.
2. Bootstrap credentials don't contain the certificate for S3 servers and will therefore fail on staging environment (and on production when redirection to AWS is merged into production).

It is therefore convenient to have an option to suppress output of root.crt.

### DIFF
--- a/src/implicit_writer/main.cc
+++ b/src/implicit_writer/main.cc
@@ -31,7 +31,8 @@ bpo::variables_map parse_options(int argc, char *argv[]) {
       ("credentials,c", bpo::value<std::string>()->required(), "zipped credentials file")
       ("config-input,i", bpo::value<std::string>()->required(), "input sota.toml configuration file")
       ("config-output,o", bpo::value<std::string>()->required(), "output sota.toml configuration file")
-      ("prefix,p", bpo::value<std::string>(), "prefix for root CA output path");
+      ("prefix,p", bpo::value<std::string>(), "prefix for root CA output path")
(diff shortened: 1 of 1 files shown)


patrickvacek:
How do we got about solving those two problems?

OYTIS:
Temporary solution is just to use root.crt provided by cert_provider. Permanent solution would be to pack everything necessary in the root.crt we get as a part of implicit provisioning credentials package.

patrickvacek:
Okay, that sounds like what I was thinking. Kind of a server-side issue in generating the package, right? I suppose we could do some fiddling with the root.crt in implicit_writer, but I'm not too excited about that.

OYTIS:
> I suppose we could do some fiddling with the root.crt in implicit_writer, but I'm not too excited about that.

Not really. The missing data is really missing, it can only be found in the package we receive as a result of device registration.

cajun-rat:
@koshelev Where should we be getting the root CA from?

patrickvacek:
#2 is going to be a bug for the non-implicit case, too, right? How will a device bootstrap itself if it needs the S3 cert and doesn't have it?

OYTIS:
@patrickvacek No, we only need S3 credentials for pulling from treehub. For device registration what we have in bootstrap.p12 is enough.

PR #375, opened 2017-10-12
Run clang-check and clang-format in parallel

cajun-rat:


patrickvacek:
This generates a crap-ton of formatting output. Could we maybe... reduce that?

Otherwise this is really cool!

patrickvacek:
I don't know if it's worth holding up the PR over that. It's good by me if you're ready to merge it.

PR #376, opened 2017-10-13
Support extracting client certificate from PKCS#11 device

OYTIS:
device_id file is used in end-to-end tests.

### DIFF
--- a/src/crypto.cc
+++ b/src/crypto.cc
@@ -247,6 +247,24 @@ bool Crypto::parseP12(FILE *p12_fp, const std::string &p12_password, std::string
   return true;
 }
 
+bool Crypto::extractSubjectCN(const std::string &cert, std::string *cn) {
+  BIO *bio = BIO_new_mem_buf(const_cast<char *>(cert.c_str()), (int)cert.size());
+  X509 *x = PEM_read_bio_X509(bio, NULL, 0, NULL);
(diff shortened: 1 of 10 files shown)


patrickvacek:
This compiles and passes all tests, but should we have a specific test case for testing this functionality? And is there a convenient way to test this end-to-end, or is that still underway as part of the rest of the p11 work? Generally I think this looks fine, though.

OYTIS:
@patrickvacek 
Yes, testing can be a bit tricky, I suspect that when we import a certificate into SoftHSM only first one really gets imported, so we can't just compare the textual versions. Extracting a device ID can be a good test, I'll write that one.

OYTIS:
@patrickvacek 
Oh, yes how you can see it's using pkcs11. The only thing that is currently in the log is the listing of a token's properties, it can be found in the beginning of the log. Other than that you can just distingush two working modes by looking into sota.toml. Do you need any additional information in the logs?

patrickvacek:
I have aktualizr sort of working and I see the log message now. That should be fine. However, now I'm seeing errors like this:
```
* <url> malformed
* Closing connection -1
[2017-10-16 09:53:18.294106]: curl error 3: URL using bad/illegal format or missing URL
```
...and `The director repo is missing.`

PR #377, opened 2017-10-13
Add support of certs authentication and tests

patriotyk:


src/sota_tools/CMakeLists.txt
```diff
@@ -11,7 +11,12 @@ set (GARAGE_PUSH_SRCS
   ostree_object.cc
   request_pool.cc
   treehub_server.cc
+  ${PROJECT_SOURCE_DIR}/src/utils.cc
+  ${PROJECT_SOURCE_DIR}/src/logger.cc
+  ${PROJECT_SOURCE_DIR}/third_party/jsoncpp/jsoncpp.cpp
 )
+set_source_files_properties(${PROJECT_SOURCE_DIR}/third_party/jsoncpp/jsoncpp.cpp PROPERTIES COMPILE_FLAGS -w)
+
```
cajun-rat: I'm not keen on having a CMakeLists pull files from another random directory--can't we use a library (libaktualizr) to share this between projects, rather than having one C++ file that is built into several binaries?

src/sota_tools/authenticate.cc
```diff
@@ -18,7 +18,27 @@ using boost::property_tree::json_parser::json_parser_error;
...
+  const char *buff = nullptr;
+  std::stringstream result;
+  size_t size;
+  int64_t offset;
+  for (;;) {
+    r = archive_read_data_block(a, (const void **)&buff, &size, &offset);
+    if (r == ARCHIVE_EOF) {
+      break;
+    } else if (r != ARCHIVE_OK) {
+      std::cerr << archive_error_string(a) << std::endl;
```
cajun-rat: What about making this an exception?

patrickvacek:
Looks good to me other than the aforementioned.

src/sota_tools/treehub_server.cc
```diff
@@ -35,8 +44,32 @@ void TreehubServer::InjectIntoCurl(const string& url_suffix, CURL* curl_handle)
     curl_easy_setopt(curl_handle, CURLOPT_PASSWORD, password_.c_str());
   }
 
-  if (ca_certs_ != "") {
-    curl_easy_setopt(curl_handle, CURLOPT_CAINFO, ca_certs_.c_str());
+  boost::filesystem::path tmp_dir = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
```
cajun-rat: Leaving junk around in a user's tmp directory isn't great--can't we use `TemporaryFile` from utils.h to manage this for us (but we'll have to make the TemporaryFile a member variable of TreehubServer since the curl handle outlives this function call.

tests/sota_tools/auth_test.cc
```diff
@@ -50,6 +79,8 @@ TEST(authenticate, invalid_file) {
 #ifndef __NO_MAIN__
 int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
+  TestHelperProcess server_process("fake_http_server/ssl_server.py", "");
+  sleep(2);
```
cajun-rat: Can we use getFreePort() from test_utils.h to make this pick a port automatically?

src/sota_tools/authenticate.cc
```diff
@@ -131,6 +143,10 @@ int authenticate(const string &cacerts, string filepath, TreehubServer &treehub)
       }
       break;
     }
+    case CERT: {
+      treehub.SetCerts(root_cert, client_cert, client_key);
```
patrickvacek: Is it worth checking at this point that all three cert strings are non-empty? I think it's better to fail early and explicitly in that case.

src/sota_tools/CMakeLists.txt
patriotyk: Yes, we can

src/sota_tools/authenticate.cc
patriotyk: Ok, I will do it, but it is old code.

src/sota_tools/authenticate.cc
```diff
@@ -72,10 +87,6 @@ int authenticate(const string &cacerts, string filepath, TreehubServer &treehub)
       std::cerr << "Error closing zipped credentials file: " << filepath << std::endl;
       return EXIT_FAILURE;
     }
-    if (!found) {
-      std::cerr << "treehub.json not found in zippled credentials file: " << filepath << std::endl;
-      return EXIT_FAILURE;
-    }
```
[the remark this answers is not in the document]
patriotyk: Yes, You are right, I removed it because I thought that for cert authentication we don't need that file, but we need it to get treehub url.

src/sota_tools/authenticate.cc
```diff
@@ -96,10 +107,7 @@ int authenticate(const string &cacerts, string filepath, TreehubServer &treehub)
       method = AUTH_BASIC;
       auth_user = ba_pt->get<string>("user", "");
       auth_password = ba_pt->get<string>("password", kPassword);
-    } else {
-      std::cerr << "Unknown authentication method " << std::endl;
```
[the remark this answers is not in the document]
patriotyk: Ok

src/sota_tools/treehub_server.cc
```diff
@@ -2,10 +2,11 @@
 #include <iostream>
 
 #include "treehub_server.h"
+#include "utils.h"
 
 using std::string;
 
-TreehubServer::TreehubServer() : using_oauth2_(false) {
+TreehubServer::TreehubServer() : using_oauth2_(false), using_certs_(false) {
```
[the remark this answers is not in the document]
patriotyk: I did it in the same way as it was for another auth methods

src/sota_tools/treehub_server.cc
patriotyk: Good Idea

tests/sota_tools/auth_test.cc
patriotyk: It is not so easy, because url with port number is inside auth archive.

src/sota_tools/authenticate.cc
patriotyk: Ok

src/sota_tools/authenticate.cc
```diff
@@ -108,6 +116,10 @@ int authenticate(const string &cacerts, string filepath, TreehubServer &treehub)
     return EXIT_FAILURE;
   }
 
+  if (method == AUTH_NONE && client_cert.size() && client_key.size()) {
+    method = CERT;
+  }
+
```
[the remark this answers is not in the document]
patriotyk: Ok

src/sota_tools/authenticate.cc
patrickvacek: Looks good, thanks.

tests/sota_tools/auth_test.cc
cajun-rat: OK

PR #378, opened 2017-10-16
PRO-4082 garage deploy

cajun-rat:


cajun-rat:
Yep, no problem with the merge, thank you for reviewing

PR #379, opened 2017-10-16
Remove test directory after use.

patrickvacek:
This is really helpful if switching between running tests as root and
non-root. Otherwise, root will leave a directory that non-root cannot
use.


cajun-rat:
Nice one.  We might want to think about a RIAA temporary directory class sometime...

patrickvacek:
> We might want to think about a RIAA temporary directory class sometime...

That would be wonderful to do alongside moving the tests to the build directory. I'll add that to my list of test case cleanups...

PR #380, opened 2017-10-17
Bugfix/pro 4187/secondary ecu paths

patrickvacek:
This is really just a pretty basic path handling fixup, but along the way I fixed up a bunch of other stuff. This was one of those bugs where I think I had to learn everything the hard way.

### DIFF
--- a/src/uptane/secondary.cc
+++ b/src/uptane/secondary.cc
@@ -37,8 +37,8 @@ Json::Value Secondary::genAndSendManifest(Json::Value custom) {
   manifest["previous_timeserver_time"] = "1970-01-01T00:00:00Z";
   manifest["timeserver_time"] = "1970-01-01T00:00:00Z";
 
-  std::string public_key = Utils::readFile((config.full_client_dir / config.ecu_public_key).string());
-  std::string private_key = Utils::readFile((config.full_client_dir / config.ecu_private_key).string());
+  std::string public_key = Utils::readFile(config.ecu_public_key());
+  std::string private_key = Utils::readFile(config.ecu_private_key());
   // TODO: support PKCS11 for secondaries. Maybe.
   Json::Value signed_ecu_version = Crypto::signTuf(NULL, private_key, Crypto::getKeyId(public_key), manifest);
   return signed_ecu_version;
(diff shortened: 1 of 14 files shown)


tests/CMakeLists.txt
```diff
@@ -64,7 +78,6 @@ function(add_aktualizr_test)
                      ${WD})
         endif()
     add_dependencies(build_tests t_${AKTUALIZR_TEST_NAME})
-    aktualizr_source_file_checks(${AKTUALIZR_TEST_SOURCES})
```
cajun-rat: What about adding an option to `add_aktualizr_test` to skip adding the sources?  That would remove one more place for people to forget. Adding the .h files is a good catch, though

patrickvacek: Possible, but this gets tricky fast. The problematic file is fake_ostree.cc. It is used as a supplemental file by two tests, both of which also have an individual .cc as well. Hence, if you skipped adding the sources of one, you'd still have to manually add the individual .cc.

But I was sad to have to break your logic there, so maybe that is worth it. (Or is there a way to skip duplicates? Or could `add_aktualizr_test` maybe just add the files to a list, which can be de-dup'd, and then run `aktualizr_source_file_checks` on the list at the end?)

patrickvacek:
I managed to fix the issue with `aktualizr_source_file_checks` about as ideally as I think is possible. Files are automatically added to a list in `add_aktualizr_test` and can be automatically added as well, and the list is de-duped to prevent ugly CMake errors before handing off the list to `aktualizr_source_file_checks`.

src/uptane/secondaryconfig.h
```diff
@@ -1,17 +1,23 @@
...
-    boost::filesystem::path firmware_path;
+namespace Uptane {
+class SecondaryConfig {
+ public:
+  std::string ecu_private_key() const;
+  std::string ecu_public_key() const;
+
+  std::string ecu_serial;
+  std::string ecu_hardware_id;
+  std::string ecu_private_key_;
```
patriotyk: ecu_private_key_ and ecu_public_key_ should be private?

patrickvacek: I have wondered about that. For now, I'm following the convention we set with the TlsConfig, which is that you can get/set the strings directly, but if you want to get the path with the directory prefixed automatically, you have to use the function. We could perhaps use a different name for the function if that would help make it clearer.

src/uptane/secondary.cc
```diff
@@ -59,18 +58,24 @@ Json::Value Secondary::newTargetsCallBack(const std::vector<Uptane::Target> &tar
 }
 
 void Secondary::setKeys(const std::string &public_key, const std::string &private_key) {
-  if (!boost::filesystem::exists(config.full_client_dir)) {
-    boost::filesystem::create_directories(config.full_client_dir);
+  boost::filesystem::path private_parent_dir = boost::filesystem::path(config.ecu_private_key()).parent_path();
+  if (!private_parent_dir.empty()) {
+    boost::filesystem::create_directories(private_parent_dir);
```
cajun-rat: We might want to think about adding an optional enum second parameter to Utils::writeFile to optionally create any necessary parent directories.

patrickvacek: I am in favor of that!

PR #383, opened 2017-10-21
Feat/pro 4196/garage deploy copying

patriotyk:


cajun-rat:
with BUILD_SOTA_TOOLS=OFF, a local build with (with 'make -j8 qa') fails with the following error:

```
The following tests FAILED:
	 33 - garage-deploy-option-help (Not Run)
	 34 - garage-deploy-nonexistent-option (Not Run)
```

With BUILD_SOTA_TOOLS=ON, I get the following errors:

```
../src/sota_tools/ostree_ref.cc: In constructor ‘OSTreeRef::OSTreeRef(const OSTreeRepo&, std::__cxx11::string)’:
../src/sota_tools/ostree_ref.cc:18:93: error: no matching function for call to ‘std::basic_ifstream<char>::basic_ifstream(std::__cxx11::basic_string<char>, std::_Ios_Openmode)’
     std::ifstream f(repo.root() + "/refs/heads/" + ref_name, std::ios::in | std::ios::binary);
```

I'm running Ubuntu 16.04 LTS

patrickvacek:
I'm getting this again: `make[2]: *** No rule to make target 'CMakeFiles/package.dir/all', needed by 'CMakeFiles/qa.dir/all'.  Stop.` This came up before with https://github.com/advancedtelematic/aktualizr/pull/378. Phil's fix (https://github.com/advancedtelematic/aktualizr/pull/378/commits/866505f1b05b5c994556e3a84dbb541011e81ba6) was to remove the dependency of `qa` on `package`. Apparently since `package` is a built-in thing, cmake doesn't play happily with it being a dependency. (Seems like a bug in cmake TBH.) I suggest re-incorporating that fix and thus doing a `make package` in `test.sh`.

scripts/test_deb_in_docker.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+if [ "$1" == "Dockerfile.deb-stable" ]; then
+  echo "Building docker for test deb package inside it."
+  docker build -t advancedtelematic/garage_deploy_test -f src/sota_tools/Dockerfile.debian.stable .
+  echo "Running docker conatiner with debian packege inside."
+  docker run -v /persistant:/persistant -t advancedtelematic/garage_deploy_test
```
patrickvacek: Please excuse my pedantry, but I think it's worth getting right for things at this level. The correct spellings are "package" and "persistent".

src/sota_tools/CMakeLists.txt
```diff
@@ -11,19 +10,100 @@ set (GARAGE_PUSH_SRCS
...
+add_executable(garage-push ${GARAGE_PUSH_SRCS})
 
-add_executable(garage-push main.cc)
-add_library(garage_push_static_lib STATIC ${GARAGE_PUSH_SRCS})
-target_link_libraries(garage-push garage_push_static_lib aktualizr_static_lib
+# Export compile_commands.json for clang-check
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+target_compile_options(sota_tools_static_lib PUBLIC -std=c++11)
+
+target_include_directories(sota_tools_static_lib INTERFACE ${PROJECT_SOURCE_DIR} ${GLIB2_INCLUDE_DIRS})
```
patrickvacek: I had to change this line for a fix in meta-updater. See my https://github.com/advancedtelematic/aktualizr/pull/384/commits/a0c50801d1a9472ccbf7c8d73d9032529913ac31 for the details. You should keep the syntax from that but of course use the new `sota_tools_static_lib` name.

src/sota_tools/CMakeLists.txt
```diff
@@ -11,19 +10,100 @@ set (GARAGE_PUSH_SRCS
...
+
+install(TARGETS garage-deploy RUNTIME DESTINATION bin COMPONENT garage_deploy)
+add_custom_target(download_garage_sign COMMAND wget https://ats-tuf-cli-releases.s3-eu-central-1.amazonaws.com/cli-0.1.0-175-g9618465.tgz -O ${CMAKE_CURRENT_BINARY_DIR}/garage-sign.tgz
+COMMAND tar -xzf ${CMAKE_CURRENT_BINARY_DIR}/garage-sign.tgz -C ${CMAKE_CURRENT_BINARY_DIR}
+COMMAND chmod +x ${CMAKE_CURRENT_BINARY_DIR}/garage-sign/bin/*)
+FILE(GLOB garage_sign_bin_files ${CMAKE_CURRENT_BINARY_DIR}/garage-sign/bin/*)
+FILE(GLOB garage_sign_lib_files ${CMAKE_CURRENT_BINARY_DIR}/garage-sign/lib/*)
+
+install(PROGRAMS ${garage_sign_bin_files} DESTINATION /usr/bin COMPONENT garage_deploy)
+install(FILES ${garage_sign_lib_files} DESTINATION /usr/lib COMPONENT garage_deploy)
```
patrickvacek: These should be use `RUNTIME DESTINATION bin` and `lib` instead of hardcoding `/usr/bin` and `/usr/lib`. See https://github.com/advancedtelematic/aktualizr/pull/384/commits/a0c50801d1a9472ccbf7c8d73d9032529913ac31.

patriotyk:
Processed all comments, but we have a new Issue. Someone on the server side has disabled public access to the garage-sign package, so cmake can't download package and build fails. How can we fix that?

patrickvacek:
Sorry @patriotyk, should have warned you. We need to work out a licensing issue before we can make that URL accessible again. We should get that sorted out on Wednesday.

patrickvacek:
FYI @patriotyk I rebased and updated this with a few fixes so that I can actually test it effectively. I think I'm done messing with it and I can evaluate if it actually works now.

patrickvacek:
So I think this covers PRO-4212 (get garage-sign packaged inside garage-deploy) and stage 1 of PRO-4196 (use garage-deploy to copy data from one repo to another), but stage 2 (actually make use of garage-sign) still needs to be done. Does that sound correct?

patrickvacek:
Actually, I'm seeing a problem: when I push a new update to a device and then reboot it, it appears to reprovision itself with a new device_id. Something isn't right there.

If possible, Phil and I agree that if we can get this chunk of work merged now/soon, that would be advantageous. The last bit of garage-sign integration can be down in another branch based on this one.

patrickvacek:
Hmmm. This reprovisioning bug is happening even on master... this will require investigation in the morning.

cajun-rat:
I get the following error what trying to build everything with `ninja qa`:

```
[162/276] cd /home/phil/pro/client/aktualizr/build && cpack
FAILED: src/sota_tools/CMakeFiles/garage-deploy-deb 
cd /home/phil/pro/client/aktualizr/build && cpack
CPack: Create package using DEB
CPack: Install projects
CPack: - Install project: aktualizr
CPack: -   Install component: garage_deploy
CMake Error at /home/phil/pro/client/aktualizr/build/src/sota_tools/cmake_install.cmake:58 (file):
  file INSTALL cannot find
  "/home/phil/pro/client/aktualizr/build/src/sota_tools/garage-deploy".
Call Stack (most recent call first):
  /home/phil/pro/client/aktualizr/build/cmake_install.cmake:41 (include)


CPack Error: Error when generating package: aktualizr
[165/276] Building CXX object src.../garage-push.dir/garage_push.cc.o
ninja: build stopped: subcommand failed.
```

patriotyk:
I will look to it today later, after pkcs11 task

patrickvacek:
For the record, the re-provisioning bug is not related to this. I'm working on addressing that separately. Sorry for the confusion.

patriotyk:
Hm, when I was fixing it, wget complained to self signed certificate. But now it works.

cajun-rat:
OK, I'm going to merge this. It has turned into a rather giant changeset, with a couple of overlapping changes in it.  In the future we should try and keep these separate.

PR #393, opened 2017-11-07
Add keygeneration on device

patriotyk:


Dockerfile
```diff
@@ -56,9 +55,13 @@ RUN apt-get update && apt-get -y install \
...
   python3-venv \
-  softhsm2 \
   valgrind \
   wget
 
+RUN wget http://mirrors.kernel.org/ubuntu/pool/universe/s/softhsm2/softhsm2-common_2.2.0-3_amd64.deb && dpkg -i softhsm2-common_2.2.0-3_amd64.deb
+RUN wget http://mirrors.kernel.org/ubuntu/pool/universe/s/softhsm2/libsofthsm2_2.2.0-3_amd64.deb && dpkg -i libsofthsm2_2.2.0-3_amd64.deb
+RUN wget http://mirrors.kernel.org/ubuntu/pool/universe/s/softhsm2/softhsm2_2.2.0-3_amd64.deb && dpkg -i softhsm2_2.2.0-3_amd64.deb
+RUN wget http://mirrors.kernel.org/ubuntu/pool/main/libp/libp11/libp11-2_0.4.7-1_amd64.deb && dpkg -i libp11-2_0.4.7-1_amd64.deb
+RUN wget http://mirrors.kernel.org/ubuntu/pool/main/libp/libp11/libp11-dev_0.4.7-1_amd64.deb && dpkg -i libp11-dev_0.4.7-1_amd64.deb; test 1
```
patrickvacek: Just curious: what makes these wgets necessary instead of the regular apt-gets? And why the `test 1` at the end?

scripts/setup_hsm.sh
```diff
@@ -2,10 +2,11 @@
...
+echo "Initialized token in slot: $SLOT"
+softhsm2-util --import ./src/tests/test_data/implicit/pkey.pem --label "pkey" --id 02 --slot $SLOT --pin 1234
 openssl x509 -outform der -in ./src/tests/test_data/implicit/client.pem -out ./src/tests/test_data/implicit/client.der
-pkcs11-tool --module=/usr/lib/x86_64-linux-gnu/softhsm/libsofthsm2.so --id 1 --write-object ./src/tests/test_data/implicit/client.der --type cert --login --pin 1234
+pkcs11-tool --module=/usr/lib/softhsm/libsofthsm2.so --id 1 --write-object ./src/tests/test_data/implicit/client.der --type cert --login --pin 1234
 
 openssl pkcs8 -topk8 -inform PEM -outform PEM -nocrypt -in ./src/tests/test_data/priv.key -out ./src/tests/test_data/priv.p8
-softhsm2-util --import ./src/tests/test_data/priv.p8 --label "uptane" --id 03 --slot 0 --pin 1234
-
+softhsm2-util --import ./src/tests/test_data/priv.p8 --label "uptane" --id 03 --slot $SLOT --pin 1234
```
patrickvacek: I like this look of this!

src/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 set_source_files_properties(${PROJECT_SOURCE_DIR}/third_party/jsoncpp/jsoncpp.cpp PROPERTIES COMPILE_FLAGS -w)
+set_source_files_properties(${PROJECT_SOURCE_DIR}/src/p11engine.cc PROPERTIES COMPILE_FLAGS -w)
```
patrickvacek: Dare I ask what necessitated hiding warnings from that file?

src/uptane/initialize.cc
```diff
@@ -128,28 +128,25 @@ bool Repository::initEcuKeys(const UptaneConfig& uptane_config) {
...
+    if (p11.readPublicKey(uptane_config.public_key_path, &public_key_content)) {
+      primary_public_id = Crypto::getKeyId(public_key_content);
+      setEcuKeysMembers(primary_public, primary_private, primary_public_id, kPkcs11);
+      return true;
     }
-    primary_public_id = Crypto::getKeyId(public_key_content);
-
-    setEcuKeysMembers(primary_public, primary_private, primary_public_id, kPkcs11);
-    return true;
+    if (!p11.generateRSAKeyPair(uptane_config.private_key_path)) return false;
```
OYTIS: We need to make sure that `private_key_path` and `public_key_path` are the same in the pkcs#11 case. Or just use one of these variables everywhere.

patriotyk: It is keypair, and when we ganarate it we need to pass the single id

src/CMakeLists.txt
patriotyk: The problem here is that one api call is marked as deprecated but they have not added not deprecated call for keypair generation

Dockerfile
patriotyk: I will explain on the call

src/CMakeLists.txt
patriotyk: Ok, maybe it is better to hide only deprecated warnings. I will update it

src/uptane/initialize.cc
patriotyk: Yes You are right, I will add some checks

patriotyk:
Now it should be ok

patrickvacek:
I'm seeing what looks like the same error on my device after I changed `uptane.key_source` to `05`.

OYTIS:
@patrickvacek It might be a library issue, yes.
I'm still having some weird unstable issues with the latest version of libp11, trying to pin them up.

OYTIS:
OK, this issue (empty signature in /director/manifest while registering) only happens on the first run, if one restarts aktualizr update works e2e (modulo https://github.com/advancedtelematic/aktualizr/pull/394)

src/uptane/initialize.cc
OYTIS: OK, I can see where the problem is. After we generate keys with generateRSAKeyPair we don't initialize members. Correct logic would be:
1. Try to readPublicKey.
2. On success skip, otherwise generateRSAKeyPair.
3. readPublicKey, on success set primary_public_id and setEcuKeyMembers, otherwise fail.

Can't make a patch right now unfortunately, let's do that tomorrow.

OYTIS:
@patriotyk 
I've made a PR to your PR, please pull to this branch if you have no objections: https://github.com/advancedtelematic/aktualizr/pull/395

cajun-rat:
Squashed merge commit and fixed up formatting

PR #394, opened 2017-11-08
Don't prepend '/' to pkcs11 URI

OYTIS:
A minor bug that currently breaks e2e with pkcs11


cajun-rat:
What is the test procedure for this bug?

OYTIS:
@cajun-rat 
I've noticed the bug when `ostree pull` failed with pkcs#11. Maybe such things can only be tested with e2e tests.

patrickvacek:
@cajun-rat We might be able to test this with what I'm doing with oe-selftest. We could also write unit tests to verify that the paths and URIs are what we expect them to be, but that seems a bit pedantic.

cajun-rat:
Leaving the test to the e2e tests is fine--there is no need to build a unit test if it can't test anything meaningful.

PR #400, opened 2017-11-15
Added dummy external flasher

patriotyk:


src/external_secondaries/ecuinterface.cc
```diff
@@ -0,0 +1,59 @@
...
+        std::cout << desc;
+      }
+
+      return EXIT_SUCCESS;
+    } else {
+      std::cout << "You must provide command: \n";
+      std::cout << desc;
+      return EXIT_FAILURE;
+    }
+    po::notify(vm);
```
cajun-rat: This line is unreachable.  The example at http://www.boost.org/doc/libs/1_65_1/doc/html/program_options/tutorial.html suggests that it should be called right after po::store()

src/external_secondaries/ecuinterface.h
```diff
@@ -0,0 +1,18 @@
+#include <string>
+
+class ECUInterface {
+ public:
+  enum InstallStatus {
+    InstallationSuccessfull = 0,
```
cajun-rat: Minor typo: This should be `InstallationSuccessful` (yes, English spelling is ridiculous)

PR #401, opened 2017-11-16
Add garage-check utility

OYTIS:
TreehubServer is responsible for authentication how it's done on the build machine (i.e. without a registered device), so I've made it do the job for accessing TUF repo as well. Which is of course a questionable solution.

### DIFF
--- a/src/sota_tools/garage_check.cc
+++ b/src/sota_tools/garage_check.cc
@@ -1,0 +1,167 @@
+
+#include <curl/curl.h>
+#include <boost/filesystem.hpp>
+#include <boost/program_options.hpp>
+#include <iomanip>
+#include <iostream>
+
+#include "accumulator.h"
+#include "authenticate.h"
+#include "json/json.h"
+#include "logging.h"
+#include "treehub_server.h"
+
+namespace po = boost::program_options;
+using std::string;
+
+// helper function to download data to a string
+static size_t writeString(void *contents, size_t size, size_t nmemb, void *userp) {
+  assert(userp);
+  // append the writeback data to the provided string
+  (static_cast<std::string *>(userp))->append((char *)contents, size * nmemb);
+
+  // return size of written data
+  return size * nmemb;
+}
+
+int main(int argc, char **argv) {
+  logger_init();
+
+  string repo_path;
+  string ref = "uninitialized";
+  string pull_cred;
+
+  string credentials_path;
+  string home_path = string(getenv("HOME"));
+  string cacerts;
+
+  int verbosity;
+  po::options_description desc("Allowed options");
+  // clang-format off
+  desc.add_options()
+    ("help", "produce a help message")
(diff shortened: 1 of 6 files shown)


src/sota_tools/garage_check.cc
```diff
@@ -0,0 +1,167 @@
...
+    LOG_FATAL << "Error initializing curl";
+    return EXIT_FAILURE;
+  }
+  curl_easy_setopt(curl_handle, CURLOPT_VERBOSE, get_curlopt_verbose());
+  curl_easy_setopt(curl_handle, CURLOPT_NOBODY, 1);  // HEAD
+
+  treehub.InjectIntoCurl("objects/" + ref.substr(0, 2) + "/" + ref.substr(2) + ".commit", curl_handle);
+
+  CURLcode result = curl_easy_perform(curl_handle);
+  curl_easy_cleanup(curl_handle);
```
cajun-rat: I think this should be after curl_easy_getinfo() a few lines down?

src/sota_tools/garage_check.cc
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  // check if the ref is present on treehub
+  CURL *curl_handle = curl_easy_init();
+  if (!curl_handle) {
+    LOG_FATAL << "Error initializing curl";
+    return EXIT_FAILURE;
+  }
+  curl_easy_setopt(curl_handle, CURLOPT_VERBOSE, get_curlopt_verbose());
+  curl_easy_setopt(curl_handle, CURLOPT_NOBODY, 1);  // HEAD
```
cajun-rat: Do we need a `1L` to force a long here?

src/sota_tools/garage_check.cc
```diff
@@ -0,0 +1,167 @@
...
+    LOG_FATAL << "OSTree commit " << ref << " is missing in treehub";
+    return EXIT_FAILURE;
+  } else if (http_code != 200) {
+    LOG_FATAL << "Error " << http_code << " getting commit " << ref << " from treehub";
+    return EXIT_FAILURE;
+  }
+  LOG_INFO << "OSTree commit " << ref << " is found on treehub";
+
+  // check if the ref is present in targets.json
+  curl_handle = curl_easy_init();
```
cajun-rat: ...actually I think we can reuse the curl_handle instead

cajun-rat:
A couple of minor issues, but otherwise good to go

src/sota_tools/garage_check.cc
OYTIS: Yes, it's a long, better to specify explicilty.

src/sota_tools/garage_check.cc
OYTIS: Makes sense, yes.

src/sota_tools/garage_check.cc
OYTIS: I'll test that.

PR #406, opened 2017-11-21
Refactor virtual secondaries to implement SecondaryInterface

OYTIS:
1. TimeMeta is deleted. It requires backend support and will be probably added in a different sprint.
2. ECU serial/hw_id generation is deleted. For testing it should be present in secondary config files, in future it can be reimplemented properly if needed.
3. Minor changes to SecondaryInterface.

### DIFF
--- a/src/uptane/secondaryconfig.h
+++ b/src/uptane/secondaryconfig.h
@@ -23,13 +23,12 @@ class SecondaryConfig {
   std::string ecu_serial;
   std::string ecu_hardware_id;
   bool partial_verifying;
-  std::string ecu_private_key_;
-  std::string ecu_public_key_;
+  std::string ecu_private_key;
+  std::string ecu_public_key;
 
   boost::filesystem::path full_client_dir;     // kVirtual, kLegacy
   boost::filesystem::path firmware_path;       // kVirtual
-  boost::filesystem::path time_path;           // kVirtual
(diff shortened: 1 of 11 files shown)


src/uptane/secondaryconfig.h
```diff
@@ -23,13 +23,12 @@ class SecondaryConfig {
   std::string ecu_serial;
   std::string ecu_hardware_id;
   bool partial_verifying;
-  std::string ecu_private_key_;
-  std::string ecu_public_key_;
+  std::string ecu_private_key;
+  std::string ecu_public_key;
```
patrickvacek: Looks good and I'm going to go ahead and merge this in, but the original reason for the duplicated `ecu_*_key_` members and `ecu_*_key` functions was to have functions that would automatically handle the case where the complete path was included without needing to use `full_client_dir`. This came up in one of Robert's tests. Not a big deal for now, but I like having one place where we normalize paths.

OYTIS: Yes, makes sense. I'd prefer to get rid of full_client_dir at all, it only makes stuff more complicated. I hope backwards compatibility is not an issue yet.

patrickvacek: I won't argue with that. Anything to make it simpler.

I manually rebased this into the main branch, but since there were (trivial) conflicts with my branch, github wasn't smart enough to automatically recognize what I did.

PR #407, opened 2017-11-21
Implement legacy secondaries

OYTIS:
Common functionality of legacy and virtual secondaries is isolated into ManagedSecondary class

### DIFF
--- a/src/uptane/legacysecondary.cc
+++ b/src/uptane/legacysecondary.cc
@@ -7,70 +7,34 @@
 #include "logger.h"
 
 namespace Uptane {
-LegacySecondary::LegacySecondary(const LegacySecondaryConfig &sconfig_in, Uptane::Repository *primary)
-    : sconfig(sconfig_in), transport(primary), wait_image(false) {
+LegacySecondary::LegacySecondary(const LegacySecondaryConfig &sconfig_in) {
   boost::filesystem::create_directories(sconfig.firmware_path.parent_path());
 }
 
-bool writeImage(const uint8_t *blob, size_t size) {
-  if (!wait_for_target) return false;
+bool LegacySecondary::storeFirmware(const std::string& target_name, const std::string& content) {
+  // reading target hash back is not currently supported, so primary needs to save the firmware file locally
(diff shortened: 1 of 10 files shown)


src/utils.cc
```diff
@@ -270,6 +271,27 @@ void Utils::copyDir(const boost::filesystem::path &from, const boost::filesystem
   }
 }
```
patriotyk: I added the similar method in my new PR )))

src/uptane/legacysecondary.cc
```diff
@@ -7,70 +7,34 @@
...
-      }
-    }
-  }
-  detected_attack = "";
-  Utils::writeFile(sconfig.target_name_path.string(), content);
-  return true;
+  std::string output
+  int rs = Utils::shell(sconfig.flasher.string() + " --hardware-identifier " + sconfig.ecu_hardware_id + " --ecu-identifier " + sconfig.ecu_serial + " --firmware " + sconfig.firmware_path.string(), &output);
+
+  if(rs != 0)
```
patriotyk: It is better always use {}. If someone will need to add some code inside if statement, he could not notice that there is no {} symbols.

src/uptane/legacysecondary.h
```diff
@@ -1,33 +1,21 @@
-#ifndef UPTANE_LEGACYSECONDARY_H_
-#define UPTANE_LEGACYSECONDARY_H_
+#ifndef UPTANE_VIRTUALSECONDARY_H_
+#define UPTANE_VIRTUALSECONDARY_H_
```
patrickvacek: That seems like a typo, right? It still should be `UPTANE_LEGACYSECONDARY_H_`.

patrickvacek:
I fixed the include guard issue and merged. While I generally agree with Serhiy about preferring brackets around single-clause branches, I don't think it is strictly necessary nor an impediment to moving forward with the merge and further development. (Nor is it a strict requirement in the Google style guide.) But perhaps before this is all finally merged into master we can look more closely at things like that.

PR #408, opened 2017-11-21
Added tests to test ecuinterface utility

patriotyk:


src/external_secondaries/validate_secondary_interface_test.cc
```diff
@@ -0,0 +1,111 @@
...
+  std::regex ecu_regex("\\w+(?: \\w+)?\\n?");
+
+  unsigned int matched_symbols = 0;
+  std::string::const_iterator search_start(output.cbegin());
+  while(std::regex_search(search_start, output.cend(), ecu_match, ecu_regex)) {
+    matched_symbols += ecu_match.length();
+    search_start += ecu_match.position() + ecu_match.length();
+  }
+  return matched_symbols == output.size();
+}
```
cajun-rat: Since the goal here is to check that the output of the user's tool matches what Aktualizr is expecting when it calls the tool would it be better to put this parsing code somewhere it can be shared with the code inside aktualizr that will parse the data 'for real'?

src/external_secondaries/CMakeLists.txt
```diff
@@ -6,6 +6,14 @@ set(MAIN_ECU_INTERFACE_SRC ecuinterface.cc)
 add_executable(myecuinterface ${MAIN_ECU_INTERFACE_SRC} dummy.cc)
 target_link_libraries(myecuinterface ${Boost_LIBRARIES})
 
+if(NOT FLASHER_VERSION)
+    add_definitions(-DFLASHER_VERSION="1")
+endif(NOT FLASHER_VERSION)
```
cajun-rat: For now this can be hard coded--the API version information logically belongs with the code that calls the API and parses the results.  If we switch to version 2, then the parser will have to change too. Therefore I think it is better to put the version number in with the parser.

src/external_secondaries/dummy.cc
```diff
@@ -2,13 +2,29 @@
...
+  return "1";
+}
 
-std::string ECUInterface::listEcus() { return "msp430\n"; }
+std::string ECUInterface::listEcus() {
+  if (loglevel_ == 4) {
+    std::cerr << "Displaying list of ecus:\n";
+  }
+  return "msp430 123456\n";
+}
```
cajun-rat: Good.

src/external_secondaries/validate_secondary_interface_test.cc
patriotyk: Good point, but where to put it? In utils.cc? Maybe later when we will need to use the same functionality we can move it to the appropriate place?

src/external_secondaries/CMakeLists.txt
patriotyk: OK

src/external_secondaries/validate_secondary_interface_test.cc
patrickvacek: Putting it in utils.cc is probably fine for now. Anton and I have a big secondaries branch that is still in progress, and until we merge that it's going to be difficult to pinpoint the right place.

PR #409, opened 2017-11-22
Don't set is_registered if the device really isn't

OYTIS:
'is_registered' is meant to indicate that ECUs are registered on the
backend, so no (other) local file can indicate the same. Current
behavior breaks certain scenarios where all these files are already
present at the time where aktualizr is called first time and also
decreases initialization robustness against interruptions. This commit
is basically reverting 4a9fb2b1b5d5d067b1bcd4675ce79433cd193039, so
migration from very old versions of aktualizr might be broken.

### DIFF
--- a/src/fsstorage.cc
+++ b/src/fsstorage.cc
@@ -12,22 +12,8 @@ FSStorage::FSStorage(const Config& config) : config_(config) {
   boost::filesystem::create_directories(config_.tls.certificates_directory);
   boost::filesystem::create_directories(config_.uptane.metadata_path / "repo");
   boost::filesystem::create_directories(config_.uptane.metadata_path / "director");
-
-  // migrate from old aktualizr behavior if we already provisioned
-  boost::filesystem::path public_key_path = config_.tls.certificates_directory / config_.uptane.public_key_path;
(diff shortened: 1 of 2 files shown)


OYTIS:
Oops, there was a test for it.

cajun-rat:
Isn't that test directly covering the migration behavior?

stevana:
Does it make sense to bump meta-updater also?

cajun-rat:
Anton and I talked about this, and decided that merging (and unblocking several other issues) was more important than temporarily breaking this old backwards compatibility code.  PRO-4406 covers the work to add the test back.

patrickvacek:
@stevana Do you need it for your tests? If so, then yes, we can certainly bump meta-updater.

stevana:
Technically no, but it would make things easier.

patrickvacek:
@stevana Rarely is it a problem for us to bump the version in pyro (that's what you need, right?). Anton already has a PR (https://github.com/advancedtelematic/meta-updater/pull/188) that I'm checking right now.

stevana:
Cool. Yep, pyro.

PR #413, opened 2017-11-24
Fix uptane test

OYTIS:
Restored registering secondary ECUs on director


OYTIS:
BTW, I can run coverage test locally now, so it might make sense to make a pull request for master at this point.

PR #414, opened 2017-11-27
Refactor/pro 4320/secondary interface

patrickvacek:
This is still a bit of a work-in-progress (as evidenced by all the TODOs) but this at least compiles and passes all tests (thanks @OYTIS!). It certainly covers redoing the interface, but the actual implementation still has some details to work out.


src/sotauptaneclient.cc
```diff
@@ -1,17 +1,33 @@
...
+#include "uptane/secondaryconfig.h"
+#include "uptane/secondaryfactory.h"
 #include "utils.h"
 
 SotaUptaneClient::SotaUptaneClient(const Config &config_in, event::Channel *events_channel_in, Uptane::Repository &repo)
-    : config(config_in), events_channel(events_channel_in), uptane_repo(repo), last_targets_version(-1) {}
+    : config(config_in), events_channel(events_channel_in), uptane_repo(repo), last_targets_version(-1) {
+  std::vector<Uptane::SecondaryConfig>::iterator it;
+  for (it = config.uptane.secondary_configs.begin(); it != config.uptane.secondary_configs.end(); ++it) {
+    std::map<std::string, boost::shared_ptr<Uptane::SecondaryInterface> >::const_iterator map_it =
```
patriotyk: We can use std:set for unique elements

src/uptane/legacysecondary.cc
```diff
@@ -0,0 +1,41 @@
...
+
+#include "crypto.h"
+#include "logger.h"
+
+namespace Uptane {
+LegacySecondary::LegacySecondary(const SecondaryConfig& sconfig_in) : ManagedSecondary(sconfig_in) {
+  boost::filesystem::create_directories(sconfig.firmware_path.parent_path());
+}
+
+bool LegacySecondary::storeFirmware(const std::string& target_name, const std::string& content) {
```
patriotyk: Firmware might be very big and store it in memory may consume all of free memory.

src/uptane/managedsecondary.h
```diff
@@ -0,0 +1,61 @@
...
+  std::string detected_attack;
+  std::string expected_target_name;
+  std::vector<Hash> expected_target_hashes;
+  int64_t expected_target_length;
+
+  MetaPack current_meta;
+
+  virtual bool storeFirmware(const std::string& target_name, const std::string& content) = 0;
+  // TODO: bool getFirmwareInfo(std::string* target_name, size_t& target_len, std::vector<Uptane::Hash>* hashes) {};
+  virtual bool getFirmwareInfo(std::string* target_name, size_t& target_len, std::string* sha256hash) {
```
patriotyk: Why it is not a pure virtual?

src/uptane/virtualsecondary.cc
```diff
@@ -0,0 +1,35 @@
...
+  Utils::writeFile(sconfig.firmware_path.string(), content);
+  return true;
+}
+
+bool VirtualSecondary::getFirmwareInfo(std::string* target_name, size_t& target_len, std::string* sha256hash) {
+  if (!boost::filesystem::exists(sconfig.target_name_path) || !boost::filesystem::exists(sconfig.firmware_path)) {
+    *target_name = std::string("noimage");
+    target_len = 0;
+    *sha256hash = std::string("");
+    return true;
```
patriotyk: maybe return false?

src/sotauptaneclient.cc
OYTIS: The point here is to be able to quickly find a secondary by its serial number.

src/uptane/virtualsecondary.cc
OYTIS: No, false is reserved for connection failures, which don't happen in virtual secondaries.

src/uptane/managedsecondary.h
patrickvacek: You are right, probably should be a pure virtual function. That was leftover cruft.

cajun-rat:
When building this in meta-updater, the device doesn't register correctly in the UI.  I'll do some investigation...

patrickvacek:
@cajun-rat Frankly I'm not surprised that there are still some problems. Although do you mean even the primary doesn't show up?

cajun-rat:
Yeah, no primary and no packages

cajun-rat:
Are you able to make this work against ATS Garage? I still haven't got a device to provision correctly in the GUI

patrickvacek:
@cajun-rat No, still working on it. I think I found the problem but still sorting out the details. But I think I resolved the one issue @patriotyk brought up that @OYTIS didn't already answer.

cajun-rat:
Tested in with an update to Qemu inside meta-updater

PR #417, opened 2017-11-27
Enforce C++98/03 where it matters.

patrickvacek:
* Do not tolerate GNU or other extensions.
* Build sota_tools with C++11.
* Revert 8aee6ac8e9dbef2e2dd290f4e74c2872c502177e and instead build aktualizr-validate-secondary-interface with C++11.
* Do not link with boost regex since we don't use it.


cajun-rat:
I'm happy with the this PR, subject to CI passing

patrickvacek:
Why is this decreasing code coverage so much? Oh well.

PR #419, opened 2017-11-27
Move updateSecondaries function to SotaUptaneClient

OYTIS:
The client now appends operation_result to every manifest, backend's reaction should be tested. If it's OK with it, the second call to putManifest() can be removed.

### DIFF
--- a/src/sotauptaneclient.cc
+++ b/src/sotauptaneclient.cc
@@ -119,8 +119,7 @@ data::InstallOutcome SotaUptaneClient::OstreeInstall(const Uptane::Target &targe
   }
 }
 
-Json::Value SotaUptaneClient::OstreeInstallAndManifest(const Uptane::Target &target) {
-  Json::Value operation_result;
+void SotaUptaneClient::OstreeInstallAndManifest(const Uptane::Target &target) {
   if (isInstalled(target)) {
     data::InstallOutcome outcome(data::ALREADY_PROCESSED, "Package already installed");
     operation_result["operation_result"] = data::OperationResult::fromOutcome(target.filename(), outcome).toJson();
@@ -134,17 +133,6 @@ Json::Value SotaUptaneClient::OstreeInstallAndManifest(const Uptane::Target &tar
       uptane_repo.saveInstalledVersion(target);
     }
   }
-  Json::Value unsigned_ecu_version =
-      OstreePackage(target.filename(), target.sha256Hash(), "").toEcuVersion(target.ecu_identifier(), operation_result);
-
-  ENGINE *crypto_engine = NULL;
-#ifdef BUILD_P11
-  if ((uptane_repo.key_source == kPkcs11)) crypto_engine = uptane_repo.p11.getEngine();
-#endif
-
-  Json::Value ecu_version_signed = Crypto::signTuf(crypto_engine, uptane_repo.primary_private_key,
-                                                   uptane_repo.primary_public_key_id, unsigned_ecu_version);
-  return ecu_version_signed;
 }
 
 void SotaUptaneClient::reportHwInfo() {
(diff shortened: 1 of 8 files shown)


src/sotauptaneclient.cc
```diff
@@ -249,3 +236,34 @@ void SotaUptaneClient::runForever(command::Channel *commands_channel) {
...
+      if (!boost::filesystem::exists(firmware_path)) continue;
+
+      if (!sec->second->putMetadata(meta)) {
+        // connection error while putting metadata, can't upload firmware
+        continue;
+      }
+
+      std::string firmware = Utils::readFile(firmware_path);
+
+      sec->second->sendFirmware((const uint8_t *)firmware.c_str(), firmware.size());
```
patrickvacek: What's the benefit of using `uint8_t`? Currently (for `ManagedSecondary`), we're going file -> `string` -> C-string -> `uint8_t` -> C-string -> `string` -> file. I know that we may want to rethink that anyway because of large file sizes, but I think we can do without some of these layers of casts. Or we should just read into a binary stream with something like `fopen`.

src/sotauptaneclient.cc
```diff
@@ -119,8 +119,7 @@ data::InstallOutcome SotaUptaneClient::OstreeInstall(const Uptane::Target &targe
   }
 }
 
-Json::Value SotaUptaneClient::OstreeInstallAndManifest(const Uptane::Target &target) {
-  Json::Value operation_result;
+void SotaUptaneClient::OstreeInstallAndManifest(const Uptane::Target &target) {
```
patrickvacek: This function name seems like a misnomer at this point. This is just like a pre-install step or status check before and after `OstreeInstall` now.

src/sotauptaneclient.cc
```diff
@@ -165,7 +153,7 @@ Json::Value SotaUptaneClient::AssembleManifest() {
     (refname += "unknown-") += hash;
   }
   Json::Value unsigned_ecu_version =
-      OstreePackage(refname, hash, "").toEcuVersion(uptane_repo.getPrimaryEcuSerial(), Json::nullValue);
+      OstreePackage(refname, hash, "").toEcuVersion(uptane_repo.getPrimaryEcuSerial(), operation_result);
 
   result.append(uptane_repo.getCurrentVersionManifests(unsigned_ecu_version));
```
patrickvacek: This also seems like a misnomer (even if it isn't your fault or related specifically to this PR!). It is just getting one manifest, namely for the primary. The name implies that it is getting multiple (presumably from the secondaries), but that is in fact done by this function (`AssembleManifest`). This is the only place that `getCurrentVersionManifests` is called.

patrickvacek:
Looks good to me other than my minor trifles.

src/sotauptaneclient.cc
OYTIS: Yes, maybe the interface should be just `sendFirmware(std::string data)`

src/sotauptaneclient.cc
OYTIS: It mostly sets operation result now. `OstreeInstallSetResult`?

src/sotauptaneclient.cc
OYTIS: In fact `getCurrentVersionManifests` doesn't even get anything, it just signs its argument.

src/sotauptaneclient.cc
patrickvacek: Sure. Or maybe `SetOstreeInstallResult` to put the verb first?

OYTIS: Well, it doesn't just set result, it makes OstreeInstall first.

PR #420, opened 2017-11-28
Add targets with garage-sign

patriotyk:


src/sota_tools/deploy.cc
```diff
@@ -168,24 +168,34 @@ bool copy_repo(const std::string &cacerts, const std::string &src, const std::st
   if (errors) {
     LOG_ERROR << "One or more errors while pushing";
     return EXIT_FAILURE;
-  } else {
-    return EXIT_SUCCESS;
   }
 
+  if (sign) {
+    sign_repo(dst, ostree_ref, hardwareids);
+  }
```
Danten: Is there a reason why `copy_repo` calls `sign_repo`?

src/sota_tools/deploy.cc
```diff
@@ -168,24 +168,34 @@ bool copy_repo(const std::string &cacerts, const std::string &src, const std::st
...
       throw std::runtime_error("Could not initilaize tuf repo for sign");
     }
   }
+
   if (system("garage-sign targets  pull --repo aktualizr") != 0) {
     throw std::runtime_error("Could not pull targets");
   }
 
+  std::string cmd(
+      "garage-sign targets  add --repo aktualizr --format OSTREE --length 0 --url \"https://example.com/\"");
```
Danten: Extra space between `targets` and `add`.

patriotyk: Ok

src/sota_tools/deploy.cc
patriotyk: It is because `ostree_ref` argument is creating inside of `copy_repo` and it contains branch name and hash calculation function that is needed inside sign_repo.  Such design doesn't allow to do that better.

Danten: Ah right, I see now that `ref` is of a different type in either cases.

cajun-rat:
Tested with a SW update in qemu

PR #422, opened 2017-11-28
Bugfix/secondarymetapath

OYTIS:
Secondaries are very close to working e2e now, I'll check that again tomorrow, when I (hopefully) get garage-sign working again.


OYTIS:
OK, doesn't work, still reporting hash for empty string instead of actual image.

OYTIS:
OK, now virtual secondaries work end-to-end (modulo serial number generation).

PR #423, opened 2017-11-29
Add vscode and Libreoffice files to .gitignore

cajun-rat:
Change-Id: I5f461befcf0d4352211f422ac60b1cc3db5509aa


patriotyk:
I think it is not possible because aktualizr will not compile on windows.

cajun-rat:
I'm using vscode, which is cross platform and based on atom/electron.  It is actually pretty good, and I've been using https://github.com/jacobdufault/cquery for C++ code indexing

patrickvacek:
Fascinating. I had no idea. Obviously I have no problem appending to gitignores.

PR #424, opened 2017-11-29
Move *.toml config files from meta-updater to aktualizr

OYTIS:
This will let us make non backwards compatible changes to configs


patrickvacek:
Wait, why haven't we been doing this the whole time?

OYTIS:
Well, the primary reason is because it already works the old way. But now I really want to introduce some breaking changes in the hope to make config a bit cleaner.

OYTIS:
What happened to codecov by the way?

PR #425, opened 2017-11-29
Feat/pro 3792/legacy init

patrickvacek:
This isn't really fully tested yet, but it does pass all the unit tests!

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -394,18 +394,18 @@ void Config::updateFromCommandLine(const boost::program_options::variables_map& 
       if (rs != 0) {
         LOGGER_LOG(LVL_error, "Legacy external flasher list-ecus command failed: " << output);
       } else {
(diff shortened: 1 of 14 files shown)


patrickvacek:
This doesn't actually work in practice yet. I'm looking into why.

patrickvacek:
Cool, this works now! To test, you'll need to copy the myecuinterface built incidentally by bitbaking aktualizr onto the device. We should think about formalizing that. We also need to think more about unit testing this code.

cajun-rat:
I'm going to do a qemu test run then should be good to go

patrickvacek:
Updated with some fixed/improved tests.

patrickvacek:
@cajun-rat  this is theoretically good to go now... if it still works for you.

cajun-rat:
Works on My Machine™

PR #426, opened 2017-11-30
Disable refs push if server supports offline signing

patriotyk:


src/sota_tools/deploy.cc
```diff
@@ -77,15 +77,15 @@ bool copy_repo(const std::string &cacerts, const std::string &src, const std::st
       fetch_server.ca_certs(cacerts);
     } else {
       LOG_FATAL << "--cacert path " << cacerts << " does not exist";
-      return EXIT_FAILURE;
+      return false;
```
cajun-rat: Good spot

src/sota_tools/authenticate.cc
```diff
@@ -8,142 +10,24 @@
 #include <iostream>
```
cajun-rat: I think these includes are no-longer needed--can you remove them, please?

src/sota_tools/deploy.cc
```diff
@@ -67,7 +67,7 @@ void uploaded_ev(RequestPool &p, OSTreeObject::ptr h) {
 }
 
 bool copy_repo(const std::string &cacerts, const std::string &src, const std::string &dst, const std::string &ref,
-               const std::string &hardwareids, bool sign, bool dryrun) {
+               const std::string &hardwareids, bool sign, bool is_offline_signed, bool dryrun) {
```
cajun-rat: I don't think `is_offline_signed` is necessary: Whether to use offline signing can be determined by ServerCredentials::CanSignOffline(), and there isn't (to my knowledge) a reason for the user to override that choice.

src/sota_tools/deploy.cc
```diff
@@ -143,37 +144,40 @@ bool copy_repo(const std::string &cacerts, const std::string &src, const std::st
...
 
   if (errors) {
     LOG_ERROR << "One or more errors while pushing";
-    return EXIT_FAILURE;
+    return false;
   }
 
   if (sign) {
     sign_repo(dst, ostree_ref, hardwareids);
   }
```
cajun-rat: I'd be happier if sign_repo() was the `else` of the ìf` above that calls `ostree_ref.PushRef`

The error handling here is also done in two different ways. `sign_repo()` uses `std::runtime_error` exceptions to communicate failure, but the rest of the nearby code is using error codes.  I would prefer if we changed sign_repo() to return an error code (and use the `errors` local variable).  (In keeping with "exceptions should be exceptional")

src/sota_tools/deploy.cc
```diff
@@ -143,37 +144,40 @@ bool copy_repo(const std::string &cacerts, const std::string &src, const std::st
...
-  return EXIT_SUCCESS;
+  return true;
```
cajun-rat: Good

src/sota_tools/garage_push.cc
```diff
@@ -31,6 +31,7 @@ int main(int argc, char **argv) {
     ("help", "produce a help message")
     ("verbose,v", accumulator<int>(&verbosity), "Verbose logging (use twice for more information)")
     ("quiet,q", "Quiet mode")
+    ("is-offline-signed", "Skips pushing refs to the repo if the server supports offline signing")
```
cajun-rat: (I think this isn't needed)

src/sota_tools/server_credentials.h
```diff
@@ -0,0 +1,42 @@
...
+        std::string client_key_;
+        std::string root_cert_;
+        std::string repo_url_;
+        std::string auth_user_;
+        std::string auth_password_;
+        std::string auth_server_;
+        std::string ostree_server_;
+        std::string client_id_;
+        std::string client_secret_;
+        bool can_sign_offline_;
```
cajun-rat: It would be better to calculate this in the CanSignOffline() method, rather than cache it from the ctor.  That way there will never be any problems keeping the can_sign_offline_ member variable correctly updated, for example if someone adds some new code paths through the constructor is forgets to set can_sign_offline_ (or adds code after that point that sets variables that can_sign_offline_ depends on.

cajun-rat:
Overall this is good, and I like the refactor.  Can you polish it up in a few places, please, and it'll be good to go?

src/sota_tools/deploy.cc
patriotyk: Done, but we need sign flag for garage-push so just added check for push_creds.CanSignOffline()

src/sota_tools/server_credentials.h
patriotyk: I think it is better to cover that with tests.

patriotyk: I have submitted tests that tests  CanSignOffline functionality. But If You assure we really need to move that functionality I will do that. Anyway the tests is useful here.

cajun-rat: Tests are great. I'm definitely keen on keeping those!  It is really just the redundant information that we are storing that I'm worried about.  If we just calculate `CanSignOffline` every time the answer is needed, then there is no chance that the two can get out of sync.

patriotyk: Ok. done

patrickvacek:
@patriotyk So @Danten and I are seeing a problem where garage-deploy isn't finding the requested ref in the source repo. We can see it there, though. Any ideas?

src/sota_tools/deploy.cc
```diff
@@ -94,18 +94,19 @@ bool copy_repo(const std::string &cacerts, const std::string &src, const std::st
 
   if (!src_repo->LooksValid()) {
     LOG_FATAL << "The OSTree src repo does not appear to contain a valid OSTree repository";
-    return EXIT_FAILURE;
+    return false;
   }
 
   OSTreeRef ostree_ref(src_repo->GetRef(ref));
   if (!ostree_ref.IsValid()) {
     LOG_FATAL << "Ref " << ref << " was not found in repository " << src;
```
patrickvacek: `src` is just the name of the zipped credentials file; this should probably print the repo name or location instead.

patriotyk:
It is ready.

src/sota_tools/deploy.cc
patriotyk: No src it is folder or url.

patriotyk: I just veryfied and everything works for me. Are You sure You pass the exists refname?

patrickvacek: It does work for me, but @Danten was trying to put this into an end-to-end test and saw the above problem. I checked with him that everything was where we expected it to be.

cajun-rat:
I've tested this end-to-end using an account on the prod deployment for the initial upload, then using garage-deploy to send it to an account on staging (that does offline signing).

PR #427, opened 2017-11-30
Refactor config to isolate storage settings into StorageConfig

OYTIS:
The most interesting part is in src/config.h, everything else is more or less adaptation to these changes. This refactoring will hopefully make migration to SQLite a bit easier.
Downloading targets in tufrepository.cc still needs more work.

Final version needs to be shown to the people working on e2e tests before merging.


OYTIS:
FIX: regexp magic in implicit_writer_test.py

patrickvacek:
This seems mostly fine to me so far (although I haven't actually tested yet). However, I think this will break implicit_writer, which depends on being able to write absolute paths to the config, which this change breaks support for. I know the getXyzPath() functions are ugly, but it seems to be useful for many cases to have some way to intelligently and consistently produce an absolute path. Traditionally we used directory + filename, but we should also support complete path stored in the filename and skipping the directory.

config/basic.toml
```diff
@@ -1,12 +1,11 @@
-[tls]
-certificates_directory = "sota-prov"
+[storage]
+path = "sota-prov"
```
patriotyk: Nice, because certificates_directory is confusing

OYTIS:
@patrickvacek As a second thought, I'd prefer not to have these functions (ca_file() etc.) in the StorageConfig class. The idea here was that all files of FSStorage should live under config.storage.path directory. If we want to support credentials that are stored on arbitrary (possibly read-only) locations, we need to import them. For FSStorage it looks like a mere annoyance, but it's what we'll have to do anyway once aktualizr swithes to using sqlite for storing stuff.

tests/uptane_vector_tests.cc
```diff
@@ -40,14 +40,15 @@ bool run_test(const std::string& test_name, const Json::Value& vector, const std
...
   config.uptane.director_server = "http://127.0.0.1:" + port + "/" + test_name + "/director";
   config.uptane.repo_server = "http://127.0.0.1:" + port + "/" + test_name + "/image_repo";
-  config.uptane.metadata_path = uptane_vector_test_dir + "/" + port + "/aktualizr_repos";
+  config.storage.path = uptane_vector_test_dir;
+  config.storage.uptane_metadata_path =  port + "/aktualizr_repos";
   config.ostree.os = "myos";
   config.ostree.sysroot = "./sysroot";
-  boost::filesystem::remove_all(config.uptane.metadata_path / "director");
-  boost::filesystem::remove_all(config.uptane.metadata_path / "repo");
+  boost::filesystem::remove_all(config.storage.path / config.storage.uptane_metadata_path / "director");
```
patriotyk: What about add getAbsoluteMetadataPath()?

patrickvacek:
@OYTIS I can see why putting the functions in StorageConfig would be ugly. Is putting them in FSStorage viable? If instead we say that in general that you simply have to install files into the specified directory, that's probably fine, although having the option to put files elsewhere is kind of nice. More important, though, is what to do with implicit_writer. Currently it writes the root CA to `/usr/lib/sota/root.crt` despite that `certificates_directory = "/var/sota/"`. If I recall correctly, we wanted to have the root CA be something that was baked into the image and updateable via bitbake, which makes `/var/` untenable.

OYTIS:
@patrickvacek Yes, true. I think I'll have to add importing logic right away.

tests/uptane_vector_tests.cc
OYTIS: Yes, you're right, I'll add some helper functions.

src/config.cc
```diff
@@ -477,8 +483,16 @@ void Config::writeToFile(const std::string& filename) {
   writeOption(sink, storage.uptane_metadata_path, "uptane_metadata_path");
   writeOption(sink, storage.uptane_private_key_path, "uptane_private_key_path");
   writeOption(sink, storage.uptane_public_key_path, "uptane_public_key_path");
-  writeOption(sink, storage.tls_cacert_path, "tls_ca_path");
+  writeOption(sink, storage.tls_cacert_path, "tls_cacert_path");
```
patrickvacek: +1 for the more accurate name change!

patrickvacek:
I'm testing this with implicit provisioning (with an HSM) but my device isn't provisioning. Not quite sure why. I haven't tried with softhsm yet.

OYTIS:
@patrickvacek Just tried it with SoftHSM. Provisions fine for me.

patrickvacek:
Turns out it was just a naming issue with sota.toml vs what cert_provider uses. I'll fix cert_provider in a separate PR. This looks fine.

OYTIS:
Now waiting for https://github.com/advancedtelematic/meta-updater/pull/192 to be merged in order not to break things.

PR #431, opened 2017-12-05
Improve test coverage

cajun-rat:
See #429 for the original PR.  This is a rebase of that, which we had to revert because of a conflict with #b51749304c


cajun-rat:
@patriotyk Thank you for this fix.  It is nice to see a PR with code coverage going up :)

PR #436, opened 2017-12-06
Preliminary work on SQLStorage

OYTIS:
Creates a DB from a script, then checks it against a schema, no migration logic yet.


patrickvacek:
I don't understand codecov. It only shows green blocks and yet it claims coverage will go down. What gives?

More importantly, building is failing for me:
```
/home/patrick/Code/pro/client/aktualizr/src/sqlstorage.h:8:21: fatal error: sqlite3.h: No such file or directory
 #include <sqlite3.h>
```
Not sure why `FindSQLite3.cmake` isn't figuring that out automatically.

OYTIS:
@patrickvacek I suspect sqlstorage.cc is green, because it is a new file, that is previously the coverage was 0% (although one could argue that it was 100%) and now it's 24.8% which is a positive change. The overall coverage has decreased, because sqlstorage.cc has a lot of code copied and pasted from fsstorage.cc, but for the moment it is not tested separately. I want to refactor *storage tests so that there is a set of implementation independent test cases.

PR #439, opened 2017-12-07
Add virtual secondary config example

OYTIS:
Can be used for testing


OYTIS:
@patrickvacek Yes, secondary_serial.json is probably redundant.

PR #440, opened 2017-12-07
 Add flag to create directories for writeFile

patriotyk:


patrickvacek:
Looks good, but the next step is to actually make use of the option throughout the code base! I know there are many places where we check if a directory exists and make it if it doesn't already, and that can now be simplified with this option.

patriotyk:
Do You suggest me to do that in this PR?

patrickvacek:
@patriotyk Yeah, that was my intention, although admittedly I didn't make that clear when I wrote the ticket. Just do a quick search for places that call writeFile and see if they can benefit from the new option.

patriotyk:
Ok

patriotyk:
Done, but there is no so many places. I found only two.

PR #441, opened 2017-12-08
Tolerate (ignore) whitespace in api-version output.

patrickvacek:
Just in case someone ever included it.

This is also just a bit of a refactor to clean up that section of the code.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -7,6 +7,7 @@
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
 #include <boost/algorithm/string/trim.hpp>
+#include <boost/filesystem.hpp>
 
 #include "bootstrap.h"
 #include "utils.h"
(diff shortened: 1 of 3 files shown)


src/config.cc
```diff
@@ -341,97 +343,124 @@ void Config::updateFromCommandLine(const boost::program_options::variables_map&
...
+    return ret;
+  }
+  std::stringstream command;
+  std::string output;
+  command << legacy_interface << " api-version --loglevel " << loggerGetSeverity();
+  int rs = Utils::shell(command.str(), &output);
+  if (rs != 0) {
+    LOGGER_LOG(LVL_error, "Legacy external flasher api-version command failed: " << output);
+  } else {
+    boost::trim_if(output, boost::is_any_of(", \n\r\t"));
```
patriotyk: Are You sure we need to remove comma?

patrickvacek: Need, no. But I figured it might be useful to catch that. Do you think that's likely to be a bad idea?

PR #444, opened 2017-12-11
Bugfix/miscellaneous

patrickvacek:
Assorted minor things that I caught/fixed while looking for things to add to PRO-4470.


patrickvacek:
I'm amused that I've apparently confused GitHub by ordering my commits non-chronologically. Travis CI is running on the correct version, but they are ordered incorrectly in the list here. `
7216710` should come before `1efcb05`.

PR #445, opened 2017-12-11
Migration logic and store device_id in sqlite db

OYTIS:
Need to clean excessive 'NOT NULL's


tests/sqlstorage_test.cc
```diff
@@ -6,15 +6,15 @@
 
 const boost::filesystem::path sqlstorage_test_dir = "tests/test_sqlstorage";
 
-TEST(sqlstorage, migrate0) {
+TEST(sqlstorage, migrate) {
   StorageConfig config;
   config.path = sqlstorage_test_dir;
-  config.schema_version = 0;
-  config.sqldb_path = "config/test.db";
+  config.sqldb_path = "tests/test_sqlstorage/test.db";
```
cajun-rat: This PR looks good overall, but can you make this test write in a directory under the build directory, rather than create files in the project tree?

OYTIS: @cajun-rat 
It's just how other tests (fsstorage, uptane) behave to my understanding, but it makes sense to chacnge that.

patrickvacek: For reference, I made ticket PRO-4484 for exactly that.

PR #446, opened 2017-12-12
Use boost paths and fix problems with subdirectories.

patrickvacek:
For now, just ignore subdirectories.

This is unintentionally sort of related to PRO-4525.


patrickvacek:
FYI all: it looks like debian-unstable is transitioning away from libssl1.0-dev and libp11 is currently in an unusable state. Hooray!

patrickvacek:
Merging despite the Debian unstable problems.

PR #447, opened 2017-12-12
Replace more tests with new RAII Directory.

patriotyk:


patriotyk:
Ok I see. I am removing  this PR

PR #451, opened 2017-12-14
Document credentials.zip file format

cajun-rat:


docs/credentials.adoc
```diff
@@ -1,16 +1,130 @@
...
 
-To authenticate on the ostree server, aktualizr uses cerdentials.zip archive.
+The credentials.zip file provides the link between a user's on-line account on ATS Garage or OTA+ deployment and developer's machine. It is used for:
 
-The minimal requirement for this archive is including `treehub.json` file with
-the following content:
+* Uploading the built OSTree image to treehub with garage-push
+* Offline-signing Uptane images targets metadata with garage-sign
+* Provisioning the device with Aktualizr or aktualizr_cert_provider
+* Providing the URL to the update server to Aktualizr (this is templated into sota.toml at build time, to allow software updates to modify that URL)
```
patrickvacek: Can be templated. Currently this is only done for implicit/HSM provisioning. Autoprov just gets it from the zip.

docs/credentials.adoc
```diff
@@ -1,16 +1,130 @@
...
+The following files are present in credentials.zip:
+
+[options="header"]
+|======================
+| Filename in zip | Purpose | Used by
+| treehub.json | Location and authentications for treehub and Uptane repo | garage-sign, garage-push
+| client.crt | Certificate for TLS client authentication | Aktualizr 
+| client.key | Private key for TLS client authentication | Aktualizr
+| root.crt | Root CA for TLS client authentication | Aktualizr
+| repo.url
```
patrickvacek: This is what I see:
```
tufrepo.url
autoprov.url
autoprov_credentials.p12
root.json
targets.pub
targets.sec
treehub.json
```
`client.crt`, `client.key`, and `root.crt` are all part of `autoprov_credentials.p12`.

docs/credentials.adoc
```diff
@@ -1,16 +1,130 @@
...
+Multiple types of authentication are supported, and more types may be added in the future.  Clients should ignore unrecognised authentication types, and the server may add multiple authentication types to support both old and new clients.
+
+=== No authentication (e.g. community edition)
+
+```
+{
+  "ostree": {
+    "server": "https://localhost:2443/api/v3"
+  },
+  "no_auth" : {},
```
simao: Why is this an empty object I didn't get this in the end ?

docs/credentials.adoc
```diff
@@ -1,16 +1,130 @@
...
+}
+```
+
+=== TLS Client Certificate authentication
+```
+{
+  "ostree": {
+    "server": "https://localhost:2443/api/v3"
+  },
+  "certificate_auth": True,
```
simao: `true`

docs/credentials.adoc
cajun-rat: I was thinking that if we need to add any options, then we can put something in that object

docs/credentials.adoc
cajun-rat: true.

docs/credentials.adoc
cajun-rat: Would be happy with anything else though.

cajun-rat:
Is this version OK with everyone?

docs/credentials.adoc
```diff
@@ -1,16 +1,135 @@
...
+| treehub.json | Location and authentications for treehub and Uptane repo | garage-sign, garage-push
+| client.crt | Certificate for TLS client authentication | garage-push
+| client.key | Private key for TLS client authentication | garage-push
+| autoprov_credentials.p12 | TLS client credentials for automatic device provisioning | aktualizr, aktualizr_cert_provider
+| autoprov.url | URL for automatic provisioning server | aktualizr, aktualizr_cert_provider
+| root.json | Initial Uptane root.json (for secure bootstrapping) | garage-sign
+| targets.pub | Public key for offline Uptane image signing | garage-sign
+| targets.sec | Private key for offline Uptane image signing | garage-sign
+| tufrepo.url | URL to Uptane repository | garage-sign
+| root.crt | Root CA for TLS client authentication | aktualizr
```
patrickvacek: Is there ever a time that we have `client.crt`, `client.key`, or `root.crt` separate from `autoprov_credentials.p12`?

patrickvacek: Currently implicit_writer uses the `root.crt` extracted from `autoprov_credentials.p12` as well as `autoprov.url`.

cajun-rat: I think that `client.crt` `client.key` and `root.crt` are only used by garage-push and only for `certificate_auth`

patrickvacek: Yeah, but my point is just that the are actually just contained with the p12 file, they aren't included separately.

docs/credentials.adoc
simao: I think this should be either true or false? I don't think there is anything that could be added that would make no_auth be less confusing ?

cajun-rat: `true` works for me

docs/credentials.adoc
cajun-rat: (Patrick and I talked about this face to face)

patrickvacek:
I think this is good other than that I'd like to argue for implicit_writer being listed as a user of the zip.

cajun-rat:
Good point. Now added

PR #452, opened 2017-12-14
Do not try to unpack boostrap credentials when no-root is specified

OYTIS:
Required to support _real_ implicit provisioning

### DIFF
--- a/config/sota_implicit_prov.toml
+++ b/config/sota_implicit_prov.toml
@@ -7,6 +7,3 @@ tls_pkey_path = "pkey.pem"
 uptane_metadata_path = "metadata"
 uptane_private_key_path = "ecukey.der"
 uptane_public_key_path = "ecukey.pub"
-
-[import]
-tls_cacert_path = "/usr/lib/sota/root.crt"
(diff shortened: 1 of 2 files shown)


cajun-rat:
Shall we sit down tomorrow and work through how this should work?

OYTIS:
@patrickvacek @cajun-rat 
We've agreed with @jerrytrieu on credentials.zip format that contains everything that 'normal' credentials.zip has except autoprov.p12. `Bootstrap::readServerUrl` just needs autoprov.url, which is there for implicit provisioning, but `Bootstrap::Bootstrap()` looks for autoprov.p12 qute naturally and crashes.

patrickvacek:
Yes, I'm all too familiar with that crash! We should consider some refactoring to make it actually work correctly.

OYTIS:
@patrickvacek Oops, you're right, it crashes anyway. Let's discuss that tomorrow.

OYTIS:
@patrickvacek Oh no, it works, I just forgot to add `--no-root-ca` to the command line. I will use this branch to test Spekulatius for now and we can come up with something better tomorrow.

patrickvacek:
@OYTIS Right you are. Still seems like a problem for the case without `--no-root-ca` though. Didn't we want to be able to push new root CAs as part of the image?

patrickvacek:
@OYTIS I made a small change that makes things a bit cleaner. I thought this ought to work smoothly for all cases.

PR #454, opened 2017-12-15
Update provisioning documentation.

patrickvacek:
This is a way more accurate description of the current state of affairs.
I removed a lot of outdated information. There's probably plenty more
that could be said but at least this is a start.


docs/implicit-provisioning.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+An example `.toml` configuration file can be found at `config/sota_implicit_prov.toml`. This is what is used as in input to `implicit_writer` during bitbaking with meta-updater.
+
+## Steps
+
+1. In meta-updater, set `SOTA_CLIENT_PROV = "aktualizr-implicit-prov"`. Currently, you will also need to set `SOTA_PACKED_CREDENTIALS` to provisioning credentials zip file.
+1. Build a standard image using bitbake.
+1. Boot the image.
+1. Optionally, verify that Aktualizr is not provisioning and that the device is not visible in the Garage.
+1. Load credentials onto the device, for example with `cert_provider` from the Aktualizr repo: `cert_provider -c credentials.zip -t <device> -d /var/sota`
```
OYTIS: It still looks like a description of hacky implicit provisioning: in the production case cert_provider is not used.

patrickvacek: What would you like to see to make it less hacky? After all, `cert_provider` is the only thing that works right now other than manually copying, which was the original "hacky" name referred to.

OYTIS: Real implicit provisioning should be already enabled at least on CI, it might be time (next sprint probably) to create a manual on how to create devices for it with and without HSM. Until then I'm fine with keeping instructions for cert_provider in place, but IMO some note telling that this is not how implicit provisioning really works might be helpful.

patrickvacek: > Real implicit provisioning

Do you mean using an HSM? I didn't write a document for that case, although I certainly could. This document was just meant to describe how to make use of the `aktualizr-implicit-prov` recipe.

Either way, you are probably right that some better context of the intended use case is a good idea.

OYTIS: Either HSM or storing pkey/ca/cert on the file system, but without using autoprovisioning. I mean the "normal" (at least it was our plan) process of creating root keypair, uploading public key to the backend and signing all the devices' individual certificates with the private key. It's what our testing people emulate as far as I understand.

patrickvacek:
@OYTIS I've revised the document substantially with the intended workflow. The `cert_provider` bit is now basically a footnote. I don't actually know how to upload the root CA to the server, though, and so that part is a bit vague. I still didn't mention HSMs at all, but does this look better for now?

PR #456, opened 2017-12-15
Add Anton's secondaries class diagram

cajun-rat:
Change-Id: Ib19c107ef13a2d52302ff0c52139a7eb934bb50d


patrickvacek:
Two minor suggestions: I ended up naming `ISecondary` as `SecondaryInterface` since that seemed closer to the Google recommendations, and I think instead of `EcuInterface` it'd be better to say `example_interface`.

PR #457, opened 2017-12-18
Clean generated dirs after tests

patriotyk:


patrickvacek:
The `run_vector_tests.sh` cleanup is good. Thanks! However, there is another output directory that I'm even more concerned about, namely (in the build directory): `./tests/tests/test_uptane_vector/`, in which a subdirectory gets created and named after the port used for running some of the tests.

Regarding `uptane_test`: I did some investigation yesterday and found the source of some of the problems there. I have a branch with a more complete fix for that code. Your catch-all code fixes the problem, but I think the more fine-tuned approach is better.

patriotyk:
Ok, reassigned PRO-4488 to You

patrickvacek:
All fixed up. @patriotyk Look okay to you?

PR #458, opened 2017-12-18
Upload garage-deploy.deb to github releases

patriotyk:


patrickvacek:
This is pretty cool. Is that api_key okay to have in the open? And do we want to create a tagged release every time we push something? I think we should either name the tags (to clean up the `untagged-xxx` nomenclature) or only upload when specifically want to (to prevent proliferation of redundant releases). Thoughts?

patriotyk:
Currently It produces deb file for each commit in master but It could be changed to do that only for tagged commits.

patriotyk:
Just changed to make deb packages for tagged commits only. And about secret key, documentation confirms that it is secure because key is encrypted

OYTIS:
Not sure what the value of GITHUB_TOKEN was, created GITHUB_API_KEY just in case.

patriotyk:
GITHUB_TOKEN it is my key that I tried to use in travis file, but this did not work for me.

patrickvacek:
What is the status here? Where does `GITHUB_API_KEY` come from?

@patriotyk I'm still not convinced that your original usage of your encrypted key is a good idea. Doesn't that mean that anyone can use your key to push arbitrary things in your name to github?

patriotyk:
This PR has been updated and doesn't contains any keys, instead it uses GITHUB_API_KEY variable which was added in settings page of travis - https://travis-ci.org/advancedtelematic/aktualizr/settings

PR #464, opened 2017-12-22
Add example UDS legacy flasher and the counterpart firmware

OYTIS:


OYTIS:
Probably codecov has problems with large diffs. Merging anyway.

patrickvacek:
@OYTIS I know this is already merged, but I still have some thoughts:

+1 on the external_secondaries reorganization. However, there was a problem with getting CMake to do anything without loading the submodules for iso-tp, which I think is not ideal. I made https://github.com/advancedtelematic/aktualizr/pull/470 to address that.

Regardless of that, though, is there a way that we can get the `deps/bitfield-c` submodule to get automatically loaded when we load the `isotp-c` submodule? It's annoying to have to `git submodule init` and `update` in the root directory and then move to the `isotp-c` project and repeat.

It looks like I broke the Travis CI build by trying to restart it. It's missing a ref now for some reason. I had even held back on my normal urge to immediately delete merged branches, but maybe something else I deleted is causing the problem now.

I know you've explained this before, but what is the meaning of the "partial" directory? Either way, seems like we might want it to be coverage by clang-format and everything else.

patrickvacek:
Sorry, I'd forgotten about `git submodule update --init --recursive`.

PR #465, opened 2017-12-25
Use boost::path instead of std::string

patriotyk:


patrickvacek:
@patriotyk I like the look of this so far, but it looks like there's a build failure relating to something with p11.

patrickvacek:
More places to get boostified:

1. `firmware_path` in `src/external_secondaries/interface_main.cc`
2. `path` in `src/ostree.cc`/`src/ostree.h` (maybe)
3. `objpath` in `src/sota_tools/ostree_dir_repo.cc` (maybe)
4. `root_path` in `src/sota_tools/ostree_dir_repo.h`
5. `credentials_path` in `src/sota_tools/server_credentials.cc`
6. `fetch_cred` and `push_cred` in `src/sota_tools/garage_deploy.cc`
7. several things with `path` in their name in `src/sota_tools/garage_check.cc` and `src/sota_tools/garage_push.cc`
8. `credentials_path_` in `src/sota_tools/server_credentials.h`
9. `legacy-interface` in `src/main.cc`
10. `path` (of `PublicKey`) in `src/crypto.h` (maybe)
11. several things, including four input variables, in `src/implicit_writer/main.cc`
12. `credentials` and `config` in `src/cert_provider/main.cc`

That's just what I found with some quick investigation with `grep path src/ -riIn | grep string`.

patriotyk:
1) Done
2) Done
3) Done
4) Done
5) Done
6) Done
7) Done
8) Done
9) Done
10) Done
11) Done
12) Done

patrickvacek:
Looking good so far, thanks! Apparently there is now a merge conflict, though. And my other comments just disappeared....

patrickvacek:
Strange about my other comments. Well, basically I made two further recommendations:

1. There are still some paths being passed to functions in Config that should be paths and not strings (Config constructor, `checkLegacyVersion`, `initLegacySecondaries`, `readSecondaryConfigs`, `writeToFile`).
2. Is using `boost::filesystem::path.string().c_str()` necessary? path has it's own c_str: http://www.boost.org/doc/libs/1_53_0/libs/filesystem/doc/reference.html#c_str We should remove the extraneous intermediate step if possible. I think we do that in many places.

patriotyk:
1) Ok I am working on it,
2) Yes if boost's path has it's own c_str() that returns char* we can use it.

patriotyk:
Now, it should be ok

patriotyk:
Fixed builds

PR #467, opened 2017-12-26
Implemented is_registered flag in sqlite backend

patriotyk:


patrickvacek:
This looks fine; any idea what's up with the Travis CI error?

patriotyk:
I think we need ignore that. I committed and pushed changes to the master, then I fixed it by git reset. I think some of that manipulations caused the error.

PR #469, opened 2017-12-27
Use absolutePath in fsstorage class

patriotyk:


src/utils.cc
```diff
@@ -322,6 +322,14 @@ int Utils::shell(const std::string &command, std::string *output) {
   return WEXITSTATUS(exitcode);
 }
 
+boost::filesystem::path Utils::absolutePath(const boost::filesystem::path &root, const boost::filesystem::path &file) {
+  if (file.is_absolute() || root.empty()) {
+    return root;
```
patrickvacek: I think you mean `return file`!

patrickvacek:
I think this might be slightly backwards -- I think fsstorage is probably fine as it is, but there are plenty of places with the Config and SecondaryConfig classes where this is more of an issue, like the `StorageConfig.path` along with the `uptane_metadata_path`, etc. This might make the `ImportConfig` redundant, unless that has another use that I don't know of.

src/utils.cc
patriotyk: Yes, You are right.

patriotyk:
I think You can merge this. All places where we concatenate path with uptane_metadata_path have been replaced with helper function.

PR #471, opened 2017-12-28
Construct storage from config.type

patriotyk:


src/uptane/initialize.cc
```diff
@@ -108,7 +108,7 @@ void Repository::setEcuKeysMembers(const std::string& primary_public, const std:
   key_source = source;
 }
 
-// Postcondition: (public, private) is in the storage. It should not be stored until secondaries are provisioned
+// Postcondition: (public, private) is in the storage-> It should not be stored until secondaries are provisioned
```
patrickvacek: This is an amusing typo. :)

patrickvacek:
This looks good, but I think uptane_ci_test also needs an update.

patrickvacek:
I've tested with an actual device, and it works! Very cool. However, aktualizr-info does not. It needs to be able to read from the database, too.

patrickvacek:
Actually, I can't send an update to the device. I'm seeing `Error of pulling image: Server returned HTTP 503`.

src/uptane/initialize.cc
patriotyk: That's was an  replace all  ))))))

patriotyk:
Fixed uptane_ci_test and aktualizr-info

PR #473, opened 2017-12-29
Add custom exceptions for ServerCredentials

patriotyk:


patrickvacek:
Looks good, but would you mind fixing up the `runtime_error` in `src/sota_tools/garage_push.cc` as well?

cajun-rat:
Looks good.  Can you also change the tests in `aktualizr/tests/sota_tools/auth_test.cc`  It should be possible to use gtest's `EXPECT_THROW` to simplify those tests now

patriotyk:
Updated tests with new exceptions.
Also changed garage_push.cc to handle exceptions separately, but I don't see any reason to do that.

PR #475, opened 2018-01-02
Remove the dreaded "friend class ::SotaUptaneClient".

patrickvacek:
SotaUptaneClient now requires both Storage and HttpClient as members, but at least they are just references and Storage is even const.

This isn't necessarily the best solution, but it is better than using friend. I'm open to further discussion on where to go from here.

### DIFF
--- a/src/sotauptaneclient.h
+++ b/src/sotauptaneclient.h
@@ -8,6 +8,7 @@
 #include "config.h"
 #include "events.h"
 #include "httpclient.h"
+#include "invstorage.h"
 #include "ostree.h"
 #include "uptane/secondaryinterface.h"
 #include "uptane/tufrepository.h"
@@ -15,7 +16,8 @@
 
 class SotaUptaneClient {
  public:
-  SotaUptaneClient(const Config &config_in, event::Channel *events_channel_in, Uptane::Repository &repo);
+  SotaUptaneClient(const Config &config_in, event::Channel *events_channel_in, Uptane::Repository &repo,
+                   const boost::shared_ptr<INvStorage> &storage_in, HttpInterface &http_client);
   void OstreeInstallSetResult(const Uptane::Target &package);
   void runForever(command::Channel *commands_channel);
 
@@ -39,6 +41,8 @@ class SotaUptaneClient {
   const Config &config;
   event::Channel *events_channel;
   Uptane::Repository &uptane_repo;
+  const boost::shared_ptr<INvStorage> &storage;
+  HttpInterface &http;
   int last_targets_version;
   Json::Value operation_result;
 };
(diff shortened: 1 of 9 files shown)


src/sotauptaneclient.h
```diff
@@ -39,6 +41,8 @@ class SotaUptaneClient {
   const Config &config;
   event::Channel *events_channel;
   Uptane::Repository &uptane_repo;
+  const boost::shared_ptr<INvStorage> &storage;
```
cajun-rat: Would this be better as a bare `boost::shared_ptr<INvStorage> storage`, rather than a reference to a shared pointer?

src/uptane/uptanerepository.h
```diff
@@ -72,13 +76,9 @@ class Repository {
   std::string primary_private_key_uri;
   std::string primary_public_key_id;
 
-  std::string pkcs11_tls_keyname;
-  std::string pkcs11_tls_certname;
-
   // ECU serial => (ECU hardware ID, ECU public_key)
   std::map<std::string, std::pair<std::string, std::string> > secondary_info;
 
-  friend class ::SotaUptaneClient;
```
cajun-rat: Happiness!

src/sotauptaneclient.h
patrickvacek: Yeah, if I keep it const, that should be fine. Reference to a pointer is not good style!

patrickvacek:
Rebased and got through the codecov gate.

src/sotauptaneclient.h
patrickvacek: @cajun-rat I pointed out a similar issue on Serhiy's PR (https://github.com/advancedtelematic/aktualizr/pull/485#discussion_r159892094) and he shared this link: https://stackoverflow.com/questions/3310737/shared-ptr-by-reference-or-by-value

I think there's room to argue about non-const references to a shared pointer, but if it is const, it would seem that passing as a reference is the right way to go after all.

cajun-rat: @patriotyk @patrickvacek  Interesting!  Turns out I was wrong about ref to shared_ptr.

cajun-rat: Ah, but hold on--this is a member variable in an object, where we want to use shared_ptr to manage the object lifecycle.  Those links are discussing function parameters, where I can see the benefit.

patrickvacek: So my impression is that if we are passing a non-const shared_ptr to a class constructor, it should probably not be passed by reference. But if it is a const shared_ptr, it probably is better a reference. That only affects SotaUptaneClient in this PR; the other two cases are fine as they are (non-const, non-reference shared_ptrs).

PR #485, opened 2018-01-04
Add migration from filesystem to sql backend

patriotyk:


src/fsstorage.cc
```diff
@@ -279,7 +292,7 @@ bool FSStorage::loadEcuSerials(std::vector<std::pair<std::string, std::string> >
 
 void FSStorage::clearEcuSerials() {
   boost::filesystem::remove(Utils::absolutePath(config_.path, "primary_ecu_serial"));
-  boost::filesystem::remove(Utils::absolutePath(config_.path, "primary_hardware_id"));
+  boost::filesystem::remove(Utils::absolutePath(config_.path, "primary_ecu_hardware_id"));
```
patrickvacek: Good catch!

patriotyk:
I have updated PR, now it should be ok

src/invstorage.cc
```diff
@@ -52,9 +52,63 @@ void INvStorage::importData(const ImportConfig& import_config) {
...
       return boost::shared_ptr<INvStorage>(new SQLStorage(config));
-    case kFile:
+    case kFileSystem:
     default:
       return boost::shared_ptr<INvStorage>(new FSStorage(config));
   }
+}
+
+void INvStorage::FSSToSQLS(const boost::shared_ptr<INvStorage>& fs_storage,
+                           boost::shared_ptr<INvStorage>& sql_storage) {
```
patrickvacek: Reference-to-pointer is probably unnecessary. This came up recently for some of my code as well and I'm not sure there's a good reason for the reference.

patriotyk: It is reference to object(instance of shared_ptr) that holds pointer, not to pointer itself.  So I don't see any reason to make a copy of that instance. 
https://stackoverflow.com/questions/3310737/shared-ptr-by-reference-or-by-value

patrickvacek: Good point. Seems fine to me upon reconsideration.

src/invstorage.cc
```diff
@@ -52,9 +52,63 @@ void INvStorage::importData(const ImportConfig& import_config) {
 boost::shared_ptr<INvStorage> INvStorage::newStorage(const StorageConfig& config) {
   switch (config.type) {
     case kSqlite:
+      if (!boost::filesystem::exists(config.sqldb_path)) {
+        boost::shared_ptr<INvStorage> sql_storage(new SQLStorage(config));
+        boost::shared_ptr<INvStorage> fs_storage(new FSStorage(config));
```
OYTIS: That shouldn't work to my understanding, because the new `[Storage]` configuration will be tuned for SQL storage. What we should I believe is to hardcode a config snippet for FSStorage to match what we have had before normally.

src/invstorage.cc
```diff
@@ -52,9 +52,63 @@ void INvStorage::importData(const ImportConfig& import_config) {
...
+        INvStorage::FSSToSQLS(fs_storage, sql_storage);
```
OYTIS: It seems to me that it will only work for the first time, and on subsequent calls when the FS storage has been cleared it will just populate our SQLStorage with empty values. There is also a minor problem with FSStorage constructor creating directories which we don't want to be there once we switched to SQL storage.

src/invstorage.cc
patriotyk: But how could we know the old configuration if it wasn't default?

OYTIS: We could use libostree to get previous deployment and extract configuration from there. But then there is a different set of open issues (what happens if we change sota.toml location between updates for example), and I believe using the default one should work for all our users.

src/invstorage.cc
patriotyk: No, the second time we will not run the migration because we check if database exists. SQLStorage constructor will create the databse at first time. Also we will not run FSStorage constructor second time because it runs only if database not exists.

OYTIS: Thank you, makes sense.

src/invstorage.cc
patriotyk: I just added separate commit, please look if did it right.

OYTIS: @patriotyk The FSStorage should be created with old_config, not SQLStorage. Might be the reason for segfault in tests as well.

patriotyk: Oh, yes it's just a mistake

patriotyk: I changed it  to work in another way. There was a problem with tests because in tests I need to setup "path" property. Is it ok? Or maybe You plan to replace sqldb_path with path in the future?

OYTIS:
@patriotyk Thank you. I need to fix a couple of things end-to-end and I'm going to push the changes here, so please don't overwrite them.

src/invstorage.cc
OYTIS: @patriotyk Yes, sqldb_path will be merged with path in the end. I liked the previous solution, the only thing that had to be fixed there is we need a way to override the path of old file system storage for tests. I believe it can be done either with build options (see e.g. TEST_PKCS11_MODULE_PATH) of with additional parameters to INvStorage functions.

OYTIS:
@patriotyk Sorry, didn't notice your changes to migration process. Could you please re-add old_config with necessary changes and then we're practically ready to go.

patriotyk:
Now it is ok

PR #486, opened 2018-01-07
Feat/pro 4623/aktualizr info more info

patriotyk:


src/fsstorage.h
```diff
@@ -41,8 +41,10 @@ class FSStorage : public INvStorage {
   virtual void clearEcuRegistered();
   virtual void storeInstalledVersions(const std::map<std::string, std::string>& installed_versions);
   virtual bool loadInstalledVersions(std::map<std::string, std::string>* installed_versions);
+  virtual StorageType type() { return type_; };
 
  private:
+  const StorageType type_;
```
cajun-rat: Could this be simplified to just:

~~~
virtual StorageType type() { return kFileSystem; };
~~~

patriotyk: Oh, yes You are right

src/aktualizr_info/main.cc
```diff
@@ -32,7 +32,13 @@ int main(int argc, char **argv) {
       exit(EXIT_SUCCESS);
     }
 
-    std::string sota_config_file = vm["config"].as<std::string>();
+    std::string sota_config_file = "/usr/lib/sota/sota.toml";
+    if (vm.count("config") != 0) {
+      sota_config_file = vm["config"].as<std::string>();
+    } else if (boost::filesystem::exists("/tmp/aktualizr_config_path")) {
+      sota_config_file = Utils::readFile("/tmp/aktualizr_config_path");
```
patrickvacek: What's the deal with this temp file?

patrickvacek: Oh, I see, that's the way to figure out whatever config aktualizr is using. Seems alright.

PR #487, opened 2018-01-09
Resign all json files with new expiration date

patriotyk:


cajun-rat:
Nice one

patrickvacek:
Thanks for looking into this! I was just starting to but got distracted by meetings and the new laptop. Looks like there's a new problem, but I think you're on the right track.

patriotyk:
Now it is ok, but codecov fails.

PR #491, opened 2018-01-09
Add some info about aktualizr-info utility

patriotyk:


cajun-rat:
+1 for documentation :)

PR #501, opened 2018-01-18
Bugfix/pro 4623/store misconfigured

patrickvacek:
While finally testing the last SQLite work, I found a couple issues. They were actually relatively minor, but the typo fix makes this look like a big change.


patrickvacek:
For the record, the misconfigured ecu tracking is not currently covered by any tests, but since the requirements for this are undefined and possibly nonexistent, I'm not sure how much sense it makes to bother with it until we have a more concrete idea of what we really want.

PR #502, opened 2018-01-18
Split Uptane steps

patriotyk:


patriotyk:
Just noticed that branch name is wrong. This PR is related to PRO-4478 issue.

src/aktualizr.cc
```diff
@@ -68,8 +68,12 @@ int Aktualizr::run() {
 #endif
   } else {
 #ifdef BUILD_OSTREE
-    // TODO: compile unconditionally
+// TODO: compile unconditionally
+#ifdef BUILD_P11
     boost::shared_ptr<INvStorage> storage = INvStorage::newStorage(config_.storage);
+#else
+    boost::shared_ptr<INvStorage> storage = INvStorage::newStorage(config_.storage);
+#endif
```
cajun-rat: Aren't both legs of this if the same?

patriotyk: Oh, yes, I made wrong merge conflict resolve

src/invstorage.cc
```diff
@@ -71,10 +72,10 @@ boost::shared_ptr<INvStorage> INvStorage::newStorage(const StorageConfig& config
         INvStorage::FSSToSQLS(fs_storage, sql_storage);
         return sql_storage;
       }
-      return boost::make_shared<SQLStorage>(config);
+      return boost::shared_ptr<INvStorage>(new SQLStorage(config, p11_config));
     case kFileSystem:
     default:
-      return boost::make_shared<FSStorage>(config);
+      return boost::shared_ptr<INvStorage>(new FSStorage(config, p11_config));
```
patrickvacek: Not a big deal, but probably should keep the `make_shared` in both of these.

PR #505, opened 2018-01-19
Make use of boost::make_shared helper

lbonn:
Arguably nicer than manual wrapping of a `new ...`. Also, only does one
allocation for the object and the counter altogether.

Available since boost versions from 2014 and also part of the C++11
standard


cajun-rat:
Nice

PR #507, opened 2018-01-19
Feat/prj 330/hsm not test

patrickvacek:


patrickvacek:
BTW there's a meta-updater PR coming too, but this has to get merged first anyway.

PR #509, opened 2018-01-23
Read secondary config as XER-encoded ASN.1

lbonn:
This is a kind of mergeable PoC for the integration of asn1c in the code base.

In a nutshell: asn1c generates C and header files from a set of asn1 schemas, these are then compiled as a static lib and linked against the main aktualizr lib.

Some various notes and caveats:

- the names of the generated source files are not easily predictable, so a first pass generation is done during the initial CMake run. This also means that if the schemas change sufficiently to influence the list of files, `cmake .` must be run again
- the multiple `#include <netinet/in.h>` is a workaround for some weird includes errors I got. Not really sure what is happening there, maybe a second pair of eyes would help. To reproduce the problem, just comment my includes and try to compile. The problem goes away if the `#include <netinet/in.h>` is removed manually from the generate `build/src/asn1/asn1/asn_system.h`
- asn1c generates C code we don't have much control over, so I added some helpers and wrapped the functions inside a namespace. This is mostly cosmetics since the header files are still accessible directly by `#include`. We might have to choose a common prefix for our ASN.1 objects if we don't want to risk too many collisions with our own C++ types...
- deserialization is much more comfortable than serialization right now, which requires the tedious building of the C objects and their children. I don't think there is a magic recipe to solve this...


lbonn:
Well, obviously travis needs to have asn1c installed to be happy...

patrickvacek:
This is cool. Are we confident we want to move forward with this ASN1 implementation, or is this "just" a PoC? I'm just wondering how comfortable we should be with merging now-ish.

src/asn1/helpers.h
```diff
@@ -0,0 +1,87 @@
+#ifndef ASN1_HELPERS_H
+#define ASN1_HELPERS_H
```
patriotyk: Should be ASN1_HELPERS_H_

tests/asn1_test.cc
```diff
@@ -0,0 +1,62 @@
+#include <gtest/gtest.h>
+
+#include <string>
+
+#include <boost/smart_ptr.hpp>
+
+#include <SecondaryConfig.h>
```
patriotyk: We include project heders with quotes

tests/asn1_test.cc
```diff
@@ -0,0 +1,62 @@
...
+#include "asn1/helpers.h"
+
+using namespace Uptane::ASN1;
+
+TEST(asn1, serialize_simple) {
+  SecondaryConfig *secConf = static_cast<SecondaryConfig *>(calloc(1, sizeof(*secConf)));
```
patriotyk: Why not: 
` SecondaryConfig *secConf = new SecondaryConfig;
`

src/httpclient.h
```diff
@@ -2,6 +2,7 @@
 #define HTTPCLIENT_H_
 
 #include <curl/curl.h>
+#include <sys/socket.h>
```
patriotyk: Do we really need this include?

src/uptane/managedsecondary.h
```diff
@@ -4,6 +4,7 @@
 #include <string>
 #include <vector>
 
+#include <netinet/in.h>
```
patriotyk: Why do we need this include?

tests/asn1_test.cc
patriotyk: Hm, I assume SecondaryConfig is ASN1::SecondaryConfig. Probable better to use  `using namespace Uptane;
` and `ASN1::SecondaryConfig`

src/httpclient.h
lbonn: This is covered this in the pull request description.

I admit I don't have a better solution for now...

tests/asn1_test.cc
lbonn: Can't use `new` because the deallocation is handled by the asn1c generated code which is in C and uses `free`. We most probably want to stick with their version as it frees the objects recursively.

lbonn: About the namespace comment: you're right, I forgot to update the test case to the new style of code organization I used.

lbonn:
Does not compile on Ubuntu 14.04, an include issue again...

cajun-rat:
Just a note that we need to sort out the meta-updater recipe ahead of merging this.

lbonn:
The issues were caused by the include of C headers inside a C++ namespace... That wasn't that smart.

So, for now, no way to isolate the generated C code from our code base. Except if we're ready to hack deep in asn1c.

cajun-rat:
Is that going to pollute the namespace of the implementation, or will it also end up in the namespace of the headers?

src/asn1/helpers.h
```diff
@@ -0,0 +1,85 @@
+#ifndef ASN1_HELPERS_H_
+#define ASN1_HELPERS_H_
+
```
patriotyk: You need to add all source files to "aktualizr_source_file_checks" function in cmake. Now, `make format` and `make check` don't checks this file. Also implementation should be in cc file, only inline implementation is allowed in header files.

src/asn1/helpers.h
```diff
@@ -0,0 +1,85 @@
...
+ public:
+  EncodeError()
+      : std::domain_error("Invalid ASN.1 serialized input") {}
+  virtual ~EncodeError() throw() {}
+};
+
+
+// Deserialization
+template<typename T>
+boost::shared_ptr<T> xer_parse(asn_TYPE_descriptor_t &descr, const std::string &in) {
```
patriotyk: Instead of refferenses we use pointers for out arguments

lbonn:
Let's close these experiments

PR #512, opened 2018-01-23
Refactor Ostree class.

patrickvacek:
There is now a hierarchy with PackageManagerInterface on top alongside PackageInterface. OstreeManagerInterface and OstrePackageInterface inherit from that, and there are real and fake versions inheriting from that.

This seems to work but might still need some tidying.

I'm not expecting to merge this quite yet, but I'd like to solicit some feedback at this point.


src/config.cc
```diff
@@ -282,9 +303,20 @@ void Config::updateFromPropertyTree(const boost::property_tree::ptree& pt) {
 
   // uptane.secondary_configs currently can only be set via command line.
 
+  std::string package_manager = "none";
```
cajun-rat: What was your thinking behind making this "none" rather than "ostree"?

patrickvacek: That's just the default if nothing is read out of ostree.type. Could be ostree if you'd prefer. In fact, that probably makes more sense.

cajun-rat:
Looks good to me.  I tried running this up in qemu, and ran into some issues with sqlite schemas.  With a tweaked sota.toml it would update fine.

It think there are some places where we can push this refactor further, but I'd be keen to get this merged and then carry on the work.

patrickvacek:
This is definitely still a work in progress, and there's more work to do, but I'm going to go ahead and merge this for now since it is in a working state.

PR #513, opened 2018-01-24
Docker file for garage-push

jochenschneider:


Dockerfile.garage-push
```diff
@@ -0,0 +1,10 @@
+FROM advancedtelematic/aktualizr
+LABEL Description="garage-push"
+
+ADD . src
+
+RUN apt-get install libsqlite3-dev
```
cajun-rat: I think you need apt-get update && apt-get install libsqlite3-dev here.  The reason is that if the advancedtelematic/aktualizr image is old then the apt metadata cache will be old, and the apt-get install will fail because it will try and download a package that is now obsolete.

The two commands need to be on one line, because Docker caches intermediate results at the line level.

Dockerfile.garage-push
```diff
@@ -0,0 +1,10 @@
+FROM advancedtelematic/aktualizr
```
cajun-rat: Are we planning on some process to keep this image up to date?  The last push to that image was a month ago.

jochenschneider: Good question. I guess libsqlite3-dev could be added and the name could be changed to "aktualizr-build-env" or something.

cajun-rat: The other option would be to start from one of the Ubuntu images and include all the packages required for the build

patrickvacek: I was under the impression that some part of the Travis CI process pushed the docker image, but that does not appear to be the case. How did this ever get updated before?

jochenschneider:
Updated.

cajun-rat:
Nice

PR #514, opened 2018-01-24
Switch to C++11 by default

cajun-rat:
Now is the time to move Aktualizr to C++11.  At the moment the core of the product compiles against C++98, although we require C++11 for some of the support utilities.

C++11 came in gcc around version 4.7, and Debian oldstable is currently at gcc 4.9, so I hope that we should be able to assume C++11 now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,11 @@ function(aktualizr_source_file_checks)
     aktualizr_clang_format(${ARGN})
 endfunction()
 
+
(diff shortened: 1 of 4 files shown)


cajun-rat:
@codders wants to check with the product management org to be sure this is a sane thing to do.  Please don't merge until that is done

cajun-rat:
(Waiting for a meeting on 1st Feb for a decision on this)

cajun-rat:
@codders If you are happy with this decision, can you approve, please?

PR #515, opened 2018-01-24
Add systemd service and control scripts for debian/ubuntu

OYTIS:
Looks like having systemd work in docker is a whole piece of black magic and the result still doesn't look exactly like a target system. So the test is just testing that aktualizr gets installed and to test for restarts after updates we'll need to figure out something else (e2e probably once all pieces of the puzzle are in place).


src/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ set(CPACK_DEBIAN_PACKAGE_MAINTAINER "anton.gerasimov@here.com")
 SET(CPACK_CMAKE_GENERATOR "Unix Makefiles")
 set(CPACK_PACKAGE_VERSION 1.0.0)
 set(CPACK_DEBIAN_PACKAGE_DEPENDS "libsodium18, libboost-filesystem1.58.0, libboost-log1.58.0, libboost-thread1.58.0, libboost-chrono1.58.0, libboost-random1.58.0, libboost-system1.58.0, libcurl3, openssl (>=1.0.1), libsqlite3-0, libssl1.0.0, libarchive13")
+set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA "${PROJECT_SOURCE_DIR}/config/systemd/debian-control/preinst;${PROJECT_SOURCE_DIR}/config/systemd/debian-control/postinst;${PROJECT_SOURCE_DIR}/config/systemd/debian-control/prerm;")
```
patrickvacek: How does it know where to install these files? Is this just CMake magic?

src/CMakeLists.txt
```diff
@@ -120,8 +120,8 @@ endif(BUILD_GENIVI)
 aktualizr_source_file_checks(main.cc ${SOURCES} ${HEADERS})
 
 ################## INSTALL RULES
-#configure_file(distribution/sota.service.cmake distribution/sota.service)
-#install(FILES distribution/sota.service DESTINATION /etc/systemd/system  COMPONENT init)
```
patrickvacek: Can we get rid of `distribution/sota.service`? We don't appear to use it anymore.

patrickvacek:
I think this looks fine, but I'm having trouble with the interaction between Docker and systemd, seemingly as expected. I don't think I'm seeing the debian-control files getting installed, though.

src/CMakeLists.txt
OYTIS: Sorry, I didn't notice that it was a real file. I'll delete it, sure.

OYTIS:
@patrickvacek To see the control files you can unpack the package with `ar x aktualizr.deb`, the contol scripts will be in `control.tar.gz` then. Systemd doesn't normally work in docker, yes, that's why I needed all these `|| true` in control scripts.

patrickvacek:
I'm seeing two problems.
1) I still don't see the control files anywhere in the .deb. I'm using dpkg-deb to examine it, since `ar x` doesn't seem to be doing anything for me. Am I doing something wrong?
2) If I run cmake with `DBUILD_SOTA_TOOLS=ON` and then run `make package`, it only builds the garage-deploy package and not the aktualizr package. That seems a bug.

OYTIS:
@patrickvacek 
Surprisingly (not really) I had to rewrite everything to fix the SOTA_TOOLS issue. Components support is a bit flaky, so I couldn't find a way to specify different package versions and maintainers for aktualizr and garage_deploy.

OYTIS:
OK, now the behavior depends on cmake version.

CMakeLists.txt
```diff
@@ -15,11 +15,10 @@ if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
     message(FATAL_ERROR "Aktualizr does not support building in the source tree. Please remove CMakeCache.txt and the CMakeFiles/ directory, then create a subdirectory to build in: mkdir build; cd build; cmake ..")
 endif()
 
-if(AKTUALIZR_VERSION)
-    add_definitions(-DAKTUALIZR_VERSION="${AKTUALIZR_VERSION}")
-else(AKTUALIZR_VERSION)
-    add_definitions(-DAKTUALIZR_VERSION="dev")
+if(NOT AKTUALIZR_VERSION)
```
lbonn: CMake warning here because the `endif` content doesn't match with `if`

lbonn:
Other CMake warning: https://github.com/advancedtelematic/aktualizr/blob/feat/PRO-4751/systemd-ubuntu/tests/CMakeLists.txt#L40

`garage-deploy-deb` is not a valid target anymore

OYTIS:
Thanks, @lbonn, should be fixed now.

PR #516, opened 2018-01-25
move all p11 dependant code to one place

patriotyk:
I did it in another way that Phil suggested. The reason for this is that we have source flag for each key. So currently I just moved all pkcs11 related code to one class.

### DIFF
--- a/src/uptane/initialize.cc
+++ b/src/uptane/initialize.cc
@@ -5,13 +5,13 @@
 
 #include "bootstrap.h"
 #include "logging.h"
+#include "uptane/cryptokey.h"
 #include "uptane/uptanerepository.h"
 
 namespace Uptane {
 
 // Postcondition: device_id is in the storage
-bool Repository::initDeviceId(const ProvisionConfig& provision_config, const UptaneConfig& uptane_config,
-                              const TlsConfig& tls_config) {
+bool Repository::initDeviceId(const ProvisionConfig& provision_config, const UptaneConfig& uptane_config) {
   // if device_id is already stored, just return
   std::string device_id;
   if (storage->loadDeviceId(&device_id)) return true;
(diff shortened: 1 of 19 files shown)


src/uptane/initialize.cc
```diff
@@ -5,13 +5,13 @@
...
 #include "logging.h"
+#include "uptane/cryptokey.h"
 #include "uptane/uptanerepository.h"
 
 namespace Uptane {
 
 // Postcondition: device_id is in the storage
-bool Repository::initDeviceId(const ProvisionConfig& provision_config, const UptaneConfig& uptane_config,
-                              const TlsConfig& tls_config) {
+bool Repository::initDeviceId(const ProvisionConfig& provision_config, const UptaneConfig& uptane_config) {
```
patrickvacek: Not relevant to your changes, but we are we bothering passing in UptaneConfig if we have a reference to the entire Config object in the Repository class already?

patriotyk: Yes, It looks strange, I don't understand this to. You may ask it Anton maybe he can answer.

tests/ostree_test.cc
```diff
@@ -27,10 +29,11 @@ TEST(OstreePackage, ToEcuVersion) {
 
 TEST(OstreePackage, InstallBadUri) {
   OstreePackage op("branch-name-hash", "hash", "bad_uri");
-  data::PackageManagerCredentials cred;
   Config config;
   config.pacman.type = kOstree;
   config.pacman.sysroot = sysroot;
+  CryptoKey keys(boost::make_shared<FSStorage>(FSStorage(config.storage)), config);
```
patrickvacek: I get an error when I test locally here because I don't have write access to `/var/sota`. We need to set `config.storage.metadata_path` to something temporary in the build directory.

patriotyk: Updated, waiting for CI finish

cajun-rat:
Just building a test image now

cajun-rat:
I wasn't able get a device to register when running on qemu using pkcs#11 yesterday. I'm not sure if that is caused by this change though.  We should write some tests for this (for example using oe-selftest), but I don't think that should be blocking this PR.

@patrickvacek if you've got time today, could you try the PKCS#11 workflow, otherwise I think we should merge this (it has been open too long already) and focus on writing the automated test

cajun-rat:
I wasn't able get a device to register when running on qemu using pkcs#11 yesterday. I'm not sure if that is caused by this change though.  We should write some tests for this (for example using oe-selftest), but I don't think that should be blocking this PR.

@patrickvacek if you've got time today, could you try the PKCS#11 workflow, otherwise I think we should merge this (it has been open too long already) and focus on writing the automated test

cajun-rat:
OK, merging

patrickvacek:
Turns out the problem is not just in rocko. I tried running this branch in pyro and am getting the same error. I tried without this branch and it was fine. I'm still investigating why.

PR #519, opened 2018-01-25
Feat/pro 4478/uptane separate steps

patriotyk:


patrickvacek:
@patriotyk Everything okay here? I know my recent OSTree work probably complicated what you're doing.

PR #525, opened 2018-01-29
Enable -f{function,data}-sections for partial

lbonn:
Before:
```
$ arm-none-eabi-size kea128_ms1.elf
   text    data     bss     dec     hex filename
  10588      36    4052   14676    37e0 kea128_ms1.elf
```
After:
```
$ arm-none-eabi-size kea128_ms1.elf
   text    data     bss     dec     hex filename
  10216      36    4052   14304    37e0 kea128_ms1.elf
```


OYTIS:
Cool, 300 bytes can still be useful. I'll just make a short test that it actually runs on the hardware.

lbonn:
In my experience, this is pretty safe and should only break at link-time.

OYTIS:
It's been a while since I've last checked the binary size. 10K even without UPTANE and crypto is huge.

PR #528, opened 2018-01-30
Generalize garage-push docker image to garage tools

jochenschneider:
- add Java to enable garage-sign
- remove entrypoint so that any of the tools can be called the same way


patrickvacek:
Very confused by the Codecov results. Why are they even affected by a change to a Dockerfile that isn't used by CI?

cajun-rat:
I think codecov is a bit broken.  It thinks that a bunch of C++ files have changed--perhaps it is getting the base of the comparison wrong?

PR #530, opened 2018-01-30
Move 'on' section up one level

OYTIS:


OYTIS:
Sorry, I'll just merge it. It's probably not the last iteration and can't really be done on a branch,

PR #531, opened 2018-01-30
Don't wrap api key into 'secure' field

OYTIS:


OYTIS:
Just merging again, sorry.

PR #533, opened 2018-01-30
Pass secret key as (encrypted) plaintext.

OYTIS:
The way @patriotyk originally implemented passing github tokens looks like the only officially supported one. Travis docs claim that the "secure" key cannot be used by anyone outside advancedtelematic.


OYTIS:
Some hiccups because of non-existent tag, will open another PR.

PR #534, opened 2018-01-30
Pass secret key as (encrypted) plaintext.

OYTIS:
The way @patriotyk originally implemented passing github tokens looks like the only officially supported one. Travis docs claim that the "secure" key cannot be used by anyone outside advancedtelematic.


OYTIS:
@patrickvacek As far as I understand it is encrypted message that can only be decrypted by a key specific to this project. And travis only gives this key to builds launched on behalf of this project (not to PRs from other accounts e.g.)

PR #536, opened 2018-01-30
Remove GENIVI support

lbonn:
Not sure if we should also remove SocketGateway, GatewayManager and the like?


patriotyk:
And 'config/ex'

CMakeLists.txt
```diff
@@ -156,11 +150,6 @@ include_directories(${SQLITE3_INCLUDE_DIRS})
 include_directories(${LIBDBUS_INCLUDE_DIRS})
 include_directories(${sodium_INCLUDE_DIR})
 
-if(BUILD_GENIVI)
-    add_subdirectory("third_party/rvi_lib")
-    set_property(TARGET rvi PROPERTY CXX_STANDARD 11)
-endif(BUILD_GENIVI)
-
 # Setup warnings. Do this after rvi_lib is added so that it isn't affected.
```
patriotyk: Remove or modify this comment

patrickvacek:
> And 'config/ex'

Are you sure about that? I know for sure that `config/ex/secondary` is still used. Not sure about the rest. Either way, I'd like to see `ex` renamed to `example`. Or, if everything else in `config/ex` is obsolete, then we should just move the `secondary` folder one level up.

src/main.cc
```diff
@@ -53,10 +53,7 @@ bpo::variables_map parse_options(int argc, char *argv[]) {
       ("loglevel", bpo::value<int>(), "set log level 0-4 (trace, debug, warning, info, error)")
       ("config,c", bpo::value<std::string>()->required(), "toml configuration file")
       ("gateway-http", bpo::value<bool>(), "enable the http gateway")
-      ("gateway-rvi", bpo::value<bool>(), "enable the rvi gateway")
       ("gateway-socket", bpo::value<bool>(), "enable the socket gateway")
-      ("gateway-dbus", bpo::value<bool>(), "enable the D-Bus gateway")
-      ("dbus-system-bus", "Use the D-Bus system bus (rather than the session bus)")
```
patrickvacek: @cajun-rat @OYTIS do we know of anyone that might be using these options, even if it is to just set them to `false`? I think we can largely not worry about that, but I'm just trying to be careful.

tests/config_tests_device_id.toml
```diff
@@ -1,12 +1,9 @@
...
-dbus = true
 http = false
-rvi = true
 socket = true
 
 [network]
 http_server = "127.0.0.1:8888"
-rvi_edge_server = "127.0.0.1:9080"
 socket_commands_path = "/tmp/sota-commands.socket"
 socket_events_path = /tmp/sota-events.socket"
```
patrickvacek: How has this typo survived so long? There's a missing quote here (unrelated to your PR).

patriotyk: Oh, I saw that highlighted path, but could not understand why github highlights it. I saw it in two places.

CMakeLists.txt
patriotyk: I did not know that secondaries uses it.  I added this folder for rvi support only.

patrickvacek: Ah yes, `git log` confirms that. Then we should just move the `secondary` directory outside of `ex` and get rid of `ex`.

patrickvacek:
We can also remove the jansson lib at this point (currently specified in the README and Dockerfiles). If we do also get rid of the gateway stuff, then picojson can go as well.

src/main.cc
cajun-rat: Good question.  I think your concern is valid.  For people who are using these features, exploding violently is probably the best option--they should find out they they can't do stuff as soon as possible.

The case that you were thinking of though was where people had forced that value to 'off' (the default).  It would be nice to not break those guys.  The simplest solution I can think of is to document these options as 'deprecated' then check then and throw an error if they are set to true.  The cost would be that we would have to maintain that extra code in the future.

Given the balance between a mild annoyance to somewhere between 0 and 10 users and our long term maintenance costs, I think just dropping the options is fine.

cajun-rat:
oe-selftest -r updater passes

src/main.cc
patrickvacek: Agreed. I think the chance that someone set one of these options explicitly to 'off' is unlikely, and the fix is simple anyway.

PR #537, opened 2018-01-31
Allow most of uptane to build without BUILD_OSTREE

lbonn:
Thanks to all your good work, this proved to be rather easy.

Hope I haven't forgotten anything.


cajun-rat:
Nice!

patrickvacek:
Awesome! We may eventually consider running `packagemanager_test` even without BUILD_OSTREE enabled, but since it doesn't do much as it is anyway, I don't think it's worth worrying about now.

patrickvacek:
Also: not critical at all, but if you have to make changes anyway, I might suggest joining together the two `BUILD_OSTREE` sections in `tests/CMakeLists.txt`.

PR #538, opened 2018-02-01
Make P11Engine a singleton

OYTIS:


lbonn:
How did this caused the referenced bug?

src/p11engine.h
```diff
@@ -65,7 +65,10 @@ class P11SlotsWrapper {
 
 class P11Engine {
```
patriotyk: derrive it from boost::noncopyable

OYTIS:
@lbonn  Despite its name P11Engine represents a PKCS11-compatible shared library. In destructor it calls PKCS11_CTX_Unload, which deinitializes the library. If the destructor is called while some other P11Engine is alive, the latter will fail because the library was already deinitialized.

OYTIS:
Sorry all, this works, but is non-testable, I'll take another approach.

patrickvacek:
What do you mean, "non-testable"?

OYTIS:
@patrickvacek The problem is P11Engine takes config as a parameter and with a singleton where I only create an instance once in program's lifetime it is doomed to be configured with the first Config I pass to Get() function. It is fine in the real aktualizr, but in tests when a single binary can contain multiple tests with different configs that just doesn't work.

cajun-rat:
Superseded by #541

PR #539, opened 2018-02-01
Latest garage-sign (106).

patrickvacek:


patrickvacek:
Merging despite our norms because of urgent needs of production.

PR #540, opened 2018-02-01
Provide safer temporary files and directories

lbonn:
As mentioned in PRO-4721 and PRO-4722, TemporaryFile and
TemporaryDirectory were somewhat unsafe.

To solve this, we create a global temporary directory root for the
lifetime of the application, with restrictive permissions. The directory
is created in the OS' canonical temporary directory so that aktualizr
can eventually run under a lesser privileged user.

### DIFF
--- a/src/utils.cc
+++ b/src/utils.cc
@@ -334,31 +334,47 @@ boost::filesystem::path Utils::absolutePath(const boost::filesystem::path &root,
   }
 }
 
-static boost::filesystem::path safe_temp_root() {
-  static boost::filesystem::path root;
-  static boost::mutex mutex;
+class SafeTempRoot : private boost::noncopyable {
+ public:
+  // provide this as a static method so that we can use C++ static destructor
+  // to remove the temp root
+  static boost::filesystem::path &Get() {
+    {
+      // Ensure thread safety (uneeded in c++11)
+      boost::lock_guard<boost::mutex> lock(mutex);
+      static SafeTempRoot r;
 
-  {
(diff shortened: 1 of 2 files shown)


lbonn:
I added some bloat to remove the directory at the end of the program. Not critical, but nice anyway, especially in dev environments.

src/utils.cc
```diff
@@ -330,8 +334,47 @@ boost::filesystem::path Utils::absolutePath(const boost::filesystem::path &root,
...
+  }
+  ~SafeTempRoot() {
+    try {
+      boost::filesystem::remove_all(root);
+    } catch (...) {
+      // ignore this, not critical
+    }
+  }
+
+  boost::filesystem::path root;
```
patriotyk: maybe 'path' is better name?

lbonn: Yes you're right, I made the change.

src/utils.cc
```diff
@@ -330,8 +334,47 @@ boost::filesystem::path Utils::absolutePath(const boost::filesystem::path &root,
   }
 }
 
+class SafeTempRoot : private boost::noncopyable {
+ public:
+  // provide this as a static method so that we can use C++ static destructor
+  // to remove the temp root
+  static boost::filesystem::path &Get() {
+    {
+      // Ensure thread safety (uneeded in c++11)
```
patrickvacek: +1 for leaving a note so that we can clean up once we upgrade for good.

patrickvacek:
This is cool and looks good to me. I recommend that we have @heartsucker approve this before merging, though.

PR #541, opened 2018-02-01
Add P11EngineGuard to count existing P11 users

OYTIS:
The new version drops P11Engine once it has no more users (practically once Uptane::Repository gets out of scope) thus allowing reinitialization in tests. I have a feeling that I'm reimplementing a shared pointer, so if you think it can be done better I'm open to suggestions.

### DIFF
--- a/src/p11engine.cc
+++ b/src/p11engine.cc
@@ -12,6 +12,9 @@
 #define kPkcs11Path "/usr/lib/engines/libpkcs11.so"
 #endif
 
+P11Engine* P11EngineGuard::instance = NULL;
+int P11EngineGuard::ref_counter = 0;
+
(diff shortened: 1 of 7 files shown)


patriotyk:
I reopened Your singleton PR because I think it is better solution and it could work if we delete instance mannualy

OYTIS:
Might be. As far as you could make it work =)

cajun-rat:
I've been working on the 'constructing P11/CryptoKey at dependency injection' patch, and it is turning into a bit of a mammoth change.  I'm starting to prefer this one (at least in the short term).  Thoughts?

patrickvacek:
@cajun-rat I'm okay with this as a short-term fix for the existing open bugs, especially since it affects production right now. Then there's a bit less pressure on hurrying to get the final answer just right.

cajun-rat:
Done.

PR #544, opened 2018-02-02
Feat/pro 4750/functional non ostree

patrickvacek:


OYTIS:
@patrickvacek Works fine for me. There are some suspicious messages in the log though (pacman.type is set to "none"):

    Feb 02 14:33:55 qemux86-64 aktualizr[299]: libostree pull from 'aktualizr-remote' for 0 refs complete
                                           security: GPG: disabled http: CA-pinned
                                           non-delta: meta: 2 content: 0
                                           transfer: secs: 0 size: 162 bytes

Might be because I'm still on pyro.

patrickvacek:
@OYTIS I was seeing those in the uptane refactor PR from @patriotyk. I don't think that is specific to this PR. Everything seems to be working, but can you tell if that message means something sinister?

OYTIS:
@patrickvacek I've got no idea, was just surpised that libostree pull is called even when using a different package manager. Just compiled with BUILD_OSTREE=OFF and they are gone.
Still curious where they are coming from of course.

patrickvacek:
@OYTIS It's because of the `BUILD_OSTREE` guard in `tufrepository.cc`. I don't like that but @patriotyk has said he's working on it. It should be mostly harmless at present.

PR #550, opened 2018-02-05
doxygen

heartsucker:
Builds docs as part of `make all`.


docs/CMakeLists.txt
```diff
@@ -10,5 +13,17 @@ if (INKSCAPE)
         COMMAND ${INKSCAPE} --export-pdf=secondary-ecus.pdf ${CMAKE_CURRENT_SOURCE_DIR}/secondary-ecus.svg
         DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/secondary-ecus.svg)
 
-    add_custom_target(docs ALL DEPENDS overview.pdf secondary-ecus.pdf)
+    add_custom_target(docs-inkscape DEPENDS overview.pdf secondary-ecus.pdf)
+    add_dependencies(docs docs-inkscape)
 endif(INKSCAPE)
+
+if (DOXYGEN_FOUND)
```
cajun-rat: This needs an AND DOXYGEN_DOT_FOUND, in case the user has doxygen but not dot installed

heartsucker: Does that need to be in the `else` and `endif` statements too?

cajun-rat: Yeah, I think so. In a meeting at the moment

PR #552, opened 2018-02-06
Store targets in db when using sqlstorage

lbonn:
This was a bit involved, good thing we have c++ 11 now


lbonn:
The stream api is broken for now...

lbonn:
... It wasn't really broken actually (as far as I know), the bug was somewhere else.

This should be ready now

cajun-rat:
Tested with oe-selftest and a manual update in qemu

PR #553, opened 2018-02-06
Feat/pro 4740/debian pkg part2

patriotyk:


src/deb.cc
```diff
@@ -27,3 +27,16 @@ Json::Value DebianManager::getInstalledPackages() {
...
 }
+
+data::InstallOutcome DebianManager::install(const Uptane::Target &target) const {
+  std::string cmd = "dpkg -i ";
+  std::string output;
+  int status = Utils::shell(cmd + (path_ / "targets" / target.filename()).string(), &output);
+  if (status != 0) {
+    return data::InstallOutcome(data::INSTALL_FAILED, output);
+  }
+  Utils::writeFile(path_ / "targets/installed", target.sha256Hash());
```
cajun-rat: Would it be better to use INvStorage to store this information?

I suspect we'd need a new table in the SQL database, but I'm not sure how it relates to the storeInstalledVersions method we have today.

src/deb.h
```diff
@@ -12,18 +12,12 @@
 
 class DebianManager : public PackageManagerInterface {
  public:
-  DebianManager(const PackageConfig &pconfig) : config_(pconfig) {}
+  DebianManager(const PackageConfig &pconfig, const boost::filesystem::path &path) : config_(pconfig), path_(path) {}
```
cajun-rat: I think this should be a INvStorage

src/deb.cc
patriotyk: Yes, I just noticed #552, so I will pass storage to packageManagerInterface

src/deb.h
patriotyk: I agree

tests/packagemanager_deb_tests.cc
```diff
@@ -0,0 +1,47 @@
...
+#include <config.h>
+#include <packagemanagerfactory.h>
+#include <packagemanagerinterface.h>
+#include <utils.h>
+
+TEST(PackageManagerFactory, Debian_Install_Good) {
+  Config config;
+  config.pacman.type = kDebian;
+  boost::shared_ptr<PackageManagerInterface> pacman =
+      PackageManagerFactory::makePackageManager(config.pacman, boost::filesystem::path("/tmp/"));
```
cajun-rat: TemporaryDirectory (in utils.h) would be better here, because writing to fixed locations in /tmp is likely to cause trouble with parallel tests, and can cause problems because the old test runs get left around.  If we just use a fresh TemporaryDirectory every time, then there is no problem with junk getting left around and breaking tests

src/deb.cc
```diff
@@ -27,3 +27,16 @@ Json::Value DebianManager::getInstalledPackages() {
   dpkg_program_done();
   return packages;
 }
+
+data::InstallOutcome DebianManager::install(const Uptane::Target &target) const {
+  std::string cmd = "dpkg -i ";
```
cajun-rat: Does this work in the case where Aktualizr is installing an update to Aktualizr?

tests/run_debian_tests.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+PATH=$2:$PATH
```
patrickvacek: Probably worth surrounding with quotes for good measure.

tests/CMakeLists.txt
```diff
@@ -178,6 +178,14 @@ if(BUILD_OSTREE)
     endif(SOTA_PACKED_CREDENTIALS)
 endif(BUILD_OSTREE)
 
+if(BUILD_DEB)
+    add_executable(t_packagemanager_deb packagemanager_deb_tests.cc)
+    add_dependencies(build_tests t_packagemanager_deb)
+    target_link_libraries(t_packagemanager_deb aktualizr_static_lib ${TEST_LIBS})
+
+    add_test(NAME test_packagemanager_deb COMMAND ${PROJECT_SOURCE_DIR}/tests/run_debian_tests.sh ${CMAKE_CURRENT_BINARY_DIR}/t_packagemanager_deb
+    ${PROJECT_SOURCE_DIR}/tests/test_data/fake_dpkg)
```
cajun-rat: Maybe also list(APPEND TEST_SOURCES packagemanager_deb_tests.cc)

config/schemas/migrate.0.sql
```diff
@@ -6,7 +6,7 @@ CREATE TABLE device_info(device_id TEXT NOT NULL, is_registered INTEGER NOT NULL
 CREATE TABLE ecu_serials(serial TEXT UNIQUE, hardware_id TEXT NOT NULL, is_primary INTEGER NOT NULL CHECK (is_primary IN (0,1)));
 CREATE TABLE misconfigured_ecus(serial TEXT UNIQUE, hardware_id TEXT NOT NULL, state INTEGER NOT NULL CHECK (state IN (0,1)));
 CREATE TABLE primary_keys(private TEXT, public TEXT);
-CREATE TABLE installed_versions(hash TEXT UNIQUE, name TEXT NOT NULL);
+CREATE TABLE installed_versions(hash TEXT UNIQUE, name TEXT NOT NULL, is_current INTEGER NOT NULL CHECK (is_current IN (0,1)));
```
cajun-rat: You can't change existing migrations, because the devices in the field only apply migrations that they haven't seen before.  The answer here is to create a migrate.3.sql that adds the new column.

src/invstorage.h
```diff
@@ -116,8 +117,8 @@ class INvStorage {
   virtual bool loadEcuRegistered() = 0;
   virtual void clearEcuRegistered() = 0;
 
-  virtual void storeInstalledVersions(const std::map<std::string, std::string>& installed_versions) = 0;
-  virtual bool loadInstalledVersions(std::map<std::string, std::string>* installed_versions) = 0;
+  virtual void storeInstalledVersions(const Json::Value& installed_versions) = 0;
+  virtual bool loadInstalledVersions(Json::Value* installed_versions) = 0;
```
cajun-rat: What was the reason for having an untyped value here?  Would a real class be a better/safer option?

patriotyk: After I added third coulumn, we cannot use pair anymore because it holds only two values. So we need or create some struct and serealize/deserealize from json or just use json avoid those conversions. So I choosed second option. But if You insist I could create structure.

patriotyk:
Please review

scripts/test.sh
```diff
@@ -19,5 +19,5 @@ if [ -n "$BUILD_ONLY" ]; then
   fi
   echo "Skipping test run because of BUILD_ONLY environment variable"
 else
-  CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=5 make -j6 check-full
+  CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=4 make -j6 check-full
```
patrickvacek: Just curious, have you experienced differing performance with different parallel levels here?

scripts/test_aktualizr.sh
```diff
@@ -0,0 +1,15 @@
...
+UPDATE_ID=$(sed -e 's/^"//' -e 's/"$//' <<<"$UPDATE_ID")
+echo "UPDATE_ID: $UPDATE_ID"
+curl "https://app.atsgarage.com/api/v1/admin/devices/385b8a0c-0add-4b43-b153-41728b2a4ced/multi_target_update/46205055-daf6-4fe3-a1eb-7cbcdf4ea960' -X PUT -H 'Cookie: sizeVerifyHidden=1; fireworksPageAcknowledged=1; systemReady=1; PLAY_SESSION=$PLAY_SESSION" -H 'Origin: https://app.atsgarage.com' -H 'Accept-Encoding: gzip, deflate, br' -H 'Accept-Language: en-US,en;q=0.9,uk;q=0.8,de;q=0.7' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36' -H 'Accept: application/json, text/plain, */*' -H 'Csrf-Token: c7ae20514d7f5e791a7e39a763720d21224c8d7c-1518391691761-d060f16175da1bbef90e3a38' -H 'Referer: https://app.atsgarage.com/' -H 'Connection: keep-alive' -H 'Content-Length: 0' -H 'DNT: 1' --compressed
+
+aktualizr -c /persistent/config_selfupdate.toml --poll-once
```
patrickvacek: So the goal here is to launch aktualizr, fetch a specific update from the server, then start a new aktualizr to verify that the download worked? I'm a bit confused where the installation is happening. I think I might recommend naming this script something like `test_aktualizr_selfupdate` just to be a bit more clear what is really being tested.

CMakeLists.txt
```diff
@@ -68,8 +68,13 @@ if(NOT CMAKE_BUILD_TYPE)
     message(STATUS "No CMAKE_BUILD_TYPE specified, defaulting to Debug")
     set(CMAKE_BUILD_TYPE Debug)
 endif(NOT CMAKE_BUILD_TYPE)
+
 set(CMAKE_CXX_FLAGS_DEBUG "-g")
 
+SET(CMAKE_BUILD_TYPE Valgrind)
```
cajun-rat: I'm not quite sure what this does? It doesn't appear to actually set CMAKE_BUILD_TYPE, but the documentation says that CMAKE_CONFIGURATION_TYPES should be used to add a new configuration type: https://cmake.org/cmake/help/v3.0/variable/CMAKE_CONFIGURATION_TYPES.html

config/schemas/migrate.3.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN TRANSACTION;
+
+ALTER TABLE installed_versions ADD COLUMN is_current INTEGER NOT NULL CHECK ( is_current IN ( 0 , 1 ) ) DEFAULT 0;
```
cajun-rat: Good. +1 for the valid range check!

scripts/test_aktualizr.sh
```diff
@@ -0,0 +1,15 @@
...
...hCV2hxaXg5MWg5T0xxS1VLdXpydXR4TVpBWWQ2WW9nODN4ZkFzM3oxQnpOaS1pZDlUN2RSRFZuN21qeUlXVkUtS2djWUVvdGRKMkJrdGZzV3p0NGtOT1BNREVnbjBiN0Q2Zzl3a0RFSWQ3S3JvTG52SjE5RzNranJ4cnY4VnBDWVdtTHhybGxqU2YtQWNOLTNKczA2ZGNhZF84Yk5pYmFXYUFLYjMtYk9SUXVZbjdUUEh2RURPaHU4MEc3UWZReDJlQSIsImFjY2Vzc190b2tlbiI6IntcImFjY2Vzc190b2tlblwiOlwiZXlKaGJHY2lPaUpJVXpJMU5pSXNJbXRwWkNJNmJuVnNiSDAuZXlKcGMzTWlPaUpvZEhSd2N6b3ZMMkYxZEdndGNHeDFjeTVoWkhaaGJtTmxaSFJsYkdWdFlYUnBZeTVqYjIwaUxDSmhkV1FpT2xzaWRHOXJaVzRpTENKamJHbGxiblFpWFN3aWFuUnBJam9pVUd4c05HeDJiRFpZV2taU1VFbzJOVWhVWjNnaUxDSmpiR2xsYm5SZmFXUWlPaUl4TldJd01XUmlaaTFtTjJWbUxUUXpZV1F0WVdaaE1pMWhZamczWW1abE5EQmpZV0VpTENKemRXSWlPaUpoZFhSb01IdzFZVGRsWmpRMU5UazJZakl3WXpjMU1ESTRObVJrWlRVaUxDSnZkMjVsY2lJNmJuVnNiQ3dpYVdGMElqb3hOVEU0TXpreE5qa3hMQ0psZUhBaU9qRTFNVGcwTnpnd09URXNJbk5qYjNCbElqb2lkRzlyWlc0dWFXNTBjbTl6Y0dWamRDQmpiR2xsYm5RdWNtVm5hWE4wWlhJZ1kyeHBaVzUwTG5Wd1pHRjBaU0o5LmJPb3NXUHo4VV9iNTJlMHByRmZCRTJDN3l0eDBHNkxKWllhbkJ0ZFo2Z3NcIixcIm5vdEFmdGVyXCI6MTUxODQ3ODA5MH0iLCJjc3JmVG9rZW4iOiJjN2FlMjA1MTRkN2Y1ZTc5MWE3ZTM5YTc2MzcyMGQyMTIyNGM4ZDdjLTE1MTgzOTE2OTE3NjEtZDA2MGYxNjE3NWRhMWJiZWY5MGUzYTM4In0sIm5iZiI6MTUxODM5MTY5MSwiaWF0IjoxNTE4MzkxNjkxfQ.9LaqhbNXZ5HgCmMdWmilOSETur0Ec4ROvlUkS2Zii8w
```
cajun-rat: I don't think we can hard-code a session-cookie in here, for two reasons:

1) The Aktualizr repo is public, and I'm a bit concerned about letting anyone who can scrape github have access to one of our accounts.  I can't see what they would do with it right now, but it is certainly bad security hygiene.

2) The cookie will expire, and the tests will stop working until someone fixes the test up, and I think this will happen rather regularly.

Would it be better to take this in via a cmake variable?  This would mean that people could set it locally to use their own accounts, and there wouldn't be problems of it expiring, and people can access the account for debugging?

cajun-rat:
I'm keen on getting the Debian support part of this in before this PR turns into a monster. Could you maybe split the adding-Debian part of the PR from the adding-tests part, so that we can get the adding-Debian part merged?

scripts/test.sh
patriotyk: I did this because worker was failnig bcause after changes the order of tests has changed and ctest runs 5 very long running tests at once. As result  there is no output for some long time and travis kills the worker.

scripts/test_aktualizr.sh
patriotyk: Installation haapnes when we run `aktualizr -c /persistent/config_selfupdate.toml --poll-once`
If You ask about the installation of aktualizr that performs selfupdate, this is just builded debian package that we build for testing in docker. To save running tests time I added this tests to script that tests debian package build

CMakeLists.txt
patriotyk: I created this configuration type just because on valgrind manual page they claims that if sources has builded with -o1 it improves performance significantly.

http://valgrind.org/docs/manual/quick-start.html

scripts/test_aktualizr.sh
patriotyk: I knew about 2. And didn't think about 1. You are right this is security issue.
But we could use travis variable settings to setup login and password for test account, then we should in script somehow retrive a PLAY_SESSION

patriotyk:
Removed selfupdate commits and tests passed.

patrickvacek:
@patriotyk Looks like there's a merge conflict but that should be easy to fix.

cajun-rat:
Looks good to me.  Tested a manual update with OSTree, will merge and test out Debian.

PR #554, opened 2018-02-07
Do not build docs by default

lbonn:


lbonn:
Like `qa` or `format`, this feels like a special target that you would not like to run at every build when developing. It also outputs warnings right now.

cajun-rat:
@heartsucker What are your thoughts here?

heartsucker:
Sup y'all. Either way works. I don't think it's necessary for the `all` I suppose. I added it, but if that's not a normal workflow, it's def fine to remove it.

PR #557, opened 2018-02-08
Fix OSTree HSM credentials bug.

patrickvacek:
We were creating temporary files for the p11 case that we unnecessary. I
moved the logic directly into CryptoKey and eliminated
PackageManagerCredentials since no one else was using it. The HttpClient
setCerts could perhaps be refactored to take advantage of this.


patrickvacek:
I think I might do some more refactoring on this, actually.

cajun-rat:
OK, give a shout when you are ready

patrickvacek:
Rebased and such but unfortunately doesn't work anymore.

patrickvacek:
I fixed the new bug (tested on both pyro and rocko), but now I'm seeing two issues. First, after receiving an update on the device, after ten seconds or so I see something like this:
```
Feb 09 08:06:16 qemux86-64 aktualizr[482]: libostree pull from 'aktualizr-remote' for 0 refs complete
                                           security: GPG: disabled http: CA-pinned
                                           non-delta: meta: 2 content: 0
                                           transfer: secs: 0 size: 162 bytes
```
I don't think this is actually related to my changes. This seems connected to the TufRepository/Ostree::pull issue that still needs to get resolved (@patriotyk).

I'm also seeing this on CI but not when I run the same commands in Docker:
```
[ RUN      ] KeyManager.SignTufPkcs11
[2018-02-09 08:04:06.318982] [0x00007fc38fd7f740] [debug]   Slot manufacturer......: SoftHSM project
[2018-02-09 08:04:06.318997] [0x00007fc38fd7f740] [debug]   Slot description.......: SoftHSM slot ID 0x78e3052b
[2018-02-09 08:04:06.319002] [0x00007fc38fd7f740] [debug]   Slot token label.......: Virtual token
[2018-02-09 08:04:06.319007] [0x00007fc38fd7f740] [debug]   Slot token manufacturer: SoftHSM project
[2018-02-09 08:04:06.319011] [0x00007fc38fd7f740] [debug]   Slot token model.......: SoftHSM v2
[2018-02-09 08:04:06.319015] [0x00007fc38fd7f740] [debug]   Slot token serialnr....: 874b122478e3052b
[2018-02-09 08:04:06.322776] [0x00007fc38fd7f740] [error]   EVP_PKEY_get1_RSA failed with error error:0607907F:digital envelope routines:func(121):reason(127)
/aktualizr/src/tests/keymanager_test.cc:109: Failure
Expected: (signed_json["signatures"][0]["sig"].asString().size()) != (0), actual: 0 vs 0
[  FAILED  ] KeyManager.SignTufPkcs11 (6 ms)
```
I haven't figured out yet what the issue is and why it sometimes works.

patrickvacek:
The second issue I mentioned has now been fixed thanks to @OYTIS. I think the first issue should still be looked into but is probably not caused by this work and should not hold back the PR for now.

PR #561, opened 2018-02-08
Generate 4096 bits RSA keys

lbonn:
Instead of 2048


patrickvacek:
Looks fine. @heartsucker is this okay to leave hardcoded? I think the answer is yes, since allowing configuration means a higher risk of ending up with a less-secure choice, but if our concern really is security over 10+ years, is even 4096 really enough?

cajun-rat:
Have you built this and checked that the device can still register and update?  It is the sort of change that might break random things on the backend...

lbonn:
Seems to work fine: 
4096b-key

I believe 4096 is pretty conservative already, experts seem to recommend ~3000 (see 
https://www.keylength.com/en/compare/) for long term and/or switching to EC, provided that quantum computers are not around the corner.

PR #562, opened 2018-02-09
Improve resilience of example_flasher logic.

patrickvacek:
Previously it assumed if /var/sota existed it was good to use. Now it
actually checks for read-/writability before moving forward. This is
obviously still liable to fail if the filesystem changes before we
actually use it, but this is just an example that is also used for
testing.


patrickvacek:
@EpicNuts please try `make qa` locally with this branch and let me know if you have any problems. This should fix the issue I was aware of.

PR #564, opened 2018-02-12
Change default pkcs11 library.

patrickvacek:
Necessary for proper bitbaking. Thanks to Anton for the suggestion.


patrickvacek:
Only took 1700 seconds each to pass `test_uptane` and `test_uptane_network`.

PR #565, opened 2018-02-12
Send a random port to the ostreesysroot scripts.

patrickvacek:


patrickvacek:
Only 2000 seconds for `test_uptane`!

PR #567, opened 2018-02-12
Don't pull OSTree ref if it is already in repo

OYTIS:
Should help preventing trying to pull an image every 10 seconds.

### DIFF
--- a/src/ostree.cc
+++ b/src/ostree.cc
@@ -30,6 +30,20 @@ data::InstallOutcome OstreeManager::pull(const Config &config, const KeyManager 
     return data::InstallOutcome(data::INSTALL_FAILED, "Could not get OSTree repo");
   }
 
+  GHashTable *ref_list = NULL;
+  if (ostree_repo_list_commit_objects_starting_with(repo, refhash.c_str(), &ref_list, nullptr, &error)) {
+    guint length = g_hash_table_size(ref_list);
+    g_hash_table_destroy(ref_list);  // OSTree creates the table with destroy notifiers, so no memory leaks expected
+    // should never be greater than 1, but use >= for robustness
+    if (length >= 1) {
+      return data::InstallOutcome(data::ALREADY_PROCESSED, "Refhash was already pulled");
+    }
+  }
+  if (error) {
+    g_error_free(error);
+    error = NULL;
+  }
+
   if (!OstreeManager::addRemote(repo, config.pacman.ostree_server, keys)) {
     g_object_unref(repo);
     return data::InstallOutcome(data::INSTALL_FAILED, "Error adding OSTree remote");


src/ostree.cc
```diff
@@ -55,6 +55,29 @@ data::InstallOutcome OstreeManager::pull(const Config &config, const KeyManager
   return data::InstallOutcome(data::OK, "Pulling ostree image was successful");
 }
 
+data::InstallOutcome OstreeManager::find_or_pull(const Config &config, const KeyManager &keys, const std::string &refhash) {
+  OstreeRepo *repo = NULL;
+  const char *const commit_ids[] = {refhash.c_str()};
+  GCancellable *cancellable = NULL;
+  GError *error = NULL;
+  GVariantBuilder builder;
+  GVariant *options;
```
patrickvacek: `commit_ids`, `builder`, and `options` are not actually used here. But does it make sense to merge this into the `pull` function anyway? Normally I wouldn't advocate for making functions longer, but there are some duplicated calls and it's the only place `pull` is used. Up to you.

src/ostree.h
```diff
@@ -24,10 +24,12 @@ class OstreeManager : public PackageManagerInterface {
   boost::shared_ptr<OstreeDeployment> getStagedDeployment();
   static boost::shared_ptr<OstreeSysroot> LoadSysroot(const boost::filesystem::path &path);
   static bool addRemote(OstreeRepo *repo, const std::string &url, const KeyManager &keys);
-  static data::InstallOutcome pull(const Config &config, const KeyManager &keys, const std::string &refhash);
+  statuc data::InstallOutcome OstreeManager::find_or_pull(const Config &config, const KeyManager &keys,
```
patrickvacek: `statuc`?

patrickvacek: The `OstreeManager::` is redundant, too.

OYTIS: Sorry, compiled with BUILD_OSTREE=OFF before pushing.

src/ostree.cc
```diff
@@ -55,6 +55,27 @@ data::InstallOutcome OstreeManager::pull(const Config &config, const KeyManager
   return data::InstallOutcome(data::OK, "Pulling ostree image was successful");
 }
 
+data::InstallOutcome OstreeManager::find_or_pull(const Config &config, const KeyManager &keys,
```
patriotyk: CamelCase

patrickvacek:
I think this looks fine, but I'm getting an error when trying to update my device. This could be user error. Have you gotten it to work successfully yet?

OYTIS:
Wrong libostree function, sorry.

OYTIS:
Now that should work. Tested on rocko with autoprovisioning.

PR #568, opened 2018-02-13
Fix a memory leak.

patrickvacek:


cajun-rat:
Approved (subject to CI passing)

PR #570, opened 2018-02-13
Use statements + more helpers to refactor sql storage

lbonn:
This is quite big, sorry

### DIFF
--- a/src/sql_utils.h
+++ b/src/sql_utils.h
@@ -5,6 +5,8 @@
 
 #include <sqlite3.h>
 
+#include "logging.h"
+
 // Unique ownership SQLite3 statement creation
 template <typename T>
 static void bindArguments(sqlite3* db, sqlite3_stmt* statement, int cnt, int v) {
@@ -34,7 +36,7 @@ struct SQLBlob {
 
 template <typename T>
 static void bindArgument(sqlite3* db, sqlite3_stmt* statement, int cnt, const SQLBlob& blob) {
-  if (sqlite3_bind_blob(statement, cnt, blob.content.c_str(), blob.content.size(), NULL) != SQLITE_OK) {
+  if (sqlite3_bind_blob(statement, cnt, blob.content.c_str(), blob.content.size(), SQLITE_STATIC) != SQLITE_OK) {
     LOG_ERROR << "Could not bind: " << sqlite3_errmsg(db);
     throw std::runtime_error("SQLite bind error");
   }
--- a/src/sql_utils.h
+++ b/src/sql_utils.h
@@ -86,6 +86,12 @@ class SQLite3Guard {
     handle_.reset(h);
   }
 
+  int exec(const char* sql, int (*callback)(void*, int, char**, char**), void* cb_arg) {
+    return sqlite3_exec(handle.get(), sql, callback, cb_arg, NULL);
+  }
+
+  std::string errmsg() const { return sqlite3_errmsg(handle.get()); }
+
   template <typename... Types>
   SQLiteStatement prepareStatement(const char* zSql, Types... args) {
     return SQLiteStatement::prepare(handle_.get(), zSql, args...);
--- a/src/sql_utils.h
+++ b/src/sql_utils.h
@@ -22,12 +22,30 @@ static void bindArgument(sqlite3* db, sqlite3_stmt* statement, int cnt, const ch
   }
 }
 
+template <typename T>
+static void bindArgument(sqlite3* db, sqlite3_stmt* statement, int cnt, const std::string& v) {
+  bindArgument<const char*>(db, statement, cnt, v.c_str());
+}
+
+struct SQLBlob {
+  const std::string& content;
+  SQLBlob(const std::string& str) : content(str) {}
+};
+
+template <typename T>
+static void bindArgument(sqlite3* db, sqlite3_stmt* statement, int cnt, const SQLBlob& blob) {
+  if (sqlite3_bind_blob(statement, cnt, blob.content.c_str(), blob.content.size(), NULL) != SQLITE_OK) {
+    LOG_ERROR << "Could not bind: " << sqlite3_errmsg(db);
+    throw std::runtime_error("SQLite bind error");
+  }
+}
+
 struct SQLZeroBlob {
   size_t size;
 };
 
 template <typename T>
-static void bindArgument(sqlite3* db, sqlite3_stmt* statement, int cnt, SQLZeroBlob& blob) {
+static void bindArgument(sqlite3* db, sqlite3_stmt* statement, int cnt, const SQLZeroBlob& blob) {
   if (sqlite3_bind_zeroblob(statement, cnt, blob.size) != SQLITE_OK) {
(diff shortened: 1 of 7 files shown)


lbonn:
The use of `SELECT` + callback could be replaced by statements + `sqlite3_column_xxx` as well.

src/sql_utils.h
```diff
@@ -0,0 +1,123 @@
...
+    sqlite3* h;
+    rc = sqlite3_open(path, &h);
+    handle.reset(h);
+  }
+
+  int exec(const char* sql, int (*callback)(void*, int, char**, char**), void* cb_arg) {
+    return sqlite3_exec(handle.get(), sql, callback, cb_arg, NULL);
+  }
+
+  const char* errmsg() const { return sqlite3_errmsg(handle.get()); }
```
patriotyk: why return type is not std::string?

lbonn: good call, thanks

src/sql_utils.h
```diff
@@ -0,0 +1,123 @@
...
+  std::string errmsg() const { return sqlite3_errmsg(handle.get()); }
+
+  template <typename... Types>
+  SQLiteStatement prepareStatement(const char* zSql, const Types&... args) {
+    return SQLiteStatement::prepare(handle.get(), zSql, args...);
+  }
+
+ private:
+  std::unique_ptr<sqlite3, int (*)(sqlite3*)> handle;
+  int rc;
```
patriotyk: add underscore for private memebres

patrickvacek: Is that actually a strict rule for us? We don't apply it very consistently. Google's style guide has all class member variables add a trailing underscore. We certainly don't do that. I'm okay with any version, but consistency would be nice.

cajun-rat: We should certainly apply it to new code, if only because 'just following the style guide' is less effort then deciding which bits we like

lbonn: In this case, I kept the case of the original class (which was moved from one file to another).

I refreshed it to follow the right™ way

src/sql_utils.h
```diff
@@ -0,0 +1,123 @@
...
+}
+
+struct SQLBlob {
+  const std::string& content;
+  SQLBlob(const std::string& str) : content(str) {}
+};
+
+template <typename T>
+static void bindArgument(sqlite3* db, sqlite3_stmt* statement, int cnt, const SQLBlob& blob) {
+  if (sqlite3_bind_blob(statement, cnt, blob.content.c_str(), blob.content.size(), NULL) != SQLITE_OK) {
```
cajun-rat: I think NULL should be SQLITE_STATIC?  https://sqlite.org/c3ref/bind_blob.html

src/sql_utils.h
```diff
@@ -0,0 +1,123 @@
...
+  (void)cnt;
+}
+
+template <typename T, typename... Types>
+static void bindArguments(sqlite3* db, sqlite3_stmt* statement, int cnt, const T& v, const Types&... args) {
+  bindArgument<const T&>(db, statement, cnt, v);
+  bindArguments<const Types&...>(db, statement, cnt + 1, args...);
+}
+
+class SQLiteStatement {
```
cajun-rat: Should this be boost::noncopyable?

lbonn: It is actually enforced automatically because the class has a non-copyable member (unique_ptr...).

But it might be nice to use it anyway as some kind of annotation.

patrickvacek: No, it should be constructor = delete since we have C++11! No need for boost.

src/sql_utils.h
```diff
@@ -0,0 +1,123 @@
+#ifndef SQL_UTILS_H_
+#define SQL_UTILS_H_
+
+#include <memory>
+
```
cajun-rat: Could you #include "logging.h" here?  It will at least make my IDE happier :)

src/sql_utils.h
lbonn: So, should we mandate that the copy constructor and assignment operator be deleted explicitly even if that's already done by the compiler?
(see: https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types)

cajun-rat:
Do we need SQLite3Guard?

Looking at https://sqlite.org/lockingv3.html I think that sqlite handles multi-reader single writer locking automatically.  I tried the following in two shells using the sqlite3 command line tool:

### Terminal  1
~~~
sqlite3 /tmp/foo
SQLite version 3.16.2 2017-01-06 16:32:41
Enter ".help" for usage hints.
sqlite> .schema
CREATE TABLE foo (id int);
sqlite> select * from foo;
sqlite> select * from foo;
5
sqlite> begin;
sqlite> insert into foo (id) VALUES (120);
sqlite> commit;
sqlite> 
~~~

### Terminal 2
~~~
sqlite3 /tmp/foo
SQLite version 3.16.2 2017-01-06 16:32:41
Enter ".help" for usage hints.
sqlite> begin;
sqlite> create table foo (id int);
sqlite> commit;
sqlite> insert into foo (id) VALUES (5);
sqlite> select * from foo;
5
sqlite> select * from foo;
5
sqlite> select * from foo;
5
120
~~~

src/sql_utils.h
patrickvacek: I don't have a firm opinion here. If you know the compiler is already doing the right thing, then it probably doesn't matter except for making it obvious to anyone reading the code.

src/sqlstorage.cc
```diff
@@ -380,16 +359,19 @@ void SQLStorage::storeMetadata(const Uptane::MetaPack& metadata) {
...
+
+  auto statement = db.prepareStatement<SQLBlob, SQLBlob, SQLBlob, SQLBlob, SQLBlob, SQLBlob>(
+      "INSERT INTO meta VALUES (?,?,?,?,?,?);", jsons[0], jsons[1], jsons[2], jsons[3], jsons[4], jsons[5]);
 
-  if (sqlite3_exec(db.get(), req.c_str(), NULL, NULL, NULL) != SQLITE_OK) {
-    LOG_ERROR << "Can't set metadata: " << sqlite3_errmsg(db.get());
+  if (statement.step() != SQLITE_DONE) {
+    LOG_ERROR << "Can't set metadata: " << db.errmsg();
     return;
   }
```
cajun-rat: The sync() call here is unnecessary - sqlite should be handling this for us

lbonn:
@cajun-rat I'm not really sure I follow you here. In your example, your two shells use different connections, just like what we're doing right now. If we want to have clean parallel calls to the storage api, isn't it recommended? In addition to that, we'd have to get rid of the unsafe fields `request`, `req_response` along with the callback mechanism.

As I understood it, the purpose of SQLite3Guard is not to do some manual locking but just to ensure that the sql connections are closed properly in a RAII manner.

src/sql_utils.h
lbonn: your ide is right :), fixed

cajun-rat:
Tested with a update on qemu and oe-selftest.  I also manually checked that aktualizr-info gave sane responses.

PR #571, opened 2018-02-13
Selfupdate with docker

patriotyk:


patrickvacek:
I think you should leave ostreesysroot alone -- I think your PR is reverting a change I made to it, which would be annoying at best and break the tests at worst.

cajun-rat:
(rebased)

scripts/test_aktualizr_deb_and_update.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
+set -e
+
+/persistent/selfupdate_server.py 8000&
+
+dpkg-deb -I /persistent/aktualizr.deb && dpkg -i /persistent/aktualizr.deb
+aktualizr --version | grep $(cat /persistent/aktualizr-version) && aktualizr-info
```
cajun-rat: Does this build a version of Aktualizr with the correct --version?

On my machine I tried the following:

~~~
# docker build -t advancedtelematic/aktualizr-noostree -f Dockerfile.noostree  .
# docker run -v /persistent:/persistent advancedtelematic/aktualizr-noostree src/scripts/build-ubuntu.sh
# dpkg -i /persistent/aktualizr.deb
# aktualizr --version
Current aktualizr version is:
~~~

patriotyk: It should, but I found another issue, looks like after rebase dissapeared `-DBUILD_DEB=ON ` from src/scripts/build-ubuntu.sh.

patriotyk: But CI logs shows that aktualizr built with debian support otherwise it should fail

patriotyk:
Fixed in apropriate(in my opinion) way.

cajun-rat:
I can't make this work.  I tried it on a Debian VM on AWS, and wasn't able to update the package.  Does it work for you?

cajun-rat:
Thank you the explanation.  The problem I was seeing before was only with _downgrades_: The old version can install the new version OK, but not the other way round.  I think this is fine.

One thing that is missing is that the Aktualizr service needs to be manually restarted.  This is actually the same behavior that we have for OSTree updates.  I think the right solution here would be to treat both of these cases in a similar way: once the update is 'complete', we need an option to restart the system (which might be the whole OS, or just the Aktualizr service)

PR #572, opened 2018-02-13
Option to run tests in parallel

cajun-rat:
This adds an option to configure the number of tests that are run in parallel. It can be set on the cmake command line or by editing CMakeCache.txt. The default is '1', because there isn't much ram on the CI server.

Change-Id: I80898affa3607ab79e8e5f1f19574a07c9c79885

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ option(BUILD_SOTA_TOOLS "Set to ON to build SOTA tools" OFF)
 option(BUILD_ISOTP "Set to ON to build ISO-TP" OFF)
 option(BUILD_OPCUA "Set to ON to compile with OPC-UA protocol support" OFF)
 option(INSTALL_LIB "Set to ON to install library and headers" OFF)
+set(TEST_JOBS "1" CACHE STRING "Number of parallel test jobs to run in 'make qa'")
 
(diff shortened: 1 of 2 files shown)


patrickvacek:
This is only changing what happens for the uptane_vector_tests. Do you also want to change the values used in the Dockerfiles with the scripts?

cajun-rat:
The docker file stuff is run outside of cmake, so we'd have to set it up in a different way, though?

patrickvacek:
After further investigation, clearly I was wrong. This does work. But why default to 1? Soooo sloooow....

I'm just not quite convinced that the solution to "CI tests are slow" is "run in series instead of parallel".

cajun-rat:
I was worried that the existing CI process will fail if they are run in parallel

patrickvacek:
> I was worried that the existing CI process will fail if they are run in parallel

I'm skeptical that this will help. It seems like a crapshoot right now.

patriotyk:
To work it properly, You need to remove CTEST_PARALLEL_LEVEL variable from `scripts/test.sh` and `scripts/coverage.sh`

patrickvacek:
> To work it properly, You need to remove CTEST_PARALLEL_LEVEL variable from scripts/test.sh and scripts/coverage.sh

That was my thought, too, but when I actually looked into it, it didn't seem to be true.

PR #573, opened 2018-02-13
Revert "Generate 4096 bits RSA keys"

cajun-rat:
This reverts commit 66e846982494301ead72e6e940c0ed496fecb340.

See PRO-4851 for more discussion. 4096 bit RSA keys are about 4x slower than 2048, and this gets _really slow_ when the tests are run under valgrind:

time valgrind t_uptane
4096 => 4m33,349s
2048 => 0m28,607s
1024 => 0m12,138s

### DIFF
--- a/src/crypto.cc
+++ b/src/crypto.cc
@@ -257,11 +257,11 @@ bool Crypto::extractSubjectCN(const std::string &cert, std::string *cn) {
  *         false if key generation failed
  */
 bool Crypto::generateRSAKeyPair(std::string *public_key, std::string *private_key) {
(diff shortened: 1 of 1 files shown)


cajun-rat:
(We need to check that this really does make the CI build faster)

patriotyk:
But I think we need to use cmake variable to setup key length, and use the smalles for the tests

cajun-rat:
The build step that was slow is now 18minutes :)

cajun-rat:
Created PRO-4859 to fix this properly

PR #575, opened 2018-02-13
Remove Dockerfile.garage-tools since it is not used anymore

jochenschneider:
Was only used by the e2e tests, but they use the normal bitbake process
now.


cajun-rat:
Less code => +1

PR #577, opened 2018-02-14
Try to run clang-tidy

lbonn:
Outputs a lot of warnings right now. Maybe we'd like to review the list and enable/fix a subset of them.


lbonn:
Ok, trying to run with every warnings and fixing them was not scalable. I enabled a small subset and made the changes (this is already quite useful IMO).

A lot of the stuff can help us migrating to a clean c++11 style. Note that a lot of the checks can apply corrections for you! (though it's sometimes buggy)

scripts/static-checks.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+./src/scripts/setup_hsm.sh
+
+mkdir -p build-static-checks
+cd build-static-checks
+cmake -DBUILD_OSTREE=ON -DBUILD_SOTA_TOOLS=ON -DBUILD_DEB=ON -DBUILD_P11=ON -DBUILD_WITH_CODE_COVERAGE=ON ..
```
patrickvacek: Is building with code coverage here necessary? Seems like it'll just slow the build down.

lbonn:
CI should pass any minute now.

Notes:
- clang-format was rolled back to 3.8 because 6.0 is stricter and wasn't happy with our current code base
- can clang-tidy include all of clang-check stuff? Maybe we only need to run clang-tidy

patrickvacek:
Were you still doing some work here? It looks good to me so far. You should maybe make some subtasks of Sweat the Small Stuff #2 to add some of the flags you've left off for now. I don't want to lose sight of that if it's still worth doing sooner or later.

lbonn:
@patrickvacek yeah, sure

CMakeLists.txt
```diff
@@ -98,9 +98,9 @@ add_dependencies(qa format)
 add_custom_target(build_tests)
 
 # clang-check and clang-format
-# The .clang-format file requires clang-format-3.8
 find_program(CLANG_FORMAT NAMES clang-format-3.8 clang-format)
-find_program(CLANG_CHECK NAMES clang-check-3.8 clang-check)
+find_program(CLANG_CHECK NAMES clang-check-6.0 clang-check-3.8 clang-check)
+find_program(CLANG_TIDY NAMES clang-tidy-6.0 clang-tidy-3.8 clang-tidy)
```
patrickvacek: Just a word of warning: previously we didn't push past `clang-check-3.8` because we found that between 3.8 and 4 there were some annoying situations where one version forces one thing and the other forces another thing, which meant if you didn't use the same as what the server had, you could almost never get CI to pass. For now, I'm happy to give it a go and see if we have problems again.

lbonn: I see, sound scary...

The recent versions should feature more checks but if they are too painful to install on our ubuntu, we might have to roll them back at some point.

PR #578, opened 2018-02-14
Use an env file in .deb package systemd service file

tkfu:
The debian package blows away any changes to the systemd service
file, so we want to make it possible to add command line options
for secondaries or other custom configs.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -105,6 +105,12 @@ You can then build the project from the `build` directory using Make:
 make
 ----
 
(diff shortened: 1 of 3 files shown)


OYTIS:
Yes, `/usr/lib/sota` is a more appropriate place in my opinion. Otherwise looks good to me, thanks.

tkfu:
Changed. Docs ok?

cajun-rat:
Approved, assuming CI passes.

PR #579, opened 2018-02-14
OPC-UA bridge: minimal test for secondary update

riqkum:


cajun-rat:
Looks good.  Ran tests with 'ninja qa' and oe-selftest -r.

PR #580, opened 2018-02-14
Add test to check migration version

patriotyk:


patrickvacek:
I think it's also worth testing, as Phil suggested, that there are no further migration scripts to be found. As in, if the db version is 3 but there is a .4 migration available, something went wrong.

cajun-rat:
This at least detects the last bug that we hit, merging.

PR #581, opened 2018-02-14
Make name generator a standalone utility

OYTIS:
Maybe it's not a horrible idea. Would be useful in "mock real implicit provisioning" scenario done with openssl utility.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,6 +229,7 @@ add_subdirectory("src")
 add_subdirectory("src/cert_provider")
 add_subdirectory("src/implicit_writer")
 add_subdirectory("src/external_secondaries")
+add_subdirectory("src/name_generator")
 add_subdirectory("fuzz")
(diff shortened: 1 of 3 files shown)


patrickvacek:
Who else actually needs to use the name generator? Can't we just use it from inside cert_provider?

OYTIS:
@patrickvacek Put openssl logic into cert_provider is more difficult that get name generation out of it. So my current plan is just use 'openssl' utility and give it a name as a parameter. The same problem can also be solved by adding something like `--name-only`, maybe it's a better way.

patrickvacek:
I think I was thinking you could just call the openssl utility from cert_provider with a system call (...or a series of system calls). A bit awkward, but probably no worse than other options. Up to you.

OYTIS:
@patrickvacek OK, it's not a real CA anyway, so we can probably do something simple in plain C without the utility.

PR #582, opened 2018-02-15
Flush log messages

cajun-rat:
This might not be the best solution, but without it the systemd journal is useless.

Change-Id: I8c5f2ca0d76f69238bf21acbb7e224424ef8002e

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -13,6 +13,7 @@ void logger_init() {
   gLoggingThreshold = boost::log::trivial::info;
   logging::add_console_log(std::cout, boost::log::keywords::format = "%Message%");
   boost::log::core::get()->set_filter(boost::log::trivial::severity >= gLoggingThreshold);
+  std::cout << std::unitbuf;  // Ensure that log messages are flushed to the journal
 }
 
 void logger_set_threshold(severity_level threshold) {


patrickvacek:
It's complaining about unformatted code.

src/logging.cc
```diff
@@ -13,6 +13,7 @@ void logger_init() {
   gLoggingThreshold = boost::log::trivial::info;
   logging::add_console_log(std::cout, boost::log::keywords::format = "%Message%");
```
patriotyk: It is better to add true(auto_flush) as third parameter of `add_console_log`
http://www.boost.org/doc/libs/1_54_0/libs/log/doc/html/boost/log/add_console_lo_idp21543664.html

patriotyk: I think should be `logging::add_console_log(std::cout, boost::log::keywords::format = "%Message%", boost::log::keywords::auto_flush = true);`

patriotyk: Oh, it is already merged (((

cajun-rat: I still like your solution though

cajun-rat: I've opened a PR for your better solution: #598

PR #584, opened 2018-02-15
Allow double writes of targets in sql storage

lbonn:
Theoretically, two targets with the same file names should be identical but for now, just write the file content again.

### DIFF
--- a/src/sqlstorage.cc
+++ b/src/sqlstorage.cc
@@ -726,7 +726,8 @@ class SQLTargetWHandle : public StorageTargetWHandle {
     }
 
     auto statement = db_.prepareStatement<std::string, SQLZeroBlob>(
-        "INSERT INTO target_images (filename, image_data) VALUES (?, ?);", filename_, SQLZeroBlob{expected_size_});
+        "INSERT OR REPLACE INTO target_images (filename, image_data) VALUES (?, ?);", filename_,
+        SQLZeroBlob{expected_size_});
 
     if (statement.step() != SQLITE_DONE) {
       LOG_ERROR << "Statement step failure: " << db_.errmsg();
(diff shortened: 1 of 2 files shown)


src/sqlstorage.cc
```diff
@@ -725,6 +725,26 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
+    if (nfiles != 0) {
+      int ret = db_.prepareStatement<std::string>("DELETE FROM target_images WHERE filename = ?", filename_).step();
+      if (ret != SQLITE_DONE) {
+        LOG_ERROR << "Statement step failure: " << db_.errmsg();
+        throw exc;
+      }
+    }
+
     auto statement = db_.prepareStatement<std::string, SQLZeroBlob>(
         "INSERT INTO target_images (filename, image_data) VALUES (?, ?);", filename_, SQLZeroBlob{expected_size_});
```
cajun-rat: @OYTIS  suggests that ON CONFLICT REPLACE might simplify this https://sqlite.org/lang_conflict.html

tests/storage_common_test.cc
```diff
@@ -240,6 +240,15 @@ TEST(storage, store_target) {
...
 
+  // write again
+  {
+    std::unique_ptr<StorageTargetWHandle> fhandle = storage->allocateTargetFile(false, "testfile", 2);
+    const uint8_t wb[] = "ab";
+    fhandle->wfeed(wb, 1);
+    fhandle->wfeed(wb + 1, 1);
+    fhandle->wcommit();
+  }
+
```
cajun-rat: +1 For a test case :)

cajun-rat:
(Approved subject to CI build...)

PR #590, opened 2018-02-15
Partial fix for PRO-4872

cajun-rat:
This is a partial fix that makes self-update on Aktualizr work, although we
should build a proper solution soon.  It does at least work end-to-end, and
unblocks the remaining testing.

Change-Id: I03d8c1dd461fd21c244145f641d808e8bbdfd450

### DIFF
--- a/src/deb.cc
+++ b/src/deb.cc
@@ -30,16 +30,25 @@ Json::Value DebianManager::getInstalledPackages() {
 
 data::InstallOutcome DebianManager::install(const Uptane::Target &target) const {
   LOG_INFO << "Installing " << target.filename() << " as Debian package...";
-  std::string cmd = "dpkg -i ";
-  std::string output;
-  TemporaryDirectory package_dir("deb_dir");
-  std::stringstream sstr;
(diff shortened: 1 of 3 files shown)


patriotyk:
Looks like it is wrong solution. Have You tried to remove prerm and preinst scripts? It requires to remove them from both  just builded deb and uploaded to ats garage deb

cajun-rat:
The problem with removing prerm scripts is that we will still need a way to restart the Aktualizr service.  Have you tried the prerm approach, and does it work?

I'm happy with a better solution, but the problem we have at the moment is that we need to ship this code out tomorrow, and so we need something _soon_.

patriotyk:
Oh sorry, now I understand Your email. I am working on ED25519 and could switch to this issue, but I didn't know that it is the highest priority

patriotyk:
Just noticed, that the current CI build doesn't run selfupdate test. It runs in docker where we test aktualizr.deb generation

patriotyk:
Oh sorry, my PR has not merged yet

PR #591, opened 2018-02-15
Add option to sign certificates with CA

OYTIS:
meta-updater code to test that is following.
Changes to cert_provider/aktualizr to write/read autoprov.url to/from a file are following as well


patrickvacek:
I think this looks good. Is it worth doing some basic tests inside the aktualizr repo? The only reason I didn't before was because I considered a mere test utility itself.

OYTIS:
@patrickvacek Yes, sounds like a good idea, I'll add some.

src/utils.h
```diff
@@ -53,4 +53,19 @@ class TemporaryDirectory : boost::noncopyable {
   boost::filesystem::path tmp_name_;
 };
 
+// helper template for C (mostly openssl) data structured
+//   user should still take care about the order of destruction
+//   by instantiating StructGuard<> in a right order.
+//   BTW local variables are destructed in reverse order of instantiation
+template <typename T, void (*destroy)(T *)>
```
lbonn: `unique_ptr` can serve the same purpose and is also non-copyable (see the second optional template argument http://en.cppreference.com/w/cpp/memory/unique_ptr).

`StructGuard` could be a simple wrapper to `unique_ptr` as well.

OYTIS: @lbonn Cool, thank you.

OYTIS:
No tests (yet).

patrickvacek:
While playing around, I tried this: `aktualizr_cert_provider --bits 15 --days 0 -l temp --credentials /home/patrick/Documents/creds/patrick+test-prod.zip --certificate-c abc --cerficate-st def --certificate-o ghi --certificate-cn jkl` and all I got was the list of commandline options, but no indication of what actually went wrong. I think we should also restrict the bit size options, although maybe we can wait on that until that has also been implemented in aktualizr. Settings `days` to `0` is absurd but probably doesn't need a special check.

patrickvacek:
I'm seeing an error when I try to use this:
```
./src/cert_provider/aktualizr_cert_provider --credentials /home/patrick/Documents/creds/patrick+test-prod.zip --device-ca ~/Code/root-ca/ca.cert.pem --device-ca-key ~/Code/root-ca/ca.key.pem --bits 15 --days 0 --certificate-c 123 --certificate-st 456 --certificate-o 789 --certificate-cn abc --local temp
PEM_read_bio_PrivateKey failed: error:0906B072:lib(9):func(107):reason(114)
```
Could this be because my CA key has a password?

OYTIS:
@patrickvacek I love OpenSSL error messages.
Yes, I was testing without a password, probably need an option for it as well. Some sanity check on C/ST/O/CN fields may also be required, at least I found out they can't be set to empty strings.

patrickvacek:
In my last example, I had a key without a password and all the strings were non-empty; any idea what went wrong?

OYTIS:
@patrickvacek No. Maybe they require that C is a proper country code or something. I'll check.

patrickvacek:
Ha, I didn't actually consider that sane input might be obligatory. I suppose that would make sense (if that's the problem).

OYTIS:
@patrickvacek Yes, country code can't be longer that 2 characters. Will add a check.

patrickvacek:
I tried a two-letter country code and now I'm getting another error:
```
./src/cert_provider/aktualizr_cert_provider --credentials /home/patrick/Documents/creds/patrick+test-prod.zip --device-ca ~/Code/root-ca/ca2.cert.pem --device-ca-key ~/Code/root-ca/ca2.key.pem --bits 15 --days 0 --certificate-c AA --certificate-st 456 --certificate-o 789 --certificate-cn abc --local temp
RSA_generate_key failed: error:04081078:lib(4):func(129):reason(120)
```

OYTIS:
Added simple test for cert_provider, writing/reading server url from/to file and human-readable OpenSSL errors.

OYTIS:
@patrickvacek Oh yes, key size should be at least 31bit long. We could try to fuzz it automatically as well.

PR #592, opened 2018-02-15
Remove sota-tools dockerfiles.

stevana:
The end-to-end tests are no longer using these dockerfiles, are you?


cajun-rat:
I don't think so.

patrickvacek:
@stevana For now I think it's best to keep the one Dockerfile we are using for testing. I reorganized things to hopefully make it a bit clearer what's actually going on. Hopefully it will also pass CI.

patrickvacek:
Rebased. Does this look okay to you @stevana and anyone else interested?

PR #593, opened 2018-02-15
Add support and switch between of rsa2048, rsa4096 and ed25519 keys

patriotyk:
Running all uptane related tests with command `ctest -R test_uptane` gives me the following results:
```
RSA2048 : 120 secs
ED255519: 60 secs
```
So looks like ED25519 is much faster then RSA2048. 
But server doesnt support it, so we can not use it by default yet.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -14,6 +14,7 @@
 #include <boost/uuid/uuid_io.hpp>
 
 #include "logging.h"
+#include "types.h"
 #include "uptane/secondaryconfig.h"
 
 enum ProvisionMode { kAutomatic = 0, kImplicit };
@@ -80,7 +81,8 @@ struct UptaneConfig {
         primary_ecu_hardware_id(""),
         director_server(""),
         repo_server(""),
-        key_source(kFile) {}
+        key_source(kFile),
+        key_type(kRSA2048) {}
   bool polling;
   unsigned long long polling_sec;
   std::string device_id;
@@ -89,6 +91,8 @@ struct UptaneConfig {
   std::string director_server;
   std::string repo_server;
   CryptoSource key_source;
+  KeyType key_type;
+  std::string getKeyTypeString() const { return (key_type == kED25519) ? "ED25519" : "RSA"; }
   std::vector<Uptane::SecondaryConfig> secondary_configs;
 };
 
--- a/src/crypto.cc
+++ b/src/crypto.cc
@@ -95,6 +95,22 @@ std::string Crypto::RSAPSSSign(ENGINE *engine, const std::string &private_key, c
   return retval;
 }
 
+std::string Crypto::Sign(KeyType key_type, ENGINE *engine, const std::string &private_key, const std::string &message) {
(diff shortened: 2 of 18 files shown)


src/crypto.cc
```diff
@@ -95,6 +95,22 @@ std::string Crypto::RSAPSSSign(ENGINE *engine, const std::string &private_key, c
   return retval;
 }
 
+std::string Crypto::Sign(KeyType key_type, ENGINE *engine, const std::string &private_key, const std::string &message) {
+  if (key_type == kED25519) {
+    std::cout << "PRIVKEY:" << private_key;
```
lbonn: Forgot a debugging print here

cajun-rat:
Let's hold off merging this until the Debian stuff is done and we have a working release of that.  I don't want to merge anything risky until we have this out the door.

src/uptane/secondaryconfig.h
```diff
@@ -22,6 +23,7 @@ class SecondaryConfig {
   bool partial_verifying;
   std::string ecu_private_key;
   std::string ecu_public_key;
+  KeyType key_type{kRSA2048};
```
patrickvacek: This is basically hardcoding all secondaries to use RSA 2048. That's not terrible, but we should maybe instead add the ability to set this via `Config::readSecondaryConfigs` and change our examples and test data to try it out.

src/config.h
```diff
@@ -79,7 +80,8 @@ struct UptaneConfig {
         primary_ecu_hardware_id(""),
         director_server(""),
         repo_server(""),
-        key_source(kFile) {}
+        key_source(kFile),
+        key_type(kRSA2048) {}
```
patrickvacek: I think we need to be able to set this via the config file. It needs to be handled in `Config::updateFromPropertyTree` and `Config::writeToFile`.

patriotyk:
Looks like there is no any benefits in speedin CI. So I will update PR with default ED keys for all tests to check if something change

patriotyk:
I think it is ok now. However there is no any speedup profit comparing to RSA2048 even when run all tests with ED25519. But on my local machine ED25519 performs much faster

cajun-rat:
Benchmark: 

## With these changes (3 runs)

~~~
ninja clean && time ninja qa

Total Test time (real) = 189.70 sec
real	4m26,985s
user	15m30,724s
sys	0m33,272s

Total Test time (real) = 170.85 sec
real	4m8,790s
user	15m10,832s
sys	0m34,140s

Total Test time (real) = 185.23 sec
real	4m23,467s
user	15m23,628s
sys	0m34,220s

~~~

## without (master): 

~~~
Total Test time (real) = 191.73 sec
real	4m29,087s
user	15m26,884s
sys	0m33,652s

Total Test time (real) = 198.52 sec
real	4m36,718s
user	15m38,312s
sys	0m34,048s

Total Test time (real) = 207.30 sec
real	4m45,438s
user	15m34,548s
sys	0m34,204s
~~~

PR #596, opened 2018-02-16
CMakeLists.txt: Add lshw as dependency of the deb

leon-anavi:
Add lshw as a dependency of Aktualizr deb package.


cajun-rat:
Can you diff the results of the package dependencies that dpkg reports with and without this change?  @lbonn did some magic that makes it autodetect some of this stuff, and I'll like to make sure that this doesn't interfere with that.

lbonn:
According to the docs it shouldn't but better be safe...: 

> Note
> If CPACK_DEBIAN_PACKAGE_SHLIBDEPS or more specifically CPACK_DEBIAN_<COMPONENT>_PACKAGE_SHLIBDEPS is set for this component, the discovered dependencies will be appended to CPACK_DEBIAN_<COMPONENT>_PACKAGE_DEPENDS instead of CPACK_DEBIAN_PACKAGE_DEPENDS. If CPACK_DEBIAN_<COMPONENT>_PACKAGE_DEPENDS is an empty string, only the automatically discovered dependencies will be set for this component. 

(https://cmake.org/cmake/help/latest/module/CPackDeb.html#variable:CPACK_DEBIAN_%3CCOMPONENT%3E_PACKAGE_DEPENDS)

The dependencies and general package infos will appear in the CI logs:
https://github.com/advancedtelematic/aktualizr/blob/master/Dockerfile.test-install.xenial#L17

leon-anavi:
Hi @cajun-rat, @lbonn,

With the change:
```
 Depends: libarchive13, libc6 (>= 2.22), libcurl3 (>= 7.16.2), libgcc1 (>= 1:3.0), libsodium18 (>= 0.6.1), libsqlite3-0 (>= 3.5.9), libssl1.0.0 (>= 1.0.0), libstdc++6 (>= 5.2), lshw
```

Without the change:
```
Depends: libarchive13, libc6 (>= 2.22), libcurl3 (>= 7.16.2), libgcc1 (>= 1:3.0), libsodium18 (>= 0.6.1), libsqlite3-0 (>= 3.5.9), libssl1.0.0 (>= 1.0.0), libstdc++6 (>= 5.2)
```

As you can see the only difference is that lshw is added to the list of dependencies.

Best regards, Leon

cajun-rat:
This is failing the CI because the lshw package isn't installed.  That is easy to fix, but we need to figure out how we will present this to users. @tkfu What do the docs say about installing prerequisites at the moment?

cajun-rat:
@leon-anavi Talking at the stand-up today, we think the best solution is to add lshw to the dependencies in Dockerfile.test-install.xenial.  Can you make that change (which should fix the CI build), please?

leon-anavi:
Hi @cajun-rat, I have added lshw to Dockerfile.test-install.xenial. Best regards, Leon

patrickvacek:
@leon-anavi Looks like there's a merge conflict that came up somehow.

leon-anavi:
@patrickvacek ah, yes, actually the 2nd patch is not required anymore. Git rev e2a79b19d47628f0b2846b48daef8435367dc198 has already added lshw to the list of dependencies in Dockerfile.test-install.xenial and it has been merged in master.

I've deleted my 2nd commit. Leaving just the one that adds lshw as a dependency of the deb.

Best regards,
Leon

PR #601, opened 2018-02-20
Make cert_provided use the storage path in config

lbonn:
And disallow specifying a target directory along with a config!


src/cert_provider/main.cc
```diff
@@ -319,6 +319,16 @@ int main(int argc, char* argv[]) {
     // Strip any relative directories. Assume everything belongs in one
     // directory for now.
     // TODO: provide path to root directory in `--local` parameter
+
+    if (!commandline_map["directory"].empty() && !commandline_map["config"].empty()) {
+      std::cerr << "Directory (--directory) and config (--config) options cannot be used together" << std::endl;
+      return 1;
+    }
+    if (!commandline_map["directory"].empty())
+      directory = commandline_map["directory"].as<boost::filesystem::path>();
```
patrickvacek: The --config flag works nicely now, but the --directory flag doesn't, because this line is unreachable!

lbonn: Wow, thanks!

lbonn:
@patrickvacek should be ready now

PR #605, opened 2018-02-21
Fix Uptane::Target comparison operator

lbonn:
Was considering two targets with a common hash as equals

I'm not entirely sure this was a bug. I'll push tests for this if you can confirm this is what it should do.

### DIFF
--- a/src/packagemanagerfactory.h
+++ b/src/packagemanagerfactory.h
@@ -4,6 +4,7 @@
 #include <boost/make_shared.hpp>
 #include <boost/shared_ptr.hpp>
 
+#include "invstorage.h"
 #include "logging.h"
 #include "packagemanagerfake.h"
 #include "packagemanagerinterface.h"
--- a/src/uptane/tuf.h
+++ b/src/uptane/tuf.h
@@ -124,14 +124,21 @@ class Target {
   };
 
   bool operator==(const Target &t2) const {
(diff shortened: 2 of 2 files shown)


cajun-rat:
I'm not sure if this is the behavior that we need.  I'll need to check, but I think we use the == operator to mean 'The same target according to Uptane', which means that I think we need only one hash (that we trust) to match.

OYTIS:
I think the original behaviour made sense. We can imagine repo and director metadata having a slightly different set of hashes. As long as one of them matches we're fine.
I can imagine an attack though where adversary controls the director, but not repo or primary. Then there are two hashes in repo/targets.json for a secondary firmware: sha256 and sha512. Adversary knows that the secondary can only check sha512, so he just forges the director metadata where sha256 matches that of repo metadata and replaces sha512 with whatever he needs. Primary will consider it valid, since there is one matching hash and the partial secondary will only check sha512 on director's metadata and install whatever adversary wants.
So we probalby need a smarter check, like all hashes of the same type should match, and there should be at least one hash type that is present in both director and repo targets.json.

cajun-rat:
Good point @OYTIS .  I had been thinking about a check that would require one matching hash, but your attack breaks that.

We should document these requirements and the attacks we've thought about in the code for operator==()

OYTIS:
Looks good to me, thank you.

PR #606, opened 2018-02-22
garage-check: Handle \n's in tufrepo.url

doanac:
The tufrepo.url file in my credentials.zip somehow had a trailing \n
making my bitbake's call to garage-check very difficult to debug.


cajun-rat:
I'm not sure why the CI build failed the first time.  It was complaining about some expired metadata, which was a bug that was fixed a long time ago.  I guess that Travis was building something other than your PR.

Thank you for the fix!

PR #608, opened 2018-02-22
Fix some atomicity issues

lbonn:
Try to avoid the unsafe pattern:

```
remove(file)
<-- bad things can happen here
write(file, newContent)
```

Both in SQL db and on the file system.

Note that problems can still occur if (for example) a public/private key pair is updated partially. This is difficult to solve on the file system but it could be desirable to have a safe `storePair()` in the SQL case and removing `storePublic()` and `storePrivate()`

### DIFF
--- a/src/sql_utils.h
+++ b/src/sql_utils.h
@@ -119,7 +119,24 @@ class SQLite3Guard {
     return sqlite3_exec(handle_.get(), sql, callback, cb_arg, NULL);
   }
 
+  template <typename... Types>
+  SQLiteStatement prepareStatement(const std::string& zSql, const Types&... args) {
+    return SQLiteStatement(handle_.get(), zSql, args...);
+  }
+
+  std::string errmsg() const { return sqlite3_errmsg(handle_.get()); }
+
+  // Transaction handling
+  //
+  // A transactional series of db operations should be realized between calls of
+  // `beginTranscation()` and `commitTransaction()`. If no commit is done before
+  // the destruction of the `SQLite3Guard` (and thus the SQLite connection) or
+  // if `rollbackTransaction()` is called explicitely, the changes will be
+  // rolled back
+
   bool beginTransaction() {
+    // Note: transaction cannot be nested and this will fail if another
+    // transaction was open on the same connection
     int ret = exec("BEGIN TRANSACTION;", NULL, NULL);
     if (ret != SQLITE_OK) {
       LOG_ERROR << "Can't begin transaction: " << errmsg();
@@ -141,13 +158,6 @@ class SQLite3Guard {
       LOG_ERROR << "Can't rollback transaction: " << errmsg();
     }
     return ret == SQLITE_OK;
-  }
-
-  std::string errmsg() const { return sqlite3_errmsg(handle_.get()); }
-
-  template <typename... Types>
-  SQLiteStatement prepareStatement(const std::string& zSql, const Types&... args) {
-    return SQLiteStatement(handle_.get(), zSql, args...);
   }
 
  private:
--- a/src/sql_utils.h
+++ b/src/sql_utils.h
@@ -119,6 +119,30 @@ class SQLite3Guard {
     return sqlite3_exec(handle_.get(), sql, callback, cb_arg, NULL);
   }
 
+  bool beginTransaction() {
+    int ret = exec("BEGIN TRANSACTION;", NULL, NULL);
+    if (ret != SQLITE_OK) {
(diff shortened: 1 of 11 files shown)


patrickvacek:
I think this looks good, but I'm no SQLite expert, and @heartsucker should probably review as well.

src/sql_utils.h
```diff
@@ -109,6 +109,30 @@ class SQLite3Guard {
     return sqlite3_exec(handle_.get(), sql, callback, cb_arg, NULL);
   }
 
+  bool beginTransaction() {
```
heartsucker: This is fine, but it doesn't allow nesting since you're not using named savepoints allowing for `ROLLBACK TRANSACTION TO SAVEPOINT <savepoint>`.

heartsucker: And you might not need that, but just noting it here.

lbonn: Thanks for the info. I did a quick test and SQLite indeed shouts when two `BEGIN TRANSACTION` are called in a row.

I'll add a comment about that

lbonn:
I also removed `storePublic` and `storePrivate` and made `storePrimaries` atomic for sql storage (might as well do that now)

src/sqlstorage.cc
```diff
@@ -43,45 +38,18 @@ void SQLStorage::storePrimaryPublic(const std::string& public_key) {
...
   const char* req;
   if (boost::lexical_cast<int>(req_response["result"])) {
-    req = "UPDATE OR REPLACE primary_keys SET public = ?;";
+    req = "UPDATE OR REPLACE primary_keys SET (public, private) = (?,?);";
   } else {
-    req = "INSERT INTO primary_keys(public) VALUES (?);";
+    req = "INSERT INTO primary_keys(public,private) VALUES (?,?);";
   }
-  auto statement = db.prepareStatement<std::string>(req, public_key);
+  auto statement = db.prepareStatement<std::string>(req, public_key, private_key);
```
cajun-rat: Can't we use INSERT OR REPLACE INTO here?

cajun-rat: Ah, I see. Having a dummy column in the schema feels a bit weird (not say it is wrong, just weird), so the current solution is probably fine, for now at least

lbonn: Yes but encoding that the table has only one row in the schema could be nice for added safety/self documentation.

The second example here https://stackoverflow.com/a/33104119 looks nice.

src/sqlstorage.cc
```diff
@@ -347,7 +315,17 @@ bool SQLStorage::loadTlsPkey(std::string* pkey) {
...
+  if (db.get_rc() != SQLITE_OK) {
+    LOG_ERROR << "Can't open database: " << db.errmsg();
+    return;
+  }
+
+  if (!db.beginTransaction()) return;
+
+  if (db.exec("DELETE FROM meta;", NULL, NULL) != SQLITE_OK) {
+    LOG_ERROR << "Can't clear meta: " << db.errmsg();
+    return;
```
patrickvacek: What happens in a place like this when there is an early return? Does the state created by the beginTransaction get cleared despite that the commitTransaction never gets called?

Otherwise I think this looks fine.

patrickvacek:
Also, if there's any way to test that the transactional business works as you expect, that would be a bonus. It might be nice to try to break the guarantees and check that they hold. I realize that might not really be practical.

src/sqlstorage.cc
lbonn: When `SQLite3Guard` is destructed, it calls the destructor of `handle_` which calls `sqlite3_close()`.

> If an sqlite3 object is destroyed while a transaction is open, the transaction is automatically rolled back.

https://www.sqlite.org/c3ref/close.html

`SQLiteStatement` are freed similarly with `sqlite3_finalize`.

I'll add some comments to clarify this.

lbonn:
Hum... Possible "probabilistic" test: launch a number of threads in a tight loop of data updates and interrupt them randomly to see if the state is consistent. It should detect errors on the prior versions of the code.

patrickvacek:
I like tests that can verify that something that should break does break, but that's basically impossible if the system can't break. The probabilistic test you suggest could be interesting, but don't sweat it if it isn't practical.

lbonn:
Ok I wrote this kind of test for the public/private key pair.

As expected, it fails on FS Storage (and also on master for both storages if backported).
Fs could be fixed if the key pair would be stored in the same file.

lbonn:
It looks like the test might too brutal for Travis

PR #610, opened 2018-02-23
Describe release process for Debian packages

cajun-rat:
Change-Id: I6a9a5a7144648ee9fb315f3997e784958ca1365d

### DIFF
--- a/docs/deb-package-install.adoc
+++ b/docs/deb-package-install.adoc
@@ -40,3 +40,14 @@ cd build
 cmake  -DCMAKE_BUILD_TYPE=Debug -DBUILD_DEB=ON ..
 make package
 ----
+
+=== Making a Release on github
+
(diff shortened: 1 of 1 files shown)


docs/deb-package-install.adoc
```diff
@@ -40,3 +40,14 @@ cd build
 cmake  -DCMAKE_BUILD_TYPE=Debug -DBUILD_DEB=ON ..
 make package
 ----
+
+=== Making a Release on github
+
+Releases need to be of the form `nnnn.nn` to be automatically picked up by Travis.  To create a release, checkout the revision you want to bless, then:
```
patrickvacek: Your example `2018.4` seems to conflict with your stated requirement of `nnnn.nn`. Maybe it would be better to specify it as something like `<year>.<counter>`? It's also just a bit confusing to mention `nnnn.nn`, `2018.n`, and `2018.4`. Not that it should be hard to figure out the intention, but I'd suggest removing the interim placeholder.

cajun-rat:
@patrickvacek I've re-worded it to hopefully make more sense.

PR #612, opened 2018-02-26
Check if aktualizr runs within another user than before

patriotyk:


patriotyk:
Updated with warning about non-root user

PR #613, opened 2018-02-26
Ensure that OstreeRepo is destroyed after use

OYTIS:
Should fix the 'Couldn't load sysroot bug'. I really hope that no user was affected.

Basically there was a leak of OstreeRepo objects that caused the process to run out of file descriptors after 15-30 minutes. I'm not sure if we can test for that.

This particular error would be not so fatal if aktualizr (SotaUptaneClient to be precise) just crashed upon receiving an exception, but if it is always the best thing to do is questionable.

### DIFF
--- a/src/ostree.cc
+++ b/src/ostree.cc
@@ -16,7 +16,6 @@
 #include "utils.h"
 
 data::InstallOutcome OstreeManager::pull(const Config &config, const KeyManager &keys, const std::string &refhash) {
-  OstreeRepo *repo = NULL;
   const char *const commit_ids[] = {refhash.c_str()};
   GCancellable *cancellable = NULL;
   GError *error = NULL;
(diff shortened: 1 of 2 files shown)


src/ostree.cc
```diff
@@ -16,22 +16,22 @@
...
     g_error_free(error);
     return data::InstallOutcome(data::INSTALL_FAILED, "Could not get OSTree repo");
   }
 
   GHashTable *ref_list = NULL;
-  if (ostree_repo_list_commit_objects_starting_with(repo, refhash.c_str(), &ref_list, nullptr, &error)) {
+  if (ostree_repo_list_commit_objects_starting_with(repo.get(), refhash.c_str(), &ref_list, nullptr, &error)) {
     guint length = g_hash_table_size(ref_list);
     g_hash_table_destroy(ref_list);  // OSTree creates the table with destroy notifiers, so no memory leaks expected
     // should never be greater than 1, but use >= for robustness
```
lbonn: The leak was here?

OYTIS: It was just that not all calls to `ostree_sysroot_get_repo` were matched with `g_object_unref` on the acquired OstreeRepo. If you see more leaks here, please shout.

patrickvacek:
Good catch, I'm testing now.

PR #614, opened 2018-02-26
Aktualizr secondary skeleton

lbonn:
Mostly a copy&paste of `src/main.cc`.

I opened this PR early, without much content, because the requirements are still vague.

Also: should we share the `Config` class with aktualizr and add more fields for our needs? I would think it's better to have a separate config but maybe it would need splitting the content of `config.cc` to extract some common parts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,6 +269,8 @@ if(BUILD_SOTA_TOOLS)
     add_subdirectory("src/sota_tools")
 endif(BUILD_SOTA_TOOLS)
 
+add_subdirectory("src/aktualizr_secondary")
+
 add_subdirectory("src/aktualizr_info")
 
 include(CTest)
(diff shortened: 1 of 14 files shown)


src/aktualizr_follow/main.cc
```diff
@@ -0,0 +1,111 @@
...
+  LOG_DEBUG << "Current directory: " << boost::filesystem::current_path().string();
+  // Initialize config with default values, the update with config, then with cmd
+  boost::filesystem::path follow_config_path(commandline_map["config"].as<std::string>());
+  if (!boost::filesystem::exists(follow_config_path)) {
+    std::cout << "aktualizr-follow: configuration file " << boost::filesystem::absolute(follow_config_path)
+              << " not found. Exiting." << std::endl;
+    exit(EXIT_FAILURE);
+  }
+
+  // TODO: do something
```
OYTIS: `return 0;` maybe?

patrickvacek:
I know it's just a start, but a couple thoughts:
1. How about naming it something like aktualizr-secondary? Is "uptane-secondary" too vague?
2. Some basic tests would be nice.
3. It should get added to code coverage.

PR #617, opened 2018-02-28
Misc SQL + test fixes

lbonn:
The biggest thing is to copy the data used in SQL statements inside the object because they need to be accessed later by sqlite step functions.

This will add a bunch of string copies but will make it harder to make mistakes.

`SQLiteStatement` is also a bit more readable now.


lbonn:
I bundled a bunch of other stuff that were sitting in another branch

cajun-rat:
Looks good, just running oe-selftest...

PR #619, opened 2018-03-01
Fix 'make qa'

cajun-rat:
Change-Id: I2043ebaca9aa442e59cf86e2d3df07c369f5581f

### DIFF
--- a/tests/ostree_test.cc
+++ b/tests/ostree_test.cc
@@ -94,11 +94,12 @@ TEST(OstreeManager, BadSysroot) {
 }
 
 TEST(OstreeManager, ParseInstalledPackages) {
+  TemporaryDirectory temp_dir;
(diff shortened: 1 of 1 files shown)


lbonn:
Also stumbled on that yesterday and included the fix into https://github.com/advancedtelematic/aktualizr/pull/617. Maybe I should do smaller PRs...

lbonn:
Also: this passed on Travis because tests are run as root inside the docker container.

To catch these kinds of test bugs, we might want to run as an unprivileged user inside the container.

cajun-rat:
Yes, we should fix that run as root thing.  I'm going to close this PR in preference to #617.  But yes, smaller PRs are bettr :)

patrickvacek:
+1 to not running tests as root.

PR #621, opened 2018-03-02
Feat/pro 4952/partial verification secondary

patriotyk:


src/uptane/tufrepository.h
```diff
@@ -18,30 +18,17 @@
...
-};
-
 class TufRepository {
  public:
   TufRepository(const std::string& name, const std::string& base_url, const Config& config,
-                boost::shared_ptr<INvStorage>& storage, HttpInterface& http_client);
-  Json::Value verifyRole(Role role, const TimeStamp& now, const Json::Value&, Uptane::Root* root_used = NULL);
+                boost::shared_ptr<INvStorage>& storage);
+  Json::Value verifyRole(Role role, const Json::Value&, Uptane::Root* root_used = NULL) const;
+  std::string getBasUrl() const { return base_url_; }
```
lbonn: `getBaseUrl()`?

src/fsstorage.cc
```diff
@@ -165,33 +165,21 @@ bool FSStorage::loadMetadata(Uptane::MetaPack* metadata) {
...
   metadata->image_targets = Uptane::Targets(json);
 
   json = Utils::parseJSONFile(image_path / "timestamp.json");
-  // compatibility with old clients, which store the whole metadata, not just the signed part
-  if (json.isMember("signed") && json.isMember("signatures")) json = json["signed"];
   metadata->image_timestamp = Uptane::TimestampMeta(json);
 
   json = Utils::parseJSONFile(image_path / "snapshot.json");
-  // compatibility with old clients, which store the whole metadata, not just the signed part
-  if (json.isMember("signed") && json.isMember("signatures")) json = json["signed"];
```
cajun-rat: I'm worried about this.  Are we changing the on-disk format here?

src/uptane/partialverificationsecondary.cc
```diff
@@ -0,0 +1,100 @@
...
+#include "uptane/secondaryinterface.h"
+#include "uptane/tufrepository.h"
+
+namespace Uptane {
+
+PartialVerificationSecondary::PartialVerificationSecondary(const SecondaryConfig &sconfig_in)
+    : SecondaryInterface(sconfig_in), root_(Root::kAcceptAll) {
+  boost::filesystem::create_directories(sconfig.metadata_path);
+
+  // FIXME Probably we need to generate keys on the secondary
```
cajun-rat: Yes we do :)

src/uptane/partialverificationsecondary.cc
```diff
@@ -0,0 +1,100 @@
...
+  public_key_id_ = Crypto::getKeyId(public_key_);
+}
+
+bool PartialVerificationSecondary::putMetadata(const MetaPack &meta) {
+  TimeStamp now(TimeStamp::Now());
+  detected_attack_.clear();
+
+  root_.UnpackSignedObject(now, "director", Role::Root(), meta.director_root.original());
+  root_ = meta.director_root;
+  root_.UnpackSignedObject(now, "director", Role::Targets(), meta.director_targets.original());
```
cajun-rat: I'm worried about this pattern from a security point of view. The idea behind UnpackSignedObject is that it is the boundary between the untrusted and trusted worlds, so we have a risk here that we check one thing then look at another.

tests/test_data/prov/metadata/director/root.json
```diff
@@ -1,64 +1,77 @@
 
 {
-	"_type" : "Root",
```
cajun-rat: Hold on, aren't these here to check that the migrations are backwards compatible?

tests/uptane_test.cc
```diff
@@ -58,12 +58,12 @@ TEST(Uptane, VerifyDataBad) {
...
   config.storage.path = temp_dir.Path();
   boost::shared_ptr<INvStorage> storage = boost::make_shared<FSStorage>(config.storage);
-  Uptane::TufRepository repo("director", http.tls_server + "/director", config, storage, http);
-  Json::Value data_json = repo.getJSON("root.json");
+  Uptane::TufRepository repo("director", http.tls_server + "/director", config, storage);
+  Json::Value data_json = http.get(http.tls_server + "/director/root.json").getJson();
   data_json.removeMember("signatures");
 
   try {
-    repo.verifyRole(Uptane::Role::Root(), now, data_json);
```
cajun-rat: This 'now' should really be used, it is there to make the tests continue to work in the future.  I remember now that we had a problem where the tests suddenly started failing--I think the root cause was that repo.verifyRole was ignoring the 'now' argument.

The problem is that is tests depend on the current system time, then it is hard to simulate things like metadata expiry.

src/fsstorage.cc
patriotyk: Yes. I will change the metadata load, to work with old format

tests/test_data/prov/metadata/director/root.json
patriotyk: No, we just move everything we have from file system to sql and in comments above I wrote that I will add the old format support. So everything should work.

tests/uptane_test.cc
patriotyk: Understand, I will hardcode today date

src/uptane/partialverificationsecondary.cc
patriotyk: Hm, I don't understand, what do You mean

cajun-rat: Hi @patriotyk,  

The goal with `UnpackSignedObject()`  is to provide a clean, secure separation between metadata that is trusted, and untrusted metadata that has been downloaded from the internet. If we are not careful, then an attacker may be able to exploit a corner case to evade the signature checks.

The rule is that json blobs of the form `{'signatures': ... 'signed':...}` are untrusted, and once you are holding the contents of the signed portion, then you can consider it trusted. I'm sure @stevana will not be happy with the type system implications of this rule :)

`UnpackSignedObject` is intended to be the only gateway between these two worlds, and will only extract the `signed:` portion if signature validation is succeeds.  The reason I'm worried about this change is that is bridges these two worlds (that of  `{'signatures': ... 'signed':...}` and trusted metadata), and that will make it hard to audit for security.

Long term, @heartsucker has a nicer design here, where the validation is global, i.e. it takes an entire set of metadata and asks 'is this collection of metadata valid according to the Uptane rules', but for now I'd prefer if we can keep `UnpackSignedObject` as the only airlock between trusted and untrusted metadata.

tests/test_data/prov/metadata/director/root.json
cajun-rat: but isn't the purpose of the metadata in `tests/test_data/prov/metadata` to be a sample 'old metadata' snapshot to test that backwards compatibility is maintained?

My understanding is that  `tests/test_data/prov/metadata` is there to mimic a old device just coming back online, and to check that everything still works.  Is that wrong?

patriotyk: Yes You are right, I just reverted to old metadat, and evrything works well

scripts/test_aktualizr_deb_and_update.sh
```diff
@@ -7,6 +7,7 @@ dpkg-deb -I /persistent/aktualizr.deb && dpkg -i /persistent/aktualizr.deb
 aktualizr --version | grep $(cat /persistent/aktualizr-version) && aktualizr-info
 
 
+date --set "25 Sep 2017 15:00:00"
```
cajun-rat: Isn't this going to fail CI?

cajun-rat:
I've just tried building this into a Yocto image and got an error running aktualizr-info.  This might be a problem with my local environment, so I'm rebuilding on a test box right now.

cajun-rat:
I'm afraid this is still not working. If I take the current rocko meta-updater and add it, aktualizr-info fails:

~~~
root@qemux86-64:~# aktualizr --version
Current aktualizr version is: 1.0+gitAUTOINC+c67dbc69ea
root@qemux86-64:~# aktualizr-info 
Storage backend: Sqlite
terminate called after throwing an instance of 'Uptane::InvalidMetadata'
  what():  The  metadata failed to parse:Invalid timestamp
Aborted (core dumped)
~~~

To reproduce this, add the following lines to local.conf:

~~~
BRANCH_pn-aktualizr = "feat/PRO-4952/PartialVerificationSecondary"
BRANCH_pn-aktualizr-native = "${BRANCH_pn-aktualizr}"
require classes/sota_bleeding.inc
~~~

then

~~~
bitbake core-image-minimal
../meta-updater/scripts/run-qemu-ota
~~~

or 

~~~
oe-selftest -r updater
~~~

cajun-rat:
What tests have you done on this?  Does oe-selftest -r updater pass?

patriotyk:
I was able to reproduce this issue by running aktualizr-info. Now it works.

cajun-rat:
(passes oe-selftest)

PR #622, opened 2018-03-06
Aktualizr secondary config

lbonn:
Notes:

- the name of the executable was changed back to `aktualizr-follower`
- some utility functions were stripped back from `config.h` and put in `config_utils.h`
- config structures use C++11 member initializers

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -27,8 +27,7 @@ std::ostream& operator<<(std::ostream& os, CryptoSource cs);
 // updateFromPropertyTree() in config.cc.
 
 struct GatewayConfig {
-  GatewayConfig() : http(true), socket(false) {}
-  bool http;
+  GatewayConfig() : socket(false) {}
   bool socket;
 };
 
--- a/src/config.h
+++ b/src/config.h
@@ -112,7 +112,6 @@ struct ImportConfig {
 class Config {
  public:
   Config();
-  Config(const boost::property_tree::ptree& pt);
   Config(const boost::filesystem::path& filename, const boost::program_options::variables_map& cmd);
   Config(const boost::filesystem::path& filename);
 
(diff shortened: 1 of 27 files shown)


src/aktualizr_follower/follower_config.h
```diff
@@ -0,0 +1,28 @@
...
+#include <boost/program_options.hpp>
+#include <boost/property_tree/ini_parser.hpp>
+
+struct FollowerNetConfig {
+  int port{9030};
+};
+
+class FollowerConfig {
+ public:
+  FollowerConfig(){};
```
patriotyk: Should be space between braces. Do You added this file to clang-format target?

src/config.h
```diff
@@ -27,129 +27,91 @@ std::ostream& operator<<(std::ostream& os, CryptoSource cs);
...
 };
 
 struct ProvisionConfig {
-  ProvisionConfig() : server(""), p12_password(""), expiry_days("36000"), provision_path(""), mode(kAutomatic) {}
-  std::string server;
-  std::string p12_password;
-  std::string expiry_days;
-  boost::filesystem::path provision_path;
-  ProvisionMode mode;
+  std::string server{};
```
patriotyk: Default constuctor for server will automatically called on ProvisionConfig construction. So I think it would be nicer to remove empty {}

lbonn: Yes, I kept them for consistency, because it's easy to forget the braces for integers and other primitive types. But I don't feel strongly about this, so we can omit them for non-primitive types.

src/aktualizr_follower/follower_config.h
lbonn: I indeed forgot to add the header to the source file checks. But for some reason, clang-format still removes the spaces. I'll look into that.

lbonn: Ok, that's because of the trailing `;` which shouldn't be there. There are a few other instances of these in the code base.

src/aktualizr_follower/follower_config.cc
```diff
@@ -0,0 +1,44 @@
...
+
+void FollowerConfig::updateFromCommandLine(const boost::program_options::variables_map& cmd) {
+  if (cmd.count("server-port") != 0) {
+    network.port = cmd["server-port"].as<int>();
+  }
+}
+
+void FollowerConfig::updateFromPropertyTree(const boost::property_tree::ptree& pt) {
+  // Keep this order the same as in follower_config.h and writeToFile().
+  CopyFromConfig(network.port, "network.port", boost::log::trivial::info, pt);
```
OYTIS: Shouldn't we just accept the socket systemd gives us?

lbonn: That's if we don't want to be tied to systemd in all cases, eg for testing.
But that should not be hard to add it back when we'll need it, so yes we could omit this part.

PR #624, opened 2018-03-06
OPC-UA bridge: ostree repo sync

riqkum:


cajun-rat:
This looks great. With BUILD_OPCUA=on the build is fine but a saw a test failing. Since we don't have BUILD_OPCUA=on in the CI at the moment, that isn't a problem for merging.  When I get a moment I'll investigate some more.

Builds in meta-updater continue to work, so this is good to go.

PR #630, opened 2018-03-13
Add ASN.1 schemas for uptane primary<->secondary communication messages

OYTIS:


tests/asn1-reference/uptane/ipuptane_message.asn1
```diff
@@ -0,0 +1,112 @@
...
+	AKManifestReqPayload ::= SEQUENCE {
+		...
+	}
+
+	AKManifestRespPayload ::= SEQUENCE {
+		format AKIpUptaneMetadataFormat,
+		manifest ANY DEFINED BY format -- OCTET string for json of ASN.1 sequence (to be defined) for ber
+	}
+	
+	AKImageMeta ::= SEQUENCE {
```
lbonn: Here, `DEFINED BY format` should be `DEFINED BY imageTimestampFormat`

(same issue in other fields and in `AKPutMetaReqPayload`)

tests/CMakeLists.txt
```diff
@@ -47,7 +47,8 @@ add_custom_target(check-full COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COM
 
 # Reference asn1c libraries
 add_asn1c_lib(config asn1-reference/common asn1-reference/config/cryptosource asn1-reference/config/tlsconfig)
-list(APPEND TEST_LIBS config_asn1)
+add_asn1c_lib(ipuptane asn1-reference/common asn1-reference/uptane/ipuptane_message)
+list(APPEND TEST_LIBS config_asn1 ipuptane_asn1)
```
lbonn: Originally, it was intended to have only one asn1c_lib in the code base, because the common generated C files could not link with each others. Also, cmake now outputs twice the garbage.

tests/asn1-reference/uptane/ipuptane_message.asn1
OYTIS: @lbonn Good catch, thank you.

tests/CMakeLists.txt
OYTIS: @lbonn Yes, might work that way. I'll change FindAsn1.cmake a bit and we'll figure out if that works.

lbonn:
Too bad that the `AKIpUptane` payloads are not tightly coupled with the list of existing messages and their ids.

I briefly tried to use Information Object Classes: https://tools.ietf.org/html/rfc6025#section-2.1 because apparently `ANY DEFINED BY` is deprecated since 1994 but couldn't make it work with asn1c...

But since we'll not use code generation that's not such a big issue.

OYTIS:
@lbonn As of metadata serializations I replaced `ANY DEFINED BY` with `OCTET STRING` as originally planned. I've started working on primary<->secondary communication and it seems to me it doesn't make sense to parse metadata on the fly.
For `AKIpUptane` we can just use `CHOICE`. I prefer specifying tags manually to make parsing without a parser generator easier.

OYTIS:
And of course we can delete message identifiers now.

PR #632, opened 2018-03-13
Feat/pro 5018/ip discovery

patriotyk:


src/config.h
```diff
@@ -32,9 +32,14 @@ struct GatewayConfig {
 };
 
 struct NetworkConfig {
-  std::string socket_commands_path{"/tmp/sota-commands.socket"};
-  std::string socket_events_path{"/tmp/sota-events.socket"};
-  std::vector<std::string> socket_events{"DownloadComplete", "DownloadFailed"};
+  NetworkConfig() : socket_commands_path("/tmp/sota-commands.socket"), socket_events_path("/tmp/sota-events.socket") {}
+  std::string socket_commands_path;
+  std::string socket_events_path;
+  std::vector<std::string> socket_events = {"DownloadComplete", "DownloadFailed"};
```
cajun-rat: Is there a reason to revert these changes, or is this a rebase error?

patriotyk: No sorry, I did not noticed that when resolving conflicts for rebase

cajun-rat: Heh, no worries

tests/CMakeLists.txt
```diff
@@ -158,6 +160,9 @@ add_aktualizr_test(NAME uptane_key SOURCES uptane_key_test.cc PROJECT_WORKING_DI
 
 add_aktualizr_test(NAME uptane_network SOURCES uptane_network_test.cc PROJECT_WORKING_DIRECTORY)
 
+add_aktualizr_test(NAME discovery_secondary SOURCES ipsecondary_discovery_test.cc PROJECT_WORKING_DIRECTORY)
+
+
```
cajun-rat: +1 for a test, but it fails on my machine :(

~~~
[ RUN      ] ipsecondary_discovery.test_discovery
starting up on 127.0.0.1 port 60443

waiting to receive message
/home/phil/pro/client/aktualizr/tests/ipsecondary_discovery_test.cc:25: Failure
      Expected: secondaries.size()
      Which is: 0
To be equal to: 2
~~~

cajun-rat:
Overall this looks good, and we can change the broadcast address to a suitable subnet at some point

src/ipsecondarydiscovery.cc
```diff
@@ -0,0 +1,94 @@
...
+    tv.tv_sec = left_seconds;
+    setsockopt(socket_fd, SOL_SOCKET, SO_RCVTIMEO, (const char *)&tv, sizeof(tv));
+  }
+  return secondaries;
+}
+
+void IpSecondaryDiscovery::sendRequest() {
+  int socket_fd = socket(AF_INET, SOCK_DGRAM, 0);
+  if (socket_fd == -1) {
+    LOG_ERROR << "Error of creating socket: " << std::strerror(errno);
```
lbonn: The function should return or throw in case of errors (also in subsequent `setsockopt`).

To make that easier, I was about to move `SocketHandle` to `utils.h`:
https://github.com/advancedtelematic/aktualizr/blob/05eeb37d2bfaa0e31cc26a4a6c4d610efd7f8ba9/src/aktualizr_secondary/aktualizr_secondary.h#L17
But no need to use it right away.

src/ipsecondarydiscovery.cc
```diff
@@ -0,0 +1,94 @@
...
+  tv.tv_usec = 0;
+  setsockopt(socket_fd, SOL_SOCKET, SO_RCVTIMEO, (const char *)&tv, sizeof(tv));
+
+  char rbuf[2000] = {};
+  int recieved = 0;
+  while ((recieved = recv(socket_fd, rbuf, sizeof(rbuf) - 1, 0)) != -1) {
+    std::string data(std::move(rbuf), recieved);
+    try {
+      int type = 0;
+      Uptane::SecondaryConfig conf;
```
lbonn: The secondary config should probably contain the device ip so that we can communicate with it on the regular interface later.

src/ipsecondarydiscovery.cc
```diff
@@ -0,0 +1,94 @@
...
+      asn1::Deserializer des(data);
+      des >> asn1::seq >> asn1::implicit<kAsn1Integer>(type) >> asn1::implicit<kAsn1Utf8String>(conf.ecu_serial) >>
+          asn1::implicit<kAsn1Utf8String>(conf.ecu_hardware_id) >> asn1::restseq;
```
lbonn: @OYTIS do you think it would be worth it to also transmit the secondary server tcp port, to make the discovery more fail-safe? So that we only need to change the port in one place. We can worry about it later but it's pretty straightforward.

patriotyk: We can't use discovery with tcp because it requires connection to hosts. But we need to find those hosts.

lbonn: Yes, I meant transmitting the tcp port number of the server (secondary) in the udp discovery response so that it doesn't need to be hard-coded on the primary side.

patriotyk: Do we need this IP now? It is not a problem to add any fields we need later.

lbonn: You're right, I think it can wait until the integration

src/ipsecondarydiscovery.cc
```diff
@@ -0,0 +1,91 @@
...
+    setsockopt(socket_fd, SOL_SOCKET, SO_RCVTIMEO, (const char *)&tv, sizeof(tv));
+  }
+  return secondaries;
+}
+
+void IpSecondaryDiscovery::sendRequest() {
+  int broadcast = 1;
+  if (setsockopt(socket_fd, SOL_SOCKET, SO_BROADCAST, &broadcast, sizeof broadcast) < 0) {
+    LOG_ERROR << "Could not setup socket for broadcast: " << std::strerror(errno);
+    close(socket_fd);
```
lbonn: Should return here.

And possibly indicate failure in the method's return type so that aktualizr can avoid waiting 10s if it failed to send the request.

patriotyk: Yes, You are right. Done

src/config.h
cajun-rat: Can you un-revert these back to the c++11 style, please?

patriotyk: Sorry, I marked this change in my mind as resolved. 10 minutes and will update PR

PR #633, opened 2018-03-13
RFC: Provide a way to make a fixed machine id

doanac:
My cmake understanding is limited, so I'm not sure this is the best
approach. However, I think I've found an issue that's worth addressing:

While testing on an rpi3 I found my machine-id being registered with
ATS Garage was showing up as the name of my device rather than its
actual hardware-id. Looking into lshw, I see they just read the value
from /etc/hostname and there's not really an easy way to probe for
this. This change allows you to define a fixed machine-id.


cajun-rat:
We have a configuration file option to set this, would that work for you?

~~~
[uptane]
primary_ecu_hardware_id = "mycoolhardware"
~~~

I'd quite like to make it possible to fetch this stuff from the hardware itself, but it isn't clear what the best way to do that is.  For example, if someone has an eeprom on the hardware that identifies the board revision, then it would be nice to support grabbing the hardware ID from that, rather than relying on information in the filesystem.

doanac:
@cajun-rat - thanks for the tip. That should work just fine for us.

cajun-rat:
If you are happy with the config option, are you happy if I close this PR?

PR #635, opened 2018-03-14
Move secondary config next to other config files

cajun-rat:
Change-Id: I894b2b9c3e1bbad5a1403fc34a4721d1f62ff55c


lbonn:
Updated the only place where the path was referenced

PR #636, opened 2018-03-15
Fix/uptane initialize

lbonn:
- one comment was not in sync with the code in `initialize`
- we don't need serial and hardware id for device id registration
- and other misc inconsistencies.


cajun-rat:
Looks good to me.  I'd like to fix that CI failure before merging though. I think is is caused by something else, though

lbonn:
Rebased, will pass CI hopefully

PR #638, opened 2018-03-15
Implement primary<->secondary message serialization and deserialization

OYTIS:
The PR is somewhat raw, but it is important to have at least preliminary version to start implementing primary<->secondary communication and I've delayed it for a long time already.
What is there: definition of communication messages, their serializations and deserializations in ASN.1.
What's not (yet) there: exhaustive tests of (de)serialization of all messages. Aktualizr secondary is in a half-broken state now.


OYTIS:
Yes, I've also accidentially run in through cmake-format (v5.0.1) on my machine and it added comments to all closing brackets for namespaces. I didn't delete them because a) It's tedious, dozens of files are affected; b) It doesn't hurt to have this comments c) One day they'll upgrade clang-format on Ubuntu and we'll have them anyway.
If you disagree I can find and delete them.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -79,23 +79,12 @@ void AktualizrSecondary::run() {
...
-
-    if (pkt->msg->variant == "Stop") {
-      // Will cause `accept()` to fail and break the loop
-      shutdown(*socket_hdl_, SHUT_RDWR);
-    }
+    // TODO: process message
   }
 
   tcp_thread.join();
 }
```
lbonn: It would be nice if we could keep the ability to stop the loop, it's useful for testing,
Perhaps, we can stop using a channel for now and move the tcp thread code directly in `run()`. `stop()` would be just a call to `shutdown()`

OYTIS: Yes, I think it's better to dedicate channel to primary<->secondary communication and use something else to stop the loop. Alternatively we may have two different channels, one for communication and another for control (like 'event' and 'command' channels in primary).

cajun-rat:
Just rebasing this now

src/asn1-cerstream.cc
```diff
@@ -67,12 +116,15 @@ Deserializer& Deserializer::operator>>(Token tok) {
...
 
-    case Token::endseq_tok: {
+    case Token::endseq_tok:
+    case Token::endexpl_tok: {
       int32_t len = seq_lengths.top();
       int32_t cons_len = seq_consumed.top();
       seq_lengths.pop();
       seq_consumed.pop();
-      if (len > 0) {
+      std::cout << "ENDSEQ LEN: " << std::dec << len << std::endl;
```
lbonn: debug print?

src/ipuptaneconnection.cc
```diff
@@ -0,0 +1,128 @@
...
+    asn1::Deserializer asn1_stream(message_content);
+    std::unique_ptr<SecondaryMessage> mes;
+    try {
+      asn1_stream >> mes;
+    } catch (deserialization_error) {
+      LOG_ERROR << "Unexpected message format";
+      continue;
+    }
+    std::unique_ptr<SecondaryPacket> pkt{new SecondaryPacket{*addr, std::move(mes)}};
+
```
lbonn: Actually, I'm not sure how we can use the in and out channel because we need to reuse the same tcp connection to transmit the response.

src/ipuptaneconnection.cc
```diff
@@ -0,0 +1,128 @@
...
+void IpUptaneConnection::handle_connection_msgs(SocketHandle con, std::unique_ptr<sockaddr_storage> addr) {
+  std::string peer_name = Utils::ipDisplayName(*addr);
+  LOG_INFO << "Opening connection with " << peer_name;
+  std::string message_content;
+  while (true) {
+    uint8_t c;
+    if (recv(*con, &c, 1, 0) != 1) {
+      break;
+    }
+    message_content.push_back(c);
```
lbonn: `message_content` is never reset to empty string.

src/asn1-cerstream.cc
OYTIS: Yes, thanks.

src/ipuptaneconnection.cc
OYTIS: Yes, that was the original idea, but now I think it was wrong. If we want to reuse a connection we either need to add additional layer of complexity by prepending sent data with its length or parse data on the fly. I now think it's better to launch these loops on both primary and secondary and use one connection per message.

src/ipuptaneconnection.cc
OYTIS: If we use one connection per message it shouldn't I believe.

PR #639, opened 2018-03-15
OPC-UA bridge integration

riqkum:


cajun-rat:
The CI build is failing because `opcuaserver_secondary_delegate.cc` hasn't been added to `aktualizr_source_file_checks`, which checks the files are being automatically formatted:

~~~
CMake Error at CMakeLists.txt:302 (message):
  src/aktualizr_secondary/opcuaserver_secondary_delegate.cc not checked
~~~

Can you add this to `aktualizr_source_file_checks` please?

cajun-rat:
Otherwise, it looks good!

riqkum:
Both configurations (BUILD_OPCUA: ON/OFF) should be good now.

cajun-rat:
Rebased to fix merge conflict

src/aktualizr_secondary/main.cc
```diff
@@ -124,12 +127,17 @@ int main(int argc, char *argv[]) {
     // storage (share class with primary)
     boost::shared_ptr<INvStorage> storage = INvStorage::newStorage(config.storage);
 
+#ifdef OPCUA_SECONDARY_ENABLED
+    AktualizrSecondaryOpcua secondary(config);
+#else
     AktualizrSecondary secondary(config, storage);
+#endif
```
cajun-rat: When we spoke about this on Monday, I thought that you were intending this to be a run-time configuration option?

src/CMakeLists.txt
```diff
@@ -123,20 +127,23 @@ set(OPCUABRIDGE_HEADERS opcuabridge/boostarch.h
...
                         opcuabridge/opcuabridge.h
+                        opcuabridge/opcuabridgeclient.h
+                        opcuabridge/opcuabridgeserver.h
                         opcuabridge/signature.h
                         opcuabridge/signed.h
                         opcuabridge/utility.h
-                        opcuabridge/versionreport.h)
-
+                        opcuabridge/versionreport.h
+                        uptane/opcuasecondary.h)
```
cajun-rat: This probably lives in a CMakeLists.txt under `src/opcuabridge`, but there is clearly some interdependency with `src/uptane` at the moment.  I think we can live with this for now, but this is something we will have to look at in the future.  Probably the answer is to make `src/uptane` purely an implementation of the Uptane spec and move the transport and installation related parts somewhere else.

cajun-rat:
Approved, assuming CI passes.

src/aktualizr_secondary/main.cc
riqkum: I used the compile time option because this is only an approach for the moment. And I was unsure about possible transport combinations, so I've risen the question. Currently, I have a little reason to set both transports simultaneously - the only one should be selected to avoid unnecessary resources utilization, I think. I see it would be suitable to introduce command line option, despite the fact it will require some handling during build image recipe processing.

riqkum:
Is there a way I could locally reproduce that issue with check-format?

lbonn:
Right now, you need to:
- install the same version of clang-format as in the Debian unstable Dockerfile ([clang-format-3.8](https://github.com/advancedtelematic/aktualizr/blob/master/Dockerfile.deb-unstable#L19))
- re-run cmake in a clean build directory
- run `make format` or `ninja format` if you use ninja

And your code should pass the format check.

cajun-rat:
This looks like a different version of clang-format is being used on Debian unstable. I've created PRO-5096 for a proper fix, but it should be possible to run the build inside the `Dockerfile.deb-unstable` environment and see the change.

riqkum:
Thanks. Downgraded my local clang-format to 3.8. Hope never run into this issue again.

cajun-rat:
@lbonn 's PR #648 might fix the build failure

PR #641, opened 2018-03-16
Secondary aktualizr discovery (secondary part)

lbonn:
Based on #632, I'll rebase it when merged.

Not entirely complete, it needs the hardware and serial ids of the secondary which, in turn, depends on the aktualizr secondary storage (should follow soon).


lbonn:
Rebased on master

PR #644, opened 2018-03-20
Feat/pro 5075/aktualizr secondary storage

lbonn:
Based on #638 (will be rebased)

I am not entirely satisfied with the fact that it currently reuses the whole `UptaneConfig` from the primary, which contains a lot of irrelevant stuff (we need it to instantiate KeyManager) + badly named fields for the secondary.

Since `KeyManager` only uses a small portion of the content of `UptaneConfig`, I'll try to decouple them more in this PR. It can still be reviewed in the meantime.


lbonn:
Got rid of `UptaneConfig` in `AktualizrSecondaryConfig`

lbonn:
Rebased + a bug fix but don't know what's going on with Travis...

lbonn:
The Travis tests passed after a couple of restarts...
It seems that it failed to pull some docker images for a while.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -1,50 +1,80 @@
...
+// Initialize the parts needed for a secondary
+// - ECU keys
+// - serial id
+bool AktualizrSecondary::uptaneInitialize() {
+  if (keys_.generateUptaneKeyPair().size() == 0) {
+    return false;
+  }
+
+  // from uptane/initialize.cc but we only take care of our own serial/hwid
+  std::vector<std::pair<std::string, std::string> > ecu_serial;
```
patriotyk: Maybe ecu_serials?

src/keymanager.h
```diff
@@ -4,20 +4,30 @@
...
 #include "invstorage.h"
-#ifdef BUILD_P11
 #include "p11engine.h"
-#endif
+#include "types.h"
 #include "utils.h"
 
+// bundle some parts of the main config together
+struct KeyManagerConfig {
+  KeyManagerConfig() = delete;
```
patriotyk: If thre is no any other constructors how it is possible to instantiate this struct?

tests/aktualizr_secondary/config_tests.toml
```diff
@@ -1,2 +1,6 @@
 [network]
 port = 9031
+
+[storage]
+type = sqlite
```
patriotyk: forgot double quotes

tests/aktualizr_secondary/config_tests.toml
```diff
@@ -1,2 +1,6 @@
...
+schemas_path = config/schemas
```
patriotyk: there too

lbonn: thanks!

src/keymanager.h
lbonn: It should be derived from one of the main configs: `Config` and `AktualizrSecondaryConfig`. I didn't want it to have default value. So the way to construct it is `KeyManagerConfig kmconfig{field_values,...}`.

But yes, deleting the default constructor is not strictly needed and can be confusing. Maybe I'll replace it with a comment.

cajun-rat:
With Serhiy's comments fixed, this looks good to me. Merging.

PR #645, opened 2018-03-21
Feat/pro 5078/ostree secondary

patriotyk:


src/aktualizr_secondary/aktualizr_secondary_config.cc
```diff
@@ -48,6 +48,30 @@ void AktualizrSecondaryUptaneConfig::writeToStream(std::ostream& out_stream) con
   writeOption(out_stream, key_type, "key_type");
 }
 
+void PackageConfig::writeToStream(std::ostream& out_stream) const {
```
lbonn: These two methods + `std::ostream& operator<<(std::ostream& os, PackageManager pm)` are already implemented in `config.cc` and `aktualizr-secondary` successfully links even when these implementations are removed.

Any specific reason to also define them here?

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -74,15 +75,43 @@ std::pair<KeyType, std::string> AktualizrSecondary::getPublicKeyResp() const {
...
+  }
+  if (target_found) {
+    if (target->format().empty() || target->format() == "OSTREE") {
+#ifdef BUILD_OSTREE
+      OstreeManager::pull(config_.pacman, keys_, target->sha256Hash());
+#else
+      LOG_ERROR << "Could not pull OSTree target. Aktualizr was built without OSTree support!";
+#endif
+    }
+    pacman->install(*target);
```
OYTIS: Installation should only happen in `sendFirmwareResp`. It gets TLS credentials as a payload and without them OSTree will not be able to download the image.

patriotyk: Ok

src/aktualizr_secondary/aktualizr_secondary_config.cc
patriotyk: Sorry, I just temporarry copied it.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -74,16 +75,43 @@ std::pair<KeyType, std::string> AktualizrSecondary::getPublicKeyResp() const {
...
+        detected_attack_ = "Duplicate entry for this ECU";
+        break;
+      }
+      target_found = true;
+      target = boost::movelib::make_unique<Uptane::Target>(*it);
+    }
+  }
+  return true;
+}
+#ifdef BUILD_OSTREE
```
lbonn: `sendFirmwareOstreeResp` instead of `sendFirmwareOstreResp` would appease me :).

patriotyk:
I think the treehub url should be passed from primary during the provisoning

cajun-rat:
Works for me

cajun-rat:
We'll add the treehub URL in a second PR

PR #646, opened 2018-03-21
IP Uptane primary<->secondary communication

OYTIS:
I had to make Config non-const in order to be able to add data to secondary_configs. Alternatively we can move secondary_configs out of Config or declare it as mutable.

"not-ready" label is there until I'm done with the tests.

### DIFF
--- a/src/aktualizr.h
+++ b/src/aktualizr.h
@@ -5,12 +5,12 @@
 
 class Aktualizr : public boost::noncopyable {
  public:
-  Aktualizr(const Config& config);
+  Aktualizr(Config& config);
 
   int run();
 
  private:
-  const Config& config_;
+  Config& config_;
 };
 
 #endif  // AKTUALIZR_H_
--- a/src/aktualizr_secondary/aktualizr_secondary.cc
+++ b/src/aktualizr_secondary/aktualizr_secondary.cc
@@ -8,8 +8,7 @@
 #include "socket_activation.h"
 #include "utils.h"
 
-AktualizrSecondary::AktualizrSecondary(const AktualizrSecondaryConfig &config,
-                                       const boost::shared_ptr<INvStorage> &storage)
+AktualizrSecondary::AktualizrSecondary(const AktualizrSecondaryConfig& config, boost::shared_ptr<INvStorage>& storage)
     : config_(config), conn_(config.network.port), storage_(storage), keys_(storage_, config.keymanagerConfig()) {
   pacman = PackageManagerFactory::makePackageManager(config_.pacman, storage_);
   // note: we don't use TlsConfig here and supply the default to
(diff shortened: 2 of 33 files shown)


src/config.cc
```diff
@@ -93,6 +93,7 @@ void NetworkConfig::updateFromPropertyTree(const boost::property_tree::ptree& pt
   CopyFromConfig(ipdiscovery_host, "ipdiscovery_host", boost::log::trivial::trace, pt);
   CopyFromConfig(ipdiscovery_port, "ipdiscovery_port", boost::log::trivial::trace, pt);
   CopyFromConfig(ipdiscovery_wait_seconds, "ipdiscovery_wait_seconds", boost::log::trivial::trace, pt);
+  CopyFromConfig(ipuptane_port, "network.ipuptane_port", boost::log::trivial::trace, pt);
```
lbonn: Same idea as here: https://github.com/advancedtelematic/aktualizr/pull/632#discussion_r174523296

By returning the port in the discovery packet, we would avoid another possibly misconfigured config item.

OYTIS: Sorry, didn't notice that. Yes, totally makes sense, I'll add it to this PR.

lbonn: And since the communication is now bidirectional, the primary's tcp port could be in the discovery request?

OYTIS: @lbonn I've also found a fatal flow in the discovery: it doesn't actually save discovered IP addresses =)

src/asn1-cerstream.cc
```diff
@@ -180,24 +190,38 @@ Deserializer& Deserializer::operator>>(const Token& tok) {
...
-      if (full_tag != cer_decode_token(data, &endpos, &seq_len, nullptr)) throw deserialization_error();
+      if (full_tag != cer_decode_token(data, &endpos, &seq_len, nullptr)) {
+        if (opt_count && opt_first) {
+          opt_present = false;
+          opt_first = false;
+          break;
+        }
+        throw deserialization_error();
+      }
+      if (opt_count && opt_first) {
```
lbonn: This doesn't look reachable

lbonn: Nevermind, I got the indentation wrong.

OYTIS:
Looks like year 2018 is too early to assume that IPv6 will work by default.

OYTIS:
@lbonn @cajun-rat What's the purpose of STATIC_CHECKS on debian:testing? Its clang-check version is different from what we use on shovel (6.0 vs 3.8) so it blows up for me.

lbonn:
You mean clang-format? I guess one rationale could be that 3.8 is the version provided by Ubuntu Xenial: https://packages.ubuntu.com/xenial/clang-format. But we can change that.

OYTIS:
@lbonn No, clang-check. The problem as I understand it is that I first format code with clang-format-3.8 and then test checks it with clang-check-6.0 and reports error. I'm not sure why it worked before.

src/aktualizr.h
```diff
@@ -5,12 +5,12 @@
 
 class Aktualizr : public boost::noncopyable {
  public:
-  Aktualizr(const Config& config);
+  Aktualizr(Config& config);
```
patriotyk: Why You removed const?

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -61,7 +61,96 @@ void AktualizrSecondary::run() {
...
+            in_msg_putroot->director);
+
+        break;
+      }
+      case kSecondaryMesSendFirmwareReqTag: {
+        SecondarySendFirmwareReq* in_msg_fw = dynamic_cast<SecondarySendFirmwareReq*>(&(*pkt->msg));
+        SecondarySendFirmwareResp* out_msg_fw = new SecondarySendFirmwareResp;
+        out_msg = std::unique_ptr<SecondaryMessage>{out_msg_fw};
+
+        out_msg_fw->result = sendFirmwareResp(in_msg_fw->firmware);
```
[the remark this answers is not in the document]
OYTIS: Why, its on line 142.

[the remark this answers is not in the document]
patriotyk: Nevermind, didn't notice that You didn't forget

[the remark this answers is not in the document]
OYTIS: Maybe I should move it after the assignment for readability.

src/aktualizr.h
OYTIS: I think we should eventually move it from Config, it doesn't seem to belong there especially as we'll have even move discovery mechanisms in aktualizr

src/uptane/ipuptanesecondary.cc
```diff
@@ -0,0 +1,91 @@
+#include "ipuptanesecondary.h"
+
+namespace Uptane {
+
+bool IpUptaneSecondary::sendRecv(std::unique_ptr<SecondaryMessage> mes, std::shared_ptr<SecondaryPacket>& resp,
+                                 std::chrono::milliseconds to) {
+  pending_messages.clear();
```
lbonn: Can two concurrent `sendRecv` run safely? It looks like one call could delete the response of the other before it can receive it.

src/config.h
```diff
@@ -40,8 +40,9 @@ struct NetworkConfig {
   std::vector<std::string> socket_events{"DownloadComplete", "DownloadFailed"};
 
   std::string ipdiscovery_host{"127.0.0.1"};
-  unsigned int ipdiscovery_port{12345};
-  unsigned int ipdiscovery_wait_seconds{10};
+  in_port_t ipdiscovery_port{12345};
+  uint32_t ipdiscovery_wait_seconds{10};
+  in_port_t ipuptane_port{12345};
```
cajun-rat: What about changing these to 9031/9030?

src/uptane/ipuptanesecondary.cc
OYTIS: @lbonn Thread safety hasn't been a goal so far, at least I can't imagine where our code could run sendRecv for the same secondary simultaneously. It also doesn't hurt to add a mutex of course.

PR #647, opened 2018-03-22
Treehub credentials and pull + install on secondary

lbonn:
Based on #645.

Gets the tls creds + server url from an archive (still needs the archive making code on the primary side).


patriotyk:
Looks like we made a double work

lbonn:
Yes that's unfortunate, should probably have pinged you before starting.

PR #648, opened 2018-03-22
Fix deb-unstable

lbonn:
From #646


cajun-rat:
Approved, assuming the build passes

PR #650, opened 2018-03-23
Archive refactor

lbonn:
Taken from #647

### DIFF
--- a/src/bootstrap.cc
+++ b/src/bootstrap.cc
@@ -1,7 +1,5 @@
 #include "bootstrap.h"
 
-#include <archive.h>
-#include <archive_entry.h>
 #include <boost/algorithm/string.hpp>
 
 #include <stdio.h>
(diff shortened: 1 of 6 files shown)


src/bootstrap.cc
```diff
@@ -10,169 +8,71 @@
...
-    if (found) {
-      url = url_stream.str();
-      boost::trim(url);
-    } else {
-      LOG_ERROR << "autoprov.url not found in provision archive: " << provision_path;
+  std::string url;
+  try {
+    std::ifstream as(provision_path.c_str(), std::ios::in | std::ios::binary);
+    if (as.fail()) {
+      LOG_ERROR << "Unable to open provided provision archive '" << provision_path << "': " << std::strerror(errno);
```
patriotyk: You could  use FatalException from  src/exceptions.h without logging message

PR #657, opened 2018-03-27
Full credentials passing from primary to secondary

lbonn:
I switched back to the archive format to stay in sync with the existing interface.

It still needs to be tested end to end with real provisioning.

### DIFF
--- a/tests/aktualizr_secondary/uptane_test.cc
+++ b/tests/aktualizr_secondary/uptane_test.cc
@@ -3,8 +3,8 @@
 #include "aktualizr_secondary/aktualizr_secondary.h"
 #include "sotauptaneclient.h"
 
-#include "../httpfake.h"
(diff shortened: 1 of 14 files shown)


tests/aktualizr_secondary/uptane_test.cc
```diff
@@ -1,6 +1,9 @@
 #include <gtest/gtest.h>
 
 #include "aktualizr_secondary/aktualizr_secondary.h"
+#include "sotauptaneclient.h"
+
+#include "../httpfake.h"
```
patrickvacek: Would it be better to tell CMake to search for header files in the tests directory (instead of using a relative path)?

lbonn: Yes thanks, I just added the directory to the include list of all the tests

PR #658, opened 2018-03-27
Add utility to check discovering

patriotyk:


cajun-rat:
Perfect, just what is needed.

PR #660, opened 2018-03-27
Fix more discovery issues

OYTIS:
With these changes provisioning mostly works for me except that I get a "Missing entity" error from the backend, which normally indicates that ECU registration was not successfull (although the registration itself returns 200 OK).
This issue can't be tracked down without @simao's help.

### DIFF
--- a/src/aktualizr_secondary/aktualizr_secondary.cc
+++ b/src/aktualizr_secondary/aktualizr_secondary.cc
@@ -67,7 +67,7 @@ void AktualizrSecondary::run() {
     auto peer_port = primaries_map.find(peer_addr);
     if (peer_port == primaries_map.end()) continue;
 
-    Utils::setSocketPort(&peer_addr, peer_port->second);
+    Utils::setSocketPort(&peer_addr, htons(peer_port->second));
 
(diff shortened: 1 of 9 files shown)


OYTIS:
I suspect it may have something to do with secondary having the same HW ID as primary. We needed to fix that anyway. I think there should be a compile flag to aktualizr setting its hardware ID and a respective Yocto variable that will also set hardware ID for garage-sign.

patrickvacek:
While we should still support being able to set the hardware IDs to whatever we want, and the server should support multiple ECUs with the same IDs, my initial attempt to provision a primary and secondary with different IDs still did not work:

```
Provisioned successfully on Device Gateway
Error registering device on Uptane, response: The request content was malformed:
Unknown KeyType: UNKNOWN: DownField(keytype),DownField(clientKey),MoveRight,DownArray,DownField(ecus)
ECU registration failed, abort initialization
```

patrickvacek:
Sorry, ignore the error from my last message. Now I'm getting the same thing as Phil: `response: {"code":"missing_entity","description":"Missing entity: DeviceCurrentTarget","cause":null,"errorId":"ab93b4bd-d202-4acf-8bbe-613d5380c367"}`

PR #661, opened 2018-03-27
treehub: Fix logic for HTTP basic auth

doanac:
The username and password methods used by the authenticate function
don't tell the TreeHubServer to update its method_. This changes makes
this a single API to work more like the SetCerts and SetToken methods
work.


cajun-rat:
Excellent, nice one!

PR #662, opened 2018-03-27
fix utils according new changes

patriotyk:


OYTIS:
@patriotyk Looks like we've attacked the same target. [Here](https://github.com/advancedtelematic/aktualizr/pull/660) this and a couple of other issues are fixed.

PR #669, opened 2018-03-28
Verify manifests from secondary on primary

patriotyk:
Also added similar functionality for aktualizr-check-discovery


patrickvacek:
I'm seeing things like this:

```
Mar 29 15:25:40 qemux86-64 aktualizr[367]: Unexpected message format
Mar 29 15:25:41 qemux86-64 aktualizr[367]: Secondary manifest is corrupted or not signed, manifest: null
```

Your code might be working, but whatever problem that I've been seeing with the message format is only getting exaggerated now. Any thoughts?

patrickvacek:
BTW there are some merge conflicts to be resolved now. Otherwise, this probably is fine to merge.

PR #673, opened 2018-03-29
Use ninja with travis

lbonn:
Let's see if it works reliably and make it run faster.

### DIFF
--- a/scripts/coverage.sh
+++ b/scripts/coverage.sh
@@ -5,9 +5,9 @@ set -e
 
 mkdir -p build-coverage
 cd build-coverage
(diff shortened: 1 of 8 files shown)


scripts/coverage.sh
```diff
@@ -7,7 +7,7 @@ mkdir -p build-coverage
 cd build-coverage
 cmake -GNinja -DBUILD_OSTREE=ON -DBUILD_SOTA_TOOLS=ON -DBUILD_DEB=ON -DBUILD_P11=ON -DBUILD_WITH_CODE_COVERAGE=ON -DTEST_PKCS11_ENGINE_PATH="/usr/lib/engines/engine_pkcs11.so" -DTEST_PKCS11_MODULE_PATH="/usr/lib/softhsm/libsofthsm2.so" ../src
 ninja -v
-CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=3 ninja -v coverage
+CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=3 travis_wait 15 ninja -v coverage
```
cajun-rat: Does this mean that we won't be able to run the tests locally without some extra hacks?

lbonn: Yes this is ugly, I was just trying to debug the failures I got.

cajun-rat: Ah, ok

PR #674, opened 2018-03-29
Fix warnings from clang-tidy

cajun-rat:


lbonn:
These should have broken the CI but apparently clang-tidy didn't read the `.clang-tidy` configuration for the tests.

clang-tidy advertises that it looks at `.clang-tidy` files in each source parent directories but it looks like it's a lie and doesn't work if the target is run from a build directory outside the repo like with Travis...

I'll try to find a way to make it work.

patrickvacek:
@lbonn can this be merged in the meantime? It looks good to me.

lbonn:
Yes!

PR #677, opened 2018-03-30
Run docker as nonroot

patriotyk:


Dockerfile.nop11
```diff
@@ -54,5 +54,8 @@ RUN apt-get update && apt-get -y install \
   valgrind \
   wget
 
+RUN useradd testuser
 WORKDIR aktualizr
-ADD . src
+RUN chown -R testuser:testuser .
+ADD --chown=testuser:testuser . src
+USER testuser:testuser
```
patrickvacek: Why is the nop11 version different from the others? It looks like the RUN and ADD commands are doing the same thing to the current source directory, but I don't know enough about what's really happening to know. If that logic really is necessary, could you leave a comment in the Dockerfile to explain it?

patriotyk: It is not easy to run coverage docker to run tests with user permissions because softhsm requires root access.

patrickvacek: I believe that, but isn't the nop11 Dockerfile the one that explicitly *doesn't* use softhsm?

patriotyk: Yes, it is. But why do You ask?

patrickvacek: If nop11 doesn't use softhsm, why is it the case that it needs the special handling for softhsm? I would've expected it to be the other way around.

PR #678, opened 2018-04-02
Use config dirs

patriotyk:


tests/config_test.cc
```diff
@@ -12,11 +12,14 @@
 #include <boost/program_options.hpp>
 
 #include "bootstrap.h"
-#include "config.h"
+
 #include "crypto.h"
 #include "test_utils.h"
 #include "utils.h"
 
+#define private public  // We need access to private members
```
patrickvacek: Woah, that seems brutal. Is that really the best way to do it? I looked into friending the test class before, but I wasn't very happy with that, so I don't really have anything else to suggest at the moment.

tests/config_test.cc
```diff
@@ -162,6 +165,22 @@ TEST(config, consistent_toml_nonempty) {
...
+  config.updateFromDirs();
+  EXPECT_EQ(config.storage.path, "/z_path");
+}
+
+TEST(config, config_dirs_two_dirs) {
+  TemporaryDirectory temp_dir;
+  Config config;
+  config.config_dirs = {"tests/test_data/config_dirs/one_dir", "tests/test_data/config_dirs/second_one_dir"};
+  config.updateFromDirs();
+  EXPECT_EQ(config.storage.path, "/latest_path");
```
patrickvacek: +1 for clever tests!

tests/config_test.cc
patriotyk: Why we worry about that? It is only tests

patrickvacek: True, but may mean our tests are not testing what we expect them to test in subtle ways. For the record, my previous reference on the subject was here: https://stackoverflow.com/questions/3676664/unit-testing-of-private-methods

The last time this came up for me, I ultimately gave up and made the method I wanted to test public. That was lazy and we ought to decide on a better strategy.

patrickvacek:
BTW there are some merge conflicts to be resolved now. We still need to decide as a team what to do about unit testing private functions, too.

src/config.h
```diff
@@ -134,6 +138,8 @@ class Config {
   ImportConfig import;
 
  private:
+  std::vector<boost::filesystem::path> config_dirs_ = {"/usr/sota/conf.d/", "/etc/sota/conf.d/"};
```
patrickvacek: Should this be `/usr/lib/sota/conf.d` to match the current directory to look for configs?

PR #680, opened 2018-04-03
Fix clang-tidy header filter regex

lbonn:
To also work when using a build directory outside the repository

Noticed the problem after #674. There was no problem in the option selection but only with the header controlling which header files to lint.

This is a bit ugly but looks like it now works in both cases.

### DIFF
--- a/src/sota_tools/accumulator.h
+++ b/src/sota_tools/accumulator.h
@@ -40,42 +40,42 @@ class accumulator_type : public boost::program_options::value_semantic {
     return this;
   }
 
-  virtual std::string name() const { return std::string(); }
+  virtual std::string name() const { return std::string(); }  // NOLINT
 
   /// There are no tokens for an accumulator_type
-  virtual unsigned min_tokens() const { return 0; }
-  virtual unsigned max_tokens() const { return 0; }
+  virtual unsigned min_tokens() const { return 0; }  // NOLINT
+  virtual unsigned max_tokens() const { return 0; }  // NOLINT
 
-  virtual bool adjacent_tokens_only() const { return false; }
+  virtual bool adjacent_tokens_only() const { return false; }  // NOLINT
 
   /// Accumulating from different sources is silly.
-  virtual bool is_composing() const { return false; }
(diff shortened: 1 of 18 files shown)


patrickvacek:
@lbonn Just checking, so does this actually run on our own header files? It looks like it is skipping all header files, but maybe I'm misunderstanding how the filters work.

lbonn:
Yes, it explicitly filters out direct runs on header files (clang-tidy expects its input files to be in `compile_commands.json`) but checks the parts included in source files from headers whose file name match the regex.

That's a bit convoluted but if a header is never included, it will indeed never be checked. It probably needs an explanatory comment.

lbonn:
Also, it looks like we'd need to run the `clang-tidy` target with a maximum of `BUILD_xxx` enabled since it only considers files with a compile command.

src/sota_tools/accumulator.h
```diff
@@ -40,42 +40,42 @@ class accumulator_type : public boost::program_options::value_semantic {
     return this;
   }
 
-  virtual std::string name() const { return std::string(); }
+  virtual std::string name() const { return std::string(); } // NOLINT
```
patrickvacek: What's the complaint here that's being bypassed?

patrickvacek:
+1 for curly brace initialization and `override` instead of `virtual`!

src/sota_tools/accumulator.h
lbonn: clang-tidy complains about virtual and recommends virtual. But some of these methods seem to only exist in the parent class in certain versions of boost (I had different warnings on my machine and CI) and marking a method as overriding while it's not is an error.

Would probably be better to ignore this file altogether since it's not our code but I took the easy route this time.

lbonn:
Yes, clang-analyzer (run by clang-tidy) found some fields to be possibly uninitialized.

patrickvacek:
> Also, it looks like we'd need to run the clang-tidy target with a maximum of BUILD_xxx enabled since it only considers files with a compile command.

Is there still a change to be made for that, or should this get merged? It's good to go from my perspective.

PR #681, opened 2018-04-03
check storage permissions

patriotyk:


tests/uptane_test.cc
```diff
@@ -311,15 +308,15 @@ TEST(Uptane, PetNameCreation) {
     EXPECT_TRUE(uptane.initialize());
 
     EXPECT_TRUE(boost::filesystem::exists(device_path));
-    EXPECT_EQ(Utils::readFile(device_path), test_name2);
+    EXPECT_EQ(test_name2, test_name2);
```
patrickvacek: I think this breaks the point of the test, and it's just testing that a variable is equal to itself.

tests/uptane_test.cc
```diff
@@ -296,8 +293,8 @@ TEST(Uptane, PetNameCreation) {
     Uptane::Repository uptane(conf, storage, http);
     EXPECT_TRUE(uptane.initialize());
 
-    EXPECT_TRUE(boost::filesystem::exists(device_path));
-    test_name2 = Utils::readFile(device_path);
+    EXPECT_TRUE(boost::filesystem::exists(temp_dir.Path() / "device_id"));
+    test_name2 = Utils::readFile(temp_dir.Path() / "device_id");
```
patrickvacek: It took me a minute to realize what was going on here. It looks fine, but I might recommend renaming the nested `temp_dir` even if it is just something like `temp_dir2`. I know logically it is fine as it is, but it makes it unnecessarily tricky to read. That said, thanks for cleaning up all these tests!

tests/uptane_test.cc
```diff
@@ -311,15 +308,15 @@ TEST(Uptane, PetNameCreation) {
...
   }
 
   // If the device_id file is removed, but the field is still present in the
   // config, re-initializing the config should still read the device_id from
   // config.
   {
-    boost::filesystem::remove_all(temp_dir.Path());
-    boost::filesystem::create_directory(temp_dir.Path());
+    TemporaryDirectory temp_dir;
+    conf.storage.path = temp_dir.Path();
```
patrickvacek: Another nested `temp_dir` that I'd suggest renaming.

patriotyk:
I run qemu image and it works well. But oe-selftest fails on some tests. Is there possibility to run those tests  not silently?

patrickvacek:
> Is there possibility to run those tests not silently?

You mean you'd like to see the logs? There's not a simple option to do so, but there are a few things I can recommend:

* You can try reproducing the test manually. The local.conf changes for each test are not hard to find.
* You can ssh into the qemu instances created during testing, so you can jump in and dump the log before the device is killed.
* You can edit the test to wait longer/indefinitely before checking aktualizr-info and/or killing the device and ssh in if you want to play around at leisure.
* You can boot a new qemu using the same image built during the test, since it should still be available.

patrickvacek:
BTW there are some merge conflicts to be resolved now.

patriotyk:
Yes, I see. I will done this PR  tomorow I am on holliday today.

patrickvacek:
Still getting `Storage directory has unsafe permissions` on an actual device.

src/fsstorage.cc
```diff
@@ -12,6 +12,11 @@
 #include "utils.h"
 
 FSStorage::FSStorage(const StorageConfig& config) : INvStorage(config) {
+  struct stat st;
+  stat(config_.path.c_str(), &st);
+  if ((st.st_mode & (S_IWOTH | S_IWGRP)) != 0) {
```
lbonn: Since we'd have to update from old versions with potentially unsafe permissions, don't we need to try to chmod first and only throw if it fails? Otherwise, it would brick updates on the device.

patriotyk: Yes, but task was not to change from unsafe to safe permissions. If I do chmod before every run, what is the reason to check the permissions?

patriotyk:
Tested on qemu now it works

patrickvacek:
@patriotyk Looks like there are some valgrind complaints found in Travis.

patriotyk:
It is another issue and this issue is in valgrind, we need to update valgrind to newer version. But it happens sometimes, so restarting fixes it.
this is issue https://stackoverflow.com/questions/37032339/valgrind-unrecognised-instruction

src/fsstorage.cc
heartsucker: If we `chmod` on boot, then we have the issue where someone does nasty things and then we lock them out *after* their changes. There is no way to ensure that current devices don't get locked out. This would have to be in the realase notes.

heartsucker: Also is this a config dir? If so, i needs to be `700`. If it's a file, `600`. I can't remember what the bit mask constants are.

src/sqlstorage.cc
```diff
@@ -16,7 +17,18 @@ SQLStorage::SQLStorage(const StorageConfig& config) : INvStorage(config) {
                              " missing");
   }
 
-  boost::filesystem::create_directories(config.sqldb_path.parent_path());
+  if (!boost::filesystem::is_directory(config.sqldb_path.parent_path())) {
+    boost::filesystem::create_directories(config.sqldb_path.parent_path());
```
heartsucker: Is there not a way to create a directory and set the permissions all at once? Baiscally `mkdir -m 700`.

src/utils.cc
```diff
@@ -564,6 +564,7 @@ TemporaryFile::~TemporaryFile() { boost::filesystem::remove(tmp_name_); }
 
 void TemporaryFile::PutContents(const std::string &contents) {
   std::ofstream out(Path().c_str());
+  chmod(tmp_name_.c_str(), S_IRUSR | S_IWUSR);
```
heartsucker: We shouldn't `chmod` a file after we create it. There could be a race between creation and access where someone can do nasty things.

src/utils.cc
```diff
@@ -574,6 +575,7 @@ std::string TemporaryFile::PathString() const { return Path().string(); }
 TemporaryDirectory::TemporaryDirectory(const std::string &hint)
     : tmp_name_(SafeTempRoot::Get() / boost::filesystem::unique_path("%%%%-%%%%-" + hint)) {
   boost::filesystem::create_directories(tmp_name_);
+  chmod(tmp_name_.c_str(), S_IRWXU);
```
heartsucker: Again, is there not a way to create the dir and set the permissions all at once?

patriotyk:
Replaced creating dirs and files with custom methods that creates files with correct permissions. But I  don't check for reading flags for group and others because it will break current devices. Also those changes will require changes to meta-updater

src/fsstorage.cc
heartsucker: Yo come to my desk when you see this. I think this will be faster to discuss IRL.

patriotyk:
Now I remove reading permissions for group and others

cajun-rat:
Before this feature is done, we also need to set the permissions on the SQL database.

patriotyk:
Why the parent directory is not enough?

cajun-rat:
I spun up an instance on on qemux86-64, and the permissions for both `/var/sota` and `sql.db` were `755` - I suspect that this is because the directory already exists (it is created at build time)

heartsucker:
This means we need to fix the build as well if that's being created with bad perms.

patriotyk:
Yes, but for old devices I added migration, that removes read permissions from group and others. I will test it again

PR #684, opened 2018-04-04
Feat/garage sign auto version

patrickvacek:
Use a script to fetch garage-sign instead of hardcoding a specific version.

Define `GARAGE_SIGN_VERSION` to control the version to download and package, e.g. `cmake ../ -DGARAGE_SIGN_VERSION=cli-0.3.0-5-g5908997.tgz`

Define `GARAGE_SIGN_SHA256` to provide a hash to check the downloaded file, e.g. `cmake ../ -DGARAGE_SIGN_SHA256=a24efff06bbe1617947d4a08a74f617af9a306d7013f0881407df9d630253b2b`
    
The default is the version whose name sorts highest, which should be the greatest version number. If the layout or content of the tuf-cli releases page changes substantially, this may break.

### DIFF
--- a/scripts/get_garage_sign.py
+++ b/scripts/get_garage_sign.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import argparse
 import hashlib
 import os.path
 import requests
@@ -7,8 +8,18 @@ import shutil
 import sys
 import xml.etree.ElementTree as ET
 
+from pathlib import Path
+
 
 def main():
+    parser = argparse.ArgumentParser(description='Download a specific or the latest version of garage-sign')
+    parser.add_argument('-n', '--name', help='specific version to download')
+    parser.add_argument('-o', '--output', type=Path, default=Path('.'), help='download directory')
+    args = parser.parse_args()
+    if not args.output.exists():
+        print('Error: specified output directory ' + args.output + ' does not exist!')
+        return 1
+
     r = requests.get('https://ats-tuf-cli-releases.s3-eu-central-1.amazonaws.com')
     if r.status_code != 200:
         print('Error: unable to request index!')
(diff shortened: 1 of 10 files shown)


scripts/get_garage_sign.py
```diff
@@ -0,0 +1,110 @@
...
+
+    r = requests.get('https://ats-tuf-cli-releases.s3-eu-central-1.amazonaws.com')
+    if r.status_code != 200:
+        print('Error: unable to request index!')
+        return 1
+    tree = ET.fromstring(r.text)
+    # Assume namespace.
+    ns = '{http://s3.amazonaws.com/doc/2006-03-01/}'
+    items = tree.findall(ns + 'Contents')
+    versions = dict()
```
lbonn: nitpick but perhaps too many list comprehension:

```
for i in items:
    if not i.find(ns + 'Key').text.startswith('cli-'):
        continue
    ...
```

Or use an intermerdiary variable

```
cli_items = [i for i in items if i.find(ns + 'Key').startswith('cli-')]
for i in cli_items:
    ...
```

lbonn: They should be equivalent functionally except if I made a fatal mistake, this is only about readability, which is often subjective. I tested the first version by running the program without argument and it successfully downloaded the archive.

Other than that, it's only a matter of style, really... The first version looks more like traditional imperative programming.

patrickvacek:
Rebased to fix merge conflicts from the recent C++11 PR.

cajun-rat:
Agreed, lets go with this.  I'm having oe-selftest failures on shovel at the moment, not sure if they are due to this or not...

patrickvacek:
> I'm having oe-selftest failures on shovel at the moment, not sure if they are due to this or not...

Are you using https://github.com/advancedtelematic/meta-updater/pull/291? If so, I just realized it was slightly stale, so I've fixed that now. If not, you probably need to, since I changed the location of garage-sign. If that doesn't fix the problem, let me know.

And for the record, I'm not opposed to looking at other/better ways of doing this as well. But for now I think this should be an improvement.

cajun-rat:
I'd missed the meta-updater change, that fixes it!

PR #687, opened 2018-04-06
save/get installed packages in/from storage

patriotyk:


tests/httpfake.h
```diff
@@ -101,14 +101,15 @@ class HttpFake : public HttpInterface {
       std::string hash;
       if (manifest_count == 0) {
         // Check for default initial value of packagemanagerfake.
-        hash = boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest("0")));
+        hash = boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest("")));
         ++manifest_count;
       } else {
         hash = "tst149_ecu_serial";
       }
+      std::cout << data;
```
lbonn: print left over here

src/packagemanagerfake.h
```diff
@@ -28,17 +20,25 @@ class PackageManagerFake : public PackageManagerInterface {
...
+    }
+    return getUnknown();
   }
 
   data::InstallOutcome install(const Uptane::Target &target) const override {
-    (void)target;
-    refhash_fake = target.sha256Hash();
-    return data::InstallOutcome(data::OK, "Pulling ostree image was successful");
+    storage_->saveInstalledVersion(target);
+    return data::InstallOutcome(data::OK, "Installing debian package was successful");
```
patrickvacek: Small complaint: this should maybe say "fake package" or "file package".

patrickvacek:
I think this looks about right except for one question I have: is there any reason we would want to actually fetch and install files locally, be it from treehub or via debian? I think just working with the hashes as it is now is probably fine, but I'm trying to think if we need more for testing aktualizr running on a local machine.

patriotyk:
I think downloading it is good, because it is close to the real behaviour. Also to avoid downloading will require exceptional changes for this backend.

PR #688, opened 2018-04-06
Link socket_activation as an object library

lbonn:
Avoid issues with static libraries linking with each others halfway


lbonn:
Creating another static library was simpler and I didn't know about this CMake feature. I changed it because @koshelev wanted to link an executable with aktualizr_static_lib but it depended on socket_activation which was not installed...

If we use this technique with other aktualizr components, it would be possible to link our secondary executables with a minimum of objects (aktualizr-secondary, aktualizr-info...)

PR #689, opened 2018-04-06
Fix tests/uptane_ci_test

lbonn:


lbonn:
Looks like it fails due to this bug: https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1501545

PR #691, opened 2018-04-10
Do not use saved metadata if there is no root.json

patriotyk:


patrickvacek:
Looks like this should fix the problem. Could you make a test to simulate the issue that came up and prove that we can recover/re-initialize correctly?

patriotyk:
Yes, I will try

PR #692, opened 2018-04-10
Update / re-enable vector tests

heartsucker:
Currently the tests fail because of unimplemented features in `aktualizr`. We need to abort on incorrect hardware IDs.

This PR is **not** ready for merge. There are two `FIXME`s in the runner script. We need to be able to init the `aktualizr` client in the test cases with a hardware ID and ECU ID so that tests that purposely violate that constraint can be correctly passed/failed.


patrickvacek:
FYI @patriotyk

cajun-rat:
The more I think about my changes in `Repository::verifyMeta()` the less I like this solution. It is a bodge, and is only making the eventual fix harder to implement.

I'm going to close this PR for now (we can keep the branch around) and we can look at this again when we work on PRO-4766

heartsucker:
Fair enough.

PR #694, opened 2018-04-11
Bugfix/garage sign prov hang

patrickvacek:
Fix hanging tests from running unit tests with SOTA_PACKED_CREDENTIALS set. Also fixed some other minor stuff.


lbonn:
How did you reproduce your issue with systemd? I just tried a build with `-DBUILD_SYSTEMD=off` on master without libsystemd-dev and it seemed to work. Maybe this option is a bit surprising because it's enabled by default unlike the others (we could change that).

Also, I thought that if the package detection is not run, the various cmake variables (`SYSTEMD_LIBRARIES`, `LIBDPKG_LIBRARY`...) are just left empty and don't make any difference when linked with a target. So I don't believe they need to be moved inside a if clause.

patrickvacek:
@lbonn I have this problem even with -DBUILD_SYSTEMD=OFF. Eric did as well yesterday. I also thought you were right about those variables generally not getting filled if the package wasn't found, but that doesn't seem to be the case for systemd. I'll remove the deb/dpkg one since that appears to be unnecessary.

lbonn:
Oh ok I think I understand, this is a caching issue. If you run `cmake` with the usual options, it will try to find systemd by default and set the variables to NOTFOUND.
Then, if `BUILD_SYSTEMD` is set to off, these variables keep their value from the cache (NOTFOUND).

This can happen with the other libs as well (going from `BUILD_xx=on` to `BUILD_xx=off` is incorrect) but this is only a common case here because systemd is on by default.

Whtat do you think about adding `else() unset(SYSTEMD_LIBRARY CACHE)` there: https://github.com/advancedtelematic/aktualizr/blob/4bb948528f0add49074cbeab2b4da5856c8418ef/CMakeLists.txt#L82
And do the same for the other `BUILD_xxx`?

We can also (additionally) turn systemd off by default and not forget to enable it in the meta-updater recipe.

patrickvacek:
@lbonn Good call, I think you're right about it being a caching problem. I was also thinking about the unset idea, so let me try that.

patrickvacek:
@lbonn `unset` worked.

PR #696, opened 2018-04-12
Better targets metadata comparison check

patriotyk:


patrickvacek:
FYI the test failure is do to a bug that is fixed by https://github.com/advancedtelematic/aktualizr/pull/697, so that's not a blocker here.

patrickvacek:
This looks promising. What's the deal with hardware_identifier and release_counter? Have we just not implemented those yet to be able to compare them here?

@heartsucker I think this needs your approval before we merge.

src/uptane/uptanerepository.cc
```diff
@@ -74,13 +74,29 @@ Json::Value Repository::signVersionManifest(const Json::Value& primary_version_m
 // Check for consistency, signatures are already checked
 bool Repository::verifyMeta(const Uptane::MetaPack& meta) {
   // verify that director and image targets are consistent
-  for (std::vector<Uptane::Target>::const_iterator it = meta.director_targets.targets.begin();
-       it != meta.director_targets.targets.end(); ++it) {
-    std::vector<Uptane::Target>::const_iterator image_target_it;
-    image_target_it = std::find(meta.image_targets.targets.begin(), meta.image_targets.targets.end(), *it);
+  if (meta.director_targets.version() != meta.image_targets.version()) {
```
cajun-rat: @heartsucker  Is this check really correct?

src/uptane/uptanerepository.cc
```diff
@@ -74,13 +74,29 @@ Json::Value Repository::signVersionManifest(const Json::Value& primary_version_m
...
+                                          if (director_target.hashes().size() != image_target.hashes().size())
+                                            return false;
+                                          // TODO: hardware_identifier and release_counter?
+                                          for (auto director_hash : director_target.hashes()) {
+                                            auto found = std::find(image_target.hashes().begin(),
+                                                                   image_target.hashes().end(), director_hash);
+                                            if (found == image_target.hashes().end()) return false;
+                                          }
+                                          return true;
+                                        });
```
cajun-rat: Can you move this logic out into somewhere that can be isolated and unit tested?  This logic is subtle, and without being able to easily drive it, we are going to have a hard time testing and maintaining it.

A free function that compares a director and images target would be fine, because the comparison is asymmetric.

patriotyk: I don't think the free function it is good. We could create Repository in tests, then call verifyMeta, passing MetaPack as argument. 
I will add unit tests for this

cajun-rat: The free function I was thinking of would just replace the lambda function that had the subtle logic in it. It would be a pure function of the two targets objects (I think?)

src/uptane/uptanerepository.cc
```diff
@@ -71,16 +71,38 @@ Json::Value Repository::signVersionManifest(const Json::Value& primary_version_m
   return ecu_version_signed;
 }
 
+bool comapreForConsistency(const Uptane::Target& director_target, const Uptane::Target& image_target) {
+  if (director_target.length() > image_target.length()) return false;
+  if (director_target.filename() != image_target.filename()) return false;
+  if (director_target.hashes().size() != image_target.hashes().size()) return false;
```
cajun-rat: @heartsucker: is this correct?  Can you describe the desired logic for comparing Images and Director targets, please?

I think the algorithm should be:

    for each (hash algorithm known and trusted by aktualizr:
        check if that hash is in both images and director targets:
            both present & match => good
            both present & no match => reject this target
            either missing => continue to the next algorithm

heartsucker: For hashes, it should be this:

```python
# python

for hash_alg in director_hashes:
  if hash_alg not in image_repo_hashes:
    abort()

for hash_alg in image_repo_hashes:
  if hash_alg not in director_hashes:
    abort()
  if image_repo_hashes[hash_alg] != director_hahes[hash_alg]:
    abort()
```

Aktualizr doesn't need to know about all hashes. E.g., aktualizr knows `sha256` and `sha512` but the backend sends `sha512/224` along with the other two. This is ok.

cajun-rat: OK, so: 

* The director needs to copy over exactly the set of hashes that were used to sign the image repository
* Aktualizr checks that the set of hashes match exactly between images and director
* Aktualizr then picks the best hash algorithm that it trusts.

@jerrytrieu @simao does this match your beliefs as to how the server should operate?

cajun-rat:
(rebased)

PR #698, opened 2018-04-12
Pull newer valgrind that understands C++11 instructions.

patrickvacek:
This was causing an annoying tendency for our nop11 builds to fail in CI. More information:
https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1501545
https://stackoverflow.com/questions/37032339/valgrind-unrecognised-instruction

### DIFF
--- a/Dockerfile.nop11
+++ b/Dockerfile.nop11
@@ -51,8 +51,14 @@ RUN apt-get update && apt-get -y install \
   python3-gi \
   python3-openssl \
   python3-venv \
-  valgrind \
   wget
+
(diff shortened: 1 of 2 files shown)


patrickvacek:
This is obnoxious. If I add the command to dump the suppressions, there are no errors. Of course I haven't been able to reproduce this locally. I'll keep trying to sort this out but I'll take suggestions if you have them.

patrickvacek:
Seems to work now that I made the suppression slightly more general to include the issue I was seeing before.

PR #701, opened 2018-04-13
OPC-UA bridge: secondary update support

riqkum:


src/opcuabridge/opcuabridgeclient.cc
```diff
@@ -163,14 +165,25 @@ VersionReport Client::recvVersionReport() const {
...
+    original_manifest.ClientRead(client_);
+  }
+  return original_manifest;
+}
+
 bool Client::sendMetadataFiles(std::vector<MetadataFile>& files) const {
   MetadataFiles metadatafiles;
   bool retval = true;
   if (UA_Client_getState(client_) != UA_CLIENTSTATE_DISCONNECTED) {
+    std::srand(std::time(nullptr));
```
patriotyk: it is very slow and should be run only once, but You run it every time you send metadata files

patriotyk: if You run this method on primary You should not run srand because  because we already call it at aktualizr startup.

src/opcuabridge/metadatafile.h
```diff
@@ -18,6 +18,11 @@ class MetadataFile {
   std::vector<unsigned char>& getMetadata() { return metadata_; }
   const std::vector<unsigned char>& getMetadata() const { return metadata_; }
   void setMetadata(const std::vector<unsigned char>& metadata) { metadata_ = metadata; }
+  void setMetadata(const Json::Value& metadata) {
+    Json::FastWriter fw;
+    std::string s = fw.write(metadata);
+    metadata_ = std::vector<unsigned char>(s.begin(), s.end());
```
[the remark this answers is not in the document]
riqkum: Bound binary data has more general purpose than string, so vector container is used.

riqkum:
Updated and rebased to get #698. Should pass now.

PR #703, opened 2018-04-13
Feat/prov tests

patrickvacek:
First draft of an automatic provisioning test in python.

Requires SOTA_PACKED_CREDENTIALS to be set, so this still won't be run
by default in CI, but hopefully that will change soon.

One can easily imagine versions for implicit and perhaps even HSM
provisioning. The latter half of the python code was borrowed straight
from the oe-selftest code in meta-updater, so it shouldn't be hard to
figure out what to do.


cajun-rat:
(rebased)

PR #704, opened 2018-04-13
Test/jenkins docker

lbonn:
Changes to make a part of the tests run by a Jenkins instance.

Note that the full coverage tests with p11 can now be run as an unprivileged user.


Jenkinsfile
```diff
@@ -0,0 +1,36 @@
+pipeline {
+  agent none
+    environment {
+      JENKINS_RUN = '1'
+    }
+  stages {
```
OYTIS: Cool feature! I mean pipelines in general.
What we probably want is execute tests (coverage, valgring, deb-unstable etc.) in parallel and for that as far as I could understand we need one mega-stage with multiple parallel steps (see https://jenkins.io/doc/book/pipeline/jenkinsfile/#parallel-execution). It can be fixed later though.

lbonn: Yes, I'm currently trying to run the debian install test in parallel in the branch `test/jenkins-debinstall`. It should be close to be working now.

tests/run_vector_tests.sh
```diff
@@ -7,8 +7,9 @@ fi
 
 . venv/bin/activate
 
-pip install wheel
-pip install -r "$1/requirements.txt"
+# use `python -m pip` to avoid problem with long shebangs on travis
+python -m pip install wheel
+python -m pip install -r "$1/requirements.txt"
```
patrickvacek: @lbonn Should these be `python3`?

PR #705, opened 2018-04-16
Continue to split out code into modules

cajun-rat:


cajun-rat:
OK, I think I have it now. I moved the config out into a separate module, which make it possible to remove `aktualizr_static_lib` from aktualizr-secondary

src/package_manager/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+set(SOURCES packagemanagerfactory.cc
```
lbonn: What about this style?

```
set(SOURCES packagemanagerfactory.cc
            packagemanagerfake.cc
)

set(HEADERS packagemanagerfactory.h
            packagemanagerfake.h
            packagemanagerinterface.h)

aktualizr_source_file_checks(${SOURCES} ${HEADERS})
add_library(package_manager OBJECT ${SOURCES})

# debian backend
if(BUILD_DEB)
    target_sources(package_manager PUBLIC debianmanager.cc)
endif(BUILD_DEB)
aktualizr_source_file_checks(debianmanager.cc debianmanager.h)

# ostree backend
if(BUILD_OSTREE)
    target_sources(package_manager PUBLIC ostreemanager.cc ostreereposync.cc)
    target_include_directories(package_manager PUBLIC ${LIBOSTREE_INCLUDE_DIRS})
endif(BUILD_OSTREE)
aktualizr_source_file_checks(ostreemanager.cc ostreereposync.cc
                             ostreemanager.h ostreereposync.h)
```

cajun-rat: Nice idea. Will push an update to this one and src/utilities/CMakeLists.txt too

lbonn: Yes I think it streamlines the CMakeLists a little bit. After a few tries, it seems that `target_sources(... PRIVATE ...)` works better than `PUBLIC` actually.

`PUBLIC` modifies `INTERFACE_SOURCES` which we don't use and it caused some strange error when I tried to do the same in `src/aktualizr_secondary`. I'll push the changes on top of the PR.

src/config/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(SOURCES bootstrap.cc
+            config.cc)
+
+set(HEADERS bootstrap.h
```
lbonn: I am not sure about the config component since I think it should only be used by the primary aktualizr and not shared with the secondary executable which has its own config. It's almost possible now to link the secondary without config now (only need to move `PackageConfig` to `package_manager`).

config_utils is for everyone though (in utilities..?). 

Perhaps `Config` should be named `AktualizrConfig` now that we have other configs. That might be a big change though.

cajun-rat: Is the question whether we want the config object to be at the top or the bottom of the dependency graph?

If we put it at the top (therefore the implementation modules depend on it), then either it will need to contain both primary and secondary config, or we have to duplicate the implementation inheritance tree (config_common/config_primary/config_secondary).

The other option would be to put the config object at the bottom of the dependency graph (i.e. it depends on everything). This would let us put the configuration inside the modules that use it, but would tangle the configuration with the implementation.

The model I have in my head is that configuration goes through a set of transformations:

- sota.toml (byte string)
-  Boost property tree (untyped tree-ish thing)
- Typed configuration holder object (`Config`). This could/should be validated for consistency.
- Dependency injection then uses `Config` to instantiate the actual system.

lbonn: It used to follow the first model (everything depends on it) but I've moved it to the bottom (almost completely) when creating the secondary config since I felt it was a bit cleaner and make it easier to have self-contained components.

We can still go back though.

cajun-rat: Summary of a F2F discussion: We will put the config at the bottom of the dependency tree. This means that:

* I'll undo the change to move config into a module
* I'll move config_utils into utilities
* In future refactorings we will break up the config object into smaller pieces, and move these into the modules.

PR #706, opened 2018-04-17
Feat/pro 4915/clang tidy

patriotyk:


CMakeLists.txt
```diff
@@ -176,6 +176,16 @@ if(CLANG_TIDY)
     add_dependencies(qa clang-tidy)
     function(aktualizr_clang_tidy)
         file(RELATIVE_PATH SUBDIR ${CMAKE_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
+        set(CHECKS "\
```
lbonn: We're currently using `.clang-tidy` to define the list of checks: https://github.com/advancedtelematic/aktualizr/blob/53626c2e2ddae139216fda53523cef7ea687729a/.clang-tidy

We can switch to using CMake if it makes sense but we'll probably want to keep the whole list in one place.

cajun-rat:
This look good.  The `-fix` option in clang-tidy feels a bit dangerous, but it will be interesting to see if it helps or not.

Can you move the list of checks into `.clang-tidy` and fix the CI error, then this should be good to go

```
/aktualizr/src/src/asn1-cer.cc: In function 'uint8_t cer_decode_token(const string&, int32_t*, int32_t*, std::__cxx11::string*)':
/aktualizr/src/src/asn1-cer.cc:163:81: error: suggest parentheses around comparison in operand of '==' [-Werror=parentheses]
         auto sign = static_cast<int>(static_cast<int>(((((content[0] & 0x80)))) == 0 == 0));
```

patriotyk:
I use -fix on my local machine to fix warnings, but some warnings it fixes wrongly. This PR is not ready yet so in  the final I will remove it

src/aktualizr_secondary/opcuaserver_secondary_delegate.h
```diff
@@ -11,15 +11,15 @@ class AktualizrSecondaryCommon;
...
-  void handleDirectoryFilesSynchronized(opcuabridge::ServerModel*) override;
-  void handleOriginalManifestRequested(opcuabridge::ServerModel*) override;
-  void handleDirectoryFileListRequested(opcuabridge::ServerModel*) override;
+  void handleServerInitialized(opcuabridge::ServerModel* /*unused*/) override;
+  void handleVersionReportRequested(opcuabridge::ServerModel* /*unused*/) override;
+  void handleMetaDataFileReceived(opcuabridge::ServerModel* /*unused*/) override;
+  void handleAllMetaDataFilesReceived(opcuabridge::ServerModel* /*unused*/) override;
+  void handleDirectoryFilesSynchronized(opcuabridge::ServerModel* /*unused*/) override;
+  void handleOriginalManifestRequested(opcuabridge::ServerModel* /*unused*/) override;
+  void handleDirectoryFileListRequested(opcuabridge::ServerModel* /*unused*/) override;
```
patrickvacek: I don't understand these `unused` comments. The variables in question look pretty well used to me. (This isn't critical and shouldn't hold up the PR.)

src/asn1/asn1-cer.cc
```diff
@@ -72,76 +73,94 @@ std::string cer_encode_string(const std::string& contents, ASN1_UniversalTag tag
...
+  int len_len = (static_cast<uint8_t>(content[0])) & 0x7F;
   *endpos = len_len + 1;
-  if (len_len > 4) return -2;
+  if (len_len > 4) {
+    return -2;
+  }
 
   int32_t res = 0;
-  for (int i = 0; i < len_len; i++) {
+  for (uint64_t i = 0; i < len_len; i++) {
```
patrickvacek: If `len_len` is an `int`, then so should `i`. Isn't it best to compare two values of the same type?

src/asn1/asn1-cerstream.cc
```diff
@@ -154,51 +166,61 @@ Deserializer& Deserializer::operator>>(const Token& tok) {
...
+              ++nestedness;
+            }
           } else if (ret == kAsn1EndSequence) {
-            if (nestedness == 0)
+            if (nestedness == 0) {
               break;
-            else
-              --nestedness;
+            }
+            { --nestedness; }
```
patrickvacek: Why is this in brackets?

src/bootstrap.cc
```diff
@@ -2,7 +2,7 @@
 
 #include <boost/algorithm/string.hpp>
 
-#include <stdio.h>
+#include <cstdio>
```
patrickvacek: Isn't this asking for trouble since we aren't prefixing the C functions with `std::`? (See https://stackoverflow.com/questions/10460250/cstdio-stdio-h-namespace and the `fmemopen` below.)

src/asn1/asn1-cerstream.cc
patriotyk: because i"f else" should be always with brackets. Without breckets it is easy to write an incorrect logic

patrickvacek: So it should look like this?

```
if (nestedness == 0) {
  break;
}
{ --nestedness; }
```

That just seems odd to me. It won't hurt anything, but I don't get it. If there was an `else` it would make sense.

patrickvacek: (For the record, I'm 100% in favor of adding the brackets for all ifs and elses, but this case is a little weirder to me.)

src/cert_provider/main.cc
```diff
@@ -363,19 +384,23 @@ int main(int argc, char* argv[]) {
   boost::filesystem::path url_file = "gateway.url";
   if (!config_path.empty()) {
     Config config(config_path);
-    // TODO: provide path to root directory in `--local` parameter
+    // TODO(lbonn): provide path to root directory in `--local` parameter
```
patrickvacek: Why is lbonn's name getting injected in every TODO? I would've thought it would inject whoever last edited that line, but in this case, that would be Anton.

src/httpinterface.h
```diff
@@ -10,14 +11,14 @@
...
+  HttpResponse(std::string body_in, const int64_t http_status_code_in, CURLcode curl_code_in,
+               std::string error_message_in)
+      : body(std::move(body_in)),
         http_status_code(http_status_code_in),
         curl_code(curl_code_in),
-        error_message(error_message_in) {}
+        error_message(std::move(error_message_in)) {}
   std::string body;
-  long http_status_code;
+  int64_t http_status_code;
```
patrickvacek: I recently changed this to `long` because that is what the curl interface returns and that's the only place we get it from. Why would we want to use anything other than that?

src/socketgateway.cc
```diff
@@ -70,15 +71,16 @@ void SocketGateway::commandsWorker(int socket, std::shared_ptr<command::Channel>
 
 void SocketGateway::eventsServer() {
   events_socket = socket(AF_UNIX, SOCK_STREAM, 0);
-  struct sockaddr_un cli_addr, addr;
+  struct sockaddr_un cli_addr {
+  }, addr{};
```
patrickvacek: This is a typo, right? Should be all on one line.

src/socketgateway.cc
```diff
@@ -92,15 +94,16 @@ void SocketGateway::eventsServer() {
 void SocketGateway::commandsServer() {
   commands_socket = socket(AF_UNIX, SOCK_STREAM, 0);
 
-  struct sockaddr_un cli_addr, addr;
+  struct sockaddr_un cli_addr {
+  }, addr{};
```
patrickvacek: And again.

src/sota_tools/deploy.h
```diff
@@ -13,9 +13,8 @@
  *                 are fetched from disk), or OSTreeHttpRepo (in which case
  *                 the objects will be pulled over https).
  */
-bool UploadToTreehub(const OSTreeRepo::ptr src_repo, const ServerCredentials& push_credentials,
-                     const OSTreeHash& ostree_commit, const std::string& cacerts, const bool dryrun,
-                     const int max_curl_requests);
+bool UploadToTreehub(const OSTreeRepo::ptr& src_repo, const ServerCredentials& push_credentials,
+                     const OSTreeHash& ostree_commit, const std::string& cacerts, bool dryrun, int max_curl_requests);
```
patrickvacek: What's wrong with `const`?

src/sota_tools/garage_check.cc
```diff
@@ -124,12 +126,13 @@ int main(int argc, char **argv) {
     return EXIT_FAILURE;
   }
 
-  long http_code;
+  int64_t http_code;
```
patrickvacek: Again...

src/sota_tools/ostree_hash.cc
```diff
@@ -15,7 +15,7 @@ OSTreeHash OSTreeHash::Parse(const std::string& hash) {
   for (int i = 0; i < 32; i++) {
     char byte_string[3];
     byte_string[2] = 0;
-    unsigned long byte_holder;
+    uint64_t byte_holder;
```
patrickvacek: There is some major irony in using a `uint64_t` to store a byte.

src/sota_tools/ostree_object.cc
```diff
@@ -212,7 +221,7 @@ void OSTreeObject::CurlDone(CURLM *curl_multi_handle) {
   refcount_--;            // Because curl now doesn't have a reference to us
   assert(refcount_ > 0);  // At least our parent should have a reference to us
 
-  long rescode = 0;
+  int64_t rescode = 0;
```
patrickvacek: again!

src/sota_tools/ostree_object.h
```diff
@@ -40,7 +40,7 @@ class OSTreeObject {
 
   void MakeTestRequest(const TreehubServer& push_target, CURLM* curl_multi_handle);
 
-  void Upload(const TreehubServer& push_target, CURLM* curl_multi_handle, const bool dryrun);
+  void Upload(const TreehubServer& push_target, CURLM* curl_multi_handle, bool dryrun);
```
patrickvacek: Alas, poor const.

src/sota_tools/request_pool.cc
```diff
@@ -67,12 +69,12 @@ void RequestPool::LoopListen() {
   FD_ZERO(&fdread);
   FD_ZERO(&fdwrite);
   FD_ZERO(&fdexcept);
-  long timeoutms = 0;
+  int64_t timeoutms = 0;
   mc = curl_multi_timeout(multi_, &timeoutms);
```
patrickvacek: `curl_multi_timeout` also is defined to use `long`.

src/uptane/managedsecondary.cc
```diff
@@ -50,16 +50,18 @@ bool ManagedSecondary::putMetadata(const MetaPack &meta_pack) {
...
 }
 
 int ManagedSecondary::getRootVersion(const bool director) {
-  if (director)
+  if (director) {
     return current_meta.director_root.version();
-  else
-    return current_meta.image_root.version();
+  }
+  { return current_meta.image_root.version(); }
```
patrickvacek: Another unnecessary bracket situation.

src/utilities/httpclient.cc
```diff
@@ -94,18 +93,17 @@ void HttpClient::setCerts(const std::string& ca, CryptoSource ca_source, const s
 
   if (ca_source == kPkcs11) {
     throw std::runtime_error("Accessing CA certificate on PKCS11 devices isn't currently supported");
-  } else {  // ca_source =kFile
-    std::unique_ptr<TemporaryFile> tmp_ca_file = std::make_unique<TemporaryFile>("tls-ca");
-    tmp_ca_file->PutContents(ca);
-    curl_easy_setopt(curl, CURLOPT_CAINFO, tmp_ca_file->Path().c_str());
-    tls_ca_file = std::move_if_noexcept(tmp_ca_file);
-  }
+  }  // ca_source =kFile
```
patrickvacek: This comment either belongs on a separate line or can be removed entirely.

src/utilities/keymanager.cc
```diff
@@ -53,7 +54,7 @@ void KeyManager::loadKeys(const std::string *pkey_content, const std::string *ce
     }
     if (!ca.empty()) {
       if (tmp_ca_file == nullptr) {
-        tmp_ca_file = std::make_unique<TemporaryFile>("tls-ca");
+        tmp_ca_file = std_::make_unique<TemporaryFile>("tls-ca");
```
patrickvacek: This is weird. I get that adding stuff to the `std` namespace is asking for trouble, but in this case, the point is that we won't need to do it anymore after we move to C++14, and we want the switch to be seamless. In this case, it will be anything but, since we will have to change everything back from `std_` to `std`.

patrickvacek:
There is a lot of good stuff here. I'd like to get this merged as soon as possible. Most of my comments are fairly trivial, so sorry that there are so many. I stopped keeping track of the `const bool`, curl-related `long`, and `TODO(lbonn)` changes because there were so many, but all three of those seem odd/unnecessary/wrong to me across the board. I may stand to be corrected.

src/httpinterface.h
patriotyk: because long is not size defined. So it is not crossplatform. But int64_t is always 64 bit integer

src/sota_tools/deploy.h
patriotyk: Because parameter is passed by copy, not reffernce.

src/httpinterface.h
patrickvacek: Right, but if the curl headers use long, the safest option for us is to use the same thing, since on one machine, they will compile to the same thing. If long != int64_t on some machine, then we may have compatibility problems. Unlikely to cause an actual problem, but possible.

src/sota_tools/deploy.h
patrickvacek: There is still a semantic value in having a copied-by-value variable be const, as in, it's worth specifying that the value should not change.

src/aktualizr_secondary/opcuaserver_secondary_delegate.h
patriotyk: That's mean that the variables in declarations is never used. Only in implementation, but I think we could remove this check and restore old names

src/asn1/asn1-cer.cc
patriotyk: Yes, You are right

src/aktualizr_secondary/opcuaserver_secondary_delegate.h
patrickvacek: Not a big deal on this one, it's just confusing, since it the `unused` comment makes it seem like the variable is unused in the implementation. I know the names can be left out of the declaration, but it would probably be best if they matched the definition name (in this case, `model`).

src/socketgateway.cc
patriotyk: Yes, it is better  to do it in one line

src/sota_tools/deploy.cc
```diff
@@ -175,11 +175,11 @@ bool PushRootRef(const ServerCredentials &push_credentials, const OSTreeRef &ref
...
     ref.PushRef(push_server, easy_handle);
     CURLcode err = curl_easy_perform(easy_handle);
-    if (err) {
+    if (err != 0u) {
       LOG_ERROR << "Error pushing root ref:" << curl_easy_strerror(err);
       errors++;
     }
-    long rescode;
+    int64_t rescode;
     curl_easy_getinfo(easy_handle, CURLINFO_RESPONSE_CODE, &rescode);
```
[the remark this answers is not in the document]
patriotyk: Yes will fix it

src/sota_tools/deploy.h
patriotyk: Hm, I think it is controversial issue. I did it because some check gives me a warning

src/sota_tools/garage_check.cc
patriotyk: will done

src/sota_tools/ostree_hash.cc
patriotyk: )))

src/sota_tools/deploy.h
patrickvacek: Indeed, I did not realize how controversial it was! This SO answer shares my perspective: https://stackoverflow.com/a/117557/289099. Since it doesn't programmaticallly matter, I'm not going to argue hard on this one.

src/sota_tools/request_pool.cc
patriotyk: ok

src/uptane/managedsecondary.cc
patriotyk: ok

src/utilities/httpclient.cc
patriotyk: ok

src/utilities/keymanager.cc
patriotyk: We could just use "replace all" std_:: to std:: Or You suggest to replace it back now, and disbale the check?

patrickvacek: I know, replace all is fine, but it's just tedious that we have to change all the code twice. Easier to just let it be, right? I can kind of see it both ways, though, since it is confusing to be adding custom functions to the std namespace. I don't have a strict opinion here.

.clang-tidy
```diff
@@ -1,5 +1,5 @@
 ---
-Checks:          'clang-analyzer-*,modernize-use-override,readability-inconsistent-declaration-parameter-name'
+Checks:          'clang-analyzer-*, performance-*,readability-*,-readability-identifier-naming,-readability-container-size-empty,-readability-redundant-declaration,modernize-*,-modernize-avoid-bind,-modernize-loop-convert,-modernize-use-using,llvm-namespace-comment,cppcoreguidelines-pro-*,-cppcoreguidelines-pro-type-reinterpret-cast,-cppcoreguidelines-pro-type-vararg,-cppcoreguidelines-pro-bounds-pointer-arithmetic,-cppcoreguidelines-pro-type-const-cast,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-pro-bounds-constant-array-index,cert-*,-cert-err58-cpp,-cert-env33-c,-cert-err09-cpp,-cert-err60-cpp,-cert-err61-cpp, -cert-err61-cpp, bugprone-*, google-*,-google-runtime-references,-google-runtime-member-string-references, -google-explicit-constructor, misc-*, -misc-unused-parameters, -misc-macro-parentheses, -misc-throw-by-value-catch-by-reference, -misc-suspicious-enum-usage'
 WarningsAsErrors: ''
 AnalyzeTemporaryDtors: false
 User:            lbonn
```
patrickvacek: I wonder if this `User: lbonn` line is the source of the `TODO(lbonn)` problems. Is it necessary or wise to keep that line?

src/asn1/asn1-cer.cc
patriotyk: changed len_len to uin64_t

.clang-tidy
patriotyk: Yes You are right, but with removed `User: lbonn` it suggest to add everywhere my username, so it doesn't use git blame. I will remove this check

src/bootstrap.cc
patriotyk: yes You are right - https://clang.llvm.org/extra/clang-tidy/checks/modernize-deprecated-headers.html

patriotyk: We need to prepend all c functions with std::

patriotyk: I replaced all changes back

src/sotauptaneclient.cc
```diff
@@ -14,12 +16,12 @@
 #include "utilities/utils.h"
 
 SotaUptaneClient::SotaUptaneClient(Config &config_in, std::shared_ptr<event::Channel> events_channel_in,
-                                   Uptane::Repository &repo, const std::shared_ptr<INvStorage> storage_in,
+                                   Uptane::Repository &repo, std::shared_ptr<INvStorage> storage_in,
                                    HttpInterface &http_client)
     : config(config_in),
-      events_channel(events_channel_in),
+      events_channel(std::move(std::move(events_channel_in))),
```
patrickvacek: I know this is already merged, but is this really intentional to have a nested `std::move`? I've never seen that before!

CMakeLists.txt
```diff
@@ -186,7 +186,7 @@ if(CLANG_TIDY)
             endif()
             string(REPLACE "/" "_" TARGETNAME "aktualizr_clang_tidy-${SUBDIR}-${FILE}")
             add_custom_target(${TARGETNAME}
-                COMMAND ${CLANG_TIDY} -quiet -header-filter=\(${CMAKE_SOURCE_DIR}|\\.\\.\)/src/.* --extra-arg-before=-Wno-unknown-warning-option -format-style=file -warnings-as-errors=* -p ${CMAKE_BINARY_DIR} ${FILE}
+                COMMAND ${CLANG_TIDY} -quiet -header-filter=\${CMAKE_SOURCE_DIR}/src/.* --extra-arg-before=-Wno-unknown-warning-option -format-style=file -warnings-as-errors=* -p ${CMAKE_BINARY_DIR} ${FILE}
```
lbonn: This probably broke the checks run on our header files. The regex looks horrible but was made to work on Travis and local dev environments: https://github.com/advancedtelematic/aktualizr/pull/680

PR #707, opened 2018-04-17
Test/jenkins debinstall

lbonn:
Tests the install of the deb packaged version inside docker.

Based on #704


cajun-rat:
@lbonn What is required to get this in?

lbonn:
It's basically done it tests the debian package install and stores the .deb files, I was just using this branch to make trigger tests lately.

I'll just clean it a bit and remove the `not-ready` label when I'm done.

lbonn:
Ok I think I found the solution for the small permission issue I still had. It should be ready if our CI tests pass.

lbonn:
It should be fine now, I had a difficult to track issue with serial id which was provisioned directly in the fs storage for the test.

src/sotauptaneclient.cc
```diff
@@ -73,6 +73,7 @@ std::vector<Uptane::Target> SotaUptaneClient::findForEcu(const std::vector<Uptan
 }
 
 data::InstallOutcome SotaUptaneClient::PackageInstall(const Uptane::Target &target) {
+  LOG_INFO << "Installing package using " << pacman->name() << " package manager";
```
patrickvacek: +1

tests/config/selfupdate.toml
```diff
@@ -5,19 +5,20 @@ cert_source = "file"
...
-tls_cacert_path = "ca.pem"
-tls_pkey_path = "pkey.pem"
-tls_clientcert_path = "client.pem"
+path = "/tmp/aktualizr-storage"
+uptane_metadata_path = "/tmp/aktualizr-storage/metadata"
+uptane_private_key_path = "/persistent/prov_selfupdate/ecukey.pem"
+uptane_public_key_path = "/persistent/prov_selfupdate/ecukey.pub"
+tls_cacert_path = "/persistent/prov_selfupdate/ca.pem"
+tls_pkey_path = "/persistent/prov_selfupdate/pkey.pem"
+tls_clientcert_path = "/persistent/prov_selfupdate/client.pem"
```
patrickvacek: Not that it really matters, but you shouldn't need to specify the paths for each of these if you've already specified it in `path`.

lbonn: The main point is that they come from outside the storage path. Is there a more succinct way to define them?

patrickvacek: Oh crap, for some reason I read this wrong. Only `uptane_metadata_path` is in `path`, so that's the only one my comment would apply to. You can ignore my comment, sorry!

lbonn: No worries!

PR #710, opened 2018-04-19
Trigger poll event in poll event processor to avoid accumulation of events

OYTIS:
Accumulation gets especially annoying when debugging aktualizr, but can also slow everything down if network connection is slow/unstable.

### DIFF
--- a/src/aktualizr.cc
+++ b/src/aktualizr.cc
@@ -31,9 +31,9 @@ Aktualizr::Aktualizr(Config &config) : config_(config) {
 
 int Aktualizr::run() {
   std::shared_ptr<command::Channel> commands_channel{new command::Channel};
-  event::Channel events_channel;
+  std::shared_ptr<event::Channel> events_channel{new event::Channel};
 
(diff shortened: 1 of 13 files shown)


cajun-rat:
I think the CI error with this is real--I get the following from test_uptane:


    terminate called after throwing an instance of 'std::logic_error' 
    what():  Attempt to write to closed channel.

OYTIS:
@cajun-rat 
It is. Looks like nobody is really closing it, it just gets out of scope with unpredictable result. I made it a shared pointer everywhere, hope that helps. Also I removed this exception, because if somebody desides to use close() it will be subject to run condition (or we'll need an atomic operation 'check if not closed and write' on a Channel).

OYTIS:
Maybe it's also worth making events_channel a shared pointer just for symmetry (and to avoid future problems of course).

cajun-rat:
(Rebased)

cajun-rat:
Approved, subject to CI passing.

PR #712, opened 2018-04-20
Config documentation

heartsucker:
I'm sitting here with @OYTIS and we can't figure out what some of the options do. This PR is an attempt to sort out what everything does. @OYTIS says everything has a default, but it's not obvious where those come from. Also, some things are relative paths and some are absolute paths. That's also not obvious. And somethings are misnomers like `ipdiscovery_host` is actually a subject mask, not a host. Anyway. Here's as start. Someone else should push to this branch on top and fill in the rest.

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -1,0 +1,184 @@
+= Configuration
+
+Aktualizr is configured via `.toml` config file and passed to the application via the `--config` flag.
+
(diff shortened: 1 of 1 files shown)


cajun-rat:
Let's just put this in.  Although not complete it is at least (I hope) correct, and gives us something to build on.

PR #713, opened 2018-04-23
Check for empty database

patriotyk:


src/sqlstorage.cc
```diff
@@ -1006,6 +1006,9 @@ bool SQLStorage::dbMigrate() {
   int schema_version = getVersion();
   if (schema_version == kSqlSchemaVersion) {
     return true;
+  } else if (schema_version == -2) {
+    LOG_ERROR << "We point to the wrong sqlite database file.";
```
lbonn: When should this happen? Corrupted db or really weird configuration?

It would be nice to use an enum instead of magic numbers, it would help readability (for example DBOPEN_SUCCESS: 0, DBOPEN_NOEXIST: -1, DBOPEN_INVALID: -2)

cajun-rat: 'it would be nice to use an enum instead of magic numbers'  Or a class...

tests/sqlstorage_test.cc
```diff
@@ -126,6 +127,21 @@ TEST(sqlstorage, MigrationVersionCheck) {
...
+  config.sqldb_path = temp_dir.Path() / "test.db";
+  config.schemas_path = "config/schemas";
+  SQLite3Guard db(config.sqldb_path.c_str());
+  if (db.exec("CREATE TABLE some_table(somefield INTEGER);", NULL, NULL) != SQLITE_OK) {
+    FAIL();
+  }
+
+  SQLStorage storage(config);
+  EXPECT_EQ(storage.getVersion(), -2);
+}
```
cajun-rat: +1 for a test :)

src/sqlstorage.cc
cajun-rat: Something like this?

~~~
class SqlSchemaVersion {
  public:
    static SqlSchemaVersion EmptyDatabase() {
      return SqlSchemaVersion(-1);
    }

    static SqlSchemaVersion InvalidDatabase() {
      return SqlSchemaVersion(-2);
    }
    explicit SqlSchemaVersion(int v) : version_(v) {}

    SqlSchemaVersion& operator++() {
      if( version_ == -2) {
        throw std::logic_error("InvalidDatabase");
      }
      version_++;
      return *this;
    }

    bool operator==(const SqlSchemaVersion) {}  // etc
  private:
    int version_;
    friend std::ostream &operator<<(std::ostream &os, const SqlSchemaVersion &v);
};
~~~

PR #714, opened 2018-04-23
Macos compile

koshelev:
Should allow to compile project without ostree and systemd under macos

### DIFF
--- a/src/aktualizr_info/CMakeLists.txt
+++ b/src/aktualizr_info/CMakeLists.txt
@@ -11,8 +11,12 @@ target_link_libraries(aktualizr-info aktualizr_static_lib
   ${GLIB2_LIBRARIES}
   ${LibArchive_LIBRARIES}
   ${LIBOSTREE_LIBRARIES}
-  ${SQLITE3_LIBRARIES}
-  gcov)
+  ${SQLITE3_LIBRARIES})
+
+
+if(BUILD_WITH_CODE_COVERAGE)
+    target_link_libraries(aktualizr-info gcov)
+endif(BUILD_WITH_CODE_COVERAGE)
 
 install(TARGETS aktualizr-info
         COMPONENT aktualizr
--- a/src/aktualizr_secondary/CMakeLists.txt
+++ b/src/aktualizr_secondary/CMakeLists.txt
@@ -48,8 +48,11 @@ target_link_libraries(aktualizr-secondary
     ${LIBDPKG_LIBRARIES}
     ${LIBP11_LIBRARIES}
     ${SYSTEMD_LIBRARY}
-    gcov
     )
+
+if(BUILD_WITH_CODE_COVERAGE)
+    target_link_libraries(aktualizr-secondary gcov)
+endif(BUILD_WITH_CODE_COVERAGE)
 
 install(TARGETS aktualizr-secondary
         COMPONENT aktualizr
--- a/src/cert_provider/main.cc
+++ b/src/cert_provider/main.cc
@@ -230,37 +230,35 @@ bool generate_and_sign(const std::string& cacert_path, const std::string& capkey
 
   // serialize private key
   char* privkey_buf;
-  size_t privkey_len;
-  FILE* privkey_file = open_memstream(&privkey_buf, &privkey_len);
+  BIO* privkey_file = BIO_new(BIO_s_mem());
   if (!privkey_file) {
     std::cerr << "Error opening memstream" << std::endl;
     return false;
   }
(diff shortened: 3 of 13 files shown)


src/aktualizr_secondary/CMakeLists.txt
```diff
@@ -48,9 +48,12 @@ target_link_libraries(aktualizr-secondary
     ${LIBDPKG_LIBRARIES}
     ${LIBP11_LIBRARIES}
     ${SYSTEMD_LIBRARY}
-    gcov
     )
 
+if(BUILD_WITH_CODE_COVERAGE)
+    target_link_libraries(aktualizr_secondary gcov)
```
lbonn: Typo here: `aktualizr-secondary` (executable names are hyphen-separated)

src/aktualizr_info/CMakeLists.txt
```diff
@@ -11,8 +11,12 @@ target_link_libraries(aktualizr-info aktualizr_static_lib
   ${GLIB2_LIBRARIES}
   ${LibArchive_LIBRARIES}
   ${LIBOSTREE_LIBRARIES}
-  ${SQLITE3_LIBRARIES}
-  gcov)
+  ${SQLITE3_LIBRARIES})
+
+
+if(BUILD_WITH_CODE_COVERAGE)
+    target_link_libraries(aktualizr_info gcov)
```
lbonn: Here as well: `aktualizr-info`

src/cert_provider/main.cc
```diff
@@ -230,37 +230,35 @@ bool generate_and_sign(const std::string& cacert_path, const std::string& capkey
...
     return false;
   }
-  ret = PEM_write_X509(cert_file, certificate.get());
-  fclose(cert_file);
+  ret = PEM_write_bio_X509(cert_file, certificate.get());
   if (!ret) {
     std::cerr << "PEM_write_X509" << std::endl;
     return false;
   }
+  auto cert_len = BIO_get_mem_data(cert_file, cert_buf);
```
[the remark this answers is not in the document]
cajun-rat: Wow, openssl *is* special!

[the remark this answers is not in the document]
patrickvacek: Brilliant design. I really wish we could do without macro magic at this point.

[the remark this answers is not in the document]
koshelev: Good catch! Second argument should be a pointer to a pointer` long BIO_get_mem_data(BIO *b, char **pp)`

lbonn:
All this manual memory management makes me anxious. I'm gonna work on using `StructGuard` everywhere when we work with openssl objects.

tests/crypto_test.cc
```diff
@@ -233,7 +233,7 @@ TEST(crypto, parsep12_FAIL) {
     EXPECT_TRUE(false) << " could not open tests/test_data/priv.key";
   }
   bool result = Crypto::parseP12(p12src, "", &pkey, &cert, &ca);
-  fclose(bad_p12file);
+  BIO_free(p12src);
```
koshelev: Ouch... Good catch, thanks!

src/socketgateway.cc
```diff
@@ -8,6 +8,10 @@
 
 #include "logging.h"
 
+#ifndef MSG_NOSIGNAL
```
patriotyk: Why do we need this defines?

lbonn: Used to not get a `SIGPIPE` after a broken pipe during `send()`. Note that we also incorrectly use it for `recv` but according to the man page it doesn't apply there. We'd probably want to use it for our secondary ipc implem as well...

It's not available on OSX: https://nwat.xyz/blog/2014/01/16/porting-msg_more-and-msg_nosigpipe-to-osx/.

src/utilities/crypto.cc
```diff
@@ -175,70 +175,69 @@ bool Crypto::VerifySignature(const PublicKey &public_key, const std::string &sig
...
+  ca_len = BIO_get_mem_data(ca_sink, &ca_buf);
   *out_ca = std::string(ca_buf, ca_len);
-  free(ca_buf);
+  BIO_free(ca_sink);
 
-  fclose(cert_file);
+  cert_len = BIO_get_mem_data(cert_sink, &cert_buf);
   *out_cert = std::string(cert_buf, cert_len);
-  free(cert_buf);
+  BIO_free(cert_sink);
```
cajun-rat: We should think about a openssl BIO wrapper class with methods to convert to std::string etc, but I think @lbonn is already looking into that.

lbonn: Right now, the wrappers are just unique_ptr and do not provide any additional helper methods. But we can discuss suggestions in #717

PR #717, opened 2018-04-24
Use automatic memory management for OpenSSL objects

lbonn:
Freeing the right objects at the right time is tedious...

Based on #714.

### DIFF
--- a/src/utilities/crypto.cc
+++ b/src/utilities/crypto.cc
@@ -34,29 +34,28 @@ std::string Crypto::sha512digest(const std::string &text) {
 }
 
 std::string Crypto::RSAPSSSign(ENGINE *engine, const std::string &private_key, const std::string &message) {
-  EVP_PKEY *key;
-  RSA *rsa = nullptr;
+  StructGuard<EVP_PKEY> key(nullptr, EVP_PKEY_free);
+  StructGuard<RSA> rsa(nullptr, RSA_free);
   if (engine != nullptr) {
-    key = ENGINE_load_private_key(engine, private_key.c_str(), nullptr, nullptr);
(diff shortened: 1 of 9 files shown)


cajun-rat:
@lbonn I've just merged the PR this was built on. Can you rebase, please?

lbonn:
Rebased

cajun-rat:
What about something like:

~~~
class BioBuffer {
   BioBuffer() : bio_(BIO_new(BIO_s_mem())); {}
   explicit BioBuffer(const std::string&);
   // move assignment/ctor
   BIO* bio();
   std::string str() const;
 private:
  BIO* bio_;  
};
Crypto::parseP12(const BioBuffer&, ...);
~~~

src/utilities/crypto.cc
```diff
@@ -275,72 +269,56 @@ bool Crypto::generateRSAKeyPair(KeyType key_type, std::string *public_key, std::
...
   if (!pubkey_sink) {
     return false;
   }
-  ret = PEM_write_bio_PUBKEY(pubkey_sink, pkey);
+  ret = PEM_write_bio_PUBKEY(pubkey_sink.get(), pkey.get());
   if (ret != 1) {
-    RSA_free(r);
-#if AKTUALIZR_OPENSSL_AFTER_11
-    BN_free(bne);
-#endif
```
cajun-rat: Is this logic definitely not needed any more?  I can't see where it moved to. I think @OYTIS understands the reasons for it.

lbonn:
@cajun-rat  yes actually it looks like it could work. I'll give it a go

src/utilities/crypto.cc
cajun-rat: Ah, OK. I'd missed that

PR #718, opened 2018-04-24
Separate libaktualizr from binaries in the source code tree

OYTIS:
The change is not purely aesthetic, the idea here is to prepare the ground for Android port, which will need libaktualizr, but not the binaries. Sorry for a lot of affected files, most are just renames or renamed includes (I really need to learn coccinelle at some point). Meaningful changes are only in various CMakeLists.


lbonn:
Unrelated to the PR but I believe that coccinelle only targets C code unfortunately. Super cool though.

OYTIS:
Yes, unfortunately it seems you're right.
BTW, has anyone tried to compile with BUILD_OPCUA so far? I've got a couple of compilation issues on shovel.

OYTIS:
I believe @lbonn's suggestion to to get rid of aktualizr_static_lib as a whole is what we should implement. It's a bit of fake modularity, because some modules depend on the others (esp. the one that will contain sotauptaneclient.cc), but will still be an improvement to our current structure.
Anyway, I think it can be built on top of this PR. Thoughts/objections?

patrickvacek:
> get rid of aktualizr_static_lib

Sounds fine to me, I think we originally only did that as a shortcut to allow garage-push to reuse some aktualizr code. @koshelev may suffer some but that seems to be the norm.

OYTIS:
Yes, I also think it's totally fine.

patrickvacek:
@OYTIS Do you want to merge this now and continue your work on a new PR, or keep this open for you to add to before merging?

OYTIS:
@patrickvacek I'd prefer to merge it, yes. Further rearrangement includes what's been made here.

CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ option(BUILD_OPCUA "Set to ON to compile with OPC-UA protocol support" OFF)
 option(BUILD_SYSTEMD "Set to ON to compile with systemd additional support" ON)
 option(INSTALL_LIB "Set to ON to install library and headers" OFF)
 
+set(logger_type "boost" CACHE STRING "")
```
lbonn: nitpick: should we use upper case names, like all the options that we have?

CMakeLists.txt
```diff
@@ -228,7 +231,8 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 ############### BUILD RULES
-include_directories(${PROJECT_SOURCE_DIR}/src)
+#include_directories(${PROJECT_SOURCE_DIR}/src)
+include_directories(${PROJECT_SOURCE_DIR}/src/libaktualizr)
```
patrickvacek: nitpick: why the commented out line?

OYTIS: Ooops. Yes, should just be deleted.

CMakeLists.txt
OYTIS: Yes, probably. Aesthetically I like lower case for individual module configuration better, but there is no real reason to introduce inconsistency.

src/CMakeLists.txt
```diff
@@ -1,78 +1,12 @@
...
+add_subdirectory("libaktualizr")
 
 set(OPCUABRIDGE_SOURCES opcuabridge/common.cc
                         opcuabridge/filelist.cc
                         opcuabridge/opcuabridgeclient.cc
                         opcuabridge/opcuabridgeserver.cc
                         opcuabridge/opcuabridgediscoveryclient.cc
                         opcuabridge/opcuabridgediscoveryserver.cc
-                        uptane/opcuasecondary.cc)
+                        libaktualizr/uptane/opcuasecondary.cc)
```
patriotyk: I think opcuabridge should link with libaktualizr and dont share sources.

OYTIS:
More changes, sorry. libaktualizr is still there, but is mostly just a collection of modules now and is ready to go soon.

src/CMakeLists.txt
OYTIS: Fixed

tests/CMakeLists.txt
```diff
@@ -127,7 +129,7 @@ add_aktualizr_test(NAME commands SOURCES commands_test.cc)
 
 add_aktualizr_test(NAME config SOURCES config_test.cc ARGS ${PROJECT_BINARY_DIR} PROJECT_WORKING_DIRECTORY)
 
-add_aktualizr_test(NAME crypto SOURCES crypto_test.cc PROJECT_WORKING_DIRECTORY)
+add_aktualizr_test(NAME crypto_test SOURCES crypto_test.cc PROJECT_WORKING_DIRECTORY)
```
patrickvacek: Minor complaint: this means the test will be named "test_crypto_test". A bit ugly, but not worth holding this up for.

OYTIS: Hm... Probably changed that fighting with some cmake issues and forgot to revert. Thanks!

OYTIS:
OK, so I'm merging it. If anybody needs help rebasing their work, please tell.

PR #719, opened 2018-04-24
Feat/jenkins test more

lbonn:
The goal is to have the same tests (or smarter but equivalent) on Jenkins than on Travis.

First step is a refactor of our tests scripts.


scripts/coverage.sh
```diff
@@ -1,32 +0,0 @@
...
-if [ "$JENKINS_RUN" = 1 ]; then
-    echo "Run as jenkins"
-    CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=3 make -j6 coverage || true
-else
-    CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL=3 make -j6 coverage
-fi
-)
-
-if [ -n "$TRAVIS_COMMIT" ]; then
-    bash <(curl -s https://codecov.io/bash -p src -s build-coverage)
```
lbonn: The options here were supplied to curl :) and our `codecov.yml` was never actually read by the tool (not on recent PRs at least)

PR #721, opened 2018-04-25
Fix link error due to taking a reference to kMaxSignatures

cajun-rat:
Change-Id: Ia75a5298c1e3390c7969ca8d7b4bd97a3ae285c9

### DIFF
--- a/src/uptane/root.cc
+++ b/src/uptane/root.cc
@@ -48,13 +48,19 @@ Root::Root(const std::string &repository, const Json::Value &json) : policy_(kCh
     // Threshold
     int64_t requiredThreshold = (*it)["threshold"].asInt64();
     if (requiredThreshold < kMinSignatures) {
+      // static_cast<int> is to stop << taking a reference to kMinSignatures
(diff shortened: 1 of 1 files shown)


src/uptane/root.cc
```diff
@@ -48,12 +48,16 @@ Root::Root(const std::string &repository, const Json::Value &json) : policy_(kCh
     // Threshold
     int64_t requiredThreshold = (*it)["threshold"].asInt64();
     if (requiredThreshold < kMinSignatures) {
+      // (int) is to stop << taking a reference to kMinSignatures
+      // http://www.stroustrup.com/bs_faq2.html#in-class
       LOG_DEBUG << "Failing with threshold for role " << role << " too small: " << requiredThreshold << " < "
-                << kMinSignatures;
+                << (int)kMinSignatures;
```
patriotyk: C style cast is not allowed now. clang-tidy will complain.

cajun-rat: Ah, thank you!

PR #722, opened 2018-04-26
Feat/pro 5308/config match cmdline

patrickvacek:


patrickvacek:
@lbonn I can't wait for the final transition to Jenkins. It's like 30-45 minutes faster at this point.

patrickvacek:
I'm stuck in the Travis timeout trap with test_uptane_vectors (and test_log_invalid!?). Now we are at the point of Jenkins being over a 1.5 hours faster (assuming it actually is satisfied after the latest kick).

cajun-rat:
Looks like it might be a real failure:

~~~
+echo 'Running docker container with aktualizr debian package inside.'
Running docker container with aktualizr debian package inside.
+docker run -v /persistent:/persistent -t advancedtelematic/aktualizr_ubuntu_test
/bin/sh: 1: /persistent/test_aktualizr_deb_and_update.sh: not found
~~~

patrickvacek:
> Looks like it might be a real failure:

That wasn't happening before... certainly interesting but hard to understand how my changes could have caused that.

lbonn:
It's highly probable that it came from my last merge PR. I'll look into it.

patrickvacek:
@lbonn I just had the same thought. Perhaps because of the argument fix? Anyway, I rebased my PR on the latest master (with your changes) and it appears to have made it past the `test_aktualizr_deb_and_update.sh: not found` error.

patrickvacek:
Finally passed CI... but I have a device failing to boot with `could not create directory:`. (Also, oe-selftest doesn't appear to be working on master.)

patrickvacek:
Finally fixed the device initialization problem and CI is happy too!

cajun-rat:
Looks good!

PR #724, opened 2018-04-26
Feat/pro 5333/jenkins test all

lbonn:


lbonn:
After a lot of trials and error, it should be ready.

Two issues (valgrind related) I couldn't fix:

- PRO-5338: leak with P11 tests => suppressed in valgrind for now
- leaks in `test_uptane`, only on debian testing, I'll open an issue for them => test result ignored on jenkins for now

cajun-rat:
Can we disable these on Travis somehow?

lbonn:
`test_uptane leaks` tracked in PRO-5358

lbonn:
Passing both CIs now, waiting for your reviews :)

cajun-rat:
Nice.  Good to see that some of those were real leaks too :)

PR #725, opened 2018-04-26
Read configs in systemd style

patriotyk:


cajun-rat:
This is looking good so far. Don't forget we'll need the same logic in the other command line tools that take a --config option too.

src/main.cc
```diff
@@ -137,23 +137,21 @@ int main(int argc, char *argv[]) {
...
-    Config config(sota_config_path, commandline_map);
+    Config config(commandline_map);
     if (geteuid() != 0) {
       LOG_WARNING << "\033[31mAktualizr is currently running as non-root and may not work as expected! Aktualizr "
                      "should be run as root for proper functionality.\033[0m\n";
     }
-    boost::filesystem::path saved_config_path = "/tmp/aktualizr_config_path";
-    Utils::writeFile(saved_config_path, boost::filesystem::absolute(sota_config_path).string());
+    // boost::filesystem::path saved_config_path = "/tmp/aktualizr_config_path";
+    // Utils::writeFile(saved_config_path, boost::filesystem::absolute(sota_config_path).string());
```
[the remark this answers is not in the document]
patrickvacek: I have mixed feelings about this. On one hand you are probably right, but on the other, if aktualizr is run with a nonstandard config, aktualizr-info won't have any knowledge of that and will then report incorrect information.

src/libaktualizr/config/config.cc
```diff
@@ -276,21 +278,34 @@ void Config::postUpdateValues() {
...
 
-void Config::updateFromDirs() {
-  for (const auto& dir : config_dirs_) {
-    for (const auto& config : Utils::glob((dir / "*.conf").string())) {
-      updateFromToml(config);
+void Config::updateFromDirs(const std::vector<boost::filesystem::path>& configs) {
+  std::map<std::string, boost::filesystem::path> configs_map;
+  for (const auto& config : configs) {
+    if (boost::filesystem::is_directory(config)) {
+      for (const auto& config_file : Utils::glob((config / "*.conf").string())) {
```
patrickvacek: Shouldn't this be finding files with a `.toml` extension? That's what all of our current config files use.

src/libaktualizr/config/config.cc
```diff
@@ -276,21 +278,34 @@ void Config::postUpdateValues() {
...
+        configs_map[config_file.filename().string()] = config_file;
```
patrickvacek: Isn't the opposite of the desired ordering? The first config dir specified on the command line should take first precedence over anything specified after that. This ordering (unless I'm missing something) is the reverse of that. I think the `for (const auto& config : configs)` needs to be reversed.

src/libaktualizr/config/config.h
```diff
@@ -103,11 +103,11 @@ struct DiscoveryConfig {
 class Config {
  public:
   Config();
-  Config(const boost::filesystem::path& filename, const boost::program_options::variables_map& cmd);
+  explicit Config(const boost::program_options::variables_map& cmd);
   explicit Config(const boost::filesystem::path& filename);
-  explicit Config(const std::vector<boost::filesystem::path>& config_dirs) {
-    config_dirs_ = config_dirs;
-    updateFromDirs();
+  explicit Config(std::vector<boost::filesystem::path> config_dirs) {
```
patrickvacek: Why remove the `const`? `Config::updateFromDirs()` accepts a const argument.

tests/config_test.cc
```diff
@@ -170,14 +176,18 @@ TEST(config, consistent_toml_nonempty) {
...
 }
 
 TEST(config, config_dirs_two_dirs) {
   std::vector<boost::filesystem::path> config_dirs{"tests/test_data/config_dirs/one_dir",
                                                    "tests/test_data/config_dirs/second_one_dir"};
   Config config(config_dirs);
-  EXPECT_EQ(config.storage.path, "/latest_path");
+  EXPECT_EQ(config.storage.path.string(), "path_z");
+  EXPECT_EQ(config.pacman.sysroot.string(), "sysroot_z");
+  EXPECT_NE(config.pacman.os, "os_a");
```
patrickvacek: This should be true, right? `pacman.os` doesn't get overwritten anywhere. The `one_dir` version of `a.conf` should get read and the `second_one_dir` skipped.

@cajun-rat Can you confirm that my understanding is correct?

tests/test_data/config_dirs/second_one_dir/y.conf
```diff
@@ -1,2 +1,2 @@
 [storage]
-path = "/latest_path"
\ No newline at end of file
+path = "latest_path"
```
patrickvacek: I recommend adding something in here so that we can test that `y.conf` is actually read and processed if `second_one_dir` is provided, even if it comes after `one_dir`, as long as `one_dir` doesn't have it's own `y.conf`.

tests/config_test.cc
patriotyk: No, `a.conf` from `second_one_dir` should be read and should hide `a.conf` from `one_dir`

src/libaktualizr/config/config.cc
patriotyk: It is not reversed. First specified directory will be processed first.

patrickvacek:
Looks like you addressed all of my concerns, and as far as the ordering goes, the commandline options are in the same order as the defaults, so I think that's fine. Sorry for being confusing on that point. So far so good!

I'm trying to run this on a device now, though, and running into two problems. First, if I just run aktualizr with no arguments (to use the defaults), I get an error that `/etc/sota` doesn't exist. I think it's probably a fatal error if *no* configuration file exists, but if there is at least one config file found, it's probably okay if one of the directories is empty, especially if it's one of the defaults. Second, it looks like subdirectories are not being found. Currently meta-updater puts extra config files in `/usr/lib/sota/conf.d/`, but the main one is still in `/usr/lib/sota`, and running with `aktualizr --config /usr/lib/sota` the stuff in `conf.d` is not picked up.

Also there's a merge conflict now, too, but hopefully an easy one.

patrickvacek:
It does appear to be common to have a xxx.conf file and then a xxx.conf.d directory alongside it. I haven't found any examples of subdirectories within a conf.d directory, though.

patriotyk:
If You run aktualizr without arguments You should get `/etc/sota/conf.d/ does not exist.` and I think it is ok because /etc/sota/conf.d/ does not exists. Also we don't read config files from sub directories we read only config files in provided directory.

patrickvacek:
After some discussion, it probably makes sense not to worry about subdirectories. In meta-updater, we will move the installation of `sota.toml` into `/usr/lib/sota/conf.d`, which should simplify some things. However, it still seems odd to have the defaults require that `/etc/sota/conf.d` exists. The solution there should perhaps be that we should throw a warning if a configuration directory doesn't exist, but we should only abort if one of the command line config entries is nonexistent. Presumably that is a separate check done just by iterating over the commandline input.

patrickvacek:
Something else is going wrong. I have a bunch of config fragments that I'm working with (https://github.com/advancedtelematic/meta-updater/pull/304). If I put them all in a directory and point aktualizr at them, at least one (the log-debug one) gets ignored. If I combine their contents into one file and point aktualizr at that, they all get processed as I would expect.

patrickvacek:
After further analysis, it looks like aktualizr is reading all four of my config files in `/usr/lib/sota/conf.d`, but only the last one read is actually used. The defaults from that one override everything that came before. Since the defaults are mostly what we want to use anyway, it took me a while to realize that's what is going on.

patrickvacek:
After yet further analysis, I appear to be wrong; aktualizr is using all of the config files, but confusingly, each time it reads one, anything set by a previous config file is called a "default" in the log. We should probably find a way to make it clearer what is going on in the log, but that can perhaps come later.

However, the reason for my prior confusion is that the loglevel set in my log-debug config file is getting ignored. I went through some hassle to get that working in https://github.com/advancedtelematic/aktualizr/pull/727, and I can't really tell why your changes would break that. Any thoughts?

patrickvacek:
@lbonn Can you help make sense out of why Jenkins claims one of the tests failed despite the lack of an apparent failure?

lbonn:
@patrickvacek timeout in `test_http_client`. I'll launch it again

patriotyk:
Just debugged the problem with logger. Currently it is not issue of this PR but I have added fix in separate commit. Please look on it.

patrickvacek:
> Just debugged the problem with logger.

Unfortunately the whole point of the `initialized` variable was so that if you put `--loglevel 0` on the commandline, it won't get overwritten by a config option for `loglevel`. This is tedious, but we need some way to make sure the commandline version overrules whatever is in the config, including while we are processing the config. I see now that the `initialized` check is what is causing my complaint with the logger when there are multiple configs, but we need some way to make sure the commandline version of `loglevel` (if set) is kept.

cajun-rat:
@patrickvacek Do you have any concerns with this as it stands today?  Will merging it break things?

patrickvacek:
> Do you have any concerns with this as it stands today? Will merging it break things?

Technically, yes, it will break the situation in which --loglevel 0 is set on the commandline in that the config processing will not be logged (but everything after that will be). This is a minor error and honestly probably not worth holding this up. The other issue is that the logging of processing multiple config files is confusing and should be revised/reorganized. That is also far from a blocker. At this point, I recommend merging and addressing these last two small issues in new tickets.

patrickvacek:
Follow-up tickets:
* https://advancedtelematic.atlassian.net/browse/PRO-5355 (--loglevel 0 on command line does not print out config processing)
* https://advancedtelematic.atlassian.net/browse/PRO-5356 (Logging of config processing is complicated and misleading)

PR #726, opened 2018-04-27
Only check the highest priority hash

OYTIS:
Should fix both [PRO-4767](https://advancedtelematic.atlassian.net/browse/PRO-4767) and [PRO-4768](https://advancedtelematic.atlassian.net/browse/PRO-4768).
@heartsucker please confirm

### DIFF
--- a/src/libaktualizr/crypto/crypto.h
+++ b/src/libaktualizr/crypto/crypto.h
@@ -57,11 +57,19 @@ struct PublicKey {
   int key_length{};
 };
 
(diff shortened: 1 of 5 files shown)


cajun-rat:
This looks OK, but we should look at refactoring DownloadMetaStruct.  It feels like it should take most of the arguments via the ctor:

    DownloadMetaStruct::DownloadMetaStruct(int64_t length, Hash hash, StorageTargetWHandle&)

PR #727, opened 2018-04-27
Bugfix/pro 5337/fix config logging

patrickvacek:
Now if loglevel is set on the commandline or in the config, it is taken into account while processing the rest of the config. This is a bit tedious, but I couldn't find a better way to make it work as expected.

### DIFF
--- a/src/libaktualizr/logging/boost_logging.cc
+++ b/src/libaktualizr/logging/boost_logging.cc
@@ -25,6 +25,11 @@ int loggerGetSeverity() { return static_cast<int>(gLoggingThreshold); }
 
 void LoggerConfig::updateFromPropertyTree(const boost::property_tree::ptree& pt) {
   CopyFromConfig(loglevel, "loglevel", boost::log::trivial::trace, pt);
+  // If not already set from the commandline, set the loglevel now so that it
+  // affects the rest of the config processing.
+  if (!initialized) {
+    setLogLevel();
+  }
 }
 
 void LoggerConfig::writeToStream(std::ostream& out_stream) const { writeOption(out_stream, loglevel, "loglevel"); }
@@ -39,6 +44,7 @@ void LoggerConfig::setLogLevel() {
     loglevel = boost::log::trivial::fatal;
   }
   logger_set_threshold(loglevel);
+  initialized = true;
 }
 
 // vim: set tabstop=2 shiftwidth=2 expandtab:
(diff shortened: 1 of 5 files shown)


src/libaktualizr/logging/boost_logging.cc
```diff
@@ -25,8 +25,26 @@ int loggerGetSeverity() { return static_cast<int>(gLoggingThreshold); }
 
 void LoggerConfig::updateFromPropertyTree(const boost::property_tree::ptree& pt) {
   CopyFromConfig(loglevel, "loglevel", boost::log::trivial::trace, pt);
+  // If not already set from the commandline, set the loglevel now so that it
+  // affects the rest of the config processing.
+  if (!initialized) {
+    setLogLevel();
```
lbonn: This bugs me a little bit that `setLogLevel` is a method of LoggerConfig and that the `updateFromPropertyTree` does more that it says here (and departs from the other `xxConfig` objects). What about moving `initialized` and `setLogLevel` out of the class?

Maybe, make it possible to call `logger_init()` several times, with an optional `LoggerConfig` argument and use it inside `Config::updateFromPropertyTree`

patrickvacek: I don't think having extra functionality in `LoggerConfig` is inherently bad, but I do see that setting the loglevel from inside `updateFromPropertyTree()` is a bit odd. What's the benefit of calling `logger_init()` several times, though? We already have `logger_set_threshold()` that can be called multiple times.

I suppose I could make setLogLevel() a non-member function (maybe even just overload `logger_set_threshold()` with an integer input). The `initialized` variable annoys me but I can't find a good way to get around it.

lbonn: `setLogLevel` should not be method for the config: it primary applies to the logging library itself and only incidently reads the config. The usual pattern is that the config objects are dumb and don't depend on the module they apply to. On the contrary, the module depends on the config.

`initialized` can be a static variable outside of the class. It's slightly ugly but logging facilities are the kind of canonical examples of acceptable uses of a singleton.

I thought about reusing `logger_init()` because it is a good name for a function that configures the library entirely from a configuration object.

patrickvacek:
@lbonn I think I've addressed your concerns, and I agree with your premises about the config. However, I didn't overload `logger_init()` since that is used to initialize globally. `logger_set_threshold()` can be called repeatedly without a problem, so that's what I overloaded to take a config object.

Also, it looks like the last Jenkins build failed, but I can't tell why. All of the tests passed. And it doesn't look like it has restarted after I pushed my latest commit. Any idea what's going on?

lbonn:
Thanks patrick, it looks good. Just the minor issue about static variables.

Jenkins failed because it couldn't kill one of the containers at the end of the test. That's the first time I see that, I'll try to look into it a bit...

patrickvacek:
@lbonn Good call, and fixed. Now how is it that Travis finished before Jenkins this time!?

patrickvacek:
I vote for a return to more parallelization in compiling and running the tests.

PR #728, opened 2018-04-30
Add more tests for garage push

patriotyk:


patrickvacek:
Looks about right to me, although this should be confirmed by @cajun-rat's code that should make these pass. I'm wondering why the `server-500_after_20` test isn't failing, though? It should be unrecoverable, but I don't see anything in the CMake code to state that it should fail.

patrickvacek:
@patriotyk Ah yes, now I see it. I had overlooked that the python script flips the garage-push return code. That makes sense to me now.

cajun-rat:
Included in #730

PR #730, opened 2018-05-02
PRO-5301 Make garage-push more robust

cajun-rat:


src/sota_tools/ostree_object.cc
```diff
@@ -215,38 +217,61 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
...
+      NotifyParents(pool);
+    } else if (rescode == 409) {
+      LOG_DEBUG << "OSTree upload reported a 409 Conflict, possibly due to concurrent uploads";
+      is_on_server_ = OBJECT_PRESENT;
+      last_operation_result_ = ServerResponse::kOk;
+      NotifyParents(pool);
     } else {
-      LOG_ERROR << "OSTree upload error: " << rescode;
-      is_on_server_ = OBJECT_BREAKS_SERVER;
+      LOG_WARNING << "OSTree upload reported an error code:" << rescode << " retrying...";
```
doanac: I found a few cases where having the body of the response would have saved me some time understanding what error took place on my treehub server. Any chance we could include that here?

cajun-rat: That should be pretty easy to add, let me have a go

cajun-rat: Does this work for you?  It will be hidden unless the -v option is used.  This PR also makes the curl verbose output only visible with verbosity up to -vv

tests/CMakeLists.txt
```diff
@@ -374,19 +374,8 @@ add_test(NAME validate-secondary-interface
...
-    target_link_libraries(t_sota_tools_auth_test sota_tools_static_lib aktualizr_static_lib ${TEST_LIBS})
-    target_include_directories(t_sota_tools_auth_test
-                               PUBLIC ${PROJECT_SOURCE_DIR}/src/sota_tools ${GLIB2_INCLUDE_DIRS})
-
-    add_aktualizr_test(NAME ostree_hash SOURCES sota_tools/ostree_hash_test.cc)
-    target_link_libraries(t_ostree_hash sota_tools_static_lib)
-    target_include_directories(t_ostree_hash
-                               PUBLIC ${PROJECT_SOURCE_DIR}/src/sota_tools ${GLIB2_INCLUDE_DIRS})
-
     # Setup coverage for sota_tools_static_lib
```
patrickvacek: Any reason not to put the entire set of sota_tools tests in the tests/sota_tools CMakeLists? I'm wouldn't hold up the PR for that, but just wondering if there was a reason.

cajun-rat: None really.  I just moved the easy/obvious ones to start with.  We really need to sort through the stuff in tests/ though

src/sota_tools/ostree_object.h
```diff
@@ -11,19 +13,20 @@
...
+enum PresenceOnServer { OBJECT_STATE_UNKNOWN, OBJECT_PRESENT, OBJECT_MISSING, OBJECT_INPROGRESS };
 
 enum CurrentOp {
   OSTREE_OBJECT_UPLOADING,
   OSTREE_OBJECT_PRESENCE_CHECK,
 };
 
+/**
+ * Broad categories for server response codes.
+ * There is no category for a permanent failure at the moment: we are able to
```
lbonn: we are or are not able to detect it?

cajun-rat: Oops. Fixed.

PR #731, opened 2018-05-02
Move global curl initialization into static data member

koshelev:


patrickvacek:
> I believe the old behavior was safe

Presumably @koshelev's problem is that `curl_global_init()` is not thread-safe. This never mattered for us, but it might for the load tests.

koshelev:
@patrickvacek is right, thread-safety is what I'm trying to solve here

cajun-rat:
I've just pushed a commit to fix the clang-format issue

lbonn:
It should solve the thread safety problem. Relevant doc for reference: http://en.cppreference.com/w/cpp/language/storage_duration. Ok if CI.

PR #735, opened 2018-05-03
Initial commit of load tests

koshelev:


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -187,30 +213,8 @@ void SotaUptaneClient::runForever(const std::shared_ptr<command::Channel> &comma
 
     try {
       if (command->variant == "GetUpdateRequests") {
+        getUpdateRequests();
```
OYTIS: I like the direction, but then we sould probably catch exceptions inside getUpdateRequests(). Otherwise a non-fatal Uptane::Exception will be able to stop the polling process.

koshelev: Oh, right. We should put `schedulePool` in front of it, as it was before...

OYTIS:
executor.cc seems to be empty.

src/load_tests/check.h
```diff
@@ -0,0 +1,11 @@
+#ifndef LT_CHECK_H
+#define LT_CHECK_H
+
+#include <string>
+#include <boost/filesystem.hpp>
+
+const std::string CHECK_FOR_UPDATES = std::string{"check_for_update"};
```
patriotyk: No need to make copies here.
Could be 
`const std::string CHECK_FOR_UPDATES{"check_for_update"};`

src/load_tests/check.h
```diff
@@ -0,0 +1,11 @@
+#ifndef LT_CHECK_H
```
patriotyk: SHould be `LT_CHECK_H_`

src/load_tests/context.h
```diff
@@ -0,0 +1,9 @@
+#ifndef LT_CONTEXT_H
```
patriotyk: Should be LT_CONTEXT_H_

CMakeLists.txt
```diff
@@ -29,12 +30,16 @@ unset(AKTUALIZR_CHECKED_SRCS CACHE)
...
 set(Boost_USE_STATIC_LIBS ON)
-set(BOOST_COMPONENTS log_setup log filesystem program_options system)
+set(BOOST_COMPONENTS log_setup log filesystem program_options system )
 
 if(BUILD_OPCUA)
     list(APPEND BOOST_COMPONENTS serialization iostreams)
 endif(BUILD_OPCUA)
 
+if(BUILD_LOAD_TESTS)
+    list(APPEND BOOST_COMPONENTS thread)
```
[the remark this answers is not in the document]
koshelev: I'm not using `boost::thread`, but `boost::latch` to coordinate worker threads

src/load_tests/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    ExternalProject_Get_Property(hdr_histogram source_dir)
+    set(HDRHISTOGRAM_SOURCE_DIR ${source_dir})
+    ExternalProject_Get_Property(hdr_histogram binary_dir)
+    set(HDRHISTOGRAM_BINARY_DIR ${binary_dir})
+
+    set(HDRHISTOGRAM_LIBS
+            ${HDRHISTOGRAM_BINARY_DIR}/src/${CMAKE_STATIC_LIBRARY_PREFIX}hdr_histogram_static${CMAKE_STATIC_LIBRARY_SUFFIX}
+            )
+
+    add_dependencies(ota-load-tests aktualizr hdr_histogram)
```
lbonn: should be `add_dependencies(ota-load-tests aktualizr hdr_histogram_static)` I think.

Also, there is a gcc warning when compiling hdr_histogram (`-Wmissing-prototypes`). Would be better if we could suppress it.

koshelev: I believe it is dependency on a project, and the project is added as 
```
ExternalProject_Add(
            hdr_histogram
```

cajun-rat: The other option would be to use git submodules to pull it in.  I'm kind-of allergic to fetching stuff from the internet as during the build process, and since the library is built with cmake anyway, it is easy to just reference it directly. See #737

lbonn:
I pushed two additional commits to (hopefully) fix CI

cajun-rat:
Can whoever merges this please rebase the branch first. No need to wait for a 2nd CI run, though.

PR #736, opened 2018-05-03
Move initialization out of Uptane::Repository

OYTIS:
This is the first step in the long refactoring process:

1. Strip Uptane::Repository from everything that is not related to Uptane protocol.
2. Split Uptane::Repository into ImageRepository and DirectorRepository, move coordinating logic into SotaUptaneClient.
3. Refacror ImageRepository and DirectorRepository in accordance with PRO-4769 and PRO-4770


OYTIS:
A message from clang-tidy:

```
/var/lib/jenkins/workspace/aktualizr_PR-736-U5AN25GZPE4ONABYGN63YRTP2AFGNFMF4CJP3W6BPZOKNISBLKDA@3/src/libaktualizr/primary/
initializer.cc:203:85: error: the parameter 'storage' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]

Initializer::Initializer(const ProvisionConfig& config, std::shared_ptr<INvStorage> storage, HttpInterface& http_client,
```

Passing shared_ptr as a reference doesn't seem idiomatic for me, but the check itself looks reasonable. Any idea about how to disable it for that line specifically?

OYTIS:
Found it!

`// NOLINT(performance-unnecessary-value-param)`

lbonn:
You can add a `// NOLINT` comment at the end of the line.

But the message can also be interpreted as "you don't need a smart pointer here". `Initializer` could take a `const *INvStorage`

OYTIS:
Sorry, don't quite get the connection here. The message seems to tell me that I don't change to contents of the pointer, while smart pointer is used to manage life time.
BTW, we do pass `shared_ptr` as `const&` in some other places in code, so maybe it's fine. Or maybe not, I beleive it will not work as a shared pointer if passed by reference.

lbonn:
Hum yes, I talked probably too fast.
In our use of Initializer, it doesn't look like it needs to hold on to a shared pointer since it's only a temporary objects (cf `sotauptaneclient.cc` and tests) but using a shared pointer for future-proofness makes sense.

However, I think I understand the spirit of the lint message. When the shared pointer is passed by value and then stored inside the object, its reference count is incremented twice (one for the parameter passing and one for the assignment) and decremented once at the exit of the constructor.

So the advice is to take a `const std::shared_ptr<...>&` in the constructor argument and keep a full smart pointer inside.

OYTIS:
> So the advice is to take a const std::shared_ptr<...>& in the constructor argument and keep a full smart pointer inside.

Got it, totally makes sense, thank you.

cajun-rat:
Rebased.

lbonn:
Needs changes in `uptane_ci_test` which is only built when `SOTA_PACKED_CREDENTIALS` is set

tests/initializer_implicit_test.cc
```diff
@@ -0,0 +1,153 @@
...
+  config.storage.tls_clientcert_path = "client.pem";
+  config.storage.tls_pkey_path = "pkey.pem";
+  config.postUpdateValues();
+
+  auto storage = INvStorage::newStorage(config.storage);
+  HttpFake http(temp_dir.Path());
+  KeyManager keys(storage, config.keymanagerConfig());
+  std::map<std::string, std::shared_ptr<Uptane::SecondaryInterface> > secondaries;
+
+  Initializer initializer(config.provision, storage, http, keys, secondaries);
```
lbonn: Minor: every time we give an empty map to `Initializer`, we can just write `Initializer initializer(config.provision, storage, http, keys, {});` without having to define an empty secondaries map.

src/libaktualizr/config/config.cc
```diff
@@ -275,6 +277,12 @@ void Config::postUpdateValues() {
     }
   }
 
+  if (!uptane.director_server.empty()) {
+    if (provision.ecu_registration_endpoint.empty()) {
+      provision.ecu_registration_endpoint = uptane.director_server + "/ecus";
+    }
```
patrickvacek: General question (not actually specific to this PR): do we expect to ever have some use custom server URLs that aren't just the server URL from the credentials with static names ('/directory', '/repo', '/ecus', etc.) appended?

src/libaktualizr/primary/initializer.cc
```diff
@@ -180,23 +167,22 @@ InitRetCode Repository::initEcuRegister() {
...
     primary_ecu["ecu_serial"] = ecu_serials[0].first;
-    primary_ecu["clientKey"]["keytype"] = config.uptane.getKeyTypeString();
+    primary_ecu["clientKey"]["keytype"] = keyTypeToString(keys.getUptaneKeyType());
     primary_ecu["clientKey"]["keyval"]["public"] = primary_public;
     all_ecus["ecus"].append(primary_ecu);
   }
 
-  std::vector<std::shared_ptr<Uptane::SecondaryInterface> >::const_iterator it;
-  for (it = secondary_info.begin(); it != secondary_info.end(); it++) {
+  for (auto it = secondary_info.begin(); it != secondary_info.end(); it++) {
```
patrickvacek: Not actually a big deal, but is it possible to use cbegin() and cend() here?

OYTIS:
@lbonn 

> Needs changes in uptane_ci_test which is only built when SOTA_PACKED_CREDENTIALS is set

Is it something that this PR has broken or a general problem?

src/libaktualizr/config/config.cc
OYTIS: I don't think so. We're relying on mutual TLS for authentication and it is much easier to do it if we always connect to the same server.
It might happen though that ota-commuinity-edition get real users and aktualizr will want to support multiple repo/director deployments with slightly different paths.

tests/initializer_implicit_test.cc
OYTIS: Thanks!

src/libaktualizr/primary/initializer.cc
OYTIS: Yes, makes sense, thank you.

lbonn:
@OYTIS yes, it's related to your PR: the test still calls `Uptane::Repository::initialize`. I've tweaked #724 so that we'll now build these tests on CI even if they are not run (yet).

OYTIS:
@lbonn Thank you, updated.

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -29,6 +30,19 @@ class SotaUptaneClient {
...
+  FRIEND_TEST(aktualizr_secondary_uptane, credentialsPassing);
+  FRIEND_TEST(Uptane, RandomSerial);
+  FRIEND_TEST(Uptane, ReloadSerial);
+  FRIEND_TEST(Uptane, LegacySerial);
+  FRIEND_TEST(Uptane, AssembleManifestGood);
+  FRIEND_TEST(Uptane, AssembleManifestBad);
+  FRIEND_TEST(Uptane, PutManifest);
+  FRIEND_TEST(Uptane, UptaneSecondaryAdd);
+  FRIEND_TEST(UptaneCI, OneCycleUpdate);
+  FRIEND_TEST(UptaneCI, CheckKeys);
```
patrickvacek: I looked into this before, glad that it is working for you here.

cajun-rat:
Have you tried this though oe-seftest?  If it passes that, then I'm happy

patrickvacek:
@heartsucker This is ready now, so I can try to answer any questions you have.

PR #739, opened 2018-05-06
Move unit tests to source dirs

patriotyk:


CMakeLists.txt
```diff
@@ -287,6 +287,28 @@ include(CPack)
 include(CPackComponent)
 
 
+set (LIBS
```
lbonn: This is cool and should help us solve all these linking issues we periodically get.

However, I'd suggest renaming it to a longer and less generic name. Maybe something like `AKTUALIZR_EXTERNAL_LIBS`?

patrickvacek: Do we want all of our executables to link with every library we use? Maybe it's harmless, especially if it is dynamically linked, but that seems like a lot of dependencies for some of the simpler executables.

src/cert_provider/CMakeLists.txt
```diff
@@ -23,4 +15,6 @@ add_dependencies(build_tests aktualizr_cert_provider)
 
 install(TARGETS aktualizr_cert_provider RUNTIME DESTINATION bin)
 
+add_test(NAME certprovider-ca COMMAND ${PROJECT_SOURCE_DIR}/tests/run_certprovider_test.sh ${PROJECT_BINARY_DIR}/src/cert_provider/aktualizr_cert_provider ${PROJECT_SOURCE_DIR}/tests/test_data/credentials.zip ${PROJECT_SOURCE_DIR}/tests/test_data/CAcert.pem ${PROJECT_SOURCE_DIR}/tests/test_data/CApkey.pem)
```
patrickvacek: I recommend splitting this over multiple lines. Not a big deal, though.

CMakeLists.txt
lbonn: You're right. But I feel that every time we tried to link something with `aktualizr_static_lib` without the full list, it led to some linking errors in the end (sooner or later).

If we stop linking our executables with the full static lib, we'll probably be able to select our external dynamic libraries.

src/libaktualizr/package_manager/CMakeLists.txt
```diff
@@ -13,13 +13,35 @@ aktualizr_source_file_checks(${SOURCES} ${HEADERS})
 # Debian backend
 if(BUILD_DEB)
     target_sources(package_manager PRIVATE debianmanager.cc)
+    add_executable(t_packagemanager_deb debianmanager_test.cc)
+    add_dependencies(build_tests t_packagemanager_deb)
+    target_link_libraries(t_packagemanager_deb aktualizr_static_lib ${TEST_LIBS})
+
+    add_test(NAME test_packagemanager_deb COMMAND ${PROJECT_SOURCE_DIR}/tests/run_debian_tests.sh ${CMAKE_CURRENT_BINARY_DIR}/t_packagemanager_deb
+    ${PROJECT_SOURCE_DIR}/tests/test_data/fake_dpkg)
```
patrickvacek: This too would probably be best split on separate lines, with the following ones indented in some fashion. NBD.

src/libaktualizr/uptane/CMakeLists.txt
```diff
@@ -31,4 +31,33 @@ if (BUILD_OPCUA)
...
+
+add_aktualizr_test(NAME uptane_network SOURCES uptane_network_test.cc PROJECT_WORKING_DIRECTORY)
+set_tests_properties(test_uptane_network PROPERTIES LABELS "crypto")
+
+add_aktualizr_test(NAME uptane_secondary SOURCES uptane_secondary_test.cc PROJECT_WORKING_DIRECTORY)
+
+add_aktualizr_test(NAME uptane_serial SOURCES uptane_serial_test.cc ARGS ${PROJECT_BINARY_DIR}
+                   PROJECT_WORKING_DIRECTORY)
+
+aktualizr_source_file_checks(${SOURCES} ${HEADERS} opcuasecondary.cc opcuasecondary.h uptane_ci_test.cc ${TEST_SOURCES})
```
patrickvacek: Would it make more sense to append opcuasecondary.cc, opcuasecondary.h, and uptane_ci_test.cc to SOURCES, HEADERS, and TEST_SOURCES, respectively? Seems easier to keep track of that way. The first two obviously need to be in the if(BUILD_OPCUA) guard.

patrickvacek:
On the whole I think this is about right. There are some minor formatting things, some merge conflicts, and a few errors in Travis/Jenkins that need resolving, but I don't see any major problems.

cmake-modules/AddAktualizrTest.cmake
```diff
@@ -0,0 +1,30 @@
...
+    if(AKTUALIZR_TEST_PROJECT_WORKING_DIRECTORY)
+        set(WD WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+    else()
+        set(WD )
+    endif()
+        # Running tests that are both instrumented by gcov and valgrind at
+        # the same time is very expensive.  Disable valgrind when running
+        # under gcov, or when the test is explicitly flagged that it fails
+        # under valgrind (these should be fixed)
+        if(AKTUALIZR_TEST_NO_VALGRIND OR BUILD_WITH_CODE_COVERAGE)
```
lbonn: We've made this possible (running valgrind and coverage at the same time) again in #724.
It's the current setup on Jenkins at the moment.

Can we keep master's version of `add_aktualizr_test`?

third_party/googletest/googletest/cmake/internal_utils.cmake
```diff
@@ -192,7 +192,7 @@ function(cxx_executable_with_flags name cxx_flags libs)
   endif()
   # To support mixing linking in static and dynamic libraries, link each
   # library in with an extra call to target_link_libraries.
-  foreach (lib "${libs}")
+  foreach (lib "${AKTUALIZR_EXTERNAL_LIBS}")
```
cajun-rat: What is the reason for changing 3rd party libraries?

cajun-rat:
@patriotyk Can you look at the merge conflict here, please?

tests/CMakeLists.txt
```diff
@@ -54,14 +39,9 @@ foreach(label ${TESTSUITE_EXCLUDE})
     set(CTEST_EXTRA_ARGS ${CTEST_EXTRA_ARGS} -LE ${label})
 endforeach()
 
-add_custom_target(check COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND} ${CTEST_EXTRA_ARGS}
+add_custom_target(check COMMAND CTEST_OUTPUT_ON_FAILURE=1 ${CMAKE_CTEST_COMMAND}
```
lbonn: Why did it go away?

patriotyk: I don't know, maybe after merge conflicts resolve

PR #740, opened 2018-05-07
Do not read loglevel from config if it present in cmdline

patriotyk:


patriotyk:
Looks like something has been changed in python pip repository because obtain-vector-tests began failing

PR #741, opened 2018-05-07
Fix garage-push test-verbose-logging.

patrickvacek:
Only runs when credentials are enabled.


cajun-rat:
Good catch.  Didn't know we had a test for that even :)

patrickvacek:
> Didn't know we had a test for that even

It's old, and it used to use dummy credentials, but at some point I changed it to use real credentials so that it actually tested something real. However, that means it only runs if you give it real credentials, which means it doesn't run in CI. Yet.

PR #742, opened 2018-05-07
Robustify tests/run_vector_tests.sh

lbonn:
Make sure that the tuf-test-vectors server is running before running our tests. Bail out after 10 seconds

I'm not entirely sure it was the cause of the recent CI instabilities (couldn't figure how to find the Travis logs again...) but it should not hurt anyway.

### DIFF
--- a/tests/get_open_port.py
+++ b/tests/get_open_port.py
@@ -6,7 +6,8 @@ import socket
 
 def main():
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    s.bind(('', 0))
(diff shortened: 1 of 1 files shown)


cajun-rat:
Hmm, looks like the CI error is real @lbonn ?

lbonn:
Yes, looks annoying. Working on it

lbonn:
The recent failures definitely look Travis related. Possible culprit here: https://www.traviscistatus.com/incidents/0ch1w22m3lwc, but somehow weird because it does not seem we're using their "`sudo: required` infrastructure"

PR #743, opened 2018-05-07
Add explicit option for SOTA_PACKED_CREDENTIALS

cajun-rat:
This makes it appear in the GUI, and should help avoid bugs like #741, where the tests are run without the option enabled

Change-Id: Id7b2aecd56a32004f1ed0141c27eabf97d44248e


cajun-rat:
The only thing that is will do is to make the option appear a bit more.  The GUI will have an option for it, and there will be an empty line in the CMakeCache.txt where it is easy to set a value.  It would be nice to create a warning when 'ninja qa' is run without SOTA_PACKED_CREDENTIALS set, but I don't really know how to report a warning as build time rather than configure time.

PR #747, opened 2018-05-08
Do not build load tests by default

lbonn:
They are currently not used by CI and add extra compilation time.


patrickvacek:
We should be building this in at least one CI run, right? The whole point of adding the code to the repo was to make sure we didn't accidentally break it again.

patrickvacek:
Looks better, thanks. We don't actually have any test cases to run the load tests, right? We should probably change that just to run a basic version of whatever @koshelev normally runs, just to make sure it works as expected. (That's presumably for another ticket/PR, though.)

patrickvacek:
Can we merge despite the remote Jenkins failure?

PR #750, opened 2018-05-08
Don't run clang-check on `make qa` and CI

lbonn:
I've become convinced that it's made redundant by clang-tidy.

Investigation: https://stackoverflow.com/questions/49943781/does-clang-tidy-make-clang-check-redundant/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,7 +160,6 @@ endif()
 
 if(CLANG_CHECK)
     add_custom_target(clang-check)
-    add_dependencies(qa clang-check)
     function(aktualizr_clang_check)
         file(RELATIVE_PATH SUBDIR ${CMAKE_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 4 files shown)


lbonn:
They both end up running [`Tool.run`](https://github.com/llvm-mirror/clang-tools-extra/blob/0ae5fe435fe0cc658ef0bca8769b022d963b4498/clang-tidy/ClangTidy.cpp#L566) (also [here](https://github.com/llvm-mirror/clang/blob/2e23251d3e91db37d19bf876f91b85059ead2491/tools/clang-check/ClangCheck.cpp#L185)) after having pushed the `AnalysisASTConsumer` inside somehow.

And we can see that clang-check only does that and exits afterwards (it's a very simple wrapper). So I conclude that `clang-check` only runs analyze and `clang-tidy` runs it, among other things. The only doubt I have is which options are selected when it's run from clang-check and if it could run more checks than clang-tidy. I didn't manage to find the source of truth for that but I would assume that all clang-analyzer checks are available to run from `clang-tidy`.

PR #751, opened 2018-05-09
Remove attempt to squash a memory leak

cajun-rat:
It caused a double-free bug


cajun-rat:
The lifecycle of `key` should be managed by the StructGuard on line 37: this line was to attempt a fix for a memory leak that was potentially caused by P11 doing wierd stuff. I think @lbonn has a better idea of the original intention here.

lbonn:
Yes, I just tracked the reference count and it didn't go to zero at the end on the version I tried (our coverage environment). `ENGINE_load_private_key` incremented the counter IIRC.

But I'm not a fan of crashing either so let's revisit that when somebody will try to squash the P11 leaks again...

lbonn:
By the way, sorry about the Jenkins notifications from auto-ota.jenkins.xxx, we can ignore them for now and this should be good to merge.

PR #752, opened 2018-05-09
Prevent aktualizr-info from probing legacy secondaries.

patrickvacek:
That should only be done by aktualizr (the primary). aktualizr-info only
needs the config in order to read data out of storage. In the long term,
it may be better to give it its own config (much like
aktualizr-secondary) that basically only includes storage and maybe
logging.

### DIFF
--- a/src/aktualizr_info/main.cc
+++ b/src/aktualizr_info/main.cc
@@ -36,7 +36,8 @@ int main(int argc, char **argv) {
       exit(EXIT_SUCCESS);
     }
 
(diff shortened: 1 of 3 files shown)


cajun-rat:
@patrickvacek thinks that there might be a better solution to this. Let's hold off on this for a bit

patrickvacek:
Closing as I believe https://github.com/advancedtelematic/aktualizr/pull/754 is the proper solution.

PR #753, opened 2018-05-09
Disallow pointer arithmetic with clang-tidy

cajun-rat:
Also, move WarningsAsErrors into .clang-tidy


CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ if(CLANG_TIDY)
             endif()
             string(REPLACE "/" "_" TARGETNAME "aktualizr_clang_tidy-${SUBDIR}-${FILE}")
             add_custom_target(${TARGETNAME}
-                COMMAND ${CLANG_TIDY} -quiet -header-filter=\${CMAKE_SOURCE_DIR}/src/.* --extra-arg-before=-Wno-unknown-warning-option -format-style=file -warnings-as-errors=* -p ${CMAKE_BINARY_DIR} ${FILE}
+                COMMAND ${CLANG_TIDY} -quiet -header-filter=${CMAKE_SOURCE_DIR}/src/.* --extra-arg-before=-Wno-unknown-warning-option -format-style=file -p ${CMAKE_BINARY_DIR} ${FILE}
```
lbonn: Oh this looks still broken. Might work on CI and not on usual dev envs (or the converse, I don't remember).
Can we try to restore `header-filter` from 098353eb0319e768ccf22e069e84e9f8191c05e3?

lbonn: Ok, I just tried that and it will need some adjustment because of our new directory structure.
I'll work on that and try to push a new corrected regex.

lbonn:
Just a heads-up: I ran `ninja clang-tidy` with `BUILD_OSTREE=on` and it found another instance of pointer arithmetic.

That's because it only runs on source files it knows how to build (which means, compiled with the current CMake settings). This is another argument for keeping compile options compatible with each others and be able to run tests with everything enabled.

In the meantime, I'll add the NOLINT annotation.

PR #754, opened 2018-05-09
Create new minimalist AktualizrInfoConfig.

patrickvacek:
This should keep aktualizr-info simpler and prevent some potential
conflicts between aktualizr and aktualizr-info usage of the config.

### DIFF
--- a/src/aktualizr_info/aktualizr_info_config.cc
+++ b/src/aktualizr_info/aktualizr_info_config.cc
@@ -1,0 +1,89 @@
+#include "aktualizr_info_config.h"
+
+#include "utilities/config_utils.h"
+
+AktualizrInfoConfig::AktualizrInfoConfig(const boost::program_options::variables_map& cmd) {
+  // Redundantly check and set the loglevel from the commandline prematurely so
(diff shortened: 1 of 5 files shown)


src/aktualizr_info/aktualizr_info_config.cc
```diff
@@ -0,0 +1,89 @@
...
+}
+
+AktualizrInfoConfig::AktualizrInfoConfig(const boost::filesystem::path& filename) {
+  updateFromToml(filename);
+  postUpdateValues();
+}
+
+void AktualizrInfoConfig::postUpdateValues() { logger_set_threshold(logger); }
+
+void AktualizrInfoConfig::updateFromDirs(const std::vector<boost::filesystem::path>& configs) {
```
lbonn: We might want to factor this kind of methods to some common utilities or (maybe!) base class before it goes out of hand.

Not critical though

patrickvacek: Agreed, there is probably an even smarter way of solving this. I'll ponder that, but for now, I think this is at least an improvement to what was there before.

PR #756, opened 2018-05-10
Parse configs for secondary in the same way as for primary

patriotyk:
Also share common code between secondary and primary configs

### DIFF
--- a/src/libaktualizr/logging/logging.h
+++ b/src/libaktualizr/logging/logging.h
@@ -4,8 +4,6 @@
 #include <boost/log/trivial.hpp>
 #include <boost/property_tree/ini_parser.hpp>
 
-#include "utilities/config_utils.h"
-
 /** Log an unrecoverable error */
 #define LOG_FATAL BOOST_LOG_TRIVIAL(fatal)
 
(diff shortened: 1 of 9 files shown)


tests/CMakeLists.txt
```diff
@@ -525,12 +525,6 @@ set_tests_properties(aktualizr_secondary_cmdline--something
                      PROPERTIES WILL_FAIL TRUE)
 
 # Check for expected output messages given specific misconfigurations
-# test the error message when running the executable with non-existent configuration file
-add_test(NAME aktualizr_secondary_no_config_check_message
-         COMMAND aktualizr-secondary -c non-existent-config.toml)
-set_tests_properties(aktualizr_secondary_no_config_check_message
-                     PROPERTIES PASS_REGULAR_EXPRESSION "aktualizr-secondary: configuration file .* not found. Exiting.")
```
cajun-rat: Isn't this test still valid? When the user explicitly specifies a config file that isn't present, we should tell them :)

src/libaktualizr/logging/logging.h
```diff
@@ -4,8 +4,6 @@
 #include <boost/log/trivial.hpp>
 #include <boost/property_tree/ini_parser.hpp>
 
-#include "utilities/config_utils.h"
```
cajun-rat: Good.

PR #758, opened 2018-05-10
Verify metapack for consistency before saving

patriotyk:


OYTIS:
To me it seems contrary to the direction of refactoring suggested by Eric, namely isolating storage and network communication from TUF/Uptane-related logic (see for example, https://advancedtelematic.atlassian.net/browse/PRO-4771).

OYTIS:
I mean the ticket itself is OK, since we don't yet have storage for non-verified metadata. But making the check inside the class abstracting the storage just feels wrong.

patriotyk:
But ticket requires MetaPack::IsConsistent() to be created. So should I address Your comment? Or do we need to changes the ticket?

OYTIS:
Having isConsistent() as a member of MetaPack is probably OK for now. But I strongly believe that the check (`if(pack.isConsistent()) { ...`) should be made by the caller of `storeMetadata`, not from inside the storage classes.

cajun-rat:
@OYTIS Are you happy with this PR now?

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -27,6 +27,9 @@ bool ManagedSecondary::putMetadata(const MetaPack &meta_pack) {
   // No verification is currently performed, we can add verification in future for testing purposes
   detected_attack = "";
   current_meta = meta_pack;
+  if (current_meta.isConsistent()) {
```
lbonn: should be `if (!current_meta.isConsistent())`?

patriotyk: Yes, thank You.

OYTIS:
Looks like the check is only done for secondaries. Is it intended?

cajun-rat:
Rebased

patriotyk:
@OYTIS It is because we already do more advanced checks for the primaries.

PR #759, opened 2018-05-11
Cleanup keymanager.cc of most of `BUILD_P11` ifdefs

lbonn:


tests/CMakeLists.txt
```diff
@@ -383,8 +383,8 @@ add_test(NAME validate-secondary-interface
     --firmware ${PROJECT_SOURCE_DIR}/tests/test_data/firmware.txt)
 
 
-add_subdirectory(sota_tools)
 if(BUILD_SOTA_TOOLS)
+    add_subdirectory(sota_tools)
```
patrickvacek: Don't you still want to add `sota_tools` even if not building them so that the source files get checked? Or do we not care about that for the tests?

lbonn: Yes, that's a fair point. We don't enforce source checks in `tests` right now but I'll move the condition inside the directory for consistency.

src/libaktualizr/crypto/keymanager.h
```diff
@@ -42,12 +42,10 @@ class KeyManager {
...
-#ifdef BUILD_P11
-  P11EngineGuard p11_;
-#endif
-  std::unique_ptr<TemporaryFile> tmp_pkey_file;
-  std::unique_ptr<TemporaryFile> tmp_cert_file;
-  std::unique_ptr<TemporaryFile> tmp_ca_file;
+  std::unique_ptr<P11EngineGuard> p11_{nullptr};
+  std::unique_ptr<TemporaryFile> tmp_pkey_file{nullptr};
+  std::unique_ptr<TemporaryFile> tmp_cert_file{nullptr};
+  std::unique_ptr<TemporaryFile> tmp_ca_file{nullptr};
```
cajun-rat: Is the explicit initialization to `{nullptr}` needed?  Or is this a stylistic thing?

lbonn: They should not be needed but I added them to fix an "uninitialized attribute" warning and don't remember in which configuration it showed up. I'll try removing these and see what happens

lbonn:
@patrickvacek Actually, I believe it is not a special feature of constexpr. defines are commonly used but it's a matter of style I guess :)

PR #763, opened 2018-05-14
Basic explanation of systemd-style config support.

patrickvacek:
I just realized this was never updated. This is a pretty basic explanation; is there a definitive source that we should link to for systemd-style configuration?


patrickvacek:
@tkfu:
1. There are two steps. Searching by file name overrules per unique file name, then processing the options overrules per option.
2. Yes, if `--config` is specified at least once, the defaults are not used.

I made some meaning tweaks here: https://github.com/advancedtelematic/aktualizr/pull/766

PR #765, opened 2018-05-14
Fix OPC-UA build

cajun-rat:
Suppress a few pointer arithmetic warnings
Fix location of open62541 library

Change-Id: I6991d5b55e0f3bb1610a9c9dbcffea5582b88d29


cajun-rat:
I don't think clang-tidy is run automatically with either `BUILD_OPCUA` or `BUILD_ISOTP` turned on.

PR #770, opened 2018-05-16
Fix/test detection

lbonn:
Changes needed after #739


patrickvacek:
When you run with this locally, do the tests pass for you? I'm getting 6 failures, most of which I suspect are because of execution path assumptions that are no longer valid.

patrickvacek:
Never mind, after blasting some stale files it seems to be working fine except for test_uptane_vectors: `ImportError: cannot import name 'blockalgo'`. Probably just a version issue that I need to figure out, and if it passes CI I have no objection.

PR #760, opened 2018-05-13
Embed migration scripts to sources

patriotyk:

PR #772, opened 2018-05-17
Make check-formatting.sh fail when file could not be found

lbonn:
Also simplify the script in general and does not support running it on
several files at once since it seems that we don't use this mode anymore

Bug found in #760


lbonn:
Also added an early CMake check, no reason it cannot be caught at configure time

PR #774, opened 2018-05-18
Fix opcuabridge test

cajun-rat:
This was failing because run-valgrind has moved

Change-Id: Ie001af01ad74cc1ea3a3cfdf4dae992a275cb3ef


tests/run_opcuabridge_ostree_repo_sync_test.sh
```diff
@@ -25,11 +25,9 @@ fi
 sleep 3
 trap 'kill %1' EXIT
 
-if [ "$3" == "valgrind" ]; then
-  $BIN_DIR/run-valgrind $BIN_DIR/opcuabridge-ostree-repo-sync-client $TEST_SRC_REPO $PORT
-else
-  $BIN_DIR/opcuabridge-ostree-repo-sync-client $TEST_SRC_REPO $PORT
-fi
+# $3 can be set to 'valgrind' (or the run-valgrind wrapper)
```
lbonn: Should we also modify line 19 to look like this? It hard-codes arguments to valgrind.

patrickvacek: That makes sense to me.

lbonn:
I've pushed another commit that's significantly more complicated, please review again :).

Actually `$RUN_VALGRIND` was not working in this case since it was set in the `tests/` directory. So I've moved it to the top directory so that it's possible to write integration tests like this one outside of `tests/`.

Also moved the script to the same directory. It was a bit weird to declare it in `src/libaktualizr/opcuabridge` while the .sh was sitting in `tests/`.

And finally, make it follow the current practice of running every test under valgrind if possible.

cajun-rat:
Looks good.

PR #775, opened 2018-05-18
Check targets expiry date in garage-check

patriotyk:


patrickvacek:
Is there a way we can build a test case to verify that this is checking correctly? Logically it seems fine, though.

patriotyk:
Yes we need to write the tests, but to do that we need to generate a repo and serve it. Phil wrote me to create such tools. I will do that, and after that will add tests

patriotyk:
Depends on #775

patriotyk:
#775

src/sota_tools/garage_check.cc
```diff
@@ -160,6 +161,22 @@ int main(int argc, char **argv) {
...
 
   Json::Value targets_json = Utils::parseJSON(targets_str);
+  std::string expiry_time_str = targets_json["signed"]["expires"].asString();
+  if (expiry_time_str.length() != 20 || expiry_time_str[19] != 'Z') {
+    LOG_FATAL << "'expires' field has wrong format";
+    return EXIT_FAILURE;
+  }
+
+  struct tm expiry_time {};
+  strptime(expiry_time_str.c_str(), "%Y-%m-%dT%H:%M:%SZ", &expiry_time);
```
cajun-rat: We already have code to parse timestamps in `[Uptane::TimeStamp](https://github.com/advancedtelematic/aktualizr/blob/b2705bd25d3ccc9eda10421abd6a0a45125adf7b/src/libaktualizr/uptane/tuf.h#L63)`, and we should use it here--this will eliminate duplicating the implementation, and will also mean that we are checking that the timestamp is valid using exactly the same code that will we used later on when Aktualizr validates it.

patriotyk:
We use it now, I have updated this PR

tests/run_expired_test.sh
```diff
@@ -0,0 +1,35 @@
...
+zip -r $TEMP_DIR/credentials.zip  treehub.json tufrepo.url
+popd
+./tests/fake_http_server/fake_api_server.py $PORT $TEMP_DIR &
+sleep 2
+
+$2/src/sota_tools/garage-check -j $TEMP_DIR/credentials.zip -r 714581b2ffbbf7a750cb0a210fa7d74fd9128bd627cd4913e365d5bf2f66eba9
+RET=$?
+
+kill %1
+rm -rf $TEMP_DIR/
```
lbonn: For robustness and better error diagnostic on CI:

- running with `set -euo pipefail`  (http://redsymbol.net/articles/unofficial-bash-strict-mode/) and `set -x` for better logs
- using traps for `kill %1` and `rm -rf $TEMP_DIR` (http://redsymbol.net/articles/bash-exit-traps/)
- replacing `rm -rf $TEMP_DIR/` with `rm -rf $TEMP_DIR`. If `TEMP_DIR` is empty (by mistake), it would run `rm -rf /` ([shellcheck](https://www.shellcheck.net/) warned me about this!)

PR #777, opened 2018-05-22
Separate fetching of metadata from UPTANE-related checks

OYTIS:
Moving stuff around to separate storage, fetching and verification of metadata. Should close [PRO-4769](https://advancedtelematic.atlassian.net/browse/PRO-4769), [PRO-4770](https://advancedtelematic.atlassian.net/browse/PRO-4770), [PRO-4771](https://advancedtelematic.atlassian.net/browse/PRO-4771), [PRO-4772](https://advancedtelematic.atlassian.net/browse/PRO-4772), [PRO-4773](https://advancedtelematic.atlassian.net/browse/PRO-4773), and [PRO-4774](https://advancedtelematic.atlassian.net/browse/PRO-4774) except that director and image repos are still managed by one class, namely Uptane::Repository. I haven't found a good reason to separate them on the current development stage, but I can be wrong of course.
Updating root metadata is still hacky, waiting for the backend to implement root.json versioning in both director and image meta.


OYTIS:
Why clang-check sometimes wants me to pass shared pointers by value and sometimes as `const shared_ptr<T>`. I don't get the logic behind it.

patrickvacek:
> Why clang-check sometimes wants me to pass shared pointers by value and sometimes as const shared_ptr<T>&?

Just a guess, but does it matter if it's a constructor vs a function call, or if it's const or not?

OYTIS:
Seems to make no difference. Uptane::Initializer has std::shared_ptr<INvStorage> as its member and gets a `const shared_ptr<INvStorage>&` as a parameter to the constructor. Uptane::Fetcher does exactly the same and is shouted at.

OYTIS:
No, there is a difference in that whether the constructor is inline or defined in a separate file.

patrickvacek:
> No, there is a difference in that whether the constructor is inline or defined in a separate file.

Are we running different checks on source and header files?

OYTIS:
We don't, but they might be treated differently internally, at least decision on what headers to check seems to be gone inside clang_tidy:

```
if(${FILE} MATCHES "\\.h")
    # do not run clang-tidy directly on header files, since it
    # ignores them. headers will be checked if they are included by
    # a checked source file
    continue()
endif()
```

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -387,89 +380,161 @@ bool SQLStorage::loadMetadata(Uptane::MetaPack* metadata) {
...
+void SQLStorage::clearMetadataCommon(const std::string& tablename) {
   SQLite3Guard db(config_.sqldb_path.c_str());
 
   if (db.get_rc() != SQLITE_OK) {
     LOG_ERROR << "Can't open database: " << db.errmsg();
     return;
   }
 
-  if (db.exec("DELETE FROM meta;", nullptr, nullptr) != SQLITE_OK) {
+  if (db.exec("DELETE FROM " + tablename + ";", nullptr, nullptr) != SQLITE_OK) {
```
lbonn: Should use `db.prepareStatement`.

Also, providing an exec method which takes a `std::string` (https://github.com/advancedtelematic/aktualizr/pull/777/files#diff-8323b86c37db8f6551a01f8aec6344d6R119) encourages this kind of string concatenation business, so I'd also recommend not adding it.

lbonn: Hum sorry, my bad, this might be impossible. I'll take another look.

OYTIS: Yes, I believe it's impossible unfortunately. It is marginally OK in my opinion, because the function only gets called with hardcoded constants.

lbonn: yes, not the end of the world

OYTIS:
I wonder why `noostree` test doesn't fail for (most of) other PRs.

config/sql/migration/migrate.5.sql
```diff
@@ -0,0 +1,18 @@
...
+                  director_targets BLOB NOT NULL,
+                  image_root BLOB NOT NULL,
+                  image_targets BLOB NOT NULL,
+                  image_timestamp BLOB NOT NULL,
+                  image_snapshot BLOB NOT NULL);
+
+-- Nothing was stored it root_meta so far, so just drop
+DROP TABLE root_meta;
+CREATE TABLE root_meta(root BLOB NOT NULL, director INTEGER NOT NULL CHECK (director IN (0,1)), version INTEGER NOT NULL);
+CREATE TABLE root_rawmeta(root BLOB NOT NULL, director INTEGER NOT NULL CHECK (director IN (0,1)), version INTEGER NOT NULL);
```
cajun-rat: I'm always nervous about booleans, they seem to have a habit of collecting more than two values.  What about `repository CHAR NOT NULL CHECK (repository IN ('d','i'))`?

Eric was talking about having a separate TUF repo to handle key rotation, so we might have need for a 3rd type in the future.

cajun-rat:
I get a build failure with BUILD_OPCUA on

config/sql/migration/migrate.5.sql
OYTIS: Makes sense, but INTEGER is also extensible that way. If we want it to be self-descriptive, we might use TEXT field instead.

OYTIS:
@cajun-rat Thanks for finding that, I'll check.

config/sql/migration/migrate.5.sql
OYTIS: But I'd rather leave everything as is and if we want it to take more than two values eventually, then just add comments to schema.sql about which value means what and rename the field.

cajun-rat:
np.  I've created #787 to add the check to ci

patrickvacek:
> I believe you've fixed the other case when config file/directory is explicitly specified from command line

I tried to fix everything -- fail if commandline-specified files were not found, and warn if defaults were not.

OYTIS:
OK, I see, now there is just one `updateFromDirs`. I'll delete 9b1de7608cb58c9d9cfe1adb2f6f4fba8c48ba75, but 709e9dd187c12522de1d03028eb54ed4a0314555 still makes sense in my opinion.

patrickvacek:
> I'll delete 9b1de76, but 709e9dd still makes sense in my opinion.

Seems reasonable. It might make meta-updater complain, but that's easily resolved.

patrickvacek:
This seems fine to me, although I haven't studied it in depth. Should we assign to @heartsucker at this point?

OYTIS:
@patrickvacek Yes, I think a preliminary review might be useful at this point.

patrickvacek:
> Won't 709e9dd break running make install as a regular user?

Is that not already impossible? We already install plenty of things. Or do you mean just because of the `/etc` bit?

lbonn:
@patrickvacek yes, I was talking about the `/etc` which is absolute and the only path we install outside of prefix: 

```
Install the project...
-- Install configuration: "Valgrind"
-- Installing: /tmp/aktualizr/usr/local/lib/systemd/system/aktualizr.service
-- Installing: /tmp/aktualizr/usr/local/lib/sota/sota.toml
-- Installing: /tmp/aktualizr/usr/local/lib/sota/demo_secondary.json
-- Installing: /tmp/aktualizr/etc/sota/conf.d
-- Installing: /tmp/aktualizr/usr/local/lib/sota/conf.d
-- Installing: /tmp/aktualizr/usr/local/bin/garage-push
-- Installing: /tmp/aktualizr/usr/local/bin/garage-check
-- Installing: /tmp/aktualizr/usr/local/bin/garage-deploy
-- Installing: /tmp/aktualizr/usr/local/bin/garage-sign
-- Up-to-date: /tmp/aktualizr/usr/local/./lib
-- Installing: /tmp/aktualizr/usr/local/./lib/org.scalaj.scalaj-http_2.12-2.3.0.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.slf4j.slf4j-api-1.7.25.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.typelevel.cats-core_2.12-1.0.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.typelevel.macro-compat_2.12-1.1.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/ch.qos.logback.logback-core-1.2.3.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/com.github.scopt.scopt_2.12-3.7.0.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.typelevel.machinist_2.12-0.6.2.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/com.advancedtelematic.libtuf-0.4.0-18-g41dfc1a.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.spire-math.jawn-parser_2.12-0.11.0.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/ch.qos.logback.logback-classic-1.2.3.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.typelevel.cats-macros_2.12-1.0.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.bouncycastle.bcprov-jdk15on-1.59.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.typelevel.cats-kernel_2.12-1.0.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.bouncycastle.bcpkix-jdk15on-1.59.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/com.chuusai.shapeless_2.12-2.3.3.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/io.circe.circe-java8_2.12-0.9.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/io.circe.circe-numbers_2.12-0.9.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/eu.timepit.refined_2.12-0.8.7.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/io.circe.circe-parser_2.12-0.9.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.scala-lang.scala-library-2.12.4.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/com.advancedtelematic.libats_2.12-0.1.1-20-g97c26cb.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.scala-lang.scala-reflect-2.12.4.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.scala-lang.scala-compiler-2.12.4.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.scala-lang.modules.scala-xml_2.12-1.0.6.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/io.circe.circe-jawn_2.12-0.9.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/io.circe.circe-generic_2.12-0.9.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/org.scala-lang.modules.scala-async_2.12-0.9.6.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/io.circe.circe-core_2.12-0.9.1.jar
-- Installing: /tmp/aktualizr/usr/local/./lib/com.advancedtelematic.cli-0.4.0-18-g41dfc1a.jar
-- Installing: /tmp/aktualizr/usr/local/bin/aktualizr
-- Installing: /tmp/aktualizr/usr/local/bin/aktualizr-secondary
-- Installing: /tmp/aktualizr/usr/local/bin/aktualizr-info
-- Installing: /tmp/aktualizr/usr/local/bin/aktualizr-check-discovery
-- Installing: /tmp/aktualizr/usr/local/bin/aktualizr_cert_provider
-- Installing: /tmp/aktualizr/usr/local/bin/aktualizr_implicit_writer
-- Installing: /tmp/aktualizr/usr/local/bin/example-interface
```

But `make install` with `DESTDIR` still works so I guess this is a non-issue.

patrickvacek:
> yes, I was talking about the /etc which is absolute and the only path we install outside of prefix:

Yes, I agree, it looks like we should remove the leading /. I think that should work correctly for all cases then.

config/sql/migration/migrate.5.sql
heartsucker: @cajun-rat That third repo is literally "just tuf" but it serves CA cert bundles. It is completely unrelated to the main Uptane flow.

OYTIS:
@patrickvacek Removing the leading / will make it install to `/usr/etc/` which makes little sense. I can just remove it, but I'd prefer to have @heartsucker's response first to avoid meaningless CI runs.

patrickvacek:
@OYTIS ah crap, I was wondering about that. It seems like there has to be a way to get that to work relative to an arbitrary root with sticking a /usr/ in between.

src/libaktualizr/storage/fsstorage.cc
```diff
@@ -149,102 +149,98 @@ bool FSStorage::loadTlsCert(std::string* cert) { return loadTlsCommon(cert, conf
...
+bool FSStorage::loadMetadataCommon(Uptane::RawMetaPack* metadata, const std::string& suffix) {
   boost::filesystem::path image_path = Utils::absolutePath(config_.path, config_.uptane_metadata_path) / "repo";
   boost::filesystem::path director_path = Utils::absolutePath(config_.path, config_.uptane_metadata_path) / "director";
 
-  if (!boost::filesystem::exists(director_path / "root.json") ||
-      !boost::filesystem::exists(director_path / "targets.json") ||
-      !boost::filesystem::exists(image_path / "root.json") || !boost::filesystem::exists(image_path / "targets.json") ||
-      !boost::filesystem::exists(image_path / "timestamp.json") ||
-      !boost::filesystem::exists(image_path / "snapshot.json")) {
+  if (!boost::filesystem::exists(director_path / ("root.json" + suffix)) ||
```
heartsucker: Why are we aborting if not eveything is present? The only metadata we need to actually have are the two `root.json`s. Everything else can be safely fetched from those alone.

src/libaktualizr/storage/fsstorage.cc
```diff
@@ -149,102 +149,98 @@ bool FSStorage::loadTlsCert(std::string* cert) { return loadTlsCommon(cert, conf
...
 
-  json = Utils::parseJSONFile(director_path / "targets.json");
-  if (json != Json::nullValue) {
-    Json::Value fixed_json = json;
-    if (!json.isMember("signed")) {
-      fixed_json["signed"] = json;
-    }
-    metadata->director_targets = Uptane::Targets(fixed_json);
-  }
+void FSStorage::clearMetadataCommon(const std::string& suffix) {
```
heartsucker: Again, what is the `suffix`?

src/libaktualizr/storage/fsstorage.cc
```diff
@@ -149,102 +149,98 @@ bool FSStorage::loadTlsCert(std::string* cert) { return loadTlsCommon(cert, conf
...
 
-  json = Utils::parseJSONFile(image_path / "snapshot.json");
-  if (json != Json::nullValue) {
-    Json::Value fixed_json = json;
-    if (!json.isMember("signed")) {
-      fixed_json["signed"] = json;
-    }
-    metadata->image_snapshot = Uptane::Snapshot(fixed_json);
-  }
+void FSStorage::storeUncheckedMetadata(const Uptane::RawMetaPack& metadata) { storeMetadataCommon(metadata, ".raw"); }
```
heartsucker: Ok, I think I get what the suffix is. Metadata should only be stored in the exact form in which it was read from the wire. All processing has the possibility to break signatures (most notably, parsing it to a `struct` that doesn't capture a field the backend sends).

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -352,22 +352,15 @@ void SQLStorage::storeMetadata(const Uptane::MetaPack& metadata) {
...
```
heartsucker: The rule for APIs is that it should as hard as possible to abuse them. This should either use a tool to do SQL binding of params, or it should have a manual check that the table name matches `[a-zA-Z_][0-9a-zA-Z_]*` just to be super sure nothing bad happens.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -352,22 +352,15 @@ void SQLStorage::storeMetadata(const Uptane::MetaPack& metadata) {
...
-  jsons.push_back(Json::FastWriter().write(metadata.director_targets.original()));
-  jsons.push_back(Json::FastWriter().write(metadata.image_root.original()));
-  jsons.push_back(Json::FastWriter().write(metadata.image_targets.original()));
-  jsons.push_back(Json::FastWriter().write(metadata.image_timestamp.original()));
-  jsons.push_back(Json::FastWriter().write(metadata.image_snapshot.original()));
-
   auto statement = db.prepareStatement<SQLBlob, SQLBlob, SQLBlob, SQLBlob, SQLBlob, SQLBlob>(
-      "INSERT INTO meta VALUES (?,?,?,?,?,?);", SQLBlob(jsons[0]), SQLBlob(jsons[1]), SQLBlob(jsons[2]),
-      SQLBlob(jsons[3]), SQLBlob(jsons[4]), SQLBlob(jsons[5]));
+      "INSERT INTO " + tablename + " VALUES (?,?,?,?,?,?);", SQLBlob(metadata.director_root),
```
heartsucker: Again here with the string concatenation for SQL.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -387,89 +380,162 @@ bool SQLStorage::loadMetadata(Uptane::MetaPack* metadata) {
 
   request = kSqlGetTable;
   req_response_table.clear();
-  if (db.exec("SELECT * FROM meta;", callback, this) != SQLITE_OK) {
+  if (db.exec("SELECT * FROM " + tablename + ";", callback, this) != SQLITE_OK) {
```
heartsucker: SQL string concatenation.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -387,89 +380,162 @@ bool SQLStorage::loadMetadata(Uptane::MetaPack* metadata) {
...
+    LOG_ERROR << "Can't open database: " << db.errmsg();
+    return;
+  }
+
+  if (!db.beginTransaction()) {
+    return;
+  }
+
+  auto delete_statement = db.prepareStatement<int, int>(
+      "DELETE FROM " + tablename + " WHERE (director=? AND version=?);", static_cast<int>(director), version.version());
```
heartsucker: String concatenation for `tablename`.

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -0,0 +1,114 @@
...
+    root = fetchRole((director) ? config.uptane.director_server : config.uptane.repo_server, Role::Root(), version);
+  } catch (MissingRepo()) {
+    LOG_WARNING << "Can't connect to the repo";
+    return false;
+  }
+  storage->storeUncheckedRoot(director, root, version);
+  return true;
+}
+
+bool Fetcher::fetchMeta() {
```
heartsucker: This paradigm is a little odd. It's legal to fetch just one or several things. This also won't really work when the root is `N > 1` versions behind what is on the server. It would need to fetch several roots before it could fetch the others. Or, if you do the update and have no new `timestamp` metadata, you can return early because that means the rest of the metadata is unchanged.

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -0,0 +1,114 @@
...
+          storage->allocateTargetFile(false, target.filename(), target.length());
+      ds.fhandle = fhandle.get();
+      ds.downloaded_length = 0;
+      ds.expected_length = target.length();
+
+      if (target.hashes().empty()) {
+        throw Exception("image", "No hash defined for the target");
+      }
+
+      ds.hash_type = target.hashes()[0].type();
```
heartsucker: Since this is checking the 0-th index, we would have to make sure that the `target` always returns the hashes in the order of strongest to weakest. This also appears it would break if

1. Server has hashes A, B, C (with C being the "strongest")
2. Client knows of A and B

Wouldn't this raise some sort of exception? Because if so, it shouldn't.

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -23,11 +23,23 @@ ManagedSecondary::ManagedSecondary(const SecondaryConfig &sconfig_in) : Secondar
...
+
+bool ManagedSecondary::putMetadata(const RawMetaPack &meta_pack) {
   // No verification is currently performed, we can add verification in future for testing purposes
   detected_attack = "";
-  current_meta = meta_pack;
-  storeMetadata(current_meta);
+
+  current_raw_meta = meta_pack;
+  rawToMeta();
+  storeMetadata(current_raw_meta);
```
heartsucker: We don't need to be able to store a full metapack all at once. Each piece of metadata should be loaded one at at time so it's easier to pick up where an error happened. Imagine the case where we get a metapack with `snapshot` version `X`. This points to some `x_1, x_2, x_3, ...` versions of delegated metadata. If we can't load the metapack, but we get a snapshot later with version `X+1`, but versions of the delegated metadata don't match, this should be an error, but our case won't catch it (I think).

src/libaktualizr/uptane/uptanerepository.cc
```diff
@@ -117,113 +100,93 @@ bool Repository::verifyMetaTargets(const Uptane::Targets& director_targets, cons
...
-  if (new_meta.director_root.version() > meta_.director_root.version() ||
-      new_meta.image_root.version() > meta_.image_root.version() ||
-      new_meta.director_targets.version() > meta_.director_targets.version() ||
-      new_meta.image_timestamp.version() > meta_.image_timestamp.version()) {
-    if (verifyMetaTargets(new_meta.director_targets, new_meta.image_targets)) {
-      storage->storeMetadata(new_meta);
-      meta_ = new_meta;
-    } else {
-      LOG_WARNING << "Metadata image/directory repo consistency check failed.";
+  try {
```
heartsucker: Loading of each piece of metadata should be its own function. Also, this fails to check things like

- the `targets` version in `snapshot` matches the actual version in the `targets` metadata
- the `snapshot` version in `timestamp` matches the actual version in the `snapshot` metadata
- and others

src/libaktualizr/uptane/uptanerepository.cc
```diff
@@ -117,113 +100,93 @@ bool Repository::verifyMetaTargets(const Uptane::Targets& director_targets, cons
...
+        Uptane::TimestampMeta(now, "repo", Utils::parseJSON(meta_raw.image_timestamp), new_meta.image_root);
+    new_meta.image_snapshot =
+        Uptane::Snapshot(now, "repo", Utils::parseJSON(meta_raw.image_snapshot), new_meta.image_root);
+    new_meta.image_targets =
+        Uptane::Targets(now, "repo", Utils::parseJSON(meta_raw.image_targets), new_meta.image_root);
+
+    new_meta.director_targets =
+        Uptane::Targets(now, "director", Utils::parseJSON(meta_raw.director_targets), new_meta.director_root);
+
+    if (new_meta.director_root.version() < meta_.director_root.version() ||
```
heartsucker: It is legal to load a root but not a timestamp. Also, when a root is rotated, the protocol say to blow away all stored metadata, so the root **MUST** always go up, but the others **MAY** go down if and only if the root was rotated.

config/sql/migration/migrate.5.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN TRANSACTION;
+
+CREATE TABLE rawmeta(director_root BLOB NOT NULL,
```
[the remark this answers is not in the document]
OYTIS: The code is still a bit missing root and the rest of metadata, eventually they will be treated separately. Also it's just a storage, so I don't think we should have any info about the metadata internals (e.g. version) here. For root metadata we will have version information in storage of course.

src/libaktualizr/storage/fsstorage.cc
```diff
@@ -149,102 +149,98 @@ bool FSStorage::loadTlsCert(std::string* cert) { return loadTlsCommon(cert, conf
 
 bool FSStorage::loadTlsPkey(std::string* pkey) { return loadTlsCommon(pkey, config_.tls_pkey_path); }
 
-void FSStorage::storeMetadata(const Uptane::MetaPack& metadata) {
+void FSStorage::storeMetadataCommon(const Uptane::RawMetaPack& metadata, const std::string& suffix) {
```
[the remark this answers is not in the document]
OYTIS: No, it's just an implementation detail. The suffix is either ".raw" for metadata that wasn't verified yet and an empty string for verified metadata.

src/libaktualizr/storage/fsstorage.cc
OYTIS: So my idea was that it doesn't make sense for a single piece of metadata to be valid or invalid: the verification procedure involves cross-checks between metadata files, so we should treat metadata bundle as an atomic entity. I might be wrong of course, then we need a different logic.

src/libaktualizr/storage/sqlstorage.cc
OYTIS: I could add the check, but why I thought it would be fine here is because storeMetadataCommon is a private function, not a part of interface. And `tablename` can only take one of two hardcoded strings. So the check would be like our code checking data from itself, which doesn't make much sense. We could, for instance, use some enum variable instead of `tablename` and then infer tablename in this function internally in a switch, but it will make the code somewhat more verbose.

src/libaktualizr/storage/sqlstorage.cc
OYTIS: Same here.

src/libaktualizr/storage/sqlstorage.cc
OYTIS: And here.

src/libaktualizr/storage/sqlstorage.cc
OYTIS: And here as well.

src/libaktualizr/uptane/fetcher.cc
OYTIS: Yes, I've postponed proper root fecthing until root versioning is implemented on the backend. `fetchRoot` and `feedRoot` are temporary hacks to keep uptane vector tests running.

src/libaktualizr/uptane/managedsecondary.cc
OYTIS: This shouldn't happen because the metapack is stored/loaded atomically. So if we're still on snapshot version X, we'll be using the old metapack. When we download the new snapshot.X+1, we also download the whole metadata set, cross-check it and only if it's valid update the metadata. Doesn't it work?

src/libaktualizr/uptane/uptanerepository.cc
OYTIS: > the others MAY go down if and only if the root was rotated.

Is it UPTANE-compatible?

OYTIS:
@heartsucker 
OK, so I think we need to revise the process of metadata verification.
### Current process:
1. Download all of the metadata, store in untrusted area.
2. Verify all of the metadata as a whole, including cross-checks and store it as trusted.
3. Get the trusted metadata from the storage, distribute it to secondaries so that they can check it as well.
4. Installation and stuff it target for primary has been updated.

### New process:
1. Download timestamps and roots from the server and verify just them. If needed make root rotation.
2. Check the metadata files in the storage, if for some hash doesn't match, download missing files from the server and store them. All the metadata in the storage is untrusted.
3. After that aktualizr can get the metadata from the storage and verify all of it, including cross-checks. "Metadata packages" as a whole exist only in RAM (namely as contents of Uptane::Repository class).
4. Once metadata has been verified, distribute it from storage to secondaries over respective communication channels.
5. Installation and stuff.

Did I get you right?

heartsucker:
@OYTIS See here for a pedantic enumeration of the steps needed to perform an update: https://github.com/advancedtelematic/uptane-spec

src/libaktualizr/uptane/uptanerepository.cc
heartsucker: Yes, that is in the TUF spec.

src/libaktualizr/uptane/managedsecondary.cc
heartsucker: Not quite. If the new metapack has a bad `targets` metadata, the correct thing to do is still to update everything else.

src/libaktualizr/uptane/fetcher.cc
heartsucker: Ok, that's fair enough.

src/libaktualizr/storage/sqlstorage.cc
heartsucker: This could be done with an enum as an argument that maps it to the table. That would prevent it from passing a bad string on accident.

src/libaktualizr/storage/fsstorage.cc
heartsucker: Yes, treating metadata packs as atomic is incorrect. See the linked spec below.

config/sql/migration/migrate.5.sql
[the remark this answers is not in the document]
heartsucker: Ok, so this table is basically staging for a metadata pack, correct?

OYTIS:
@patrickvacek @cajun-rat @lbonn @heartsucker @patriotyk 
If there are no new objections, I'd prefer to merge it to avoid more rebasing. The tickets are of course not closed, but it's a modest step towards closing them.

OYTIS:
Some isConsistent() check had to go on partial secondaries, but the storeMetadata() now always applies to verified metadata, so it probably won't create any security breaches.

PR #778, opened 2018-05-23
Add new tool garage-repo

patriotyk:


src/aktualizr_repo/repo.cc
```diff
@@ -0,0 +1,128 @@
...
+  signed_data["signatures"].append(signature);
+  return signed_data;
+}
+
+std::string Repo::getExpirationTime(const std::string &expires) {
+  if (expires.size() != 0) {
+    std::smatch match;
+    std::regex time_pattern("\\d{4}\\-\\d{2}\\-\\d{2}T\\d{2}:\\d{2}:\\d{2}Z");  // NOLINT(modernize-raw-string-literal)
+    if (!std::regex_match(expires, time_pattern)) {
+      throw std::runtime_error("Expiration time has wrong format");
```
lbonn: It would be helpful to print the expected format sot that the user can fix it. ISO 8601 UTC?

docs/debugging-tips.adoc
```diff
@@ -137,3 +137,29 @@ sudo tcpdump tcp port 443 -w upload.pcap
 ```
 
 Fetch both of these down. In wireshark preferences set ssl.keylog_file to point to premaster.txt. If your https traffic not on port 443, then add the port to http.ssl.port. Now open upload.pcap.
+
+== Serve repo generated by aktualizr-repo
```
lbonn: It could be made clearer by giving an example of `aktualizr-repo` invocation.

For example:

```
aktualizr can be tested against a dummy repository containing fake images

First, generate a repository using `aktualizr-repo`:

    aktualizr-repo generate --path repo_dir
    {other commands?}

Then, serve the generated directory using a web server. Here is an example configuration for nginx:

{nginx config}

{here, maybe explain how to use it with aktualizr or tests}
```

src/aktualizr_repo/repo.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  signed_data["signed"] = json;
+  signed_data["signatures"].append(signature);
+  return signed_data;
+}
+
+std::string Repo::getExpirationTime(const std::string &expires) {
+  if (expires.size() != 0) {
+    std::smatch match;
+    std::regex time_pattern("\\d{4}\\-\\d{2}\\-\\d{2}T\\d{2}:\\d{2}:\\d{2}Z");  // NOLINT(modernize-raw-string-literal)
```
patrickvacek: Not a problem per se, but it is there any reason *not* to modernize this into a raw string literal?

PR #779, opened 2018-05-23
Keep going while running format and tidy on CI

lbonn:
This way, users will get a complete report of the static tests results
in the logs


lbonn:
@patrickvacek yes, that's precisely the objective. Note: when using ninja for development, it should be `ninja -k 0 clang-format` instead.

PR #780, opened 2018-05-23
Make absence of asn1c non fatal

lbonn:
It's currently used only for a test, so we don't need to require it in all cases. Example: building with yocto.


lbonn:
I could add another CMake warning in `src/libaktualizr/asn1/CMakeLists.txt` but CMake already outputs a pretty big one at configure time.

Also, it looks like there is a weird CI failure on Travis (also on other PRs)...

PR #781, opened 2018-05-23
Feat/enable address sanitizer

patrickvacek:
The primary change was made at @heartsucker's request based on a security recommendation.


CMakeLists.txt
```diff
@@ -269,22 +269,25 @@ set_source_files_properties(third_party/jsoncpp/jsoncpp.cpp PROPERTIES COMPILE_F
...
+    # for this to be too expensive.
+    add_compile_options(-fstack-protector-all)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined -fsanitize=bounds")
     # Enable maximum set of warnings. -Wno-sign-compare is required because of
     # problems in gtest. -Wswitch-default and -Wconversion would be nice as
     # well, but they also cause problems in gtest.
-    add_definitions(-Wall -Wextra -Wformat-security -Wfloat-equal -Wcast-qual -Wlogical-op -Wno-sign-compare)
+    add_compile_options(-Wall -Wextra -Wformat-security -Wfloat-equal -Wcast-qual -Wlogical-op -Wno-sign-compare)
 
     if (CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "4.9" OR MAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.9")
```
lbonn: Unrelated to your changes, but it looks like there is a typo there, it was probably intended to be `CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.9"`.

It generates a fair amount of warning when fixed!

patrickvacek: Wow, I made that type in what must have been my second week here. Oops. I vote for fixing that in a separate PR. Do you want to look into that?

lbonn: Sure :). I guess you can blame CMake weirdness and general shell script style programming.

patrickvacek:
By the way, this seems to noticeably slow down the build and bloat aktualizr from 26M to 69M. For an embedded application that is rather painful.

CMakeLists.txt
lbonn: @patrickvacek #782

patrickvacek:
My previous estimate was for Valgrind build types, which probably isn't useful. For a release build, it's actually far worse:
Current master branch: 3.1 M
This branch: 34 M
@heartsucker Are you really really sure we need this?

lbonn:
Note that it would also require some changes in meta-updater to ship ubsan.

Also, poor man's bound checks: http://en.cppreference.com/w/cpp/container/vector/at, maybe we can enforce its use.

@patrickvacek the last commit is intended to fix CI, right? Can we open another PR and fast-track it?

patrickvacek:
> the last commit is intended to fix CI, right? Can we open another PR and fast-track it?

Yes, will do. Any idea why that just started happening now? Seems like it's been a problem since that code was first introduced.

> Note that it would also require some changes in meta-updater to ship ubsan.

Yes, you are probably right. I'm actually quite unsure if this is really the way we want to go, but it seems that we have some directives that we might have little choice on. I hoped this would be simple but it might end up requiring a protracted discussion.

patrickvacek:
@lbonn https://github.com/advancedtelematic/aktualizr/pull/786

patriotyk:
What about to add this option for ci builds/tests only?

patrickvacek:
Something about my new overflow test is strange. Originally, it was failing on my machine with an error caused by the sanitizer flags. Now, both on CI and on my machine, it gets caught by the stack protector and throws some sort of exception that is not the same as failing a cmake test. I'm not trying to test the stack protector, as we already had that enabled, but rather array overflows. Now, when I test it locally with a larger array (on the stack or heap), the test passes! So the sanitizer is apparently not working. Before I waste any more time on trying to understand and/or fix this, I'd like to discuss the merits of these flags.

cajun-rat:
Yeah, 10x bloat on aktualizr is too much. I can see the value of running it on the device, but we need to find a set of options that are not so excessive.

cajun-rat:
Also, are we sure this is a sensible thing to turn on for production code: http://www.openwall.com/lists/oss-security/2016/02/17/9  https://wiki.gentoo.org/wiki/AddressSanitizer

patrickvacek:
Closing. Address sanitizer is too heavy and doesn't actually seem to catch anything that `-fstack-protector-all` isn't already getting.

PR #783, opened 2018-05-23
uptane: Change logging for unknown keys

doanac:
After doing a key rotation described here:

 https://docs.atsgarage.com/prod/rotating-signing-keys.html

We started seeing the LOG_INFO message. It seemed a bit concerning
to people until we dug into it more. I think this message is probably
more useful as a LOG_DEBUG. Additionally, if we are going to warn
about a key I think we ought to show it.


doanac:
This is somewhat subjective, so feel free to nack. But I think it make more sense this way.

cajun-rat:
Agreed. This is a good change, thank you!  We wrote that before there was a tool to do key rotation, but now there is a debug level message feels like the right thing.

PR #784, opened 2018-05-23
Fix finding the latest garage-sign version

OYTIS:
Plain `sorted()` sorts lexicographically


patrickvacek:
Good catch. That was my mistake from being lazy. Looks like the tests are failing with `ImportError: No module named 'dateutil'` though.

PR #785, opened 2018-05-24
Refactor to use PublicKey object

cajun-rat:
Change-Id: If05bec7e4c8ed45672a6722545fb8395084fb870

### DIFF
--- a/src/libaktualizr/crypto/crypto.cc
+++ b/src/libaktualizr/crypto/crypto.cc
@@ -10,15 +10,90 @@
 #include "openssl_compat.h"
 #include "utilities/utils.h"
 
-std::string PublicKey::TypeString() const {
-  switch (type) {
-    case RSA:
-      return "rsa";
-    case ED25519:
-      return "ed25519";
-    default:
-      return "UNKNOWN";
+PublicKey::PublicKey(const boost::filesystem::path &path) : value_(Utils::readFile(path)) {
+  type_ = Crypto::IdentifyRSAKeyType(value_);
+}
+PublicKey::PublicKey(Json::Value uptane_json) {
+  if (!uptane_json["keytype"].isString()) {
+    type_ = kUnknownKey;
+    return;
   }
+  if (!uptane_json["keyval"].isObject()) {
+    type_ = kUnknownKey;
+    return;
+  }
+
+  if (!uptane_json["keyval"]["public"].isString()) {
+    type_ = kUnknownKey;
+    return;
+  }
+
+  std::string keytype = uptane_json["keytype"].asString();
+  std::string keyvalue = uptane_json["keyval"]["public"].asString();
+
+  std::transform(keytype.begin(), keytype.end(), keytype.begin(), ::tolower);
+
+  KeyType type;
+  if (keytype == "ed25519") {
+    type = kED25519;
+  } else if (keytype == "rsa") {
+    type = Crypto::IdentifyRSAKeyType(keyvalue);
+    if (type == kUnknownKey) {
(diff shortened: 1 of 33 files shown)


src/libaktualizr/crypto/crypto.cc
```diff
@@ -345,3 +397,37 @@ bool Crypto::generateKeyPair(KeyType key_type, std::string *public_key, std::str
...
+  int key_length = RSA_size(rsa.get());
+
+  // It is not clear from the OpenSSL documentation if RSA_size returns
+  // exactly 2048 or 4096, or if this can vary. Instead of an exact match,
+  // just split the range into two
+  if (key_length < 3000) {
+    return kRSA2048;
+  } else {
+    return kRSA4096;
+  }
```
patrickvacek: This seems dangerous to me. From what I can tell, RSA_size returns whatever the modulus is, which should be equivalent to our expectations (exactly 2048 or 4096). Since key sizes could potentially be whatever you can imagine, this seems rife for error.

src/libaktualizr/crypto/crypto.cc
```diff
@@ -345,3 +397,37 @@ bool Crypto::generateKeyPair(KeyType key_type, std::string *public_key, std::str
...
+}
+KeyType Crypto::IdentifyRSAKeyType(const std::string &public_key_pem) {
+  StructGuard<BIO> bufio(BIO_new_mem_buf(reinterpret_cast<const void *>(public_key_pem.c_str()),
+                                         static_cast<int>(public_key_pem.length())),
+                         BIO_vfree);
+  if (bufio.get() == nullptr) {
+    throw std::runtime_error("BIO_new_mem_buf failed");
+  }
+  StructGuard<::RSA> rsa(PEM_read_bio_RSA_PUBKEY(bufio.get(), nullptr, nullptr, nullptr), RSA_free);
```
patrickvacek: I'm also quite skeptical that this is a good idea. This looks to me like it is begging for some sort of collision or prepared attack. It also privileges RSA in a way that seems less than advisable considering that I thought we wanted to move to ED25519.

patrickvacek:
Other than my RSA complaints and CI, this looks good.

src/libaktualizr/crypto/crypto.cc
cajun-rat: Reasonable. I'll check for an absolute match.

src/libaktualizr/crypto/crypto.cc
cajun-rat: The difficulty is that the Uptane protocol just talks about 'RSA' keys, and we need a way to convert those into a `PublicKey`. I can see three options here:

* Autodetection (what is currently implemented): We are given `keytype="rsa", keyvalue="abc123..."`, and look at the value to see what kind of key it is.
* Introduce a new 'kRSAANY' type to KeyType, which means 'Any length of RSA key'
* Introduce a parallel enum with just ED25519 and RSA options (feels ugly)

I don't think autodetection is so bad. It is basically part of the process of decoding a Uptane formated key, where once we know it is RSA, we take a look at the bytes and figure out how long it is.

cajun-rat:
@patrickvacek are you happy with this now?

patrickvacek:
@cajun-rat I had one unanswered question: is `IdentifyRSAKeyType` only to be used on keys that we already know are RSA, but we don't know the length? If so, then I'm happy. Otherwise, I'm still skeptical.

cajun-rat:
@patrickvacek Yes, it is just for keys that are already known to be RSA keys.

PR #786, opened 2018-05-24
Check that a config dir exists before using it.

patrickvacek:
AktualizrInfoConfig now inherits from BaseConfig just like the other configs.


patrickvacek:
Crap, do not merge. This fix breaks tests.

patrickvacek:
> assuming the Jenkins box doesn't pass out.

Travis doesn't seem to be doing much better...

lbonn:
There are some clang-tidy warnings, I'll push a fix here

lbonn:
Ooops, I merged before pushing here, sorry... I've added directly the fix to master afterwards

PR #787, opened 2018-05-24
Add missing entries to the 'build everything' test

cajun-rat:
Change-Id: Iddf0a1939f887aba1a8df345acd907f2a156bd48

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get update && apt-get -y install \
   lcov \
   libarchive-dev \
   libboost-dev \
+  libboost-iostreams-dev \
   libboost-log-dev \
   libboost-program-options-dev \
   libboost-random-dev \
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -176,9 +176,11 @@ add_test(NAME test_build_all_off
         COMMAND ${PROJECT_SOURCE_DIR}/tests/build_with_options.sh
         ${PROJECT_SOURCE_DIR} test_build_all_off "-DBUILD_OSTREE=OFF -DBUILD_DEB=OFF")
 
+# Build with everything. BUILD_OPCUA requires boost::filesystem::relative,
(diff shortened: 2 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -178,7 +178,7 @@ add_test(NAME test_build_all_off
 
 add_test(NAME test_build_all_on
         COMMAND ${PROJECT_SOURCE_DIR}/tests/build_with_options.sh
-        ${PROJECT_SOURCE_DIR} test_build_all_on "-DBUILD_WITH_CODE_COVERAGE=ON -DBUILD_OSTREE=ON -DBUILD_DEB=ON")
+        ${PROJECT_SOURCE_DIR} test_build_all_on "-DBUILD_WITH_CODE_COVERAGE=ON -DBUILD_OSTREE=ON -DBUILD_DEB=ON -DBUILD_SOTA_TOOLS=ON -DBUILD_ISOTP=ON -DBUILD_OPCUA=ON -DBUILD_LOAD_TESTS=ON")
```
patrickvacek: `BUILD_SYSTEMD`?

cajun-rat: `BUILD_SYSTEMD` is on the default. I could add it though?

patrickvacek: Oh that's right. Porbably don't need to worry about it then.

lbonn:
Since these tests take a long time, I'd be in favor of running less of these tests and instead compiling directly with all options on default CI environment. But in the meantime, it looks good.

Ok when CI is

Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN apt-get update && apt-get -y install \
   lcov \
   libarchive-dev \
   libboost-dev \
+  libboost-iostreams-dev \
```
patrickvacek: Why was this necessary if we ultimately didn't build with BUILD_OPCUA?

PR #790, opened 2018-05-25
Cancel old Jenkins builds in PRs

lbonn:
PRO-5436


lbonn:
Not particularly pretty but it seems to work

PR #791, opened 2018-05-25
Aktualizr repo(comments and tests)

patriotyk:
Processed lbon's comments and added tests.


cajun-rat:
This looks good, but can you also add some tests to ensure that the generated Uptane metadata validates correctly with Aktualizr?

PR #792, opened 2018-05-25
Drop python dateutil

cajun-rat:
Python dateutil isn't packaged for Yocto, which makes this fail when built in meta-updater

### DIFF
--- a/scripts/get_garage_sign.py
+++ b/scripts/get_garage_sign.py
@@ -8,7 +8,6 @@ import sys
 import tarfile
 import urllib.request
 import xml.etree.ElementTree as ET
-import dateutil.parser as dp
 
 from pathlib import Path
 
@@ -52,7 +51,7 @@ def main():
             else:
                 name = name_ext
     else:
-        name = sorted(versions, key=(lambda name: dp.parse(versions[name][0])))[-1]
+        name = sorted(versions, key=(lambda name: (versions[name][0])))[-1]
 
     path = args.output.joinpath(name)
     md5_hash = versions[name][1]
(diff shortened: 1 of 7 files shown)


patrickvacek:
We could package dateutil in Yocto, right? Otherwise, I think we need to find some way to do this correctly. If you want to merge this now, I understand, but we should still get it right very soon.

cajun-rat:
I think the string representation used does sort correctly in lexicographic order, so won't the two implementations have the same behaviour?

cajun-rat:
The main reason to get this in is so that we can test the build in Yocto again, though.

patrickvacek:
> I think the string representation used does sort correctly in lexicographic order, so won't the two implementations have the same behaviour?

Unfortunately, no:

```
>>> a = "cli-0.4.0-20-gc1fb488.tgz"
>>> b = "cli-0.4.0-8-g9d6664d.tgz"
>>> a < b
True
```

cajun-rat:
But this is sorting by Last modified date (the first item of the tuple that is the value of the versions dict)?

scripts/get_garage_sign.py
```diff
@@ -52,7 +51,7 @@ def main():
             else:
                 name = name_ext
     else:
-        name = sorted(versions, key=(lambda name: dp.parse(versions[name][0])))[-1]
+        name = sorted(versions, key=(lambda name: (versions[name][0])))[-1]
```
lbonn: Was there before but `max` should be equivalent and have a clearer intent

cajun-rat: Good idea. Fixed.

cajun-rat:
@OYTIS you added the dateutil code: is this change going to break something?

PR #793, opened 2018-05-25
Add fetch from ostree

koshelev:


patrickvacek:
Rebased and formatted to hopefully get tests to pass.

src/load_tests/main.cc
```diff
@@ -7,9 +7,42 @@
...
 
+#ifdef BUILD_OSTREE
+void fetchRemoteCmd(const std::vector<std::string> &opts) {
+  std::string inputDir;
+  std::string outDir;
+  std::string branchName;
+  std::string remoteUrl;
+  uint opsPerSec;
+  uint opsNr;
+  uint parallelism;
```
patrickvacek: Can we change this to a standard type? On my machine, I found `typedef unsigned int uint;` in `/usr/include/sys/types.h`, where it is marked `Old compatibility names for C types.` and is surrounded by a `#ifdef __USE_MISC` guard. I'd rather not take our chances on the portability of this type.

src/load_tests/treehub.cc
```diff
@@ -0,0 +1,103 @@
...
+  const Config &config;
+  std::shared_ptr<INvStorage> storage;
+  std::shared_ptr<KeyManager> keys;
+  const fs::path repoDir;
+  const std::string branchName;
+  const std::string remoteUrl;
+
+  OstreeRepo *repo = nullptr;
+  void initRepo() {
+    GFile *gRepoFile = g_file_new_for_path(repoDir.native().c_str());
```
patrickvacek: I know these are "just" tests, but still, for better memory management, can we use `StructGuard` here? There are some examples in `src/libaktualizr/package_manager/ostreemanager.cc`.

patrickvacek:
Rebased and simplified the GObject business just a bit. 

BTW, I noticed two other things that we should maybe think about later, but aren't worth holding this up for:
1. You tend to list `private` class members before `public` and data members before methods. According to the Google style guide that we use, `public` should come first, and methods before data members. This also means that you can't leave out the `private` keyword, which I also think is useful.
2. You have class declarations in a .cc file. This is normally rare, but since you only do use the classes within that file, it is perfectly fine. It does imply that you are trying to hide the classes from any "public" interface, though.

PR #794, opened 2018-05-27
Use custom class for ecu_id instead of std::string

patriotyk:


cajun-rat:
This is going in the right direction, but there are a couple of things that I think need looking at:

The name `EcuSerialId` doesn't match either the name used in the Scala server-side code [EcuSerial](https://github.com/advancedtelematic/libats/blob/149f2b5b3e6a1069c53683f0eecf57dc3cbeff0d/libats-messaging-datatype/src/main/scala/com/advancedtelematic/libats/messaging_datatype/DataType.scala#L40) or [EcuIdentifier](https://docs.google.com/document/d/1wjg3hl0iDLNh7jIRaHl3IXhwm0ssOtDje5NemyTBcaw/edit#heading=h.l6lkvdudrui2) which is used in the Uptane Specification. The nice thing about using the existing terminology is that it makes it easier for people who come to the code base, and means that we can refer people to the Uptane specification rather than include equivalent documentation in Aktualizr.  Given that we have two choices, and `HardwareIdentifier` and `EcuIdentifier` are rather similar, I would suggest using `EcuSerial` as the name, and include a link to the relevant line in the Scala libats codebase.

Creating a common base class for both is nice because it saves duplicated code around defining operator...() and the hash implementation, however it comes with a some costs which I think outweigh any benefits:
* They both inherit from `BaseId`, and (public) inheritance implies a 'IS A' relationship between the two.  Other than being representable as a string and having value-like semantics (equality, hash, etc) I can't think of any other common behaviour. 
* It makes error reporting harder, for example the the error message `throw std::out_of_range("Hardware Identifier too long");` will be very confusing if it is actually a EcuSerial that is throwing the error. We could change it to something generic and make the error message less useful, or provide a hook for the base class to report it's name and make the implementation more complex.
* The validity checks are wrong: while a HardwareIdentifier is a 0..200 character string, a EcuSerial is between 1 and 64 characters long ([link](https://github.com/advancedtelematic/libats/blob/149f2b5b3e6a1069c53683f0eecf57dc3cbeff0d/libats-messaging-datatype/src/main/scala/com/advancedtelematic/libats/messaging_datatype/DataType.scala#L42)) . It would be declare the limits in the base class and override these in the derived classes, but it feels like this will be a lot more complicated than duplicating the definitions of `operator...()` 
* It would be reasonable to want to change one of these to not be a string in the future, for example a int64 or a UUID. If this happened we'd have to rework any code that uses `BaseId`.

I understand the goal of not duplicating code, but I think that in this case the costs are too high to justify the complexity.

src/libaktualizr/uptane/tuf.h
```diff
@@ -80,36 +80,47 @@ class TimeStamp {
...
       throw std::out_of_range("Hardware Identifier too short");
     } */
-    if (kMaxLength < hwid.length()) {
+    if (kMaxLength < id.length()) {
       throw std::out_of_range("Hardware Identifier too long");
     }
   }
 
-  std::string ToString() const { return hwid_; }
+  const std::string &ToString() const { return id_; }
```
cajun-rat: This feels like a dangerous optimisation. If the caller keeps the std::string reference alive longer than the `BaseId` object, then they will reference an invalid pointer. Given that it is hard to statically validate these things, I think we should pray for a Sufficiently Intelligent Compiler, i.e. accept the extra cost.

patriotyk:
Some tests of oe-selftest fails. Fixing them

patrickvacek:
@patriotyk Just checking: is it the HSM/implicit tests because the input .toml isn't found? I have a series of PRs open to fix various issues in various branches (https://github.com/advancedtelematic/meta-updater/pull/331, https://github.com/advancedtelematic/meta-updater/pull/330, https://github.com/advancedtelematic/meta-updater/pull/329), but there is still a lingering issue with sysroot population that I've asked about on the Yocto mailing list.

patriotyk:
I get following fails:
`
2018-06-05 10:15:25,848 - oe-selftest - INFO - test_grub (updater.GrubTests)
2018-06-05 10:56:10,054 - oe-selftest - INFO -  ... FAIL
2018-06-05 10:56:10,057 - oe-selftest - INFO - test_provisioning (updater.HsmTests)
2018-06-05 10:56:41,531 - oe-selftest - INFO -  ... FAIL
2018-06-05 10:56:41,533 - oe-selftest - INFO - test_provisioning (updater.ImplProvTests)
2018-06-05 10:57:13,638 - oe-selftest - INFO -  ... FAIL
2018-06-05 10:57:25,750 - oe-selftest - INFO - ERROR
2018-06-05 10:57:25,750 - oe-selftest - INFO - test_rpi (updater.RpiTests)
2018-06-05 10:58:25,348 - oe-selftest - INFO -  ... FAIL
2018-06-05 10:58:26,323 - oe-selftest - INFO - ERROR
`

patrickvacek:
Seems like something is wrong. What branch are you testing? Running oe-selftest in rocko should give you a bit more output, and sumo even more. All of my branches were merged, so it may also help to run with the latest version if you aren't already.

patriotyk:
On sumo branch I got lot of output and one of the error is following:
configure: error:
`
| '/build/patriotyk/aktualizr/build-quemu/tmp/hosttools/mktemp -d' does not create temporary directories.
| If you don't have a version of mktemp that can create directories, you
| can specify '--without-mktemp' and quilt will use its own
| internal tempfile generation mechanism.
`
Is it known issue?

patrickvacek:
> configure: error: | '/build/patriotyk/aktualizr/build-quemu/tmp/hosttools/mktemp -d' does not create temporary directories.

No, I've never seen that. That sounds pretty weird. What is the OS of the machine you are building on? I struggle to imagine that a modern implementation of `mktemp` can't create directories.

patriotyk:
Oh. instead of `rm -rf tmp/` I ran `rm -rf /tmp` And removed systems tmp folder. Now I restored it and running tests again

patriotyk:
`oe-selftest -r updater` passes.

cajun-rat:
Looks great!

PR #796, opened 2018-05-29
Test/buffer overflow

patrickvacek:
Taking the best parts of https://github.com/advancedtelematic/aktualizr/pull/781. I'm still not 100% sure that the test will always fail like I expect, but this seems to be more "robust" (as in: likely to fail) than my previous version. The goal here is just to make sure `fstack-protector-all` is doing its job.


patrickvacek:
I suspect that what I'm trying to do is a bad idea. It seems like overflow errors are inconsistent and unreliable (i.e. undefined), so trying to get an overflow test to break the same way on every machine is difficult. I've tried compiling with -O0 but that just changes the test into breaking in a way that apparently is not caught by cmake's `WILL_FAIL` properties. I can't find any other way of catching that type of failure, so this might not be possible with our current tools. It's also not very important, so I'm going to move on now unless someone has some good ideas.

PR #798, opened 2018-05-30
Feat/rollback uboot

OYTIS:
I had to use an OSTree function only implemented in libostree v2017.7, so have to use upstream ostree instead of those in debian. This creates more load on CI servers. One option would be to only enable BUILD_OSTREE in some tests, and do not build it if not needed. Another could probably be an LGPL-licensed dynamic library consisting of just the function that is missing from older versions.

meta-updater code relies on aktualizr version, so I'll have to wait for this PR.


OYTIS:
Yes, so what this PR basically does is it dropping `bootcounter` variable on successful boot and setting `upgrade_available` variable on pending update. The latter is used by uboot to avoid unnecessary writes to environment storage (boot partition in case of RPi).

lbonn:
To solve the CI load problem, we could keep the ability to specify an external ostree version from CMake and the build could be done inside docker so that it will be cached.

OYTIS:
Anyway, moving build to Dockerfiles should help. There is no such thing as "INCLUDE" command in docker, right? At least I couldn't google one.

lbonn:
@OYTIS about 02a2773, it was about running the tests as a regular user (as opposed to root) but it was not entirely complete at the time. On Jenkins, this is done through the `docker run` invocation, so these are not needed anymore.

I'd say we can probably revert it now.

OYTIS:
Just realized that one feature needed for rollbacks is missing from OSTree. I'll clarify if that is a bug and what can be done about it.

OYTIS:
OK, now everything should be in place

lbonn:
What's the new minimum ostree version? Perhaps it should appear in the readme

OYTIS:
@lbonn Good catch, added.

src/libaktualizr/bootloader/bootloader.cc
```diff
@@ -0,0 +1,94 @@
...
+      break;
+    case kUbootMasked:
+      if (Utils::shell("fw_setenv bootcount 0", &sink) != 0) {
+        LOG_WARNING << "Failed resetting bootcount";
+      }
+      if (Utils::shell("fw_setenv upgrade_available 0", &sink) != 0) {
+        LOG_WARNING << "Failed resetting upgrade_available for u-boot";
+      }
+      break;
+    default:
```
patrickvacek: Shouldn't we have a `kGrub` case for `setBookOK` and `updateNotify`? Otherwise they will throw an exception no matter what if we try to use grub, which seems a bit harsh. It's true, they aren't implemented, but that shouldn't break functionality.

patrickvacek: Well, I see that there is nothing that sets the grub option, so maybe this doesn't really matter. Up to you. On the whole, this seems fine, but I'd like to test it or see a demo.

OYTIS: Yes, maybe I should delete grub option at all to avoid confusion. Grub integration is not even scheduled yet.

OYTIS:
Broken aktualizr, (hopefully) unbroken aktualizr and added documentation on rollback.

docs/rollback.adoc
```diff
@@ -0,0 +1,60 @@
+= Rollbacks
+
+Since Aktualizr is a userspace application and relies on a lot of other software which can be
+affected by an unsuccessful update, rollback functionality should be implemented in the bootloader.
+
+Some support from Aktualizr is nevertheless required. The most important thing is notifying bootloader
+on successful boot (`Bootloader::setBootOK()`). Notidying on the update (`Bootloader::updateNotify`)
```
lbonn: "Notidying"

src/sota_tools/CMakeLists.txt
```diff
@@ -24,10 +24,19 @@ set (GARAGE_PUSH_SRCS
   garage_push.cc
 )
 
+set (SOTA_TOOLS_EXTERNAL_LIBS
+  ${Boost_SYSTEM_LIBRARIES}
+  ${Boost_LIBRARIES}
+  ${CMAKE_THREAD_LIBS_INIT}
+  ${LibArchive_LIBRARIES}
+  ${CURL_LIBRARIES}
+  ${OPENSSL_LIBRARIES}
```
lbonn: Adding `${sodium_LIBRARY_RELEASE}` here seems to resolve the link issues

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -55,6 +56,7 @@ class SotaUptaneClient {
   std::shared_ptr<PackageManagerInterface> pacman;
   HttpInterface &http;
   Uptane::Fetcher uptane_fetcher;
+  const Bootloader &bootloader;
```
cajun-rat: Is there a reason not to make this a `shared_ptr`?

OYTIS: Not really. I think until now we only wrapped stuff into `shared_ptr` when we really needed to. But I agree that we might want to make it standard. Should we pass `HttpInterface` and `Uptane::Repository` as shared pointers as well or were there some reasons not to do it?

cajun-rat: Good point.  I'm in favour of std::shared_ptr over references everywhere, but for consistency with the other parameters we can/should stick with references here.  The move away from references can be a refactor :)

PR #799, opened 2018-05-30
Build garage_deploy.deb on Ubuntu Xenial

lbonn:
Same environment as aktualizr.deb, so it saves us one build environment!


lbonn:
So we had:

- ubuntu xenial with p11 for coverage + valgrind tests
- ubuntu xenial without p11 for valgrind crypto tests
- debian testing with openssl 1.1 for crypto tests and static tests
- debian stable for building garage_deploy.deb and testing installation
- ubuntu xenial for building aktualizr.deb and testing installation

This PR merges the last two so that all packages are built with xenial.

The second environment (without p11) might actually be redundant since we actually test non-p11 crypto on the first and we test building without p11 on the .deb install env.

patrickvacek:
As long as we're okay with skipping testing on debian stable (which is probably fine), this is probably (conceptually) fine -- as in, good to go once it passes CI.

You are probably right that we could axe the second environment. Didn't we do some work to get the p11 code isolated such that we aren't automatically running with p11 if aktualizr was built with it enabled? As long as we have tests run under valgrind (and codecov) covering non-p11 and p11 cases, then we are set.

It may also be time to consider moving one of the ubuntu xenials to 18.04.

patrickvacek:
> Oh and about 18.04, we could start by using it instead of debian testing for openssl 1.1 and static tests.

I think we should still have at least one debian test env around, probably debian-testing.

tkfu:
Once this is merged and a new version is on the releases page, can @lbonn (or someone else) merge in https://github.com/advancedtelematic/ats-garage-docs/pull/45?

PR #801, opened 2018-05-31
Use asn1c generated serializers

cajun-rat:
When we originally decided to use our own serialization code, we hoped this
would mean that we could reuse our existing domain objects, but in practice we
haven't been able to do that, and are going bytes -> message -> domain object
anyway in many cases.  If we are going to have an intermediate 'message'
object, then it would seem a good idea to use an auto-generated one.

We currently have code that uses tag values and dynamic casts to deal with
discriminated unions of message types, and this would be replaced with the
code generated by asn1c, which is a enum + struct--so no worse, really.

Debugging will be made easier, because we can dump a human and machine readable
representation of the message using XER.

Finally, since BER messages are self-delimiting, and asn_decode() supports
reading partial messages from a stream, we can switch to using a single TCP
connection for the entire protocol, rather start a TCP connection per message,
meaning that the secondary won't have to open a connection to the primary any
more.

This commit only moves secondary discovery over to the the system. Once this is
in we can move the rest of the protocol over.


cmake-modules/FindAsn1c.cmake
```diff
@@ -1,55 +1,55 @@
...
+
+
+function(compile_asn1_lib)
+    set(options)
+    set(oneValueArgs )
+    set(multiValueArgs SOURCES)
+    cmake_parse_arguments(AKTUALIZR_ASN1 "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+    # clean previously generated files
+    set(ASN1_GEN_DIR ${CMAKE_CURRENT_BINARY_DIR}/generated/asn1/)
```
lbonn: +1 for using the binary dir

cmake-modules/FindAsn1c.cmake
```diff
@@ -1,55 +1,55 @@
...
+    add_custom_command(
+        OUTPUT ${ASN1_GENERATED}
+        COMMAND ${ASN1C} ${ASN1C_FLAGS} ${S}
+        WORKING_DIRECTORY ${ASN1_GEN_DIR}
+        DEPENDS ${S}
+        )
+
+    list(REMOVE_ITEM ASN1_GENERATED ${ASN1_GEN_DIR}/converter-example.c ${ASN1_GEN_DIR}/converter-sample.c)
+
+    # hardcoded list of common c files
```
lbonn: this comment dates back from an old version of this file and was already obsolete

src/libaktualizr/asn1/messages/cryptosource.asn1
```diff
@@ -3,6 +3,7 @@ CryptoSource DEFINITIONS ::= BEGIN
   AKCryptoSource ::= ENUMERATED {
 	kFile,
 	kPkcs11,
-	...
+	...,
+	ddd
```
lbonn: Is this intentional?

src/libaktualizr/asn1/asn1_message.h
```diff
@@ -0,0 +1,106 @@
...
+    memmove(&msg_, *msg, sizeof(AKIpUptaneMes_t));
+    free(*msg);  // Be careful. This needs to be the same free used in der_decode
+    *msg = nullptr;
+  }
+};
+
+/**
+ * Adaptor to write output of der_encode to a string
+ */
+int WriteToString(const void* buffer, size_t size, void* priv);
```
lbonn: Could these two functions have more specific names?

src/libaktualizr/asn1/asn1_message.h
```diff
@@ -0,0 +1,106 @@
...
+ private:
+  boost::intrusive_ptr<Asn1Message> root_;
+  T* me_;
+};
+
+/**
+ * Reference counted holder for the top-level ASN1 message structure.
+ */
+
+class Asn1Message {
```
lbonn: It's called `Asn1Message` but it only covers `AKIpUptaneMes`

lbonn:
Apart for my (mostly stylistic) comments, I like it

src/libaktualizr/asn1/messages/cryptosource.asn1
cajun-rat: oops, didn't see that with the rename

src/libaktualizr/uptane/ipsecondarydiscovery.cc
```diff
@@ -55,25 +57,32 @@ std::vector<Uptane::SecondaryConfig> IpSecondaryDiscovery::waitDevices() {
...
-      conf.secondary_type = Uptane::SecondaryType::kIpUptane;
-      conf.ip_addr = sec_address;
-      Utils::setSocketPort(&conf.ip_addr, htons(sec_port));
-      secondaries.push_back(conf);
-    } catch (const deserialization_error &ex) {
-      LOG_ERROR << ex.what();
+    Uptane::SecondaryConfig conf;
+    AKIpUptaneMes_t *m = nullptr;
+    asn_dec_rval_t res = ber_decode(nullptr, &asn_DEF_AKIpUptaneMes, reinterpret_cast<void **>(&m), rbuf, recieved);
+    Asn1Message::Ptr msg = Asn1Message::FromRaw(&m);
```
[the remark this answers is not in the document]
cajun-rat: Apparently it can still allocate even if result is an error: "Upon unsuccessful termination, the *struct_ptr_ptr may contain partially decoded
data." From: https://github.com/vlm/asn1c/raw/master/doc/asn1c-usage.pdf

I'll add a note to the source code explain why.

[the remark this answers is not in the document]
cajun-rat: ... and make `Asn1Message::FromRaw` handle `nullptr`

src/libaktualizr/uptane/ipsecondarydiscovery.cc
```diff
@@ -55,25 +57,32 @@ std::vector<Uptane::SecondaryConfig> IpSecondaryDiscovery::waitDevices() {
...
+
+    if (res.code != RC_OK) {
+      LOG_ERROR << "Failed to parse discovery response message";
```
lbonn: It looks like this function handles the case of incomplete messages gracefully (it will continue on `RC_WMORE`).

But in that case, it should probably not issue a parse error log, since it's a pretty normal case.

src/libaktualizr/uptane/ipsecondarydiscovery.cc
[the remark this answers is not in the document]
lbonn: great!

src/libaktualizr/asn1/asn1_message.h
cajun-rat: I did play with the idea of making Asn1Message generic using some template magic, but it got rather out of hand.  For the purposes of RPC, I think we can keep all of the messages for the communication under a single ASN.1 message.  I suspect the answer is to rename `AKIpUptaneMes` to something more generic.

lbonn:
Just for info, [this comment](https://github.com/advancedtelematic/aktualizr/pull/801#discussion_r192113543) became hidden, because GitHub.

src/libaktualizr/uptane/ipsecondarydiscovery.cc
cajun-rat: The incoming messages are datagrams from the secondaries, so if they don't parse completely then I think the only think we can do is to consider `RC_WMORE` to be the same as `RC_FAIL`?

Was your question about how loudly we should complain on about parse errors?  I was unsure whether these should be warnings or errors. From a single application point of view, I'd consider this a warning: it shouldn't happen, but we can continue service correctly. However from a global point of view, there is probably something up: the secondary shouldn't be sending us junk like this.

lbonn: Oh ok yes, I forgot that this part uses datagram and made the comment thinking about tcp, sorry.

About the loudness, `LOG_ERROR` sounds about right then.

PR #804, opened 2018-06-01
Bugfix/implicit writer simplified config

patrickvacek:


src/aktualizr_info/CMakeLists.txt
```diff
@@ -40,6 +42,7 @@ add_test(NAME aktualizr-info-images-target
 set_tests_properties(aktualizr-info-images-target PROPERTIES
     PASS_REGULAR_EXPRESSION "image targets.json content:\n\n\\{.*\\}")
 
-aktualizr_source_file_checks(${AKTUALIZR_INFO_SRC} ${AKTUALIZR_INFO_HEADERS})
-
+set(AKTUALIZR_INFO_ALL_CHECKS ${AKTUALIZR_INFO_SRC} ${AKTUALIZR_INFO_HEADERS} ${TEST_SOURCES})
+list(REMOVE_DUPLICATES AKTUALIZR_INFO_ALL_CHECKS)
```
lbonn: idea: we could actually move `REMOVE_DUPLICATES` in `aktualizr_source_file_checks` since it seems to be a common problem (`add_aktualizr_test` does it for us and if we have conditionally-run tests it becomes a headache).

patrickvacek: Yeah, not a bad idea at all. I'll give that a try.

patrickvacek: Done. Now ready for re-review.

src/libaktualizr/storage/invstorage.cc
```diff
@@ -25,7 +25,14 @@ void StorageConfig::updateFromPropertyTree(const boost::property_tree::ptree& pt
 }
 
 void StorageConfig::writeToStream(std::ostream& out_stream) const {
-  writeOption(out_stream, type, "type");
+  std::string storage_type;
```
lbonn: Good catch!

Would it work if we moved [this function](https://github.com/advancedtelematic/aktualizr/blob/4bdb35dfbc96351ffd1e1ef9345cbb20966766e1/src/libaktualizr/config/config.cc#L19) to this file?

patrickvacek: Ah yes, that was presumably supposed to be what made that work. Let me give that a try.

patrickvacek: Done. Even better catch. Thanks!

PR #805, opened 2018-06-01
Test/prov on ci

lbonn:


patrickvacek:
Nice. Where do the credentials come from?

lbonn:
Ok it went pretty easily actually. I hope these tests won't be too flaky

patrickvacek:
So there aren't actually credentials uploaded yet, right? Makes some sense that it should still work even if you don't provide anything.

And did we ever decide where the credentials should actually come from, as in, whose account and which server?

lbonn:
Hum no, I just uploaded my credentials to Jenkins and it uses them actually. If there isn't any, the CI coverage will not run. Would be nice to still run the rest of the tests but I don't really how to do that.

But I don't know which credentials we should actually use, yes.

patrickvacek:
Couldn't you just run with `SOTA_PACKED_CREDENTIALS` set to an empty string and let CMake take care of it from there? But in any event, I'd prefer to always use credentials if possible. I vote for an embedded team CI/dummy/test account on QA for the credentials, although we probably want to be able to run with CI credentials, too. Do you think there's an easy way to make it configurable which credentials are used?

lbonn:
From where do you want to parametrize them? You can already run that on your machine or a docker env: `TEST_SOTA_PACKED_CREDENTIALS=whatever.zip ./scripts/test.sh`

Here is how to parametrize Jenkins right now:
http://192.168.1.39:8080/credentials/store/system/domain/_/credential/garage-credentials/

But it's indeed global to all CI runs. You'd like to run some CI jobs with different credentials?

patrickvacek:
Nice. I'm mostly thinking of if we are developing something that specifically does or doesn't work with a given server. We may want to temporarily use CI instead of QA (or vice versa). Looks like we can swap out if really necessary, but that shouldn't come up often.

I would still recommend using a semi-generic account that you won't mind sharing with the group in case we need to go investigating, though.

patrickvacek:
BTW I've also been having problems with the prov_test lately. I had it turned off for the last week or two, and now that I'm trying it again it keeps not working.

lbonn:
Annoying... Well this time it worked, I'd say we could merge this provisionally and reconsider if it turns out to be too unreliable.

patrickvacek:
Working for me locally now too. Huh.

PR #807, opened 2018-06-05
Refactor TCP/IP Uptane secondaries to use generated messages

cajun-rat:
This currently uses an adaptor to map the AktualizrSecondary interface to Uptane::Secondary. In the future I'd like to refactor that to use Uptane::Secondary everywhere, but I don't want this PR to grow too big.


src/libaktualizr/utilities/dequeue_buffer.h
```diff
@@ -0,0 +1,55 @@
...
+ private:
+  static constexpr size_t kBufferSize = 4096;
+  static constexpr int kSentinel = 1337161494;
+
+  /**
+   * buffer_[0..written_bytes_] contains to contents of this dequeue
+   */
+  size_t written_bytes_{0};
+  char buffer_[kBufferSize]{};  // Zero initialise as a security pesimisation
+  int sentinel_{kSentinel};     // Sentinel to check for writers overflowing buffer_
```
lbonn: With the current level of optimization going on, not sure if we can rely on this... I'd say it's not worth the complication.
Plus all calls to assert are optimized out if `NDEBUG` is not set: http://en.cppreference.com/w/cpp/error/assert

Instead, it could maybe use a `std::array` with bound-checked accessors: http://en.cppreference.com/w/cpp/container/array/at

cajun-rat: Still, switching to std::array makes sense. It shouldn't be hard

lbonn: Then it may be worth checking `span`, maybe it'd apply here: http://codexpert.ro/blog/2016/03/07/guidelines-support-library-review-spant/ (standardized in C++20, but also available in some externally supported libraries).

Maybe that's too future-y :)

cajun-rat: Span looks interesting, however the goal of DequeueBuffer is to provide an API that fits between `recv()` and `ber_decode()`, so the external interface needs to be based on raw pointers and lengths.  I can't see how using it internally would help the implementation (which is basically just a memmove (now std::copy_backwards) when data is consumed off the head.

cajun-rat:
@lbonn updated

src/aktualizr_secondary/socket_server.cc
```diff
@@ -0,0 +1,183 @@
...
+#include "utilities/sockaddr_io.h"
+
+#include <linux/tcp.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+
+void SocketServer::Run() {
+  while (true) {
+    int con_fd;
+    std::unique_ptr<sockaddr_storage> peer_sa(new sockaddr_storage);
```
lbonn: can this be allocated on the stack here?

src/libaktualizr/utilities/dequeue_buffer.cc
```diff
@@ -0,0 +1,50 @@
...
+  assert(sentinel_ == kSentinel);
+  // It would be possible to have a smarter algorithm here that
+  // only shuffles bytes down when Tail() is called and we are getting
+  // close to the end of the buffer, or when the buffer is nearly empty
+  // the memmove operation is cheap. This isn't performance critical code.
+  if (written_bytes_ < bytes) {
+    throw std::logic_error("Attempt to DequeueBuffer::Consume() more bytes than are valid");
+  }
+  size_t remaining = written_bytes_ - bytes;
+  std::copy_backward(buffer_.begin() + bytes, buffer_.begin() + written_bytes_, buffer_.begin() + remaining);
```
patrickvacek: Why `copy_backward`? Doesn't that risk undefined/dangerous behavior if `bytes < remaining`?

lbonn: [std::move](http://en.cppreference.com/w/cpp/algorithm/move) is overloaded apparently

patrickvacek: [std::copy](http://en.cppreference.com/w/cpp/algorithm/copy) should be fine, too.

cajun-rat: It has to be one of the _backward interfaces, because (unlike `memmove`) `std::copy` and `std::move` are undefined in the cases where the input and output ranges overlap.

I much prefer the `memmove` API, but I wanted to stick to the STL where possible.

cajun-rat: sorry, you are right-this should be std::copy after all

src/aktualizr_secondary/socket_server.cc
cajun-rat: Good point. Done

cajun-rat:
@lbonn @patrickvacek : updated

PR #808, opened 2018-06-05
Flatten the config dependency tree

lbonn:
Big config header files should only depend on other (dumb) config
definitions and general types.

This way, we make sure that interpendencies between modules are
explicit and not transitively inherited from an include to `config/config.h`


patrickvacek:
I think I'm in favor of this. Solves some problems I'd been peripherally aware of. Header multiplication is annoying but seems like a small cost to pay. What is the point of `EXCLUDE_FROM_ALL`, though? I don't really see what it's helping. Isn't `build_tests` still part of `all`?

cajun-rat:
Summary of our in-person chat: we want to use the default target to be the production executables, and at the moment Yocto is also building all the tests unnecessarily. `ninja build_tests` can be used to build all the tests if necessary.

lbonn:
On master, if I run `ninja`, most of the tests are built (it shows up 234 build steps). After that, `ninja build_tests` builds `aktualizr_uptane_vector_tests` `t_leak` and `t_utils`. It seems to be the tests which are still in `./tests`. So `all` definitely does not contain `build_tests`.

Before the big move of the unit tests, I think that `all` did not include any of the tests. It seems to have changed since then, though I don't know the exact reason why them being in `src/...` or `tests/...` changes this behaviour.

So in summary, this commit restores the old behaviour.

PR #811, opened 2018-06-07
Feat/implicit prov test

patrickvacek:


lbonn:
You've checked for similar config override bugs in other parts of the code base? Would be nice-to-have to abstract away this pattern of reading to an enum and warn for invalid values.

PR #814, opened 2018-06-11
Improv/better enums

lbonn:
Big but (hopefully) dumb.

Also, I've switched back to define a separate variable for the current sql schema version. I think I suggested removing it because of the duplication but in fact, it simplifies things a bit.


patrickvacek:
While personally I'm not particularly invested, for consistency, we should indeed follow the Google standard:

https://google.github.io/styleguide/cppguide.html#Enumerator_Names

So I think if we're going in and modernizing all of the variations we have, we might as well make them all consistent: `EnumClassName::kValueName`.

But otherwise, looks good!

src/sota_tools/authenticate.cc
```diff
@@ -8,17 +8,17 @@ using std::string;
 
 int authenticate(const string &cacerts, const ServerCredentials &creds, TreehubServer &treehub) {
   switch (creds.GetMethod()) {
-    case AUTH_BASIC: {
+    case AuthMethod::BASIC: {
```
patrickvacek: Sorry to keep bugging, but can we rename these all-caps ones, too?

lbonn: So, `AuthMethod::kBasic`?

patrickvacek: Yeah, I think that's best.

PR #815, opened 2018-06-11
Suppress a Boost/clang-tidy warning

cajun-rat:


src/implicit_writer/main.cc
```diff
@@ -72,6 +72,12 @@ int main(int argc, char *argv[]) {
   logger_init();
   logger_set_threshold(static_cast<boost::log::trivial::severity_level>(2));
 
+  // Suppress the following clang-tidy warning:
+  // /usr/include/boost/detail/basic_pointerbuf.hpp
+  // error: Call to virtual function during construction
+  // [clang-analyzer-optin.cplusplus.VirtualCall,-warnings-as-errors]
+  // basic_pointerbuf() : base_type() { setbuf(0, 0); }
+  // NOLINTNEXTLINE(clang-analyzer-optin.cplusplus.VirtualCall)
```
patrickvacek: The problem is that `parse_options` calls a virtual function in its constructor? Or what's the problem? I don't think I understand where this is coming from.

cajun-rat: I haven't traced the cause all the way back, but the final error message comes from boost's `basic_pointerbuf`. I think it is coming from the boost `variables_map` copy constructor.

PR #818, opened 2018-06-13
Split update to steps

patriotyk:


cajun-rat:
Hi Serhiy,

This is looking good.  My biggest concern is the interaction between eventinterpreter and sotauptaneclient that causes 'Command' and 'Event' messages to bounce between them makes reasoning about the behaviour hard. This is an artefact of the old system design, and isn't caused by this PR. It would be good to include a diagram (in Inkscape/SVG or dotty) that shows which events can cause with commands and which commands cause which events.

I also had a couple of minor things:

* BaseCommand really should be using virtual member functions for things that are overridden, like toJson(). It works at the moment, but seems to be mostly due to luck.
* The `RunningMode` enum should be made into a C++11 class enum.  @lbonn changed all the other ones over in the last couple of days, so I know this is a 'new thing'
* The StartDownload command doesn't appear to generate an event


I know this is a work in progress--it is looking good so far!

src/libaktualizr/primary/commands.cc
```diff
@@ -37,61 +34,45 @@ std::shared_ptr<BaseCommand> BaseCommand::fromPicoJson(const picojson::value& js
...
-}
-
-AbortDownload AbortDownload::fromJson(const std::string& json_str) {
-  Json::Reader reader;
-  Json::Value json;
-  reader.parse(json_str, json);
-  return AbortDownload(json["fields"][0].asString());
+  std::vector<Uptane::Target> updates;
+  for (auto it = json["fields"][0].begin(); it != json["fields"][0].end(); ++it) {
+    std::cout << "KEY: " << it.key().asString() << "\n";
```
lbonn: debug print

src/libaktualizr/primary/commands.cc
```diff
@@ -4,11 +4,17 @@
...
   json["variant"] = variant;
-  json["fields"] = Json::Value(Json::arrayValue);
-  return json;
+  if (!json.isMember("fields")) {
+    json["fields"] = Json::Value(Json::arrayValue);
+  }
+  return Json::FastWriter().write(json);
 }
 
 std::shared_ptr<BaseCommand> BaseCommand::fromPicoJson(const picojson::value& json) {
```
lbonn: Can wait but it would be nice to remove picojson and only use one library. Now there is this weird mix in `commands.cc`. @OYTIS

patriotyk: Unfortunatelly  we cannot remove it because it supports stream parsing which is needed for socket gateways.

patriotyk: But do we really need this socket gateway functionality?

patriotyk:
Currently `oe-selftest --run-tests updater` fails because of #816  PR. We removed some commnad line options but didn't change meta-updater tests.

cajun-rat:
@patriotyk I actually have a fix for that bug on my machine, but I hit some other failing tests and never managed to open a PR for it.  Let me do that now.

patrickvacek:
> Currently oe-selftest --run-tests updater fails because of #816 PR.

Yes, sorry, I have branches open for fixing that, but I still haven't gotten to the bottom of one of the HSM issues, so things are still a bit broken in oe-selftest (on sumo/master). My rocko PR should be good to go, though.

cajun-rat:
@patriotyk I actually have a fix for that bug on my machine, but I hit some other failing tests and never managed to open a PR for it.  Let me do that now.

patriotyk:
@cajun-rat  Just added aktualizr_flow.svg to docs folder

cajun-rat:
This looks great!  I ran a manual test and oe-selftest, and it all checks out.

PR #820, opened 2018-06-14
Fix/use getaddrinfo

lbonn:


cajun-rat:
Looks good. Manual primary/secondary provisioning test worked OK.

PR #821, opened 2018-06-14
Make UPTANE metadata verification in line with the spec

OYTIS:
It is probably not working, because I haven't yet rearranged the testing scaffolding, but I'd like to get Eric's (@heartsucker ) review sooner rather than later.


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -388,22 +638,74 @@ void SotaUptaneClient::verifySecondaries() {
...
+  if (sec_root_version >= 0) {
+    for (int v = sec_root_version + 1; v <= last_root_version; v++) {
+      std::string root;
+      if (!storage->loadRole(&root, repo, Uptane::Role::Root(), Uptane::Version(v))) {
+        LOG_WARNING << "Couldn't find root meta in the storage, trying remote repo";
+        if (!uptane_fetcher.fetchRole(&root, repo, Uptane::Role::Root(), Uptane::Version(v))) {
+          LOG_ERROR << "Root metadata could not be fetched, skipping to the next secondary";  // TODO: looks
+                                                                                              // problematic, robust
+                                                                                              // procedure needs to be
+                                                                                              // defined
```
patrickvacek: NBD, but this comment got formatted ugly, maybe move it to it's own line?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -185,47 +193,253 @@ bool SotaUptaneClient::initialize() {
...
+bool SotaUptaneClient::updateDirectorMeta() {
   // Uptane step 2 (download time) is not implemented yet.
   // Uptane step 3 (download metadata)
-  if (!uptane_fetcher.fetchMeta()) {
-    LOG_ERROR << "could not retrieve metadata";
-    *events_channel << std::make_shared<event::UptaneTimestampUpdated>();
-    return;
+
+  // reset director repo to initial state before starting UPTANE iteration
+  director_repo.resetMeta();
```
heartsucker: Why are we doing this?

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -340,7 +399,111 @@ bool SQLStorage::loadTlsPkey(std::string* pkey) {
...
+    LOG_ERROR << "Can't open database: " << db.errmsg();
+    return;
+  }
+
+  if (db.exec("DELETE FROM meta;", nullptr, nullptr) != SQLITE_OK) {
+    LOG_ERROR << "Can't clear metadata: " << db.errmsg();
+    return;
+  }
+}
+/*void SQLStorage::storeMetadataCommon(const Uptane::RawMetaPack& metadata, const std::string& tablename) {
```
lbonn: Should we remove all this commented code?

OYTIS: Yes, sure, I could have missed something.

src/aktualizr_info/main.cc
```diff
@@ -43,9 +43,14 @@ int main(int argc, char **argv) {
...
               << std::endl;
 
-    Uptane::RawMetaPack pack;
-
-    bool has_metadata = storage->loadMetadata(&pack);
+    std::string director_root;
+    std::string director_targets;
+    std::string images_root;
+    std::string images_targets;
+    bool has_metadata = storage->loadRole(&director_root, Uptane::RepositoryType::Director, Uptane::Role::Root());
```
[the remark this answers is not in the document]
OYTIS: aktualizr_info is used during debugging to get an overview of the state of the system. `has_meta` indicates that aktualizr has already downloaded some meta from the server, and this process normally doesn't fail partially. I.e. if you have the connection, no problems with TLS certificates and the server is up, you will have all of needed metadata, otherwise you will have none.
The same tool allows you to show individual pieces of metadata if you want more detail.

config/sql/migration/migrate.6.sql
```diff
@@ -0,0 +1,28 @@
...
+DROP TABLE rawmeta;
+DROP TABLE root_rawmeta;
+
+CREATE TABLE repo_types(repo INTEGER NOT NULL, repo_string TEXT NOT NULL);
+INSERT INTO repo_types VALUES (0, 'director'), (1, 'images');
+
+CREATE TABLE meta_types(meta INTEGER NOT NULL, meta_string TEXT NOT NULL);
+INSERT INTO meta_types VALUES (0, 'root'), (1, 'snapshot'), (2, 'targets'), (3, 'timestamp');
+
+CREATE TABLE meta_migrate(meta BLOB NOT NULL, repo INTEGER NOT NULL, meta_type INTEGER NOT NULL, version INTEGER NOT NULL);
```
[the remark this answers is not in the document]
OYTIS: Like on `repo`, `meta_type` and `version`?

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: Defensive programming probably. I think we can move what `resetMeta()` does with `root` to `initRoot` and remove the rest. Kind of wanted to be sure that we start each iteration from a clean state.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -86,22 +87,23 @@ bool AktualizrSecondary::putMetadataResp(const Uptane::RawMetaPack& meta_pack) {
       target_ = std_::make_unique<Uptane::Target>(*it);
     }
   }
-  storage_->storeMetadata(meta_pack);
+  storage_->storeRole(meta_pack.director_root, Uptane::RepositoryType::Director, Uptane::Role::Root(),
```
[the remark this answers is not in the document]
OYTIS: Right now implemenation is the same for both repos in both `FSStorage` and `SQLStorage`. What might be worth splitting is separating `storeRoot`/`loadRoot` from `storeNonRootRole`/`loadNonRootRole`, because we actually need versions only for root.

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: Or actually, maybe I'll just limit the lifetime of `director_repo` and `images_repo` to the scope of one iteration, it will look much better.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -10,6 +10,60 @@
...
+  if (db.get_rc() != SQLITE_OK) {
+    LOG_ERROR << "Can't open database: " << db.errmsg();
+    return;
+  }
+
+  if (!db.beginTransaction()) {
+    return;
+  }
+
+  request = SQLReqId::kGetTable;
```
lbonn: These two lines can go

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -352,188 +411,97 @@ void SQLStorage::storeMetadataCommon(const Uptane::RawMetaPack& metadata, const
...
-bool SQLStorage::loadMetadataCommon(Uptane::RawMetaPack* metadata, const std::string& tablename) {
+bool SQLStorage::loadRole(std::string* data, Uptane::RepositoryType repo, Uptane::Role role, Uptane::Version version) {
   SQLite3Guard db(config_.sqldb_path.c_str());
-
   if (db.get_rc() != SQLITE_OK) {
     LOG_ERROR << "Can't open database: " << db.errmsg();
     return false;
   }
 
   request = SQLReqId::kGetTable;
```
lbonn: old sql api again

patrickvacek:
I have not reviewed everything. But regarding the commented-out tests: for the most part, you are probably right, these are things that should be tested externally, as in, with the uptane vector tests. If you are confident that all of these cases are covered, then removing the redundant parts from uptane_test is fine. However, I'm pretty sure the `krejectallTest` is not covered elsewhere. That was specifically written to make sure we can recover from an ugly state that a customer reported. We hopefully have prevented that from ever happening again, but we probably still should try to make sure we can handle that case somehow.

OYTIS:
@patrickvacek Do you remember what the awful state was?

patrickvacek:
@OYTIS It was something about a tuf Root object getting constructed in an obscure and invalid way such that the policy was left as `kRejectAll`, meaning that nothing could be installed. This was caused presumably by some really unlikely timing with killing aktualizr or networking dropouts or something, such that bad objects were written to the database and were never overwritten. The git logs should have some more specifics.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -44,27 +47,21 @@ SotaUptaneClient::SotaUptaneClient(Config &config_in, std::shared_ptr<event::Cha
...
-  // LOG_ERROR << "Multiple secondaries found with the same serial: " << map_it->second->getSerial();
   return false;
 }
 
 std::vector<Uptane::Target> SotaUptaneClient::findForEcu(const std::vector<Uptane::Target> &targets,
                                                          const Uptane::EcuSerial &ecu_id) {
   std::vector<Uptane::Target> result;
   for (auto it = targets.begin(); it != targets.end(); ++it) {
-    if (it->ecu_identifier() == ecu_id) {
+    // DEBUG
```
lbonn: :warning:

OYTIS: Thanks!

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -124,13 +121,16 @@ void SotaUptaneClient::reportNetworkInfo() {
 
 Json::Value SotaUptaneClient::AssembleManifest() {
   Json::Value result;
-  Json::Value unsigned_ecu_version = pacman->getManifest(uptane_repo.getPrimaryEcuSerial());
+  installed_images.clear();
```
lbonn: If a secondary manifest verification fails, the corresponding entry in `installed_images` will be deleted at the end of this method. Is this expected behaviour? Or should it only modify the map if all verifications passed?

lbonn: Ok, great!

OYTIS:
@lbonn OK, looks like I've broken support for production server. Thank you for your help!

OYTIS:
Or is it something with the backend? I get HTTP 500 during provisioning.

src/libaktualizr/uptane/fetcher.h
```diff
@@ -30,12 +30,12 @@ class Fetcher {
  public:
   Fetcher(const Config& config_in, std::shared_ptr<INvStorage> storage_in, HttpInterface& http_in)
       : http(http_in), storage(std::move(storage_in)), config(config_in) {}
-  bool fetchMeta();
-  bool fetchRoot(bool director, Version version);
-  bool fetchTarget(const Target& target);
+  // bool fetchMeta();
+  // bool fetchRoot(bool director, Version version);
+  bool fetchVerifyTarget(const Target& target);
+  bool fetchRole(std::string* result, RepositoryType repo, Uptane::Role role, Version version = Version());
```
lbonn: API design remark: maybe not use an argument with default value here, it obfuscates a bit the part of the code that uses it (eg: `updateDirectorMeta` in `sotauptaneclient.cc`).

Defining a separate proxy method `fetchRoleLatest` can also help in this direction

OYTIS: @lbonn Thanks, I'll add it to my list for refactoring of refactoring (really small one!). Another thing it that invstorage will define two different functions to store root and non-root metadata (because we only need versioning for the former) and Uptane::DirectorRepository and Uptane::ImagesRepository will be local to uptaneIteration method (will make SotaUptaneClient constructor not wo crowded too).

OYTIS: No, looks like I can't make them local once we've separated fetching metadata from downloading images. Can make them class members anyway.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -188,47 +190,349 @@ bool SotaUptaneClient::updateMeta() {
...
+      auto hwid_it = hw_ids.find(ecu_serial);
+      if (hwid_it == hw_ids.end()) {
+        LOG_WARNING << "Unknown ECU ID in director targets metadata: " << ecu_serial.ToString();
+        is_new = false;
+        break;
+      }
+
+      if (hwid_it->second != hw_id) {
+        LOG_ERROR << "Wrong hardware identifier for ECU " << ecu_serial.ToString();
+        return false;
```
lbonn: minor, but `new_targets` still get modified if a target has an update and this condition is triggered in another loop iteration. This won't cause a problem in a regular use of this method (`new_targets` is most likely a fresh object) but it's a subtle breach of the function's contract.

We could start using `boost::optional<T>` as a return type for this kind of methods, when possible.

lbonn:
@OYTIS @patrickvacek Is the bug with kRejectAll, the one that surfaced in April: https://github.com/advancedtelematic/aktualizr/pull/695?

Other than the now missing tests, it looks good to me. We can add them back later if we feel they are needed.

patrickvacek:
@lbonn Yes, it was specifically https://github.com/advancedtelematic/aktualizr/pull/691.

cajun-rat:
I think we must wait for the matching n.root.json changes to be pushed to production before merging this.

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: Yes, it's true, I assumed that we can't trust contents of `new_targets` if the return value is false.
You're right about `boost::optional`, it seems to be around for quite a long time, so we shouldn't have compiler compatibility issues. As much as I'd prefer to just use the latest C++ standard.

OYTIS:
So kRejectAll test is back, and I also added download size/speed conditions, so I would dare to say that after it is merged PRO-4766 should be closed. Need Eric's review/confirmation on that though (probably independently of PR review process, more like acceptance criteria check on the story).

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: Let's do it in a separate PR, I'll want to change some interfaces in INvStorage in this way as well.

OYTIS:
Works fine with newly deployed prod. If someone has objections left (to the last commit for instance), please tell.

patrickvacek:
Good by me. Time to do this before another merge complaint shows up.

PR #823, opened 2018-06-15
Move config dir checking to a common function.

patrickvacek:
Also use it for the Config(vector<path>) constructor, which cert_provider uses. Previously, a nonexistent config would be silently ignored. Now it is as least not silent.

The idea is that if config files/dirs specified directly on the commandline do not exist, you should tell the user, because that probably matters.

This does mean that if you use this in meta-updater, the implicit prov test should fail just like the HSM test. This is intentional. It should fail if it can't find the config file. Currently it is sometimes using an empty config, which mostly works, but is maybe not a good idea.


lbonn:
Approved when tests are updated to take the new message format into account

patrickvacek:
> Approved when tests are updated to take the new message format into account

Oops, that's what a last minute change will do to you. Fixed.

patrickvacek:
Gotta love it when moving duplicated code into a common function makes coverage go down. If we ever want to increase our coverage, I think I know how...

PR #824, opened 2018-06-15
Refactor/less picojson

lbonn:
I ended up significantly refactoring `commands.cc` to make it more uniform, so that bugs pop up more easily.

It's a bit inheritance-y now, sorry :).


lbonn:
The CI failure looked like a Travis connectivity issue. Everything was fine on Jenkins though.

PR #827, opened 2018-06-20
Get rid of the old sql callback-based api

lbonn:
This will have some effect on #821 as well. I thought about trying to rebase it myself and put this PR on top but I think @OYTIS will be more qualified to do that without making mistakes.

If merging this now would cause too much problem, we can let that sit here until #821 is ready.

### DIFF
--- a/src/libaktualizr/storage/sql_utils.h
+++ b/src/libaktualizr/storage/sql_utils.h
@@ -38,6 +38,23 @@ class SQLiteStatement {
   inline sqlite3_stmt* get() const { return stmt_.get(); }
   inline int step() const { return sqlite3_step(stmt_.get()); }
 
+  // get results
+  inline std::string get_result_col_blob(int iCol) {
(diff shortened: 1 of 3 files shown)


patrickvacek:
I think this makes sense. It's slightly annoying that `get_result_col_str` requires sending in the exact column number, but that's probably not actually much worse than the former string matching system. Probably makes sense to sync with Anton before merging.

lbonn:
@patrickvacek I've replaced `SELECT *` instances with the list of column entries we need to read, to lessen the chance of mistakes with the index-based api.

Also, after checking with @OYTIS, he's ok with merging this PR first, since the rebase work is going to be pretty big anyway.

patrickvacek:
> I've replaced SELECT * instances with the list of column entries we need to read, to lessen the chance of mistakes with the index-based api.

Great, that makes sense to me.

PR #828, opened 2018-06-20
Document --running-mode=once

cajun-rat:


cajun-rat:
Will wait for ack from @tkfu

PR #829, opened 2018-06-20
Doc/dependencies update

patrickvacek:
Actually not much to do with docs in the end, mostly just updating dependency tracking and moving googletest to a submodule. We might want to do the same with the json tools, too.

Also not that the thirdparty.spdx is still fairly rough. A couple more items should probably be included (systemd and dkpg) and some of the copyright lines are questionable.


lbonn:
From a bit of CMake archeology, `CMAKE_EXPORT_COMPILE_COMMANDS` was documented pretty late but has been there for a long time:

https://cmake.org/gitweb?p=cmake.git;a=commitdiff;h=5674844de4b74d0b66cfc6b8237e631702c43637

`git tag --contains 5674844de4b74d0b66cfc6b8237e631702c43637` returns v2.8.5 as the earliest version.

Also, if not supported, setting the variable won't trigger an error for regular compilation jobs and only make `clang-xxx` tools fail when running them.

lbonn:
After further thought, advertising 3.5.1 as a minimum does not really hurt, since that's the version we're still testing with (shipped on Ubuntu Xenial). So, it should be guaranteed to work in the future by CI (at least compile and run tests).

Then, I don't think that it will be necessary to protect that in a `if`

patrickvacek:
I changed the minimum to 3.5, made some other improvements, and rebased. I think it is now ready.

PR #831, opened 2018-06-21
Document --running-mode=download/install

cajun-rat:


cajun-rat:
If you are going to actually try this out, then you need `require classes/sota_bleeding.inc` in your `local.conf`, because the code to do this only landed in aktualizr today.

docs/programatically-triggering-polling.adoc
```diff
@@ -20,3 +20,40 @@ With the background service disabled, trigger an update cycle by running the fol
 
     aktualizr --running-mode=once
 
+
+== Separately running individual update stages
+
+It is sometimes necessary to download an update when the is network connectivity then perform the installation later, without relying on the network. Note that aktualizr is always _safe_ against network drops. The worst that can happen if the network becomes unavailable is that the update will fail and need to be retried later.
```
patrickvacek: "when the is network connectivity" => "when there is network connectivity"

cajun-rat: Good catch. Fixed.

tkfu:
I wanted quite a few little changes to this, so I opened #832 to this branch rather than tacking on commits. @cajun-rat, if you're cool with that, feel free to merge/squash/merge.

PR #832, opened 2018-06-22
Docs/refactor/separate download

tkfu:


docs/programatically-triggering-polling.adoc
```diff
@@ -1,59 +0,0 @@
...
-    aktualizr --running-mode=download
-
-After this point the installation can proceed without network connectivity. If you are using a hardware device, unplug the cable. Under Qemu (for example using the  `meta-updater/scripts/run-qemu-ota --no-gui` script), the following commands on the Qemu console can be used to disable the network:
-
-    (qemu) info network
-    hub 0
-    \ hub0port1: e1000.0: index=0,type=nic,model=e1000,macaddr=ca:fe:6a:ee:98:fa
-    \ hub0port0: user.0: index=0,type=user,net=10.0.2.0,restrict=off
-    (qemu) set_link user.0 off
-    (qemu) set_link e1000.0 off
```
patrickvacek: This qemu bit might just be for testing and debugging, but might be worth keeping around even if just for that purpose. Could we still keep it as an appendix or something?

tkfu: I'm pretty reluctant to just scatter qemu documentation around inside . I suppose there's no great harm in leaving it; we'd just tag it and ignore it when we include in the ats garage docs...but still, it doesn't feel right. All it tells you is "this is how you can make sure it's working", which is only really useful if you're an aktualizr developer working on the feature.

patrickvacek: Fair point, most users won't care. Does it make sense to have a separate place to put things like that? Alternatively, we could just integrate it directly into an oe-selftest case. Some of those test cases were basically translated directly from existing documentation, too.

cajun-rat:
I'll squash the docs branch in a sec

PR #834, opened 2018-06-22
Fix bugs in root.json and snapshot.json formats

OYTIS:


cajun-rat:
Approved, assuming CI passes

PR #839, opened 2018-07-03
Fix images metadata update process

OYTIS:
Remove "optimization" of UPTANE traffic, resulting in decreased robustness of update process.

It will not fix the problem with `snapshot.json` hash mismatch, but will fix one that prevented us from debugging it.


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -425,16 +423,16 @@ bool SotaUptaneClient::updateImagesMeta() {
...
-    }
     if (!images_repo.verifyTargets(images_targets)) {
       last_exception = images_repo.getLastException();
       return false;
     }
-    storage->storeNonRoot(images_targets, Uptane::RepositoryType::Images, Uptane::Role::Targets());
+
+    if (local_version > remote_version) {
+      return false;
+    } else if (local_version == remote_version) {
```
patrickvacek: Why is this case an `==` but the others are all `<`?

patrickvacek:
Is there a way we can test that this is behaving differently/more correctly than before?

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: That's an error, yes, thanks.

OYTIS:
@patrickvacek 
Yes, we need a server that would send errors on certain points and then restore normal operation.

OYTIS:
@patrickvacek OK, so now there is a test for that (and `http_fake.cc` has got a bit more convoluted), and also fixed an actual bug, which might be the reason for the failed hash verification.

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -1093,6 +1094,63 @@ TEST(Uptane, krejectallTest) {
...
+  // 1st attempt, don't get anything
+  EXPECT_FALSE(sota_client.uptaneIteration());
+  EXPECT_FALSE(storage->loadLatestRoot(nullptr, Uptane::RepositoryType::Director));
+
+  // 2nd attempt, get director root.json
+  EXPECT_FALSE(sota_client.uptaneIteration());
+  EXPECT_TRUE(storage->loadLatestRoot(nullptr, Uptane::RepositoryType::Director));
+  EXPECT_FALSE(storage->loadNonRoot(nullptr, Uptane::RepositoryType::Director, Uptane::Role::Targets()));
+
+  // 4th attempt, get director targets.json
```
lbonn: probably typo, but why no 3rd?

OYTIS:
Added commit that might solve the manual check/download/install problem. In the automatic case, there's a redundant read/verification of metadata, but we're not chasing the best performance, do we? ;)

patrickvacek:
After booting and `aktualizr --running-mode=check`, I'm getting this:
```
Provisioned successfully on Device Gateway
Can't get public key: unknown error
...
Error registering device on Uptane, response: 
ECU registration failed, abort initialization
Fatal error of tls or ecu device registration
```

PR #840, opened 2018-07-04
Bugfix/pkcs11 config correctness

patrickvacek:
If only we'd heeded @lbonn's advice in https://github.com/advancedtelematic/aktualizr/pull/811...


src/libaktualizr/config/config.cc
```diff
@@ -12,10 +12,49 @@
 
 #include "bootstrap/bootstrap.h"
 #include "config.h"
-#include "utilities/config_utils.h"
 #include "utilities/exceptions.h"
 #include "utilities/utils.h"
 
+template <>
+inline void CopyFromConfig(KeyType& dest, const std::string& option_name, const boost::property_tree::ptree& pt) {
```
lbonn: Careful here:

https://github.com/advancedtelematic/aktualizr/blob/18ea7709998f2c389b09ebcebf93fd9628cfbe35/src/aktualizr_secondary/aktualizr_secondary_config.cc#L26

Should it go in a header file somewhere?

patrickvacek: Good catch! It may indeed belong in a header.

lbonn: Ideally it would be close to the definition of `KeyType` but since it's a member of `types.h` which is at the origin of everything, I don't know. Maybe it's fine anyway since it only depends on boost and the standard library.

patrickvacek:
@lbonn Nah, I can fix that here, too. I'm pretty sure this still won't pass CI anyway; there's some PKCS#11/SSL interaction that isn't working.

patrickvacek:
I cheated by silencing the broken test and we'll have to come back to it later. Other priorities are calling, so I'm merging now.

PR #842, opened 2018-07-06
Fix build on Mac OS X

cajun-rat:


src/libaktualizr/asn1/asn1_message.cc
```diff
@@ -29,7 +35,7 @@ int Asn1SocketWriteCallback(const void* buffer, size_t size, void* priv) {
 
   while (len > 0) {
     // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
-    ssize_t written = send(*sock, b + pos, len, MSG_NOSIGNAL);
+    ssize_t written = send(*sock, b + pos, len, flags_no_signal);
```
jochenschneider: You can also ignore SIGPIPE like this. I don't know if that is needed.

    int val = 1;
    setsockopt(*sock, SOL_SOCKET, SO_NOSIGPIPE, (void*)&val, sizeof(val));

src/aktualizr_secondary/socket_server.cc
```diff
@@ -7,7 +7,7 @@
 #include "utilities/dequeue_buffer.h"
 #include "utilities/sockaddr_io.h"
 
-#include <linux/tcp.h>
+#include <netinet/tcp.h>
```
jochenschneider: Does this work on Linux as well? That would be cool.

cajun-rat: It looks like it.  The CI failures were because of a clang-format problem, so hopefully this will pass.

lbonn:
Superseeded by #843 (already merged)

PR #843, opened 2018-07-06
Compile on Mac

jochenschneider:


CMakeLists.txt
```diff
@@ -16,7 +16,9 @@ option(BUILD_P11 "Support for key storage in a HSM via PKCS#11" OFF)
 option(BUILD_SOTA_TOOLS "Set to ON to build SOTA tools" OFF)
 option(BUILD_ISOTP "Set to ON to build ISO-TP" OFF)
 option(BUILD_OPCUA "Set to ON to compile with OPC-UA protocol support" OFF)
-option(BUILD_SYSTEMD "Set to ON to compile with systemd additional support" ON)
+if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
+    option(BUILD_SYSTEMD "Set to ON to compile with systemd additional support" ON)
+endif()
```
cajun-rat: Can we do this via autodetection?  Then it will work on Linux systems without libsystemd-dev installed?

docs/debugging-tips.adoc
```diff
@@ -6,7 +6,7 @@ The `sota_local.toml` configuration file sets the package manager to `PackageMan
 
     # From the build directory
     mkdir sota-prov
-    src/aktualizr --config ../config/sota_local.toml
+    src/aktualizr_primary/aktualizr --config ../config/sota_local.toml
```
cajun-rat: Good spot!

CMakeLists.txt
jochenschneider: Yeah that would be nice. I couldn't quite figure out how to do that, but I will give it another go. `find_library` maybe?

patrickvacek:
This looks good. Is it possible and/or worthwhile to have a Dockerfile/testcase for compiling on a mac? I fear this will get broken rather quickly if not.

jochenschneider:
Is there a way to easily cross-compile to a Mac? A start would be to at least compile it with clang.

patrickvacek:
> Is there a way to easily cross-compile to a Mac?

Easy, no, but some initial research leads me to believe it may be possible. Debatable whether it is worth doing.



> A start would be to at least compile it with clang.

Agreed. Maybe we should change one of our existing Dockerfiles to compile with clang. That should be reasonable. Don't worry about that for this PR, though!

PR #844, opened 2018-07-09
Compile with clang on Debian testing

lbonn:
So that it doesn't break in the future

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
 - DOCKERFILE=Dockerfile.nop11 CACHE=latest SCRIPT=src/scripts/test.sh
   DARGS="-eTEST_TESTSUITE_ONLY=crypto"
 - DOCKERFILE=Dockerfile.deb-testing CACHE=latest-deb-unstable SCRIPT=src/scripts/test.sh
-  DARGS="-eTEST_WITH_LOAD_TESTS=1 -eTEST_WITH_TESTSUITE=0 -eTEST_WITH_STATICTESTS=1"
+  DARGS="-eTEST_CC=clang -eTEST_WITH_LOAD_TESTS=1 -eTEST_WITH_TESTSUITE=0 -eTEST_WITH_STATICTESTS=1"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -11,7 +11,7 @@ env:
 - DOCKERFILE=Dockerfile.nop11 CACHE=latest SCRIPT=src/scripts/test.sh
   DARGS="-eTEST_TESTSUITE_ONLY=crypto"
 - DOCKERFILE=Dockerfile.deb-testing CACHE=latest-deb-unstable SCRIPT=src/scripts/test.sh
-  DARGS="-eTEST_WITH_LOAD_TESTS=1 -eTEST_WITH_TESTSUITE=0 -eTEST_WITH_STATICTESTS=1"
+  DARGS="-eTEST_CC=clang -eTEST_WITH_LOAD_TESTS=1 -eTEST_WITH_TESTSUITE=0 -eTEST_WITH_STATICTESTS=1"
```
jochenschneider: And that docker image has both clang and gcc installed?

jochenschneider: Ah yeah it is using clang for code checking.

scripts/test.sh
```diff
@@ -97,9 +98,16 @@ fi
...
+    if [[ $TEST_CC = gcc ]]; then
+        export CC=gcc CXX=g++
+    elif [[ $TEST_CC = clang ]]; then
+        export CC=clang CXX=clang++
+    fi
+
     set -x
     cmake "${CMAKE_ARGS[@]}" "${GITREPO_ROOT}" || add_fatal_failure "cmake configure"
-    set +x
+    )
```
patrickvacek: Intentional removal of `set +x`?

lbonn: I set it to spawn a sub-shell (with parentheses), so that both variables are only exported inside it. It also applies to shell options, so the effect of `set -x` is contained as well.

patrickvacek: Ah, I see now. That's fine, then.

PR #845, opened 2018-07-09
Fix BUILD_SYSTEMD=on option

lbonn:
Defaults to ON on Linux, OFF on other platforms

@jochenschneider: could you make a quick test on MacOS? (should configure successfully without setting `BUILD_SYSTEMD`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,12 @@ cmake_policy(SET CMP0009 NEW)
 
 project(aktualizr)
 
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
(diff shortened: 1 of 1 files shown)


patrickvacek:
A bit ugly but this looks right to me.

lbonn:
If we feel that it's not worth the trouble, defaulting to OFF in all cases can also be fine (it would need a quick update in meta-updater). I don't have a strong opinion about this.

patrickvacek:
> If we feel that it's not worth the trouble, defaulting to OFF in all cases can also be fine (it would need a quick update in meta-updater). I don't have a strong opinion about this.

Nor do I. Let's do this for now since it fixes the problem and isn't really that bad.

PR #846, opened 2018-07-09
Feat/manual installation result

patrickvacek:
This stores the operation_result that was previously just a member variable. This means that manual aktualizr control should still send it in the manifest after an install. @rdanitz can you please verify that this fixes the manual control e2e test?

### DIFF
--- a/src/libaktualizr/storage/invstorage.cc
+++ b/src/libaktualizr/storage/invstorage.cc
@@ -1,8 +1,10 @@
 #include "invstorage.h"
-#include "fsstorage.h"
-#include "sqlstorage.h"
 
+#include <unistd.h>
+
+#include "fsstorage.h"
 #include "logging/logging.h"
+#include "sqlstorage.h"
 #include "utilities/utils.h"
 
 std::ostream& operator<<(std::ostream& os, const StorageType stype) {
@@ -129,7 +131,12 @@ void INvStorage::importData(const ImportConfig& import_config) {
 std::shared_ptr<INvStorage> INvStorage::newStorage(const StorageConfig& config, const boost::filesystem::path& path) {
   switch (config.type) {
     case StorageType::kSqlite:
-      if (!boost::filesystem::exists(config.sqldb_path)) {
+      if (!boost::filesystem::exists(config.sqldb_path) && boost::filesystem::exists(path)) {
+        if (access(path.c_str(), R_OK | W_OK | X_OK) != 0) {
+          LOG_ERROR << "Cannot read prior filesystem configuration from " << path
+                    << " due to insufficient permissions.";
+          return std::make_shared<SQLStorage>(config);
+        }
         StorageConfig old_config;
         old_config.type = StorageType::kFileSystem;
         old_config.path = path;
--- a/src/libaktualizr/storage/sqlstorage.cc
+++ b/src/libaktualizr/storage/sqlstorage.cc
@@ -953,7 +953,7 @@ void SQLStorage::clearInstalledVersions() {
   }
 }
 
-void SQLStorage::storeInstallationResult(const std::string& installation_result) {
+void SQLStorage::storeInstallationResult(const data::OperationResult& result) {
   SQLite3Guard db(config_.sqldb_path.c_str());
 
   if (db.get_rc() != SQLITE_OK) {
@@ -961,15 +961,16 @@ void SQLStorage::storeInstallationResult(const std::string& installation_result)
     return;
   }
 
-  auto statement =
-      db.prepareStatement<std::string>("UPDATE OR REPLACE installation_result SET result = ?;", installation_result);
+  auto statement = db.prepareStatement<std::string, int, std::string>(
+      "UPDATE OR REPLACE installation_result SET (id, result_code, result_text) = (?,?,?);", result.id,
+      static_cast<int>(result.result_code), result.result_text);
   if (statement.step() != SQLITE_DONE) {
     LOG_ERROR << "Can't set installation_result: " << db.errmsg();
     return;
   }
 }
 
-bool SQLStorage::loadInstallationResult(std::string* installation_result) {
+bool SQLStorage::loadInstallationResult(data::OperationResult* result) {
   SQLite3Guard db(config_.sqldb_path.c_str());
 
   if (db.get_rc() != SQLITE_OK) {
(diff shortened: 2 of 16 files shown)


src/libaktualizr/storage/invstorage.cc
```diff
@@ -129,7 +131,11 @@ void INvStorage::importData(const ImportConfig& import_config) {
 std::shared_ptr<INvStorage> INvStorage::newStorage(const StorageConfig& config, const boost::filesystem::path& path) {
   switch (config.type) {
     case StorageType::kSqlite:
-      if (!boost::filesystem::exists(config.sqldb_path)) {
+      if (!boost::filesystem::exists(config.sqldb_path) && boost::filesystem::exists(path)) {
+        if (access(path.c_str(), R_OK & W_OK & X_OK) == 0) {
```
lbonn: `R_OK | W_OK | X_OK`?

lbonn: `R_OK | W_OK | X_OK`?

src/libaktualizr/storage/sqlstorage.h
```diff
@@ -59,6 +59,10 @@ class SQLStorage : public INvStorage {
                               const std::string& current_hash) override;
   std::string loadInstalledVersions(std::vector<Uptane::Target>* installed_versions) override;
   void clearInstalledVersions() override;
+  void storeInstallationResult(const std::string& installation_result) override;
```
lbonn: How about storing and getting `data::OperationResult`, so that the json conversions (in storage implementations) are guaranteed to be compatible with each others instead of relying on api user code?

patrickvacek: So you mean store each of the constituent elements of `data::OperationResult` instead of the serialized json? Not a bad idea. I'm not sure I understand what you mean about compatibility and api user code, though.

patrickvacek: Thinking about this more, I like the idea of storing actual types instead of a json string, but it gets annoying for the old filesystem storage. I'm probably going to have to store three separate files with each of the members of `data::OperationResult`. I would love it if we could get rid of filesystem storage!

lbonn: What I meant is that passing `std::string` around is close to being untyped, users of the api need to construct the correct json object manually and nothing prevents them to go out of sync.

If this is impractical, nevermind. About `FSStorage`, I also think it's worth drafting a deprecation plan and would gladly start it.

patrickvacek: I have a plan for the `data::OperationResult` storage. I think we are on the same page now.

Deprecating `FSStorage` might be impractical, but we could at least move to a state where we do not allow it as a valid config option anymore. We probably still have to maintain the existing logic to be able to convert from `FSStorage` to `SQLStorage`, though. But at least we could stop implementing new features for it. The first step of the process is converting the default implicit and HSM provisioning configs to use `SQLStorage` as well as any tests that aren't specifically designed to test `FSStorage`.

OYTIS:
Looks good. I was about to say that we need to clear the result once it was successfully sent to the backend, but apparently we haven't been doing it before, and it worked well.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -953,6 +953,72 @@ void SQLStorage::clearInstalledVersions() {
...
 
+void SQLStorage::storeInstallationResult(const data::OperationResult& result) {
+  SQLite3Guard db(config_.sqldb_path.c_str());
+
+  if (db.get_rc() != SQLITE_OK) {
+    LOG_ERROR << "Can't open database: " << db.errmsg();
+    return;
+  }
+
+  auto statement = db.prepareStatement<std::string>(
```
lbonn: annoying, the extra arguments compared to the template parameters is not caught at compile time. Only at run time, and not with every sqlite version!

patrickvacek: Foolish mistake, but yes, very strange that it compiled and all the tests passed on my machine!

lbonn: Yes, there is something with the template argument deduction I suspect. Probably not an issue with sqlite version, like I suspected, but something like different gcc interpretations depending on version...

config/sql/migration/migrate.7.sql
```diff
@@ -0,0 +1,9 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+BEGIN TRANSACTION;
+
+CREATE TABLE installation_result(id TEXT, result_code INTEGER NOT NULL DEFAULT 0, result_text TEXT);
```
lbonn: The change to `schema.sql` is also missing

patrickvacek: Fixed. Sorry, forgot about this file. Should we maybe have an automated test that the final schema after running through the migrations matches the expected schema?

patrickvacek:
> Hum ok actually the C++ part worked fine, this is about sqlite. Maybe it needs to do an INSERT when the table is empty.

I don't think I understand what is wrong at this point. There is an INSERT in `dbInit` to get things going, modeled after how the `device_info` table is treated. What am I doing differently, other than using multiple columns in the table?

patrickvacek:
Ran out of time today. I think I'll move the installation_result to the style of the primary_keys instead of the style of device_id. device_id gets initialized at launch and there always must be exactly one entry. The primary_keys can be empty or have one entry. I think that makes more sense here. Maybe that will help uncover what is wrong with the current logic.

rdanitz:
I still get on manual aktualizr e2e-test:
```
sqlite> select * from installation_result;
qemux86-64-5dee2602cd9254f4180a23927e5a604e0d149f67c68a0a672c9ed97c5aad3bb9|1|Package already installed
```

lbonn:
Rebased on #853 + made some of the planned changes. I'll look more into your issue @rdanitz

lbonn:
Source of the issue with CI: https://www.sqlite.org/lang_update.html

> Beginning in SQLite version 3.15.0 (2016-10-14), an assignment in the SET clause can be a parenthesized list of column names on the left and a row value of the same size on the right.

(xenial ships with sqlite 3.11)

We also use the same syntax for `primary_keys`, so it means that we didn't test enough.

The issue reported by @rdanitz was not actually coming from aktualizr's side and has been resolved.

So it should be ready, I'll open another PR for the `primary_keys` issue.

patrickvacek:
@lbonn Looks good to me, although can you explain the `unique_mark` bit? And what was the problem on the e2e side?

If you and @rdanitz are happy with this now, we should merge this.

lbonn:
It's a dummy column. The two constraints `unique_mark = 0` and `primary key` enforce that there is at most one row. Then, when doing `INSERT OR REPLACE`, it triggers a constraint violation which will cause the old row to be deleted and replaced by the new.

The catch is that it only works if all values are updated: if the insert modifies only a subset of the values, the others are taken from the default and not the old content.

patrickvacek:
Thanks for the clarification! Sounds like Robert is happy with this, so I'm merging.

PR #852, opened 2018-07-11
[RFC] Initial draft of external API

cajun-rat:


src/libaktualizr/aktualizr_api.h
```diff
@@ -0,0 +1,241 @@
...
+
+1. Integration with 3rd party HMI
+2. Integration with 3rd party interfaces to install software on secondary ECUs
+3. Limit network trafic and installation to specific vehicle states
+4. Progress bars
+
+Non-functional Requirements
+===========================
+
+1. Since these interfaces are often proprietry and under NDA, the implementation must be kept separate from the
```
patrickvacek: proprietary

src/libaktualizr/aktualizr_api.h
```diff
@@ -0,0 +1,241 @@
...
+- Have to build bridges to connect this 'common' IPC mechanism to whatever the end-user system has picked
+- New features may require simultanous changes to the aktualizr IPC these bridges.
+- Hard to replace built-in functionality with external functionality (e.g. Package Managers)
+- Harder to completely replace the high-level workflow.
+
+It would still be possible to write a wrapper that would expose libaktualizr
+via an IPC mechanism, but this can be written with a specific use-case and IPC
+technology in mind.
+
+\nosubgrouping
```
patrickvacek: What is the significance of this "\nosubgrouping"? (And the "\file" at the top?)

src/libaktualizr/aktualizr_api.h
```diff
@@ -0,0 +1,241 @@
...
+ * \param aktualizr Aktualizr handle
+ * \param callback The callback to call, or 'NULL' to disable notifications
+ * \param userdata Opaque user handle that is passed to the callback
+ */
+void aktualizr_set_state_change_callback(aktualizr_t aktualizr, aktualizr_callback_t callback, void* userdata);
+
+/**
+ * The current state of the update process.
+ */
+aktualizr_update_state_t aktualizr_get_update_state(aktualizr_t aktualizr);
```
doanac: Maybe this should also return a "status" value somehow. eg
~~~
 aktualizr_update_state_t aktualizr_get_update_state(aktualizr_t aktualizr, char **state_status);
~~~
Where you could then do something like:
~~~
aktualizr_update_state_t state;
char *status=NULL
state = aktualizr_get_update_state(aktualizr, &status);
if (state == AKTUALIZR_UPDATE_DOWNLOADING && status) {
    printf("Downloading update. %s\n", status);
    free(status);
}
~~~

src/libaktualizr/aktualizr_api.h
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * Set a single string configuration option.
+ */
+aktualizr_res_t aktualizr_config_set_string(aktualizr_config_t config, char* section, char* option, char* value);
+
+/**
+ * Set a single integer configuration option.
+ */
+aktualizr_res_t aktualizr_config_set_int(aktualizr_config_t config, char* section, char* option, int value);
```
patrickvacek: Is this intended for runtime usage? Seems dangerous to change a running system. We currently don't support anything like this. What is the goal?

src/libaktualizr/aktualizr_api.h
```diff
@@ -0,0 +1,241 @@
...
+void aktualizr_config_free(aktualizr_config_t config);
+
+/**
+ * Load configuration from a file or directory of configuration fragments
+ */
+aktualizr_res_t aktualizr_config_load(aktualizr_config_t config, char* path);
+
+/**
+ * Set log level.
+ * \todo Log to places other than stdout (e.g. systemd journal, callbacks)
```
patrickvacek: Should this have a file handle or a pointer for logging to some object or location? I agree that only allowing logging to stdout is not a good idea.

src/libaktualizr/aktualizr_api.h
cajun-rat: I wanted to make it easy to override small bits of config without either exposing a full set of `aktualizr_config_set_<bla>` operations, or forcing the user to write a temporary file.  The configuration would be baked in after it is handed to `aktualizr_init()`

src/libaktualizr/aktualizr_api.h
cajun-rat: Yep, eventually.

cajun-rat: The other problem we have with logging is that the logging configuration is currently evil global state, and would need to be bound to a specific aktualizr object.

src/libaktualizr/aktualizr_api.h
cajun-rat: That's a reasonable idea. Free-ing a pointer allocated on the other side of an API boundary is a bit dangerous (we'd have to expose a `aktualizr_free()` call).

Maybe the API should be:

    void aktualizr_get_update_state(aktualizr_t, aktualizr_update_state_t* char**);

doanac: I like that better.

src/libaktualizr/aktualizr_api.h
cajun-rat: Thanks!

src/libaktualizr/aktualizr_api.h
cajun-rat: It appears that `\nosubgrouping` isn't actually needed anymore, I'll remove that.

src/libaktualizr/aktualizr_api.h
cajun-rat: Thinking some more about my proposed API, I'm a bit worried about _just_ having a text string for the information.  It seems to me like we might need a 3rd piece of information soon, and adding a `aktualizr_get_update_icon` call would open up race conditions between the calls, because the aktualizr engine is running in a background thread.  

Maybe `aktualizr_update_state_t` needs to be a full blown object that represents a snapshot of the current state, and can be queried at will.

lbonn: @cajun-rat is it `void aktualizr_get_update_state(aktualizr_t, aktualizr_update_state_t*, char**);`? I don't understand how it would solve the allocation problem.

Alternatively, `aktualizr_update_state_t` could be a struct with a `char status[MAX_STATUS_LENGTH]` member so that heap allocation can be avoided entirely (with the signature `void aktualizr_get_update_state(aktualizr_t, aktualizr_update_state_t*)` )

src/libaktualizr/aktualizr_api.h
```diff
@@ -0,0 +1,241 @@
...
+ * an \ref AKTUALIZR_ALREADY_IN_PROGRESS error is returned.
+ */
+aktualizr_res_t aktualizr_run_update(aktualizr_t aktualizr, aktualizr_poll_operation_t operation);
+
+/**
+ * Report system information to the server.
+ * This includes network, hardware and installed package information.
+ * \todo Async? Queued?
+ */
+aktualizr_res_t aktualizr_report_system_information(aktualizr_t aktualizr);
```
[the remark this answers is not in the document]
cajun-rat: Good point. I'm going to nuke this API.

src/libaktualizr/aktualizr_api.h
doanac: I like the `char status[MAX_STATUS_LENGTH]` approach. Then a user could do something really simple like:
~~~
aktualizr_update_state_t update_state;

/* repeatable safe calls to */
aktualizr_get_update_state(aktualizr, &update_state)
~~~

cajun-rat: My worry there was with ABI compatibility. If we went down that route then the size of `aktualizr_update_state_t` would be fixed. We can include a handful of padding fields that can be re-purposed for small changes, and define aktualizr_get_update_state_ex in the future to support more detailed information if needed.

Making it much harder for users to leak memory is certainly a valuable benefit of this approach.

doanac: Maybe we are looking at this the wrong way. Polling for update status is actually a bit of a pain anyway. The caller's going to need some potentially thread-based logic to try and understand what's going on and will wind up polling at some random rate. Couldn't we get rid of this API function, and just make the `void aktualizr_set_state_change_callback(aktualizr_t aktualizr, aktualizr_callback_t callback, void* userdata);` API include this kind of additional information?

cajun-rat: The idea for a polling/query interface was to allow simple command line applications or for UI initialization, and I'd expect that most applications would use `aktualizr_set_state_change_callback` rather than poll `aktualizr_get_update_state`. If the polling interface can be emulated with the callback interface, then that would be a good argument for dropping it.

In terms of what data is available, the callback should receive the same information that can be fetched from  `aktualizr_get_update_state`:

    typedef void (*aktualizr_callback_t)(void* clientp, aktualizr_update_state_t);
    void aktualizr_set_state_change_callback(aktualizr_t aktualizr, aktualizr_callback_t callback, void* userdata);

(In the last update I made `aktualizr_update_state_t` an object rather than a enum)

cajun-rat:
I've updated the API to expose `struct aktualizr_update_state` directly. I've included some 'padding' fields that can be re-purposed to add new information without breaking ABI. If we need more information in the future we can add a `aktualizr_set_state_change_callback_ex` function.

src/libaktualizr/aktualizr_api.h
```diff
@@ -151,6 +159,36 @@ typedef enum {
...
+/**
+ * A snapshot of the current update state
+ */
+struct aktualizr_update_state {
+  /**
+   * The high-level category of this update state snapshot.
+   */
+  aktualizr_update_ui_category_t ui_category;
+
+  /**
```
lbonn: Alternative: just put a big `char[]` (thoughtfully sized) and insert new fields there in new api versions while reducing the array's size. The only constraint to satisfy would be that `sizeof(struct aktualizr_update_state)` will stay constant and that old fields are not moved. It can be checked with a C11 `static_assert(sizeof(xxx) == 64)`

Also, maybe we could expose an api version just in case.

src/libaktualizr/aktualizr_api.h
```diff
@@ -0,0 +1,298 @@
...
+typedef aktualizr_config_struct* aktualizr_config_t;
+
+/**
+ * \aktualizr API result code.
+ */
+typedef enum {
+  /** The operation completed successfully */
+  AKTUALIZR_OK = 0,
+  /** The operation cannot be started because an existing operation is in progress */
+  AKTUALIZR_ALREADY_IN_PROGRESS,
```
lbonn: As this is an external api, setting all enum values explicitly would be more reassuring.

src/libaktualizr/aktualizr_api.h
doanac: The latest version works for me so ignore the rest of my comment if you want :) I was just thinking removing the polling interface might be nice because you wouldn't have to worry about struct sizes. ie - if there's just a callback, then the aktualizr engine owns the memory in the struct and the user has nothing to worry about (unless they want to *copy* data from it). I'm not totally sure about ABI compatibility, but I think you could also then just append members to the struct w/o fear of breaking old code.

patrickvacek:
Shall we close this, or do you want to keep this open for further work?

PR #853, opened 2018-07-11
Add automated test for sqlite migration consistency

lbonn:
`schema.sql` should match with the result of migrations from the first version of the schema

Note that it fails right now, it would need these lines to be added to `schema.sql`:

```
INSERT INTO meta_types(rowid,meta,meta_string) VALUES(1,0,'root');
INSERT INTO meta_types(rowid,meta,meta_string) VALUES(2,1,'snapshot');
INSERT INTO meta_types(rowid,meta,meta_string) VALUES(3,2,'targets');
INSERT INTO meta_types(rowid,meta,meta_string) VALUES(4,3,'timestamp');
INSERT INTO repo_types(rowid,repo,repo_string) VALUES(1,0,'images');
INSERT INTO repo_types(rowid,repo,repo_string) VALUES(2,1,'director');
INSERT INTO version(rowid,version) VALUES(1,6);
```

Currently `schema.sql` only create tables but I don't see why it could not insert values.


patrickvacek:
Looks promising, and yes, I don't see why we couldn't insert values in `schema.sql`.

lbonn:
My attempt to use some packages' bionic version turned out to cause some valgrind issues. I'll try to stick to artful version for the time being, we'll probably want to move the testing environments to complete 18.04 LTS soon (or maybe in addition to 16.04?).

Hopefully it should work now.

src/libaktualizr/storage/schema_migration_test.sh
```diff
@@ -0,0 +1,34 @@
...
+    echo "schema.sql not found in $SQL_DIR"
+    exit 1
+fi
+
+sqlite3 -batch -init "$SQL_DIR/schema.sql" "$DB_CUR" ";"
+for f in "$SQL_DIR"/migration/migrate.*.sql; do
+    sqlite3 -batch -init "$f" "$DB_MIG" ";"
+done
+
+R=$(sqldiff "$DB_CUR" "$DB_MIG")
```
OYTIS: I didn't know that such tool exists. Probably we should replace my horrible hack with `boost::tokenizer` with a shell-out to this tool.

lbonn: Yes, I found it while reading some stackoverflow answer (https://www.sqlite.org/sqldiff.html).
Also good idea, I'll do that.

cajun-rat:
Looks great, merging...

PR #855, opened 2018-07-13
aktualizr-repo: add sha512 hashes

OYTIS:
SHA-512 are needed for small secondaries.


OYTIS:
@patrickvacek I suspect it's what backend people will do if I convince them to support sha512 hashes. If they choose to only support sha512 instead we can change it.

cajun-rat:
Looks fine.  This is only in aktualizr-repo at the moment, so we are not committed to it in the long term.

PR #859, opened 2018-07-17
SendDeviceData also for --running-mode=check.

patrickvacek:
This will send a lot of redundant data every time aktualizr runs with that flag, but that might be reasonable anyway if the data changed since the last run.

This probably isn't a perfect solution, but since there is no specific requirement at present, I'd rather keep it simple while sending the messages that were previously being skipped in manual runs.


patriotyk:
You also need to update https://github.com/advancedtelematic/aktualizr/blob/master/docs/aktualizr_flow.svg

patrickvacek:
@patriotyk I don't think I actually changed anything that would affect that diagram. I just changed at what point the manual `check` enters the diagram.

PR #860, opened 2018-07-17
Fix/prevent rel path misuse

lbonn:
And set `storage_config.sqldb_path` to be relative to `storage_config.path` by default (an absolute path is also accepted).


lbonn:
I also meddled with `ImportConfig`, this will probably need re-review (last commit only)

patrickvacek:
Just to verify: you can still use absolute paths for the import config, right? I think the assumption there was that you may get those files from a variety of sources. I don't know if it is necessary to have a base path for the import section, but if it keeps things more consistent and makes things easier, that is fine with me.

lbonn:
Yes, it's possible to use it the same way as before: keep base_path empty and specify everything as absolute paths. Absolute paths will also work if base_path is non empty.

Maybe I should add some tests for that though.

cajun-rat:
I'm happy.  Maybe add that test for "/base" + "/mypath" => "/mypath" and manually bitbake + boot to make sure this hasn't broken any assumptions that are in the config?

lbonn:
Ok it passed oe-selftest

PR #861, opened 2018-07-17
Factory methods for SotaUptaneClient class

patriotyk:


cajun-rat:
Looks like there are a couple of CI failures:

~~~
[ 93%] Building CXX object src/sota_tools/CMakeFiles/garage-push.dir/garage_push.cc.o
/ostree/aktualizr/src/src/load_tests/check.cc:50:68: error: use of undeclared identifier 'bootloader'
    SotaUptaneClient client{config, eventsIn, storage, httpClient, bootloader, report_queue};
~~~

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -16,23 +16,53 @@
...
+                                            report_queue_in, events_channel_in);
+}
+
+std::shared_ptr<SotaUptaneClient> SotaUptaneClient::newTestClient(Config &config_in,
+                                                                  std::shared_ptr<INvStorage> storage_in,
+                                                                  std::shared_ptr<HttpInterface> http_client_in,
+                                                                  std::shared_ptr<event::Channel> events_channel_in) {
+  std::shared_ptr<Uptane::Fetcher> uptane_fetcher =
+      std::make_shared<Uptane::Fetcher>(config_in, storage_in, http_client_in);
+  std::shared_ptr<Bootloader> bootloader_in = std::make_shared<Bootloader>(config_in.bootloader);
```
patrickvacek: Do we ever use a nonstandard bootloader?

patriotyk: Currently no

src/libaktualizr/crypto/keymanager.cc
```diff
@@ -189,7 +189,7 @@ std::string KeyManager::getCN() const {
   return cn;
 }
 
-void KeyManager::copyCertsToCurl(HttpInterface *http) {
+void KeyManager::copyCertsToCurl(const std::shared_ptr<HttpInterface> &http) {
```
lbonn: For this kind of function which doesn't share ownership and just use `http`, passing a simple pointer is ok. It reduces coupling with the rest of the code (doesn't depend on the flavor of memory management used by the caller).

See https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/ in section 3.:

> (a) and (b): Prefer passing parameters by * or &.

patrickvacek:
@patriotyk FYI there are some merge conflicts now; probably worth a rebase anyway.

src/libaktualizr/crypto/keymanager.cc
patriotyk: If we hold an object or pointer, yes maybe raw pointer would be ok. But currently we already have a shared_ptr and I don't see any advantages here if I pass raw pointer from shared and pass it to the method. I have red this blog post, but for me it is not clear why raw pointer is better againts refference to the shared.

lbonn: In terms of generated code it should be the same, it's just that it's better if the type of pointer you pass serves as documentation for the way the object is used.

Passing a shared pointer indicates some ownership passing, otherwise you're not using it really. On the other hand, passing a raw pointer indicates that the method only makes use of the object for the duration of its execution (eg: borrow it).

But this is only a style matter, it should not block merging the PR.

patrickvacek: I see the point, but we are doing it wrong (according to that article) all over the place. For now I'd keep it consistent and we can debate if it's worth making the change to Sutter's recommendations. In the meantime, though, it looks like another rebase is unfortunately in order. And then I think we can merge.

patrickvacek:
Now seeing this:
```
/var/lib/jenkins/workspace/aktualizr_PR-861-BCRUSJXCWBIRYEUGR7N2KRWCN7IMARN7R3MOPLM4XXLXJ7LPBBZA@2/src/libaktualizr/uptane/uptane_test.cc:568:26: error: 'class std::shared_ptr<HttpFake>' has no member named 'tls_server'

   conf.tls.server = http.tls_server;
```

PR #862, opened 2018-07-17
Uptane library for small devices

OYTIS:
So now it works, by the definition of `works` as in unit tests. Since the code is still quite convoluted, it is of limited value until fuzzing is implemented, which should follow.

Probalby worth scanning through `partial/README.adoc` first.


partial/libuptiny/json_common.c
```diff
@@ -0,0 +1,28 @@
+#include "jsmn/jsmn.h"
+
+extern jsmntok_t token_pool[];
+extern const unsigned int token_pool_size;
+
+unsigned int consume_recursive_json(unsigned int idx) {
```
lbonn: Set a maximum recursion depth here?

OYTIS: Yes, true.

OYTIS: OK, so I've just removed the recursion, it's better than processing another error case.

OYTIS:
I've got no idea, why it started complaining about my conversions now, hope this short fix helps.

patrickvacek:
> I've got no idea, why it started complaining about my conversions now, hope this short fix helps.

Sorry, probably my fault:
https://github.com/advancedtelematic/aktualizr/pull/863

To be fair, we wanted to do that like a year ago, but I finally made it happen now. I tried to disable it for the entire partial subdirectory, but I think your reorganization changed how that gets applied.

OYTIS:
@patrickvacek Got it, makes sense really.

OYTIS:
Phew

PR #867, opened 2018-07-20
Bugfix/download failed

patrickvacek:
This fixes a few bugs I caught in the Uptane processing. I bumped tuf-uptane-vectors, but not all the way to the current master. There are a couple of the new vectors that we don't pass.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -574,7 +574,8 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets) {
       auto hwid_it = hw_ids.find(ecu_serial);
       if (hwid_it == hw_ids.end()) {
         LOG_WARNING << "Unknown ECU ID in director targets metadata: " << ecu_serial.ToString();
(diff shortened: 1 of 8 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -605,20 +603,28 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets) {
...
   // deploying)
+  std::vector<Uptane::Target> downloaded_targets;
   for (auto it = targets.cbegin(); it != targets.cend(); ++it) {
     // TODO: delegations
     auto images_target = images_repo.getTarget(*it);
     if (images_target == nullptr) {
       LOG_ERROR << "No matching target in images targets metadata for " << *it;
       continue;
+    } else {
+      downloaded_targets.push_back(*it);
```
lbonn: No need for `else`, there is a `continue` in the other condition. We could keep it but then, no reason for line 614 and 618 to sit in different code blocks

patrickvacek: Fair enough. Fixed.

PR #869, opened 2018-07-24
Silence warnings in partial external code

lbonn:
Instead of just making them not fail the build (all these warnings are
distracting)


lbonn:
@patrickvacek nice, thanks. Also I missed some instances the first time

PR #871, opened 2018-07-24
Feat/ota 298/hmi api

patrickvacek:
There's a bit of cleaning up that could be done with the EventsInterpreter, and there is still some documentation to be done, but I'd like to solicit feedback on what I've got so far sooner rather than later.

### DIFF
--- a/src/libaktualizr/crypto/crypto.cc
+++ b/src/libaktualizr/crypto/crypto.cc
@@ -322,6 +322,7 @@ bool Crypto::extractSubjectCN(const std::string &cert, std::string *cn) {
 
 /**
  * Generate a RSA keypair
+ * @param key_type Algorithm used to generate the key
  * @param public_key Generated public part of key
  * @param private_key Generated private part of key
  * @return true if the keys are present at the end of this function (either they were created or existed already)
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -47,6 +47,7 @@ int Aktualizr::run() {
 
 void Aktualizr::sendCommand(const std::shared_ptr<command::BaseCommand> &command) { *commands_channel_ << command; }
 
-void Aktualizr::setSignalHandler(std::function<void(std::shared_ptr<event::BaseEvent>)> &handler) {
-  (*sig_).connect(handler);
+boost::signals2::connection Aktualizr::setSignalHandler(
+    std::function<void(std::shared_ptr<event::BaseEvent>)> &handler) {
+  return (*sig_).connect(handler);
 }
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -45,7 +45,7 @@ int Aktualizr::run() {
   return EXIT_SUCCESS;
 }
 
-void Aktualizr::sendCommand(std::shared_ptr<command::BaseCommand> command) { *commands_channel_ << command; }
+void Aktualizr::sendCommand(const std::shared_ptr<command::BaseCommand> &command) { *commands_channel_ << command; }
 
 void Aktualizr::setSignalHandler(std::function<void(std::shared_ptr<event::BaseEvent>)> &handler) {
   (*sig_).connect(handler);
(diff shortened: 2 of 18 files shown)


src/libaktualizr/primary/aktualizr.cc
```diff
@@ -28,25 +23,30 @@ Aktualizr::Aktualizr(Config &config) : config_(config) {
...
+  std::shared_ptr<SotaUptaneClient> uptane_client =
+      SotaUptaneClient::newDefaultClient(config_, storage, events_channel_);
+  uptane_client->runForever(commands_channel_);
 
   return EXIT_SUCCESS;
 }
+
+void Aktualizr::sendCommand(std::shared_ptr<command::BaseCommand> command) { *commands_channel_ << command; }
+
+void Aktualizr::setSignalHandler(std::function<void(std::shared_ptr<event::BaseEvent>)> &handler) {
```
lbonn: It could be useful for the api user to remove a signal handler.

One possibility: return the `boost::signal2::connection` returned by `connect()`, so that `disconnect()` can be called from there.

cajun-rat: How bad would it be to directly expose the boost::signals2::signal object?  We'd be binding our API to boost, but on the other hand we'd get all the logic like automatic disconnection for free.

patrickvacek: It's an option. We are already bound to boost if I return the connection object (as I have now changed it to do). Sharing the signal object would expose `std::shared_ptr<boost::signals2::signal<void(std::shared_ptr<event::BaseEvent>)>>` instead of what I currently have (`boost::signals2::connection setSignalHandler(std::function<void(std::shared_ptr<event::BaseEvent>)>& handler);`), which is already a bit of a handful. I was considering ways to `typedef`/`using` that into something nicer, but I'm not sure what that would be that would actually make anyone's life easier. (Doing it for the function object is an option, but honestly it is probably better to know upfront what the required input and outputs of the handler are.)

OYTIS:
Looks good, so just download/installation progress is left and we can start with the actual HMI.

src/libaktualizr/crypto/crypto.cc
```diff
@@ -322,6 +322,7 @@ bool Crypto::extractSubjectCN(const std::string &cert, std::string *cn) {
 
 /**
  * Generate a RSA keypair
+ * @param key_type Type of key (RSA2048, RSA4096, etc.)
```
tkfu: Can we actually enumerate the possible values here, instead of "etc."? I bet there are few enough that it's reasonable, right?

patrickvacek: Yes, that is possible, although I would argue it would be better to go to the actual definitions directly instead of duplicating the information here. (Also, this was just my quick attempt to clean up some doxygen complaints -- this isn't part of the API that I wanted to work on.)

tkfu: Ah, gotcha. And I think I agree that it's better not to.

patrickvacek:
Rebased three times, fixed the CI complaint, and fixed the doc comment. Should be ready now.

@OYTIS progress should be easy to implement (from the API perspective). It is likely that I will want to evolve the API next sprint. It should be usable and extensible now, but I think we can do even better.

src/libaktualizr/primary/eventsinterpreter.cc
```diff
@@ -18,29 +22,46 @@ EventsInterpreter::~EventsInterpreter() {
 
 void EventsInterpreter::interpret() { thread = std::thread(std::bind(&EventsInterpreter::run, this)); }
 
-std::shared_ptr<command::BaseCommand> EventsInterpreter::handle_cycle(event::BaseEvent &event, bool forever) {
+std::shared_ptr<command::BaseCommand> EventsInterpreter::handle_cycle(event::BaseEvent &event,
```
lbonn: Is something preventing us from defining a new method? I kept `once` and `full` because they are really similar but `none` seems sufficiently different and simple:

```
std::shared_ptr<command::BaseCommand> EventsInterpreter::handle_none(event::BaseEvent &event) {
  if (event.variant == "SendDeviceDataComplete") {
    return std::make_shared<command::FetchMeta>();
  } else if (event.variant == "FetchMetaComplete") {
    return std::make_shared<command::CheckUpdates>();
  }
  return nullptr;
}
```

patrickvacek: Done. Good call.

Issue #771, opened 2018-05-16
Feature Proposal: Tracking Device Update Status


doanac:
This is somewhat I spin-off of #670 with some details around what might be possible. My use case might be a bit different than normal ATS Garage usage. However, I find it a bit frightening that I hit "apply update" in the UI and then sit in the dark for some $amount_of_time_that_increases_based_on_the_ostree_delta_size until it finally shows success. With PR #767 we now have a fairly good estimate of how far along the update is. I think it would be a really slick feature to reflect that inside the UI. To help get the conversation started, I thought I'd propose one way I think could work:

## Director Changes

 Add a new text column to `device_update_targets` named "status".
 Add a new admin endpoint:
```
  PUT /api/v1/admin/devices/<deviceUuuid>/queue/<updateUuid>
```
The payloads could look like:
```
    data: {'status': '10%'}
    data: {'status': '13 metadata, 16 content objects fetched; 44683 KiB transferred in 23 seconds'}
    data: {'Performing sync()'}
```
  The director's existing ``queue`` endpoint could include this new attribute in each item

## Aktualizr Changes

Have the ``Repository::downloadTarget`` code pass some type of status callback to ``OstreeManager::pull``. This callback would be responsible for making PUTs on the new director endpoint.

I'm curious about a few things:
1) Your thoughts on the idea?
2) Your willingness to accept such a change into aktualizr and director?
3) If 1 and 2, then how much help from me would you need? ie - is this something I need to do myself, or would you guys be willing to do some amount of the work? I feel fairly comfortable doing the aktualizr changes. My Scala skills are pretty lacking, but I think there's enough code in Director that I could figure out how to copy/paste my way through it.

jerrytrieu:
Yes, we are definitely planning on adding this feature. 

The Director `/api/v1/admin/` endpoints are not accessible to devices - it should be `/api/v1/device/:deviceUuid/download_progress` which the device would access at `/director/download_progress` on the device-gateway. The device should specify the `target` `filepath` it is reporting status for, as it would be downloading multiple images for its primary and secondary ECUs as part of the update (the device doesn't care about the `updateUuid`). There should be some way to configure how often the device should report progress.  The implementation on the Director may need an additional table and `/admin/` endpoint to retrieve this info on the admin-UI. Going forward, the backend should track when each download was started, paused, resumed, or canceled.

Let's groom this feature at our next sprint planning.

doanac:
[...]

PR #872, opened 2018-07-25
Add download progress events

patriotyk:


src/aktualizr_secondary/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ add_library(aktualizr_secondary_static_lib STATIC
     $<TARGET_OBJECTS:storage>
     $<TARGET_OBJECTS:logging>
     $<TARGET_OBJECTS:uptane>
+    $<TARGET_OBJECTS:primary>
```
patrickvacek: This seems less than ideal. What does the secondary need from the primary?

patriotyk: We don't need it really, it is old artifact

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -19,47 +19,27 @@
...
-    } else {
-      guint calculated = (fetched * 100) / requested;
-      if (calculated != *percent_complete) {
-        LOG_INFO << "ostree-pull: Receiving objects: " << calculated << "% ";
-        *percent_complete = calculated;
-      }
-    }
-  } else if (outstanding_writes != 0) {
-    LOG_INFO << "ostree-pull: Writing objects: " << outstanding_writes;
+  auto *mt = static_cast<PullMetaStruct *>(data);
```
patrickvacek: In the worst case, can't we use this `calculated`? That seems to be good enough for what we need.

patriotyk:
I updated PR and I think it is final version. I have not found the way to get the overall progress in percentage for ostree, so I left the existing event handler.

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -39,9 +39,12 @@ static void aktualizr_progress_cb(OstreeAsyncProgress *progress, gpointer data)
                << " outstanding: " << outstanding_metadata_fetches;
```
doanac: This looks great. I can't remember what all is possible to communicate through the events API. However, if possible, it might be nice to add something like:
~~~
    *mt->events_channel << std::make_shared<event::DownloadProgressReport>(mt->target, "Calculating...");
~~~

patrickvacek: @doanac Currently the event only has an integer to represent progress percentage. What is the advantage to just emitting a 0% progress event at this stage?

Just FYI for the bigger picture: this events infrastructure will probably change soon to something a bit cleaner and better-typed using Boost signals. When we do that work, we can take into account your request to have text status messages included. That doesn't seem unreasonable and might be useful elsewhere, too.

patrickvacek:
Actually the segfault happens with aktualizr as well. I'm testing on my local machine with the fake package manager. Haven't tried with OSTree on qemu yet, but I'm building it now.

patrickvacek:
Still segfaulting.

src/libaktualizr/package_manager/ostreemanager.cc
doanac: 0% is clever. works for me

patriotyk:
Problem with crash was in incorrect "if" statement. Instead of  "if (ds->events_channel != nullptr) {" there was "if (ds != nullptr) {".

src/libaktualizr/primary/eventsinterpreter.cc
```diff
@@ -115,6 +115,10 @@ void EventsInterpreter::run() {
   auto running_mode = config.uptane.running_mode;
 
   while (*events_channel >> event) {
+    if (event->variant == "DownloadProgressReport") {
+      // Process event on HMI.
+      continue;
+    }
```
doanac: I just realized I mis-read this change. I was thinking these events were winding up in the ReportQueue, but that's not correct, right? Are there plans to also place these events in the ReportQueue?

patrickvacek: Correct, they are not going to the ReportQueue. We aren't planning on sending those events to the server at present, but that is something we could consider.

doanac: Oh - i thought this might related to bug #771.

patrickvacek: I think we'd like to get there eventually, but it just isn't a high enough priority yet.

PR #873, opened 2018-07-25
Remove GatewayManager and SocketGateway.

patrickvacek:
They were unused and just gathering dust.


patrickvacek:
> You also need to remove include_directories(${PROJECT_SOURCE_DIR}/third_party/picojson) from root CMakeLists.txt file.
> Also remove picojson from 'third_party' directory

Good call. Done.

PR #874, opened 2018-07-25
Refactor/rm old ipc

lbonn:
It seems that we don't need it anymore, shout if I missed something...


patrickvacek:
oe-selftest with this on master passes (except for HSM, which seems unrelated), so it's fine with me.

PR #876, opened 2018-07-25
Separate different running mode state machines

lbonn:
Flatten everything so that it's clearer and less error-prone

The cascade of ifs was confusing, waiting to break some day.

This PR is probably oe-selftest worthy

### DIFF
--- a/src/libaktualizr/primary/eventsinterpreter.cc
+++ b/src/libaktualizr/primary/eventsinterpreter.cc
@@ -5,7 +5,7 @@
 
 #include "config/config.h"
 
-EventsInterpreter::EventsInterpreter(const Config& config_in, std::shared_ptr<event::Channel> events_channel_in,
+EventsInterpreter::EventsInterpreter(const Config &config_in, std::shared_ptr<event::Channel> events_channel_in,
                                      std::shared_ptr<command::Channel> commands_channel_in)
     : config(config_in),
       events_channel(std::move(std::move(events_channel_in))),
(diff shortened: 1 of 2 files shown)


src/libaktualizr/primary/eventsinterpreter.cc
```diff
@@ -19,56 +19,117 @@ EventsInterpreter::~EventsInterpreter() {
...
-      if (config.uptane.running_mode == RunningMode::kFull) {
-        std::this_thread::sleep_for(std::chrono::seconds(config.uptane.polling_sec));
-        *commands_channel << std::make_shared<command::FetchMeta>();
-      } else {
-        *commands_channel << std::make_shared<command::Shutdown>();
-      }
+
+    if (next_command == nullptr) {
+      LOG_ERROR << "No command to run after event " << event->variant << " in mode "
+                << StringFromRunningMode(running_mode);
```
patrickvacek: I'm not sure this is really an error. In my PR (https://github.com/advancedtelematic/aktualizr/pull/871) I've created a `kNone` running mode to indicate completely manual control, i.e. you can freely add commands to the channel, and events are produced as a result. Thus it is valid that not every event is followed by another command.

lbonn: Should it be a warning or just stay silent?

lbonn:
@patrickvacek Yes, I overheard these conversations about the removal. I still did this because I'll need to add another event for a new feature and didn't want to aggravate the situation.

It's fine if it goes away soon :)

src/libaktualizr/primary/eventsinterpreter.cc
patrickvacek: Actually, we could leave it there and only silence it if `RunningMode.kNone` (which I will do if we merge this first). In my new situation, it is not an error, but for the existing modes it is.

patrickvacek:
BTW oe-selftest on master passed everything except (again) HSM, which seems to be broken ATM for other reasons.

lbonn:
Thanks patrick! I launched oe-selftest but it only had time to slow down my machine.

PR #877, opened 2018-07-26
gcc8 warnings

lbonn:
Got blessed with an automatic update from gcc 7 to 8 this morning.

    ../src/libaktualizr/crypto/crypto.cc:150:77: error: type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]
       const auto sign_size = static_cast<const unsigned int>(RSA_size(rsa.get()));

And

    ../src/libaktualizr/uptane/uptane_test.cc:128:37: error: catching polymorphic type ‘class Uptane::IllegalThreshold’ by value [-Werror=catch-value=]
       } catch (Uptane::IllegalThreshold ex) {

### DIFF
--- a/src/libaktualizr/crypto/crypto.cc
+++ b/src/libaktualizr/crypto/crypto.cc
@@ -147,7 +147,7 @@ std::string Crypto::RSAPSSSign(ENGINE *engine, const std::string &private_key, c
 #endif
   }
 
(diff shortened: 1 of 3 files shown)


lbonn:
I'll merge, the CI errors are spurious and unrelated (to investigate, still)

PR #878, opened 2018-07-26
Log uptane exception in MetaPack::isConsistent()

lbonn:


src/libaktualizr/uptane/tuf.cc
```diff
@@ -296,7 +296,8 @@ bool MetaPack::isConsistent() const {
     Uptane::Root original_root(director_root);
     Uptane::Root new_root(RepositoryType::Director, director_root.original(), new_root);
     Uptane::Targets(RepositoryType::Director, director_targets.original(), original_root);
-  } catch (...) {
+  } catch (const Uptane::Exception &exc) {
+    LOG_WARNING << "Inconsistent metadata: " << exc.what();
```
patrickvacek: Good call. Are there other exceptions worth catching or is that the only one here?

lbonn: We throw `Uptane::Exception` children everywhere in `tuf.cc` and `tuf.h`, but also `std::out_of_range` in some constructors (`tuf.h`).

I'll change it to `std::logic_error`, the common parent, just to be sure.

OYTIS:
I would say this method should be removed eventually, but once it's there, let's make the logging of course.

PR #879, opened 2018-07-26
Fixed 5 vector tests

patriotyk:
There is one issue in vector tests. The test 'director_targets_zero_threshold' expects 'The role root had an illegal signature threshold.' which is incorrect. The problem is not in the json file because 'director_targets_zero_threshold.json' contains the correct error. So the problem is that `POST /director_targets_zero_threshold/step` returns following result:

```
{
83: 	"director" : 
83: 	{
83: 		"targets" : 
83: 		{
83: 			"file.txt" : 
83: 			{
83: 				"is_success" : true
83: 			}
83: 		},
83: 		"update" : 
83: 		{
83: 			"err" : "IllegalThreshold::Root",
83: 			"err_msg" : "The role root had an illegal signature threshold.",
83: 			"is_success" : false
83: 		}
83: 	},
83: 	"image_repo" : 
83: 	{
83: 		"targets" : 
83: 		{
83: 			"file.txt" : 
83: 			{
83: 				"is_success" : true
83: 			}
83: 		},
83: 		"update" : 
83: 		{
83: 			"is_success" : true
83: 		}
83: 	}
83: }
```

### DIFF
--- a/src/libaktualizr/uptane/exceptions.h
+++ b/src/libaktualizr/uptane/exceptions.h
@@ -91,6 +91,12 @@ class NonUniqueSignatures : public Exception {
   ~NonUniqueSignatures() noexcept override = default;
 };
 
+class BadKeyId : public Exception {
+ public:
+  BadKeyId(const std::string& reponame) : Exception(reponame, "A key has an incorrect associated key ID") {}
+  ~BadKeyId() noexcept override = default;
+};
+
 class BadEcuId : public Exception {
  public:
   BadEcuId(const std::string& reponame)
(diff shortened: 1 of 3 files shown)


patrickvacek:
So the problem is that it should be this error:
```
"err": "IllegalThreshold::Targets",
"err_msg": "The role targets had an illegal signature threshold.",
```
instead? If it is a real problem with tuf-test-vectors, feel free to make a PR against that repo and then we can bump the version in aktualizr to that.

patriotyk:
Amended with new submodule commit, waiting CI

PR #881, opened 2018-07-27
Fix filepath in the manifest.

patrickvacek:
There was a missing element that caused this to always return an empty
string. I have extended a couple tests to catch this. This was also the
root of the issue in which aktualizr would continually think it found
new updates even after successfully installing them.


patrickvacek:
@lbonn Do we need to worry about the expired metadata complaint in Travis? There error is this:
```
[2018-07-27 12:14:47.107909] [0x000000000403bcc0] [trace]   Metadata expired at:2018-01-06T20:01:05Z
unknown file: Failure
C++ exception with description "The root metadata was expired." thrown in the test body.
[  FAILED  ] SotaUptaneClientTest.RunForeverInstall (3708 ms)
```
We previously wrote this json differently so it was just a typo from a few weeks back. I was actually wrong in thinking that the typo had existed longer than that.

PR #882, opened 2018-07-27
Add a new command, campaign_check

lbonn:
Accessible from the CLI + the "new" exposed api

### DIFF
--- a/src/libaktualizr/campaign/CMakeLists.txt
+++ b/src/libaktualizr/campaign/CMakeLists.txt
@@ -1,0 +1,8 @@
+set(SOURCES campaign.cc)
+set(HEADERS campaign.h)
+
+add_library(campaign OBJECT ${SOURCES})
+
+add_aktualizr_test(NAME campaign SOURCES campaign_test.cc ARGS ${CMAKE_CURRENT_SOURCE_DIR}/test)
+
+aktualizr_source_file_checks(${SOURCES} ${HEADERS} ${TEST_SOURCES})
--- a/src/libaktualizr/campaign/campaign.cc
+++ b/src/libaktualizr/campaign/campaign.cc
@@ -4,19 +4,25 @@ namespace campaign {
 
 Campaign Campaign::fromJson(const Json::Value &json) {
   try {
+    if (!json.isObject()) {
+      throw CampaignParseError();
+    }
     std::string name = json["name"].asString();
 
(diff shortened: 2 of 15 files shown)


src/libaktualizr/campaign/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+set(SOURCES campaign.cc)
+set(HEADERS campaign.h)
+
+add_library(campaign OBJECT ${SOURCES})
+
+add_aktualizr_test(NAME campaign SOURCES campaign_test.cc ARGS ${CMAKE_CURRENT_SOURCE_DIR}/test)
```
patriotyk: test -> tests? But if it works with typo, you don't need it

lbonn: Which one? `${CMAKE_CURRENT_SOURCE_DIR}/test`?

It's to get the path of `https://github.com/advancedtelematic/aktualizr/blob/7acdc4c6c90098e055dad96abed4cf3dc385dbbf/src/libaktualizr/campaign/test/campaigns_sample.json`

src/libaktualizr/campaign/campaign.cc
```diff
@@ -0,0 +1,57 @@
...
+        if (!install_message.empty()) {
+          LOG_ERROR << "Only one install message allowed";
+          throw;
+        }
+        install_message = o["value"].asString();
+      }
+    }
+
+    return {name, install_message};
+  } catch (std::runtime_error &exc) {
```
patriotyk: should be const reference

src/libaktualizr/campaign/CMakeLists.txt
patriotyk: Sorry I thought it is project source directory

src/libaktualizr/campaign/campaign.cc
```diff
@@ -0,0 +1,57 @@
...
+    LOG_ERROR << exc.what();
+    throw CampaignParseError();
+  } catch (...) {
+    throw CampaignParseError();
+  }
+}
+
+std::vector<Campaign> campaignsFromJson(Json::Value &json) {
```
patriotyk: Why json is not a const?

lbonn:
Ok, was tested properly against backend

PR #883, opened 2018-07-30
Make dependancies visible in parent scope

patriotyk:
This small change allows to reuse aktualizr repository as cmake subdirectory from another projects

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -356,6 +356,11 @@ set (AKTUALIZR_EXTERNAL_LIBS
     ${GLIB2_LIBRARIES}
     ${SYSTEMD_LIBRARY})
 
+get_directory_property(hasParent PARENT_DIRECTORY)
+if(hasParent)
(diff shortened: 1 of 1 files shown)


cajun-rat:
Approved, assuming CI passes.

patriotyk:
Hm, looks like 'PARENT_SCOPE' moves variable to the parent scope but removes from childs

cajun-rat:
Can we export a suitable list of things to link/include?

patriotyk:
I have already fixed this issue, waiting for CI

lbonn:
It issues a warning when building aktualizr the regular way:

```
CMake Warning (dev) at CMakeLists.txt:359 (set):
  Cannot set "AKTUALIZR_EXTERNAL_LIBS": current scope has no parent.
This warning is for project developers.  Use -Wno-dev to suppress it.

```

According to [internet wisdom](https://stackoverflow.com/a/25217937), it's possible to avoid it.

patriotyk:
Now it should work

PR #886, opened 2018-08-01
Apply curl bug workaround for GET requests

lbonn:
This should fix HSM support

Was forgotten because masked by another issue.

There are similar issue sitting in curl bugs: https://github.com/curl/curl/issues/2485 (this one was fixed). I'll see if the problem we have is still present on upstream and possibly fill a report.

We'll have to keep the workaround for a while though.


cajun-rat:
Fixes the Hsm test on oe-selftest.  All the oe-selftest -r updater tests pass now.

patrickvacek:
Nice. Super frustrating though. I wonder, does this fix the hsm_prov test in the aktualizr repo? I'll check if I get to chance.

PR #889, opened 2018-08-02
Run aktualizr-info in readonly mode by default

patriotyk:


src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1343,6 +1343,11 @@ bool SQLStorage::dbMigrate() {
     return true;
   }
 
+  if (readonly_) {
+    LOG_ERROR << "Dtabase is opened in readonly mode and cannot be migrated to latest version";
```
lbonn: Typo, "Database"

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -97,7 +97,7 @@ SQLStorage::SQLStorage(const StorageConfig& config) : INvStorage(config) {
 }
 
 void SQLStorage::storePrimaryKeys(const std::string& public_key, const std::string& private_key) {
-  SQLite3Guard db(dbPath());
+  SQLite3Guard db(dbPath(), readonly_);
```
lbonn: Maybe it's time to set up a private helper method, the boiler-plate is starting to get a bit heavy:

```
SQLite3Guard SQLStorage::dbConnection() {
  SQLite3Guard db(dbPath(), readonly_);
  }
  return std::move(db);
}
```

And usage:

```
    return false;
  }
```

patriotyk: Nice idea

cajun-rat:
This is great.  Removing all the duplication in SQLStorage is very good, and I like any commit that _reduces_ the number of lines on code in the codebase.

PR #890, opened 2018-08-02
API for HMI

cajun-rat:


src/libaktualizr/primary/aktualizr.h
```diff
@@ -26,18 +26,57 @@ class Aktualizr {
...
+   * the Uptane metadata, including root and targets.
+   */
+  void FetchMetadata();
+
+  /**
+   * Asynchronously load already-fetched Uptane metadata from disk.
+   * This is only needed when the metadata fetch and downloads/installation are
+   * in separate aktualizr runs.
+   */
+  void CheckUpdates();
```
patrickvacek: Is this accurate or necessary? Previously, `CheckUpdates` was internal only, since it was automatically generated after a `FetchMetaComplete`. We could expose it if you want it, but is there any advantage to just stuffing all of that functionality inside `FetchMeta`, other than that `CheckUpdates` doesn't require network connectivity? I fear unnecessary expansion of the API to support a trivial advantage in functionality.

patrickvacek:
We will also want to update the doxygen page: https://advancedtelematic.github.io/aktualizr/index.html

cajun-rat:
@tkfu Do you know how  https://advancedtelematic.github.io/aktualizr/index.html gets updated?

patrickvacek:
@cajun-rat The current steps are: 
1. ninja docs
2. copy the output (build/docs) to a separate location.
3. switch to the `gh-pages` branch.
4. move the copied docs output to the root of the repo.
5. `git push`

The one part that is unclear to me is if you can only have one commit in that branch or if you really can have history. Last time I tried to update it seemed like I had to squash the commit history to make it work.

src/libaktualizr/primary/aktualizr.h
cajun-rat: I was trying to avoid exposing it, but `CheckUpdates` is used in main.cc:

~~~
    switch (config.uptane.running_mode) {
      case RunningMode::kDownload:
      case RunningMode::kInstall:
        aktualizr.CheckUpdates();
        break;
~~~

It is a somewhat special case, because it initializes the list of targets to be downloaded/installed from the Uptane metadata on disk.  I'm not sure how be can avoid that, given the requirement for a 'download' and 'install' that don't fetch metadata.

patrickvacek: Is that a requirement? If we have to do a fetch at some point, and the fetch eventually creates a `CheckUpdates` internally, then why not just do that? If we reboot, what is the problem with fetching again before allowing a download? The download requires connectivity anyway. If we reboot between downloading and installing, I'll grant that that is trickier. Can we do a `CheckUpdates` before installation if the metadata is marked as unchecked?

cajun-rat: I think the requirement was that at least the 'install' step could be done without internet connectivity.

patrickvacek: So for now, maybe this is fine, but could we need detect if the metadata is unchecked since last reboot and then automatically do a `CheckUpdates` if we need it? This seems like something the user should not have to care about at all.

patrickvacek:
> The one part that is unclear to me is if you can only have one commit in that branch or if you really can have history. Last time I tried to update it seemed like I had to squash the commit history to make it work.

I must've done something weird last time. This time it worked fine. The key is to blast the existing docs, copy in the new docs, `git add .` and then commit and push.

PR #893, opened 2018-08-06
Scripts to use aktualizr-repo as a replacement to the backend

OYTIS:
Tested with meta-updater rocko and raspberrypi. Uptane metadata exchange works after minor fixes to cert_provider (following). OSTree not tested yet (would require more flexibility in `aktualizr-repo` utility wrt. adding packages to images repo).

### DIFF
--- a/docs/aktualizr-repo.adoc
+++ b/docs/aktualizr-repo.adoc
@@ -1,0 +1,38 @@
+= Aktualizr repo
+
+aktualizr-repo director contains a basic implementation of Uptane server. It is comprised of three tools:
+
+. `create_repo.sh` is a script to generate a new repo
+. `serve_repo.py` is a minimalistic Uptane server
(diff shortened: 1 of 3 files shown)


OYTIS:
@cajun-rat There are no tests for this server yet unfortunately. We might add some to `oe-selftest` in meta-updater though, I think the most interesting test here is end-to-end from `serve_repo.py` to `meta-updater`.

patrickvacek:
@OYTIS We ought to be able to do some high-level tests internally, much like the provisioning tests, right? It's fine with me if you want to get this merged and add some tests like that in a separate PR, though.

OYTIS:
@patrickvacek Yes, what we could do is to generate sota.toml for a standalone aktualizr and test one update cycle. Let's wait with merger until it's done.

PR #894, opened 2018-08-06
Specify root directory to postUpdateValues (needed by cert_provider)

OYTIS:
Makes `akualizr-ca-implicit-prov.bb` happy again.

### DIFF
--- a/src/libaktualizr/config/config.cc
+++ b/src/libaktualizr/config/config.cc
@@ -142,6 +142,12 @@ Config::Config(const std::vector<boost::filesystem::path>& config_dirs) {
   postUpdateValues();
 }
 
+Config::Config(const std::vector<boost::filesystem::path>& config_dirs, const boost::filesystem::path& root_dir) {
+  checkDirs(config_dirs);
+  updateFromDirs(config_dirs);
+  postUpdateValues(root_dir);
+}
+
 Config::Config(const boost::program_options::variables_map& cmd) {
   // Redundantly check and set the loglevel from the commandline prematurely so
   // that it is taken account while processing the config.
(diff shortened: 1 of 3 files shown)


src/libaktualizr/config/config.cc
```diff
@@ -166,19 +172,21 @@ KeyManagerConfig Config::keymanagerConfig() const {
...
 
   if (tls.server.empty()) {
     if (!tls.server_url_path.empty()) {
-      tls.server = Utils::readFile(tls.server_url_path);
+      tls.server = Utils::readFile(root_dir / tls.server_url_path);
     } else if (!provision.provision_path.empty()) {
-      if (boost::filesystem::exists(provision.provision_path)) {
-        tls.server = Bootstrap::readServerUrl(provision.provision_path);
+      if (boost::filesystem::exists(root_dir / provision.provision_path)) {
+        tls.server = Bootstrap::readServerUrl(root_dir / provision.provision_path);
```
patrickvacek: Won't this break anything that previously used relative paths?

OYTIS: Right now only config test is broken, which is probably not so hard to fix. Can you remember some other scenario where relative paths could be used?

patrickvacek: Not particularly, but if any user were to be using relative paths in the field, we wouldn't know and this could break their expected functionality. Unlikely but possible.

Would it be easier to change the way that cert_provider creates/uses the paths instead? I don't understand the need for the root directory yet.

OYTIS: `Config::postUpdateValues` called from the constructor tries to read server URL from `tls.server_url_path` and fails. I can't see how it can be worked around in `cert_provider`.

patrickvacek: I'm confused. If `local_dir` is an output directory, why are we reading from it in the config? I don't see how anything has written to that directory before the Config object is created.

OYTIS: OK, so I was confused as well. Looks like everything works fine if the file doesn't exists, the constuctor just leaves tls.server empty and it will later be read from `credentials.zip`. When it does crash is when it can't read the file, because of user permissions (which is the case for `/var/sota` on my machine).
So probably the fix should look differently. Some parameter to constructor which makes it skip `postUpdateValues`?

patrickvacek: I've tried aggressively to fix issues when /var/sota is inaccessible due to permissions but the config specifies a separate path. I just fixed a couple more in https://github.com/advancedtelematic/aktualizr/pull/896.

OYTIS: Alternative solution, closing this PR
https://github.com/advancedtelematic/aktualizr/pull/897

PR #896, opened 2018-08-07
Feat/ota 244/progress api

patrickvacek:


src/libaktualizr/uptane/fetcher.cc
```diff
@@ -79,21 +80,22 @@ bool Fetcher::fetchVerifyTarget(const Target& target) {
...
       }
+      result = true;
     } else {
 #ifdef BUILD_OSTREE
       KeyManager keys(storage, config.keymanagerConfig());
       keys.loadKeys();
-      OstreeManager::pull(config.pacman.sysroot, config.pacman.ostree_server, keys, target, events_channel);
+      data::InstallOutcome outcome =
+          OstreeManager::pull(config.pacman.sysroot, config.pacman.ostree_server, keys, target, events_channel);
+      result = (outcome.first == data::UpdateResultCode::kOk);
```
rsalveti: pull might also return kAlreadyProcessed, which shouldn't be considered as failure.

patrickvacek: Good call, fixed. Thanks!

PR #897, opened 2018-08-07
Don't throw on filesystem errors in Config constructor

OYTIS:


OYTIS:
@patrickvacek I'm not sure how boost will deal with missing permissions for containing directory if I ask for permissions of a file.

PR #900, opened 2018-08-08
Do not throw BadKeyId on every unknown KeyId.

patrickvacek:
Unknown signatures are fine as long as there are still enough known
signatures to pass the threshold. To keep tuf-test-vectors happy,
though, we now throw the BadKeyId exception if there is only one
signature and its KeyId is unknown.

We urgently need tests that actually test the case of multiple
signatures, some recognized and some not.


patrickvacek:
FYI @rsalveti This should fix your issue. Can you confirm?

patrickvacek:
@rsalveti Great, thanks for reporting the issue and confirming the fix!

PR #901, opened 2018-08-08
Bugfix/check storage exists before attempting migration

patrickvacek:
Attempt to fix updater.ManualControlTests.test_manual_running_mode_once in oe-selftest.

### DIFF
--- a/src/libaktualizr/storage/invstorage.cc
+++ b/src/libaktualizr/storage/invstorage.cc
@@ -138,31 +138,35 @@ void INvStorage::importData(const ImportConfig& import_config) {
                import_config.tls_pkey_path);
 }
 
-std::shared_ptr<INvStorage> INvStorage::newStorage(const StorageConfig& config, bool readonly) {
+std::shared_ptr<INvStorage> INvStorage::newStorage(const StorageConfig& config, const bool readonly) {
   switch (config.type) {
     case StorageType::kSqlite: {
       boost::filesystem::path db_path = config.sqldb_path.get(config.path);
       if (!boost::filesystem::exists(db_path) && boost::filesystem::exists(config.path)) {
-        if (readonly) {
-          throw StorageException("Migration from FS is not possible, because of readonly database");
-        }
-
-        LOG_INFO << "Starting FS to SQL storage migration";
         if (access(config.path.c_str(), R_OK | W_OK | X_OK) != 0) {
           throw StorageException(std::string("Cannot read prior filesystem configuration from ") +
                                  config.path.string() + " due to insufficient permissions.");
         }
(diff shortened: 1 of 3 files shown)


src/libaktualizr/storage/invstorage.cc
```diff
@@ -174,56 +178,97 @@ std::shared_ptr<INvStorage> INvStorage::newStorage(const StorageConfig& config,
...
 }
 
-void INvStorage::FSSToSQLS(FSStorageRead& fs_storage, SQLStorage& sql_storage) {
+bool INvStorage::FSSToSQLS(FSStorageRead& fs_storage, const std::shared_ptr<SQLStorage>& sql_storage,
+                           const bool check_exists) {
   std::string public_key;
   std::string private_key;
   if (fs_storage.loadPrimaryKeys(&public_key, &private_key)) {
-    sql_storage.storePrimaryKeys(public_key, private_key);
+    if (check_exists) {
```
patriotyk: Maybe every 
```
    if (check_exists) {
      return true;
    } else {
      sql_storage->storeTlsCa(ca);
    }
```
change to: 

```
    if (!check_exists) {
      sql_storage->storeTlsCa(ca);
    } 
```
and `return true;` at the end of method

patrickvacek: The problem is that we want to migrate if any single file from FSStorage is found, but nothing is found, we shouldn't do anything with it. Currently, the migration is attempted if `storage.path` exists even if there is nothing in it. That is wrong and it causes an error with `aktualizr-info`.

cajun-rat:
Maybe perform the migration into a temporary, in-memory database in read-only mode?

OYTIS:
@patrickvacek Is this PR still relevant?

patrickvacek:
@OYTIS Yes and no. I still think something needs to be done. The only reason that oe-selftest still passes is because of the `--allow-migration` hack. The idea here is that we should only attempt a migration if there is something to migrate, not just if `/var/sota` exists. However, that is complicated by the fact that `installed_versions` is always written to devices and thus a migration _always_ happens at present. But then our tests that rely on `aktualizr-info` working before provisioning has commenced might need reworking. I haven't had time to look into this again, but I still think it's worth getting right. Ideally we would rework the `installed_versions` bit, perhaps by moving that to an `import` option in the aktualizr config. It's debatable whether most of this PR is still useful. Phil's suggestion is good, but would take a bit of work.

patrickvacek:
Closing because of general agreement that this isn't quite the best way to go. A few of these bits might be nice, but the `installed_versions` issue stands in the way of this anyway.

PR #902, opened 2018-08-08
Fix duplicate signature validation

cajun-rat:
When role thresholds are used, it is required that the multiple signatures come
from different keys. It should not be possible for one role to just sign the
metadata twice.

Previously, the check was on the _value_ of the signature, but this is
meaningless when signatures are non-deterministic or when an attacker can craft
two signatures that serialise differently. The correct check is to compare
keyids: these come from root.json, and attackers can't create new ones at
random.

### DIFF
--- a/src/libaktualizr/uptane/root.cc
+++ b/src/libaktualizr/uptane/root.cc
@@ -89,13 +89,13 @@ void Uptane::Root::UnpackSignedObject(RepositoryType repo, const Json::Value &si
   Json::Value signatures = signed_object["signatures"];
   int valid_signatures = 0;
 
-  std::vector<std::string> sig_values;
+  std::set<std::string> used_keyids;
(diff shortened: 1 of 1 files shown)


patrickvacek:
Is this something we can test somehow? Either as a tuf-test-vector or a normal unit test?

cajun-rat:
Good question.  There are no checks at the moment for duplicate signatures--the tests pass with the `throw NonUniqueSignatures(...)` line removed.

I guess the answer is to write a tests in tuf-test-vectors, but I'm not familiar with how that works

patrickvacek:
@cajun-rat Maybe related to a ticket I just wrote (OTA-485). I agree that ideally that kind of thing belongs in tuf-test-vectors. @patriotyk, @OYTIS, and I recently did a bit of poking around in there, and it seems like it can't be that hard to add some more tests.

PR #903, opened 2018-08-09
API AddSecondary

patriotyk:


patrickvacek:
Looking good! It's just some style/optimization complaints that are failing. The logic looks like exactly what I was hoping for, though.

patrickvacek:
Looks like just `src/libaktualizr/config/config.cc:288:30: error: use emplace_back instead of push_back` at this point.

PR #908, opened 2018-08-15
libuptiny: add manifest generation and firmware verification

OYTIS:
Now it's finally fully-functional, added a test for the full update process. The demo code is coming after testing/debugging.

### DIFF
--- a/partial/libuptiny/crypto_api.h
+++ b/partial/libuptiny/crypto_api.h
@@ -2,8 +2,8 @@
 #define LIBUPTINY_CRYPTO_API_H_
 
 #include <stdbool.h>
+#include <stddef.h>
 #include <stdint.h>
-#include "sha512.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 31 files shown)


partial/libuptiny/crypto_api.h
```diff
@@ -32,23 +32,25 @@ typedef struct {
...
 void crypto_verify_init(crypto_verify_ctx_t* ctx, crypto_key_and_signature_t* sig);
 void crypto_verify_feed(crypto_verify_ctx_t* ctx, const uint8_t* data, size_t len);
 bool crypto_verify_result(crypto_verify_ctx_t* ctx);
 
+void crypto_hash_init(crypto_hash_ctx_t* ctx);
+void crypto_hash_feed(crypto_hash_ctx_t* ctx, const uint8_t* data, size_t len);
+void crypto_hash_result(crypto_hash_ctx_t* ctx, crypto_hash_t* hash);
+
 crypto_algorithm_t crypto_str_to_keytype(const char* keytype, size_t len);
 int crypto_get_keylen(crypto_algorithm_t alg);
```
lbonn: Should return a `size_t`? Also, change the fields in `alg_match_t` and `hash_match_t` to the same?

partial/libuptiny/jsmn.c
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Creates a new parser based over a given  buffer with an array of tokens
+ * available.
+ */
+void jsmn_init(jsmn_parser *parser) {
+	parser->pos = 0;
+	parser->toknext = 0;
+	parser->toksuper = -1;
+}
+
```
patrickvacek: Formatting is a bit different on this file and the .h. Is this not getting covered by clang tools?

OYTIS: Oops. I think it's still using external submodule. Thanks!

partial/libuptiny/manifest.c
```diff
@@ -0,0 +1,114 @@
...
+    default:
+      return "Unknown";
+  }
+}
+
+/*{"attacks_detected":"<attack>","ecu_serial":"<serial>","installed_image":{"fileinfo":{"hashes":{"<hash_type>":"<hash
+ * in
+ * hex>"},"length":<length>},"filepath":"<target_name>"},"previous_timeserver_time":"1970-01-01T00:00:00Z","timeserver_time":"1970-01-01T00:00:00Z"}*/
+void uptane_write_manifest(char* signed_part, char* signatures_part) {
+  /* All strings are either literals or taken from trusted environment, use 'unsafe' functions */
```
lbonn: I'd recommend still taking the output buffer size into account. Then, with a test case using the maximum expected size in all fields, we'll get a good confidence that this will not overflow unintentionally.

OYTIS:
Rebased, hope that will help with the CI.

partial/libuptiny/jsmn.c
OYTIS: And for formatting, it's imported from an external project, so I would keep it as is for now. `jsmn.c` probably generates some style warnings as well.

patrickvacek: That sounds fine. We just need to be careful not to mess with it, then.

partial/libuptiny/manifest.c
lbonn: I don't really buy the optimization argument for this kind of check (without measurement). Also, as `strcpy` and `strcat` are strongly discouraged this day, it could end up begin the only function using them, which would in fact increase the size of the binary.

But ok, that's probably low risk here.

PR #909, opened 2018-08-15
Async secondaries support

patriotyk:


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -825,19 +837,18 @@ void SotaUptaneClient::runForever(const std::shared_ptr<command::Channel> &comma
...
+        // boost::filesystem::path reboot_flag = "/tmp/aktualizr_reboot_flag";
+        // if (boost::filesystem::exists(reboot_flag)) {
+        //   boost::filesystem::remove(reboot_flag);
+        //   if (getppid() == 1) {  // if parent process id is 1, aktualizr runs under systemd
+        //     exit(0);             // aktualizr service runs with 'Restart=always' systemd option.
+        //                          // Systemd will start aktualizr automatically if it exit.
+        //   } else {               // Aktualizr runs from terminal
+        //     LOG_INFO << "Aktualizr has been updated and requires restarting to run the new version.";
+        //   }
+        // }
```
patrickvacek: What's the story here? Can you restore this?

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -104,6 +107,11 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
 }
 
 bool ManagedSecondary::sendFirmware(const std::string &data) {
+  install_thread = std::thread(&ManagedSecondary::sendFirmware_, this, std::ref(data));
+  return true;
```
patrickvacek: Are we going to need some sort of locking here to make sure we don't try to launch two `install_thread` instances? This seems just a bit dangerous.

src/libaktualizr/utilities/events.h
```diff
@@ -97,12 +100,16 @@ class DownloadComplete : public BaseEvent {
...
+ public:
+  InstallStarted(Uptane::EcuSerial serial_in);
+  Uptane::EcuSerial serial;
+};
+
 class InstallComplete : public BaseEvent {
  public:
-  explicit InstallComplete();
+  InstallComplete(Uptane::EcuSerial serial_in);
+  Uptane::EcuSerial serial;
```
patrickvacek: We should still have doxygen-style comments for each Event class. Even if they are fairly trivial, we need to have a framework of basic documentation.

patrickvacek:
BTW I'm actually in the process of trying to remove the EventsInterpreter, but I think we will keep the Events themselves. I think the current plan of using events is good and hopefully the work of integrating this branch with my work is basically just to move the logic into SotaUptaneClient. I think your branch will be ready long before mine, so that's probably my problem to solve, not yours.

It's also worth trying to prove that this works correctly with some tests and maybe with an example using hmi_stub.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -825,19 +837,18 @@ void SotaUptaneClient::runForever(const std::shared_ptr<command::Channel> &comma
...
-        if (boost::filesystem::exists(reboot_flag)) {
-          boost::filesystem::remove(reboot_flag);
-          if (getppid() == 1) {  // if parent process id is 1, aktualizr runs under systemd
-            exit(0);             // aktualizr service runs with 'Restart=always' systemd option.
-                                 // Systemd will start aktualizr automatically if it exit.
-          } else {               // Aktualizr runs from terminal
-            LOG_INFO << "Aktualizr has been updated and requires restarting to run the new version.";
-          }
-        }
+        // boost::filesystem::path reboot_flag = "/tmp/aktualizr_reboot_flag";
```
OYTIS: Is it working without this hack now? If yes, we should delete this code.

src/libaktualizr/utilities/events.h
OYTIS: That's a good point actually. Do we want a better (or at least some) documentation policy? Right now random things get documented whenever it comes to someone's mind.

src/libaktualizr/uptane/virtualsecondary.cc
```diff
@@ -7,12 +7,17 @@
 #include "utilities/utils.h"
 
 namespace Uptane {
-VirtualSecondary::VirtualSecondary(const SecondaryConfig& sconfig_in) : ManagedSecondary(sconfig_in) {}
+VirtualSecondary::VirtualSecondary(const SecondaryConfig& sconfig_in,
+                                   std::shared_ptr<event::Channel>& events_channel_in)
+    : ManagedSecondary(sconfig_in, events_channel_in) {}
 
 bool VirtualSecondary::storeFirmware(const std::string& target_name, const std::string& content) {
+  *events_channel << std::make_shared<event::InstallStarted>(getSerial());
```
OYTIS: Cool. Can we add the same to LegacySecondary?

src/libaktualizr/utilities/events.h
patrickvacek: The current requirement, as I understand it, is that anything visible through the libaktualizr API should be documented. That's basically every function in the `Aktualizr` class and every `Event` type that we expect to emit to an API user. Maybe the `Config`. Otherwise, it doesn't matter too much. My current concern is getting this API in a usable state, so anything else is just for us. I'm open to discussion on what is useful.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -16,7 +16,7 @@
 class SecondaryAdapter : public Uptane::SecondaryInterface {
  public:
   SecondaryAdapter(const Uptane::SecondaryConfig& sconfig_in, AktualizrSecondary& sec)
-      : SecondaryInterface(sconfig_in), secondary(sec) {}
```
OYTIS: Does it make sense to pass `nullptr` as event queue? The main loop looks to be waiting for the event and won't send the report until it gets a response from everyone.

src/libaktualizr/uptane/secondaryinterface.h
```diff
@@ -19,7 +20,8 @@ namespace Uptane {
 
 class SecondaryInterface {
  public:
-  explicit SecondaryInterface(SecondaryConfig sconfig_in) : sconfig(std::move(sconfig_in)) {}
+  explicit SecondaryInterface(SecondaryConfig sconfig_in, std::shared_ptr<event::Channel> events_channel_in)
```
OYTIS: Maybe I'm being pedantic, but `events_channel_in` seems misleading to me. It took me a while to understand that `_in` is here just for the syntax sake, and it's not a queue of input events (in reality it's just the opposite).

patrickvacek: Well, I'm killing the events_interpreter anyway, so that might be a moot point. But using `_in` to denote something just being passed in doesn't seem so bad, right?

OYTIS: Well, not so bad when non-ambiguous. But if we consistenly apply the style guide (i.e. naming all class members with trailing `_`), we won't need these suffixes.

OYTIS:
BTW, looks like we still don't build OPC-UA in Jenkins. It had something to do with boost version, right?

lbonn:
@OYTIS https://github.com/advancedtelematic/aktualizr/pull/911. Looks like it was just forgotten.

src/libaktualizr/utilities/events.h
OYTIS: What's the API? Just everything that is in the headers or did we define some outwards facing subset?

patrickvacek: As far as I'm concerned, it's the aktualizr.h and by extension also event.h.

OYTIS: Makes sense, thank you.

OYTIS:
@lbonn Thanks!

src/libaktualizr/primary/sotauptaneclient.cc
patriotyk: Reboot no. I think how to deal with it

src/aktualizr_secondary/aktualizr_secondary.cc
patriotyk: SecondaryAdapter dont use event queue it is just a wrapper that uses real AktualizrSecondary implementation, passed as argument to the constructor

OYTIS: Sorry, maybe I didn't quite get it. The loop at eventsinterpreter:38 will only proceed to `PutManifest` once it's collected as many `InstallComplete` events as there were installation requests. If some of the secondaries get `nullptr` for `events_channel_in` the interpreter will never get their `InstallComplete` and will aktualizr will basically hang waiting. Am I missing something?

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -44,6 +44,7 @@ int Aktualizr::Run() {
 }
 
 void Aktualizr::AddSecondary(const std::shared_ptr<Uptane::SecondaryInterface> &secondary) {
+  secondary->events_channel = events_channel_;
```
lbonn: It feels a bit confusing that this events channel is sometimes passed through the constuctor and sometimes set later (with `nullptr` passed to the constructor). Can it made it so the constructor never sets the channel and it's always set up afterwards, possibly through a dedicated method?

src/libaktualizr/primary/eventsinterpreter.cc
```diff
@@ -29,11 +29,20 @@ std::shared_ptr<command::BaseCommand> EventsInterpreter::handle_cycle(event::Bas
   } else if (event.variant == "FetchMetaComplete") {
     return std::make_shared<command::CheckUpdates>();
   } else if (event.variant == "UpdateAvailable") {
-    return std::make_shared<command::StartDownload>(dynamic_cast<event::UpdateAvailable &>(event).updates);
+    auto upadte_available = dynamic_cast<event::UpdateAvailable &>(event);
```
lbonn: `upadte_available` instead of `update_available` in multiple places

src/aktualizr_secondary/aktualizr_secondary.cc
patriotyk: Yes, you are right, but what do you suggest? Throw exception if pointer passed to constructor is nullptr?

src/libaktualizr/primary/eventsinterpreter.cc
patriotyk: Thanks, fixed

src/libaktualizr/primary/aktualizr.cc
patriotyk: Your suggestion conflicts with Antons. I think we should disallow to pass nullptr to the constructor, but for this case we would need to add new API method getEventsChannel()

src/aktualizr_secondary/aktualizr_secondary.cc
OYTIS: Yes, either that or adding a method to the interface to tell if the secondary supports events (and event queue is not `nullptr`) and only increase ecu counter if it does.

OYTIS: Also there is a work @patrickvacek is doing to deprecate eventsinterpreter, that probably should be coordinated.

src/libaktualizr/uptane/legacysecondary.cc
```diff
@@ -32,6 +34,7 @@ bool LegacySecondary::storeFirmware(const std::string& target_name, const std::s
     // 3 - Installation failure. The previous firmware was partially overwritten or erased.
     LOG_ERROR << "Legacy external flasher install-software command failed: " << output;
   }
+  *events_channel << std::make_shared<event::InstallComplete>(getSerial());
```
patrickvacek: Can we also add the events to IpUptaneSecondary and possible OpcuaSecondary?

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -104,6 +105,14 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
 }
 
 bool ManagedSecondary::sendFirmware(const std::string &data) {
+  if (install_future.wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
+    install_future = std::async(std::launch::async, &ManagedSecondary::sendFirmware_, this, std::ref(data));
+    return true;
+  }
+  return false;
+}
```
patrickvacek: Nice. Just to make sure I understand: this checks if the installation thread is available. If so, it tries to install. If not, it aborts and returns false. Is there any sort of timeout mechanism for the installation? (We don't necessarily need that now, I'm just checking.)

patriotyk:
I just found one issue, so changed to not ready

src/libaktualizr/uptane/managedsecondary.cc
patriotyk: Yes, it just checks if thread is running or not. No we don't  have any timeout mechanisms now.

src/libaktualizr/primary/eventsinterpreter.cc
```diff
@@ -29,11 +29,20 @@ std::shared_ptr<command::BaseCommand> EventsInterpreter::handle_cycle(event::Bas
   } else if (event.variant == "FetchMetaComplete") {
     return std::make_shared<command::CheckUpdates>();
   } else if (event.variant == "UpdateAvailable") {
-    return std::make_shared<command::StartDownload>(dynamic_cast<event::UpdateAvailable &>(event).updates);
+    auto update_available = dynamic_cast<event::UpdateAvailable &>(event);
+    pending_ecus = update_available.ecus_count;
```
lbonn: Counting the number of ecus on this side seems brittle.

For example if [this condition](https://github.com/advancedtelematic/aktualizr/blob/73176f224f1dc13c88151c2e06d51dfdda2ffbb3/src/libaktualizr/primary/sotauptaneclient.cc#L1023) is taken, one `InstallationComplete` event will be missing and the machine will never go to the next state. Note: it would be good to have a test for this.

We could consider having to different events, one for each ecu complete installation and another for the complete system update. The decision of whether the system is updated completely will be computed in `sotauptaneclient.cc` and `eventsinterpreter.cc` will stay simple and mechanical.

patrickvacek:
Is there anything left to discuss or do with this PR, or is it ready for merging?

patriotyk:
Hm, looks like I forget to press "Comment" button. Yes I think it is ready.

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -122,17 +136,20 @@ bool ManagedSecondary::sendFirmware(const std::string &data) {
...
       if (boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha512digest(data))) !=
           boost::algorithm::to_lower_copy(it->HashString())) {
         detected_attack = "wrong_hash";
+        *events_channel << std::make_shared<event::InstallComplete>(getSerial());
         return true;
       }
     }
   }
   detected_attack = "";
+  *events_channel << std::make_shared<event::InstallComplete>(getSerial());
```
patrickvacek: Sorry, I think I'm missing something here. Why are we sending `InstallComplete` before `storeFirmware` is called? And why are we sending `InstalllComplete` at all in any of the error cases?

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -103,16 +104,29 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
...
-bool ManagedSecondary::sendFirmware(const std::string &data) {
+bool ManagedSecondary::sendFirmwareAsync(const std::string &data) {
+  if (!install_future.valid() || install_future.wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
+    install_future = std::async(std::launch::async, &ManagedSecondary::sendFirmware, this, data);
+    return true;
+  }
+  return false;
+}
+
+bool ManagedSecondary::sendFirmware(std::string data) {  // NOLINT(performance-unnecessary-value-param)
```
patrickvacek: Why is this passed by value?

patriotyk: Because we run sendFirmware in the thread and 'data' is destroyed before we finish installation.

src/libaktualizr/uptane/managedsecondary.cc
patriotyk: About 'InstallComplete before storeFirmware' yes you are right. 
About 'InstalllComplete' at error. It is because at error we cannot continue with install, so error means that installation is really finished, but with error. And when all ecus finished, error will be included in the ecus manifest.

src/libaktualizr/uptane/managedsecondary.cc
patriotyk: Alternatievly we could use shared_ptr.

patrickvacek: That might be a better idea. Making copies of the firmware blobs is likely to be expensive.

patrickvacek:
@patriotyk We've been tasked with writing a unit test for this logic, but I'm not sure how we can do that inside aktualizr. Do you have any ideas how we might do that? (If so, I suspect it would be something that deserves its own PR.)

patriotyk:
If we just need to test the possibility to install secondary in async mode, we could inherit the secondary interface with test secondary class and try to install package for two secondaries. But for the first install test install secondary should install secondary much longer then for the second. But the second secondary should finish first and we could test it. also after that we sould check manifest and it should include new images.

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -103,37 +104,55 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
   return true;
 }
 
-bool ManagedSecondary::sendFirmware(const std::string &data) {
+bool ManagedSecondary::sendFirmwareAsync(const std::string &data) {
+  if (!install_future.valid() || install_future.wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
+    install_future =
+        std::async(std::launch::async, &ManagedSecondary::sendFirmware, this, std::make_shared<std::string>(data));
```
patrickvacek: Isn't this still making a copy of `data` while constructing the string?

patrickvacek:
> If we just need to test the possibility to install secondary in async mode, we could inherit the secondary interface with test secondary class and try to install package for two secondaries. But for the first install test install secondary should install secondary much longer then for the second. But the second secondary should finish first and we could test it. also after that we sould check manifest and it should include new images.

Presumably you could just use the virtual secondaries, but the real question is how you would generate or simulate the packages and their metadata.

patriotyk:
> Presumably you could just use the virtual secondaries, but the real question is how you would generate or simulate the packages and their metadata.

This is not a problem, we already did it. We could copy metadata from real environment, or generate it by ourself.

patriotyk:
Looks like now it should be ok

src/libaktualizr/uptane/ipuptanesecondary.cc
```diff
@@ -54,21 +54,32 @@ bool IpUptaneSecondary::putMetadata(const RawMetaPack& meta_pack) {
   return r->result == AKInstallationResult_success;
 }
 
-bool IpUptaneSecondary::sendFirmware(const std::string& data) {
-  LOG_INFO << "Sending firmware` the secondary";
+bool IpUptaneSecondary::sendFirmwareAsync(const std::shared_ptr<std::string>& data) {
+  if (!install_future.valid() || install_future.wait_for(std::chrono::seconds(0)) == std::future_status::ready) {
+    install_future = std::async(std::launch::async, &IpUptaneSecondary::sendFirmware, this, std::ref(data));
```
lbonn: Sorry, I'm late but should it really send `std::ref` here? Looks like it will just pass around a reference to the shared_ptr, without incrementing its counter. So it could go out of scope inside `sendFirmware`

patrickvacek: @lbonn You were right. It actually crashed for me while testing. I'm fixing it in my PR.

PR #913, opened 2018-08-16
Feat/ota 327/command event restructure

pattivacek:
This is NOT expected to pass all tests yet. I just wanted to solicit feedback and verify that this path forward makes sense to other people. There are still some TODOs I want to clean up, too.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -20,7 +20,7 @@ std::shared_ptr<SotaUptaneClient> SotaUptaneClient::newDefaultClient(
     Config &config_in, std::shared_ptr<INvStorage> storage_in, std::shared_ptr<event::Channel> events_channel_in) {
   std::shared_ptr<HttpClient> http_client_in = std::make_shared<HttpClient>();
   std::shared_ptr<Uptane::Fetcher> uptane_fetcher =
-      std::make_shared<Uptane::Fetcher>(config_in, storage_in, http_client_in);
+      std::make_shared<Uptane::Fetcher>(config_in, storage_in, http_client_in, events_channel_in);
   std::shared_ptr<Bootloader> bootloader_in = std::make_shared<Bootloader>(config_in.bootloader);
   std::shared_ptr<ReportQueue> report_queue_in = std::make_shared<ReportQueue>(config_in, http_client_in);
 
@@ -33,7 +33,7 @@ std::shared_ptr<SotaUptaneClient> SotaUptaneClient::newTestClient(Config &config
                                                                   std::shared_ptr<HttpInterface> http_client_in,
                                                                   std::shared_ptr<event::Channel> events_channel_in) {
   std::shared_ptr<Uptane::Fetcher> uptane_fetcher =
-      std::make_shared<Uptane::Fetcher>(config_in, storage_in, http_client_in);
+      std::make_shared<Uptane::Fetcher>(config_in, storage_in, http_client_in, events_channel_in);
   std::shared_ptr<Bootloader> bootloader_in = std::make_shared<Bootloader>(config_in.bootloader);
   std::shared_ptr<ReportQueue> report_queue_in = std::make_shared<ReportQueue>(config_in, http_client_in);
   return std::make_shared<SotaUptaneClient>(config_in, storage_in, http_client_in, uptane_fetcher, bootloader_in,
--- a/src/libaktualizr/uptane/uptane_test.cc
+++ b/src/libaktualizr/uptane/uptane_test.cc
@@ -487,6 +487,9 @@ TEST(Uptane, PutManifest) {
 
 int num_events_FetchDownloadInstall = 0;
 void process_events_FetchDownloadInstall(const std::shared_ptr<event::BaseEvent>& event) {
+  if (event->variant == "DownloadProgressReport") {
+    return;
+  }
   switch (num_events_FetchDownloadInstall) {
     case 0:
       EXPECT_EQ(event->variant, "FetchMetaComplete");


src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -22,31 +24,40 @@
 #include "uptane/uptanerepository.h"
 #include "utilities/events.h"
 
+using EventChannelPtr = std::shared_ptr<boost::signals2::signal<void(std::shared_ptr<event::BaseEvent>)>>;
```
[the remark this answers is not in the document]
patrickvacek: I hear that. We do it all the time with various OSTree pointer objects, though. Should we refactor those as well sooner or later?

lbonn:
Some of the early exiting running modes (eg: `check`, `campaign_check`) do not cause aktualizr to exit anymore. It then goes into a `fetchMeta()` loop.

patrickvacek:
> Some of the early exiting running modes (eg: check, campaign_check) do not cause aktualizr to exit anymore. It then goes into a fetchMeta() loop.

Good call, I need to figure something out for that.

src/libaktualizr/primary/sotauptaneclient.h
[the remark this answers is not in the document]
lbonn: These are not simple `shared_ptr` to a regular type. We can try to make their intent clearer as well. For example `OstreeUniquePtr<...>`

patrickvacek:
@lbonn Still not entirely ready, but getting closer. I fixed both issues that you had mentioned.

patrickvacek:
FYI this should be code-complete, but I still need to verify a few things work and fix some of the tests that I massively broke (eventsinterpreter_test.cc and commands_test.cc). So at this point, it should be ready for review even if not quite ready for merging.

src/libaktualizr/uptane/uptane_key_test.cc
```diff
@@ -57,49 +57,54 @@ void initKeyTests(Config& config, Uptane::SecondaryConfig& ecu_config1, Uptane::
...
-  EXPECT_EQ(ecu_serials.size(), 3);
-
-  std::vector<std::string> public_keys;
-  std::vector<std::string> private_keys;
-  std::map<Uptane::EcuSerial, std::shared_ptr<Uptane::SecondaryInterface> >::iterator it;
-  for (it = sota_client->secondaries.begin(); it != sota_client->secondaries.end(); it++) {
-    if (it->second->sconfig.secondary_type != Uptane::SecondaryType::kVirtual &&
-        it->second->sconfig.secondary_type != Uptane::SecondaryType::kLegacy) {
-      continue;
+// This is a class solely for the purpose of being a FRIEND_TEST to
```
lbonn: A bit annoying but fine I guess

src/libaktualizr/utilities/events.h
```diff
@@ -116,4 +115,6 @@ class CampaignCheckComplete : public BaseEvent {
 
 }  // namespace event
 
+using EventChannel = boost::signals2::signal<void(std::shared_ptr<event::BaseEvent>)>;
```
lbonn: A bit weird to define it outside the namespace which contains all other declarations in this file. Should we rename it back to the old `events::Channel`, even if it's not the same type at all?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -676,21 +674,23 @@ bool SotaUptaneClient::downloadImages(const std::vector<Uptane::Target> &targets
...
       return false;
     }
   }
   if (!targets.empty()) {
     if (targets.size() == downloaded_targets.size()) {
-      *events_channel << std::make_shared<event::DownloadComplete>(downloaded_targets);
+      sendEvent(std::make_shared<event::DownloadComplete>(downloaded_targets));
       sendDownloadReport();
+      if (config.uptane.running_mode == RunningMode::kFull || config.uptane.running_mode == RunningMode::kOnce) {
+        uptaneInstall(downloaded_targets);
```
lbonn: That's still a bit annoying that all these operations link to each others internally depending on the running mode, which obfuscates the flow control.

Do you think it would be possible to have an `uptaneCycle()` which would subscribe to events internally to determine the next operation. `RunninMode::kOnce` would just call this once, `kFull` in a loop and the other single ops would just call other operations à la carte.

End-goal: no reading of `config.uptane.running_mode` in `sotauptaneclient.cc`.

src/libaktualizr/uptane/uptane_key_test.cc
patrickvacek: Agreed, but what is the superior alternative?

src/libaktualizr/utilities/events.h
patrickvacek: > A bit weird to define it outside the namespace which contains all other declarations in this file.

Agreed. I didn't want to define it here, since nothing else in `events.h` since nothing else in that file has anything to do with signals and such, but I couldn't find a better place to do it.

> Should we rename it back to the old events::Channel, even if it's not the same type at all?

Not a bad idea. Doesn't matter if it isn't the same type; I got rid of the old `Channel` class, so it should be unambiguous after this point.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: > That's still a bit annoying that all these operations link to each others internally depending on the running mode, which obfuscates the flow control.

Agreed, and I think your end-goal is reasonable. I will consider the `uptaneCycle` option or do something similar.

src/libaktualizr/uptane/uptane_key_test.cc
lbonn: Haven't found it :), let's live with it

patrickvacek: If it makes you feel better (it won't), I had to already do the same type of thing with `uptane_test_vectors.cc`.

src/libaktualizr/primary/CMakeLists.txt
```diff
@@ -1,23 +1,21 @@
...
-            eventsinterpreter.h
             initializer.h
             reportqueue.h
             sotauptaneclient.h)
 
 
 add_library(primary OBJECT ${SOURCES})
 
-add_aktualizr_test(NAME commands SOURCES commands_test.cc)
+# TODO: remove?:
```
lbonn: Can't we remove them already?

patrickvacek: Hahaha, yeah, but I want to keep the idea of some of those tests around still, but they have to be rewritten. I'm looking into that now. Could be a job for a follow-up PR.

patrickvacek:
Urgh, hold on, something went wrong while rebasing four times...

lbonn:
@patrickvacek yes, no problem for fast-tracking this. I'll just push one last bit on top before :)

patrickvacek:
Fixed the rebase mistake. @lbonn what do you want to add on top?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -21,7 +20,7 @@ std::shared_ptr<SotaUptaneClient> SotaUptaneClient::newDefaultClient(
     Config &config_in, std::shared_ptr<INvStorage> storage_in, std::shared_ptr<event::Channel> events_channel_in) {
   std::shared_ptr<HttpClient> http_client_in = std::make_shared<HttpClient>();
   std::shared_ptr<Uptane::Fetcher> uptane_fetcher =
-      std::make_shared<Uptane::Fetcher>(config_in, storage_in, http_client_in, events_channel_in);
+      std::make_shared<Uptane::Fetcher>(config_in, storage_in, http_client_in);
```
patrickvacek: Crap, another mistake!

PR #918, opened 2018-08-21
Clean the docker mess a bit

lbonn:
Do not use `.nop11`, `noostree` docker images anymore, let's settle with
per-distro dockerfiles and trust our CMake for not linking with extra
libs. We are reasonably confident that `cmake -DBUILD_OSTREE=off` does
not try to link with Ostree and keeping all these files up to date has
been costly compared to the benefits.

Also, move everything to a dedicated directory to clean up the root of
the project.

Not ready, I want to make a ci run first, then will push some more changes.


lbonn:
Actually, I already wanted to switch to 18.04 in this PR but some valgrind errors will need to be fixed first. So, we could come back to that later.

scripts/build_ubuntu.sh
```diff
@@ -12,15 +12,16 @@ export TEST_WITH_OSTREE=0
...
+cp -rf "$GITREPO_ROOT/tests/config/selfupdate.toml" "$TEST_INSTALL_DESTDIR"
+cp -rf "$GITREPO_ROOT/scripts/selfupdate_server.py" "$TEST_INSTALL_DESTDIR"
+cp -rf "$GITREPO_ROOT/tests/test_data/fake_root" "$TEST_INSTALL_DESTDIR"
+cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/create_repo.sh" "$TEST_INSTALL_DESTDIR"
+cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/serve_repo.py" "$TEST_INSTALL_DESTDIR"
 
-git -C "${GITREPO_ROOT}" fetch --unshallow || true
-git -C "${GITREPO_ROOT}" describe > "${TEST_INSTALL_DESTDIR}/aktualizr-version"
+git -C "$GITREPO_ROOT" fetch --unshallow || true
+git -C "$GITREPO_ROOT" describe > "$TEST_INSTALL_DESTDIR/aktualizr-version"
```
patrickvacek: Any reason for removing the braces? I get that you probably don't need them here, but were they hurting something?

lbonn:
@patrickvacek It ran on a previous version of this PR and I had to make an adjustment afterwards. It's highly probable that I broke it actually, something about yaml format...

lbonn:
@patrickvacek sure, here is your one-liner:

    ./scripts/run_docker_test.sh docker/Dockerfile.ubuntu.bionic -eTEST_INSTALL_DESTDIR=$PWD -- ./scripts/build_ubuntu.sh

PR #919, opened 2018-08-21
Feat/ota 408/campaign more meta

lbonn:
`aktualizr --running-mode campaign_accept --campaign-id xxxx`
~`aktualizr --running-mode campaign_reject --campaign-id xxxx`~

### DIFF
--- a/src/libaktualizr/primary/aktualizr.h
+++ b/src/libaktualizr/primary/aktualizr.h
@@ -45,11 +45,11 @@ class Aktualizr {
   void CampaignCheck();
 
   /**
-   * Asynchronously accept or reject a campaign for the current device
+   * Asynchronously accept a campaign for the current device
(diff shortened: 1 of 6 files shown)


src/libaktualizr/primary/aktualizr.h
```diff
@@ -44,6 +44,13 @@ class Aktualizr {
    */
   void CampaignCheck();
 
+  /**
+   * Asynchronously accept or reject a campaign for the current device
```
patrickvacek: You can reject with this function?

lbonn:
@patrickvacek currently, rejection is not handled server-side. As I see it, if an user clicks Reject, the update client will have to add the affected campaign to some list in local db to avoid nagging the user all the time.

In contrast, Accept does something and makes the update visible to the client through the director server.

PR #921, opened 2018-08-22
Fix uptane_vector_tests to work properly

patriotyk:
This PR should pass all tests after rebasing tuf-vector-tests with my changes

### DIFF
--- a/src/libaktualizr/uptane/fetcher.cc
+++ b/src/libaktualizr/uptane/fetcher.cc
@@ -24,7 +24,7 @@ static size_t DownloadHandler(char* contents, size_t size, size_t nmemb, void* u
   uint64_t downloaded = size * nmemb;
   auto expected = static_cast<uint64_t>(ds->target.length());
   if ((ds->downloaded_length + downloaded) > expected) {
-    return ds->downloaded_length + downloaded;  // curl will abort if return unexpected size;
+    return downloaded + 1;  // curl will abort if return unexpected size;
   }
 
   // incomplete writes will stop the download (written_size != nmemb*size)
--- a/tests/uptane_vector_tests.cc
+++ b/tests/uptane_vector_tests.cc
@@ -45,67 +45,100 @@ class Uptane_Vector_Test {
   }
 
  public:
-  static bool run_test(const std::string& test_name, const Json::Value& vector, const std::string& port) {
-    std::cout << "VECTOR: " << vector;
+  static bool run_test(const std::string& test_name, const std::string& address) {
     TemporaryDirectory temp_dir;
     Config config;
     config.provision.primary_ecu_serial = "test_primary_ecu_serial";
     config.provision.primary_ecu_hardware_id = "test_primary_hardware_id";
-    config.uptane.director_server = "http://127.0.0.1:" + port + "/" + test_name + "/director";
-    config.uptane.repo_server = "http://127.0.0.1:" + port + "/" + test_name + "/image_repo";
+    config.uptane.director_server = address + test_name + "/director";
+    config.uptane.repo_server = address + test_name + "/image_repo";
     config.storage.path = temp_dir.Path();
-    config.storage.uptane_metadata_path = BasedPath(port + "/aktualizr_repos");
+    config.storage.uptane_metadata_path = BasedPath(address + "aktualizr_repos");
     config.pacman.type = PackageManager::kNone;
+
+    auto storage = INvStorage::newStorage(config.storage);
+    Uptane::Manifest uptane_manifest{config, storage};
+    auto uptane_client = SotaUptaneClient::newDefaultClient(config, storage);
+    Uptane::EcuSerial ecu_serial(config.provision.primary_ecu_serial);
+    Uptane::HardwareIdentifier hw_id(config.provision.primary_ecu_hardware_id);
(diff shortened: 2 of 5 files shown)


src/libaktualizr/uptane/fetcher.cc
```diff
@@ -24,7 +24,7 @@ static size_t DownloadHandler(char* contents, size_t size, size_t nmemb, void* u
   uint64_t downloaded = size * nmemb;
   auto expected = static_cast<uint64_t>(ds->target.length());
   if ((ds->downloaded_length + downloaded) > expected) {
-    return ds->downloaded_length + downloaded;  // curl will abort if return unexpected size;
+    return downloaded + 1;  // curl will abort if return unexpected size;
```
patrickvacek: What was the reason for changing this line? Just catching the case where `ds->downloaded_length` was somehow 0?

tests/uptane_vector_tests.cc
```diff
@@ -45,67 +45,100 @@ class Uptane_Vector_Test {
...
     Config config;
     config.provision.primary_ecu_serial = "test_primary_ecu_serial";
     config.provision.primary_ecu_hardware_id = "test_primary_hardware_id";
-    config.uptane.director_server = "http://127.0.0.1:" + port + "/" + test_name + "/director";
-    config.uptane.repo_server = "http://127.0.0.1:" + port + "/" + test_name + "/image_repo";
+    config.uptane.director_server = address + test_name + "/director";
+    config.uptane.repo_server = address + test_name + "/image_repo";
     config.storage.path = temp_dir.Path();
-    config.storage.uptane_metadata_path = BasedPath(port + "/aktualizr_repos");
+    config.storage.uptane_metadata_path = BasedPath(address + "aktualizr_repos");
```
patrickvacek: +1 for reusing `address` as much as possible and avoiding the duplicated parts, but does this actually work for the `uptane_metadata_path`? Would it be better if were actually be a subdir of `temp_dir`?

tests/uptane_vector_tests.cc
```diff
@@ -45,67 +45,100 @@ class Uptane_Vector_Test {
...
     config.pacman.type = PackageManager::kNone;
+
+    auto storage = INvStorage::newStorage(config.storage);
+    Uptane::Manifest uptane_manifest{config, storage};
+    auto uptane_client = SotaUptaneClient::newDefaultClient(config, storage);
+    Uptane::EcuSerial ecu_serial(config.provision.primary_ecu_serial);
+    Uptane::HardwareIdentifier hw_id(config.provision.primary_ecu_hardware_id);
+    uptane_client->hw_ids.insert(std::make_pair(ecu_serial, hw_id));
+    uptane_client->installed_images[ecu_serial] = "test_filename";
     logger_set_threshold(boost::log::trivial::trace);
```
patrickvacek: `logger_set_threshold` only needs to be set once in `main`, right? Might as well do it first thing.

tests/uptane_vector_tests.cc
```diff
@@ -45,67 +45,100 @@ class Uptane_Vector_Test {
...
-      for (Json::ValueConstIterator it = vector["director"]["targets"].begin();
-           it != vector["director"]["targets"].end(); ++it) {
-        if (!(*it)["is_success"].asBool()) {
-          std::cout << "aktualizr did not fail as expected.\n";
+      try {
+        if (!uptane_client->uptaneIteration()) {
+          if (should_fail) {
+            throw uptane_client->getLastException();
+          }
+          std::cout << "aktualizr return error but should not\n";
```
patrickvacek: Sorry to be pedantic, but just to make this extra clear: "uptaneIteration failed but should have succeeded".

tests/uptane_vector_tests.cc
```diff
@@ -45,67 +45,100 @@ class Uptane_Vector_Test {
...
+        std::vector<Uptane::Target> updates;
+        if (!uptane_client->uptaneOfflineIteration(&updates, nullptr)) {
+          std::cout << "uptaneOfflineIteration returns false\n";
           return false;
+        } else {
+          if (updates.size()) {
+            if (!uptane_client->downloadImages(updates)) {
+              if (should_fail) {
+                throw uptane_client->getLastException();
+              }
```
patrickvacek: There should be another check here, right? If `downloadImages` failed but we didn't expect it, we should say something.

src/libaktualizr/uptane/fetcher.cc
patriotyk: Here we should return value differnt to 'downloaded', to abort download. If event calls first time, ds->downloaded_length is 0, and if variable 'downloaded' is greater then 'expected' we returns  '0 + downloaded'

patrickvacek:
Rebased and bumped the tuf-test-vectors version. My remaining comments are still open, though, so this is probably not quite ready for merging.

tests/uptane_vector_tests.cc
patrickvacek: I was wrong. Since it gets reset by constructing a Config, it actually does to be reset. That means, however, that most of our tests do not print full logging information.

patrickvacek:
I am satisfied with this PR now after cleaning up just a few final things. Any objections to finally merging this?

PR #924, opened 2018-08-24
Fix the `aktualizr --version` output

lbonn:
I have no fix yet, just more helpful logging in the meantime.


lbonn:
Ok, it was actually one of these git shallow clone issues... So does not require any code change, it worked after messing with Jenkins git settings (and don't do shallow clones!)

PR #926, opened 2018-08-28
Bugfix/garage deploy dryrun

patrickvacek:


src/sota_tools/garage_deploy.cc
```diff
@@ -91,18 +98,23 @@ int main(int argc, char **argv) {
...
       return EXIT_FAILURE;
     }
 
-    if (push_credentials.CanSignOffline()) {
-      bool ok = OfflineSignRepo(ServerCredentials(push_credentials.GetPathOnDisk()), name, commit, hardwareids);
-      return static_cast<int>(!ok);
-    }
-    LOG_FATAL << "Online signing with garage-deploy is currently unsupported";
-    return EXIT_FAILURE;
+    return EXIT_SUCCESS;
```
lbonn: Was this used for debugging? The code from line 108 looks unreachable.

patrickvacek: Crap, that was supposed to be removed. Fixed.

src/sota_tools/ostree_http_repo.cc
```diff
@@ -40,23 +42,49 @@ OSTreeRef OSTreeHttpRepo::GetRef(const std::string &refname) const { return OSTr
...
   server_->InjectIntoCurl(path.string(), easy_handle);
   curl_easy_setopt(easy_handle, CURLOPT_WRITEFUNCTION, &OSTreeHttpRepo::curl_handle_write);
   boost::filesystem::create_directories((root_ / path).parent_path());
-  int fp = open((root_ / path).c_str(), O_WRONLY | O_CREAT, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH);
+  std::string filename = (root_ / path).string();
+  int fp = open(filename.c_str(), O_WRONLY | O_CREAT, S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH);
+  if (fp == -1) {
+    LOG_ERROR << "Failed to open file: " << filename;
+    ret = false;
+    goto cleanup;
```
patrickvacek: @lbonn I should complain about this goto since it only saves you one indention.

lbonn: Yes, I actually like this kind of goto sometimes but since that's the only one in the C++ part of the code base, let's remove it :).

patrickvacek: It's true, sometimes it's useful when you've got a quadruply-nested condition or something. But I don't think this situation is bad enough to warrant it.

src/sota_tools/ostree_http_repo.cc
```diff
@@ -40,23 +42,48 @@ OSTreeRef OSTreeHttpRepo::GetRef(const std::string &refname) const { return OSTr
 
 OSTreeObject::ptr OSTreeHttpRepo::GetObject(const uint8_t sha256[32]) const { return GetObject(OSTreeHash(sha256)); }
 
+bool failure = false;
```
lbonn: Is it used somewhere?

patrickvacek: My bad, another debugging/rebasing holdover. Fixed.

src/sota_tools/ostree_http_repo.cc
```diff
@@ -41,22 +43,45 @@ OSTreeRef OSTreeHttpRepo::GetRef(const std::string &refname) const { return OSTr
 OSTreeObject::ptr OSTreeHttpRepo::GetObject(const uint8_t sha256[32]) const { return GetObject(OSTreeHash(sha256)); }
 
 bool OSTreeHttpRepo::Get(const boost::filesystem::path &path) const {
+  bool ret = false;
+  CURLcode err = CURLE_OK;
   CURL *easy_handle = curl_easy_init();
```
patrickvacek: Is it worth checking if this returns `NULL`? I would expect that `curl_easy_perform` would break, but not sure about the `curl_easy_setopt` and `InjectIntoCurl` calls yet.

lbonn: Yes, any of these can fail in theory...

patrickvacek: I mean, if something causes an exception, that's sort of okay, in that we'd hear about the problem immediately. But the potential for silent failures concerns me more.

PR #930, opened 2018-08-31
Bugfix/ota 682/rm local tuf

patrickvacek:


src/sota_tools/deploy.cc
```diff
@@ -57,12 +57,18 @@ bool UploadToTreehub(const OSTreeRepo::ptr &src_repo, const ServerCredentials &p
...
-    std::string init_cmd("garage-sign init --repo aktualizr --credentials ");
-    if (system((init_cmd + push_credentials.GetPathOnDisk().string()).c_str()) != 0) {
-      LOG_ERROR << "Could not initilaize tuf repo for sign";
-      return false;
-    }
+  const boost::filesystem::path local_repo{"./tuf/aktualizr"};
+
+  // OTA-682: Do NOT keep the local tuf directory around in case the user tries
+  // a different set of push credentials.
+  if (boost::filesystem::is_directory(local_repo)) {
```
lbonn: Should it just be a temporary directory, since it's pretty ephemeral?

patrickvacek: That would be great, but that's up to garage-sign. As far as I can tell, `garage-sign init` only gives you the `--repo` option, but whatever you provide is actually used as a subdirectory of `$PWD/tuf/`. I guess we could change the working directory before running these commands. Is that worth it?

lbonn: Ok, makes sense.

PR #932, opened 2018-09-03
Minor improvements to garage-push.

patrickvacek:
* Require credentials, no longer silently accept an undocumented default
option. Fix associated tests.
* Remove an unused getenv system call.
* More const-correctness and error checking.

Note that the return code from `curl_easy_setopt` is almost never checked. Is that worth changing?

### DIFF
--- a/src/sota_tools/ostree_object.cc
+++ b/src/sota_tools/ostree_object.cc
@@ -2,6 +2,7 @@
 
 #include <assert.h>
 
+#include <cstring>
 #include <iostream>
 
 #include <glib.h>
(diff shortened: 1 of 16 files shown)


lbonn:
`curl_easy_setopt` has failure modes depending on which option (some never fail). The most usual case is failed allocation when it needs to copy a string.

We could use a quick wrapper that just throw a fatal exception in case of error.

patrickvacek:
> We could use a quick wrapper that just throw a fatal exception in case of error.

Working on it.

src/sota_tools/ostree_object.cc
```diff
@@ -235,15 +236,37 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
...
   assert(refcount_ > 0);  // At least our parent should have a reference to us
 
+  char *url = NULL;
+  curl_easy_getinfo(curl_handle_, CURLINFO_EFFECTIVE_URL, &url);
   long rescode = 0;  // NOLINT
   curl_easy_getinfo(curl_handle_, CURLINFO_RESPONSE_CODE, &rescode);
   if (current_operation_ == CurrentOp::kOstreeObjectPresenceCheck) {
-    if (rescode == 200) {
-      LOG_INFO << "Already Present:" << object_name_;
+    if (!url || !strstr(url, object_name_.c_str())) {
```
lbonn: You're trying to see if the server sent us a redirect?

patrickvacek: Nah, just verifying that the object the server found matches the object we have locally.

patrickvacek:
I'm confused... nothing I changed should affect uptane_vector_tests, and indeed, when run the tests on master now, it breaks on the exact same tests. What happened?

patrickvacek:
Never mind, my mistake. I accidentally changed the version of tuf-test-vectors.

patrickvacek:
@lbonn Rebased and added comments. Is it enough?

PR #935, opened 2018-09-06
OTA-194 removing conditional filtering for GitHub which is now unnece…

merltron:
Removing conditional filtering for GitHub users which is now unnecessary. We have stopped including content from this page in the main docs. Instead we have created a cleaner copy of that content.


tkfu:
We're maintaining it separately--basically, copy-pasted for now. We decided that it didn't really make sense to have so much content here that can only actually be used by the docs.

patrickvacek:
@merltron Sorry to keep making you work for this, but do you mind rebasing on master? Those extra four redundant commits shouldn't be there.

merltron:
@patrickvacek ah OK, sorry. Have another look. I rebased on master and pushed again.

PR #936, opened 2018-09-06
Clear installation status after sending it once

OYTIS:


OYTIS:
@patrickvacek Does anything specific come to your mind? We could add unit tests for what it is possible and leave the rest manual QA as a part of meta-updater testing,

patrickvacek:
@OYTIS I'm not too worried about unit tests for this, more just that we verify that installation of multiple packages in sequence works and that recovery after a failed installation is easy.

PR #941, opened 2018-09-10
Build/clang update

lbonn:


src/libaktualizr/crypto/crypto.cc
```diff
@@ -354,7 +354,7 @@ StructGuard<EVP_PKEY> Crypto::generateRSAKeyPairEVP(KeyType key_type) {
   ret = RSA_generate_key_ex(rsa.get(), bits, /* number of bits for the key - 2048 is a sensible value */
                             bne.get(),       /* exponent - RSA_F4 is defined as 0x10001L */
                             nullptr          /* callback argument - not needed in this case */
-                            );
+  );
```
patrickvacek: This seems ugly to me. What's the deal? (Not worth sweating about, though.)

lbonn: Yes, that's true. I tweaked the comment and line jump so clang-format don't make it so weird.

lbonn:
I've pushed another commit of documentation about the change in our development process.

PR #943, opened 2018-09-12
Libuptiny demo

OYTIS:
OK, so it's finally there. While not really spectacular, it tests the full cycle of Uptane update. Integration with RIOT bootloader follows once it is done in RIOT. ed25519 library should be replaced with something more space-effective, but works for demo purposes. Clean-up on libuptiny APIs is following.


.gitmodules
```diff
@@ -13,3 +13,7 @@
 [submodule "third_party/googletest"]
 	path = third_party/googletest
 	url = https://github.com/google/googletest.git
+[submodule "RIOT"]
+	path = partial/extern/RIOT
+	url = git@github.com:OYTIS/RIOT.git
```
lbonn: Should be an https url, so that others can freely clone it.

OYTIS: Oops. Thank you!

src/libaktualizr/isotp_conn/isotp_send_recv.cc.bu
```diff
@@ -0,0 +1,324 @@
+#include <iostream>
```
lbonn: What is this file with a `cc.bu` extension?

OYTIS: #@$#%!
Thanks! I'll delete that. I need to split the PR as well to make it observable.

src/libaktualizr/uptane/secondaryconfig.h
```diff
@@ -23,6 +23,8 @@ enum class SecondaryType {
   kIpUptane,  // Custom Uptane protocol over TCP/IP network
 
   kVirtualUptane,  // Partial UPTANE secondary implemented inside primary
+  kIsoTpUptane,    // Custom Uptane protocol over ISO/TP network
```
patrickvacek: Maybe worth specifying what makes this different than `kIpUptane`.

partial/libuptiny-demo/main.c
```diff
@@ -0,0 +1,205 @@
...
+			isotp_buf[0] = UPTANE_PUT_IMAGE_CHUNK_ACK_ERR;
+			isotp_buf[1] = 0x00;
+			conn_can_isotp_send(&conn_isotp, &isotp_buf, 2, CAN_ISOTP_TX_DONT_WAIT);
+			break;
+		default:
+			break;
+	}
+      }
+    }
+}
```
patrickvacek: Some funky formatting here.

patrickvacek:
Also, any thoughts why the LGTM C++ check failed? This bit concerns me: `Running command: 'git --git-dir=/work/working/private/src/.git fetch --no-tags --recurse-submodules=no`. Why isn't it recursing over the submodules? That seems like it might be important.

lbonn:
@patrickvacek here is a similar report: https://discuss.lgtm.com/t/submodules-cause-checkout-to-fail/1666/2

Not much we can do about that I'm afraid

patrickvacek:
At least it looks like the problem is temporary. I tried to restart the analysis, but I can't figure out how to do it. I probably don't have the rights to do it.

OYTIS:
@patrickvacek @lbonn 
I've made ISO/TP functionality optional. I'm not particularly excited about having more conditional compilation, but on the other hand reducing attack surface and binary size are probably more important. Not sure if there can be a solution not involving conditional compilation, probably not (or will require inadequate effort).

OYTIS:
@patrickvacek As for testing, there are of course unit tests for the library (please point out if something is missing). For the ISO/TP demo it would be not as easy of course. One possible solution would be to use virtual CAN device and a native RIOT build.

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "third_party/googletest"]
 	path = third_party/googletest
 	url = https://github.com/google/googletest.git
+[submodule "partial/extern/RIOT"]
+	path = partial/extern/RIOT
+	url = https://github.com/OYTIS/RIOT.git
```
patrickvacek: Is it likely that you will be able to upstream your changes? Otherwise it may be better to put this on an `advancedtelematic` fork.

CMakeLists.txt
```diff
@@ -399,6 +406,11 @@ foreach(FILE ${ALL_SOURCE_FILES})
       continue()
     endif()
 
+    string (FIND ${FILE} "/isotp_conn/" EXCLUDE_DIR_FOUND)
+    if (NOT ${EXCLUDE_DIR_FOUND} EQUAL -1)
+      continue()
+    endif()
```
patrickvacek: What is this doing?

PR #945, opened 2018-09-14
Do not rely on FS->SQL migration for everyday provisioning

lbonn:
Note that `installed_versions` is now passed through `/var/sota/import` instead of `/var/sota`. meta-updater PR on the way.


patrickvacek:
+1 for cleaning so much stuff up. Looks good, testing now.

lbonn:
Pushed another commit which takes care of an important property of the import function.

PR #947, opened 2018-09-17
Try a few different locations for finding the p11 engine library.

patrickvacek:
If a path is provided, try that. If that doesn't work, try the new path
for openssl-1.1. If that doesn't work, try the old default location.

Also add a test.

This is part of the work to get HSM support working on master again. meta-updater PR is at https://github.com/advancedtelematic/meta-updater/pull/391.


patrickvacek:
Why does clang-format-6.0 break the alphabetical ordering of external headers? I don't get it. The old clang-format didn't do that.

PR #949, opened 2018-09-18
Don't try to read installed_versions is we don't have read permission.

patrickvacek:
This fixes the case when you are testing locally and /var/sota is owned by root.


src/libaktualizr/storage/invstorage.cc
```diff
@@ -276,10 +277,10 @@ void INvStorage::FSSToSQLS(FSStorageRead& fs_storage, SQLStorage& sql_storage) {
 std::string INvStorage::fsReadInstalledVersions(const boost::filesystem::path& filename,
                                                 std::vector<Uptane::Target>* installed_versions) {
   std::string current_hash;
-  if (!boost::filesystem::exists(filename)) {
+  if (access(filename.c_str(), R_OK) != 0 || !boost::filesystem::exists(filename)) {
```
lbonn: `access(., R_OK)` also checks if the file exists, according to the man page.

PR #950, opened 2018-09-20
Test/req trace map

patrickvacek:
This is not ready for merging yet, but rather a draft for demonstration.


xaviergully:
I don't know why but OTA-603 and OTA-601 are not valid tickets (or they have been deleted)

src/libaktualizr/uptane/uptane_implicit_test.cc
```diff
@@ -118,10 +120,10 @@ TEST(UptaneImplicit, ImplicitIncomplete) {
 }
 
 /**
- * \verify{\tst{186}} Verify that aktualizr can implicitly provision with
- * provided credentials.
+ * Verify that aktualizr can implicitly provision with provided credentials.
  */
 TEST(UptaneImplicit, ImplicitProvision) {
+  RecordProperty("zephyr_key", "OTA-601,REQ-349,TST-186");
```
xaviergully: OTA-601 does not exist

patrickvacek:
I think this is ready for review now. Basically, I just reviewed the tests that already were linked to Jira tickets and verified that they did what we thought they did. In many cases, they did not, so I fixed them. I also tried to think about how to future-proof them so that they would fail instead of silently stop doing what we hoped they would do. This makes the tests more complicated but should mean we have better confidence in them. In some cases, I just added checks that after failing for a specific reason, the test would succeed when that one reason was "fixed". In other cases, I made sure that specific parts of the code were being hit.

lbonn:
I pushed an additional commit to move the results inside CMake's build directory, to stay consistent with our other generated files.

patrickvacek:
@lbonn Thanks for helping clean this up. I've rebased and added another minor administrative fix. If it looks good to you, I think it is ready to merge.

PR #951, opened 2018-09-21
Added async image downloads

patriotyk:


patrickvacek:
@patriotyk Just to reiterate what I said at the meeting: at a certain level you will of course need to update some of the tests, particularly the aktualizr_test cases. That's no problem. The goal is to make sure that no things still work to our expectations, even though from an external perspective (almost) nothing will change.

As I think you've already figured out, Laurent refactored how the async installation works just a bit, and that's probably a good model for you to follow. Although installation on the primary is currently not asynchronous (it probably should be eventually), we should make all downloads (even for the primary) now.

patriotyk:
Looks like I fixed it, now tests should pass. But there is another issue. When we write chunks in separate transaction there is not possible to rollback whole downloaded image on some error, so we need to remove corrupted image after error. So I am working with it

src/libaktualizr/utilities/events.h
```diff
@@ -117,6 +117,18 @@ class DownloadComplete : public BaseEvent {
   static DownloadComplete fromJson(const std::string& json_str);
 };
 
+/**
+ * An target has been successfully downloaded.
```
patrickvacek: Minor: "*A* target has been..."

But while I'm thinking about it, it would be nice if the event names for download and install were similar for one vs all. I like the sound of `DownloadTargetComplete`, but maybe `DownloadComplete` should be `AllDownloadsComplete` to match the installation. (Similarly, `InstallComplete` should perhaps be renamed `InstallTargetComplete`.)

eu-smirnov:
Tested with with Sachertorte, works after adding DownloadTargetComplete event.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -669,6 +669,7 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets, u
 bool SotaUptaneClient::downloadImages(const std::vector<Uptane::Target> &targets) {
   // Uptane step 4 - download all the images and verify them against the metadata (for OSTree - pull without
   // deploying)
+  download_futures_.clear();
```
lbonn: Why does it need to be a member variable? It seems to be local to one method.

Or should it actually be fully asynchronous, like the install case (calling the complete event from a thread)?

patriotyk: I think you are right. Currently it is used only locally inside this method. And in the future if we would need wait for download finish from somewhere else, we could move it back.

PR #952, opened 2018-09-21
Changes to provisioning configs for delayed provisioning

OYTIS:
- sota_implicit_prov is deprecated
- HSM provisioning should not import certificate and private key, they
belong to HSM, not to storage
- All the imported data should be under /var/sota/import
- Make cert provider respect path to import directory

### DIFF
--- a/config/sota_hsm_prov.toml
+++ b/config/sota_hsm_prov.toml
@@ -1,4 +1,5 @@
 [tls]
+server_url_path = "/var/sota/gateway.url"
 cert_source = "pkcs11"
 pkey_source = "pkcs11"
 
@@ -18,5 +19,3 @@ type = "sqlite"
 [import]
 base_path = "/var/sota/import"
 tls_cacert_path = "root.crt"
-tls_clientcert_path = "client.pem"
-tls_pkey_path = "pkey.pem"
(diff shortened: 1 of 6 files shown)


config/sota_hsm_prov.toml
```diff
@@ -18,5 +19,3 @@ type = "sqlite"
 [import]
 base_path = "/var/sota/import"
 tls_cacert_path = "root.crt"
-tls_clientcert_path = "client.pem"
-tls_pkey_path = "pkey.pem"
```
patrickvacek: You're removing these because aktualizr won't need them, but rather the testtoken script, right?

config/sota_implicit_prov_ca.toml
```diff
@@ -5,6 +5,7 @@ server_url_path = "/var/sota/gateway.url"
 type = "sqlite"
 
 [import]
-tls_cacert_path = "/var/sota/import_root.crt"
-tls_clientcert_path = "/var/sota/import_client.pem"
-tls_pkey_path = "/var/sota/import_pkey.pem"
+base_path = "/var/sota/import"
+tls_cacert_path = "root.crt"
+tls_clientcert_path = "client.pem"
+tls_pkey_path = "pkey.pem"
```
patrickvacek: +1

config/sota_hsm_prov.toml
OYTIS: Yes, absolutely. `[import]` section is for importing stuff into storage, and our certificate/private key are in HSM, not in storage.

OYTIS:
@patrickvacek I think I've fixed the Jenkins issue, waiting for the result.

PR #954, opened 2018-09-27
Use upstream version of HdrHistogram

lbonn:
Now includes Phil's fix! Using the old repo with a commit outside master was causing issues with bitbake.


patrickvacek:
+1, I was waiting for this to happen!

PR #955, opened 2018-10-01
More tests for garage-deploy

patriotyk:


lbonn:
HdrHistogram_c has been rewound to the version before #954, probably by mistake. Could you set it back to the current version?

src/sota_tools/CMakeLists.txt
```diff
@@ -142,15 +142,10 @@ if (BUILD_SOTA_TOOLS)
...
 
-    # Detect that the thing we passed in to --repo is not a repository
-    add_test(NAME invalid-repo
-        COMMAND ${PROJECT_SOURCE_DIR}/tests/sota_tools/test-invalid-repo $<TARGET_FILE:garage-push>
-        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
-
-    # Detect a --repo that isn't in 'archive-z2' mode
-    add_test(NAME bare-mode-repo
-        COMMAND ${PROJECT_SOURCE_DIR}/tests/sota_tools/test-bare-mode-repo $<TARGET_FILE:garage-push>
-        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/tests/)
```
patrickvacek: I definitely support moving some of these scripts into proper Google Test cases. But if these old test scripts are no longer necessary, you should also delete them.

src/sota_tools/ostree_dir_repo_test.cc
```diff
@@ -0,0 +1,76 @@
...
+#include "ostree_ref.h"
+
+TEST(dir_repo, invalid_path) {
+  OSTreeRepo::ptr src_repo = std::make_shared<OSTreeDirRepo>("invalid_path");
+  EXPECT_FALSE(src_repo->LooksValid());
+}
+
+TEST(dir_repo, invalid_config) {
+  TemporaryDirectory temp_dir;
+  int result = system((std::string("cp -rf tests/sota_tools/repo ") + temp_dir.Path().c_str()).c_str());
```
lbonn: `Utils::copyDir` is a bit cleaner

lbonn:
Seems great, approved when CI will be green (test_treehub_server times out)

src/sota_tools/treehub_server_test.cc
```diff
@@ -0,0 +1,74 @@
...
+  curlEasySetoptWrapper(curl_handle.get(), CURLOPT_WRITEDATA, static_cast<void *>(&response));
+  curl_easy_perform(curl_handle.get());
+
+  long rescode;  // NOLINT
+  curl_easy_getinfo(curl_handle.get(), CURLINFO_RESPONSE_CODE, &rescode);
+  if (rescode == 200) {
+    auto response_json = Utils::parseJSON(response);
+    EXPECT_EQ(response_json["Authorization"], "Bearer test_token");
+  } else {
+    FAIL();
```
patrickvacek: If you must use `FAIL()`, I recommend adding some output text so we can get some idea what went wrong. You can just do something like `FAIL() << "Failed to authenticate token with server";`. Alternatively, since we are basically requiring `rescode == 200`, you might as well just `EXPECT_EQ` it.

src/sota_tools/deploy_test.cc
```diff
@@ -0,0 +1,55 @@
...
+#include "crypto/crypto.h"
+#include "deploy.h"
+#include "ostree_dir_repo.h"
+#include "ostree_ref.h"
+#include "test_utils.h"
+std::string port = "2443";
+TemporaryDirectory temp_dir;
+
+TEST(deploy, UploadToTreehub) {
+  OSTreeRepo::ptr src_repo = std::make_shared<OSTreeDirRepo>("tests/sota_tools/repo");
```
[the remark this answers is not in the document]
patriotyk: Yes we don't change it

src/sota_tools/deploy_test.cc
```diff
@@ -0,0 +1,55 @@
...
+                      "objects/a1/f4f81612ce959883f58e83789f6c7d97b0b55b801b2a09955235f40b0f2dfb.filez"),
+      Utils::readFile(std::string(
+          "tests/sota_tools/repo/objects/a1/f4f81612ce959883f58e83789f6c7d97b0b55b801b2a09955235f40b0f2dfb.filez")));
+
+  EXPECT_EQ(
+      Utils::readFile(temp_dir.Path() /
+                      "objects/16/ef2f2629dc9263fdf3c0f032563a2d757623bbc11cf99df25c3c3f258dccbe.commit"),
+      Utils::readFile(std::string(
+          "tests/sota_tools/repo/objects/16/ef2f2629dc9263fdf3c0f032563a2d757623bbc11cf99df25c3c3f258dccbe.commit")));
+}
```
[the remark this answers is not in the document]
patriotyk: Sorry I just missed it.  About compare, `diff -r folder1 folder2` probably should work.

tests/sota_tools/test-garage-deploy-missing-credentials
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -eu
+TARGET="Unable to read a-non-existent-file as archive or json file"
+# File not found and invalid/corrupt files are treated the same
+$1 --commit 16ef2f2629dc9263fdf3c0f032563a2d757623bbc11cf99df25c3c3f258dccbe -f a-non-existent-file -p a-non-existent-file --name testname -h hwids 2>&1 | grep "$TARGET"
```
patrickvacek: I think for this to be a valid test, you actually need to test the fetch and push credentials independently. It can still be done in one script if you run it twice and check both outputs.

src/sota_tools/deploy.cc
```diff
@@ -30,7 +30,7 @@ bool UploadToTreehub(const OSTreeRepo::ptr &src_repo, const ServerCredentials &p
   }
 
   RequestPool request_pool(push_server, max_curl_requests);
-
+  std::cout << "-----------------------------------------------------\n\n\n";
```
patrickvacek: This should probably be removed, right?

lbonn:
Could you give a bit of context on your last commit (`Fix tests hung`). Why removing `set -eu` was needed to make the test work?

src/sota_tools/ostree_http_repo_test.cc
```diff
@@ -0,0 +1,108 @@
...
+  boost::filesystem::path cert_path = "tests/fake_http_server/client.crt";
+
+  const uint8_t hash[32] = {0x16, 0xef, 0x2f, 0x26, 0x29, 0xdc, 0x92, 0x63, 0xfd, 0xf3, 0xc0,
+                            0xf0, 0x32, 0x56, 0x3a, 0x2d, 0x75, 0x76, 0x23, 0xbb, 0xc1, 0x1c,
+                            0xf9, 0x9d, 0xf2, 0x5c, 0x3c, 0x3f, 0x25, 0x8d, 0xcc, 0xbe};
+  UploadToTreehub(src_repo, ServerCredentials(filepath), OSTreeHash(hash), cert_path.string(), false, 1);
+
+  int result = system(
+      (std::string("diff -r ") + (src_repo->root() / "objects/").string() + " tests/sota_tools/repo/objects/").c_str());
+  EXPECT_EQ(result, 0);
```
patrickvacek: This looks great. My one suggestion is to force even more failures, like every other try instead of every third.

tests/sota_tools/test-garage-deploy-online-signing
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -eu
```
patrickvacek: I'm also curious about why this `set -eu` had to be removed.

lbonn:
Also this is not critical but could you rebase your changes? They are based on a two weeks old commit before we pushed a fix for coverage reports on Jenkins. It would be cool to have a new accurate coverage rate.

patrickvacek:
Rebased and fixed some minor issues. Perhaps the reason that `set -e` doesn't work in most of the tests is probably because of killing the python server. Maybe traps can do that better? I'm not very skilled with those. I also tried changing the failure frequency of the bad_connection test to 2 and it consistently fails, which doesn't seem right. I'd like to understand that.

patrickvacek:
@lbonn Good call, look at that coverage!

patrickvacek:
@patriotyk We're merging this for now because it looks good, but could you still look into those last issues (`set -e` and increasing the failure frequency of the bad_connection test)?

patriotyk:
The problem with `set -e` is because bash exits immediatly when some process returns non 0 code and we dont kill server, but ctest waits for server execution finish. In those tests garage-deploy returns non zero return code and it is okey, because we test such case(test-garage-deploy-offline-signing works with set -e, I just verified it.). I think yes, `trap` command should resolve this problem. I will create new PR with this improvement.

patriotyk:
About bad_connection test. Problem here is that we pull repo asynchronously and it happens that we close connection for one url 3 times. But I just got an idea how to improve it. We could use dictionary with url  and counter of made requests.

patriotyk:
About `garage-check` it calls `/api/v1/user_repo/targets.json` but we don't push changes to the real server so there is no any reason to call it. Yes we could fake this API call, but this is actual for garage-check tests not garage-deploy. Correct me if I am wrong.

patrickvacek:
> About `garage-check`. It calls `/api/v1/user_repo/targets.json` but we don't push changes to the real server so there is no any reason to call it. Yes we could fake this API call, but this is actual for garage-check tests not garage-deploy. Correct me if I am wrong.

You're not wrong, but I still think it would be useful to do. We currently don't have any tests that use garage-check except for run_expired_test.sh, and that is testing something fairly specific. I think it shouldn't be too hard to fake the API call for one of the other garage-deploy or garage-push tests we already have and insert the garage-check calls to verify that it does what we expect. Does that make sense and sound reasonable?

patriotyk:
This API call should depend on deploy process, so it is not enought to serve perdefined targets.json. Otherwise this check don't have any sense. But implement `/api/v1/user_repo/targets.json` api call that returns correct targets is not so easy and. How do think we should implement(fake) this call?

patrickvacek:
> This API call should depend on deploy process, so it is not enough to serve perdefined targets.json. Otherwise this check don't have any sense.

Okay, maybe I didn't think about this thoroughly enough and this isn't viable. I'll have to look into this and get back to you.

PR #957, opened 2018-10-04
Add lshw to xenial Dockerfile

lbonn:
It is now needed for the full run of the test suite (test_uptane)


lbonn:
We do not run these tests on Xenial with CI anymore, I found these when running them manually.

PR #961, opened 2018-10-05
Doc/ota 1241/actions

patrickvacek:
Created an initial list of actions for aktualizr, meta-updater, and related tools. This is based on the original version of the similar document created by @stevana for the e2e tests.

Currently, this mostly only covers the actions/requirements that are currently covered by tests. The interesting part is probably what is *not* yet covered. I would like to ask for help in thinking about what is not yet covered. (Help could be comments, patches, PRs, or whatever you want.)

The level of detail here is occasionally quite deep and sometimes rather superficial. This is probably not ideal but is caused by my relative understanding of the system and my perceived relative importance of the various parts. It could probably use a lot of work to be smoothed out and better organized. This is just a first draft to get a conversation going and find our weak spots.

Also note that this PR has some minor other bits found while doing the investigation.


lbonn:
Thanks for the hard work! Do we have a plan to make sure that this list stays as up-to-date as possible?

patrickvacek:
> Do we have a plan to make sure that this list stays as up-to-date as possible?

No, not yet. It's a very good question, but at the moment, I'd just call this a draft to get the conversation moving forward. How we might actually integrate this into our process is yet to be determined.

I don't expect this format to be long-term, either. It was just the quickest way I could assemble the information I had at hand. I might like to move it to something that can be automatically transformed into LaTeX as Stevan did for the e2e doc, but he used Haskell for that, which probably isn't ideal for us. I have experience with using python to convert a text document into LaTeX, so I'm not opposed to doing that again.

patrickvacek:
Updated the list substantially.

The external actions and their subitems now loosely reflect the normal order of the system. There are still some higher-level items (mostly from aktualizr and uptane tests) scattered about. I may try to reorganize those in some better fashion.

Any item listed as TODO requires review to verify that the stated coverage is correct.

Some common activities have been moved to the internal section and are referenced from within the external actions with a "see below" note.

patrickvacek:
Rebased and updated once more after further discussion. I think this is ready for merging now.

PR #966, opened 2018-10-16
Fixes for garage-deploy tests

patriotyk:


patrickvacek:
This looks good. Two thoughts, though:

1. I don't see (unexpected) failures on the HEAD requests in the bad connection test. The GET part looks good, though.
2. Can we also `set -o pipefail` on most of these bash scripts?

patriotyk:
1) We don't do HEAD requests to the fetch server, only to the deploy server 
2) Done

patriotyk:
Update.
2) I removed `pipefail` in `test-missing-commit` because we test missing commit, so garage-deploy returns non zero return code, then we use grep to match the missing commit. But all following commands after first non zero return code will always return this nonzero return code if we use `pipefail`. So we cannot match missing commit and this test doesn't work with this option.

patrickvacek:
1. You are correct, but then why do we have support for dropping HEAD requests in the test server if we aren't using it? We should move that to the test deploy/push server so we can prove that we can recover from a spotty connection at that part, too. We sort of cover this with the rate controller test, but I think we can do just a bit more, since we've already done most of the work.
2. Looks good, thanks!

PR #967, opened 2018-10-16
Test/ota 815/ostree object

patrickvacek:
Unit tests for OSTreeObject class.

Really only covers the constructor, MakeTestRequest, and Upload. These are the components that are easiest to verify and have the most external interaction. Testing CurlDone or the various OSTree parsing bits might be interesting, but it is quite difficult to decouple from other parts of the code, and the whole of it is already covered by higher-level tests. I'll take suggestsions for improvements, but I think this is a good start.

### DIFF
--- a/src/sota_tools/deploy_test.cc
+++ b/src/sota_tools/deploy_test.cc
@@ -6,6 +6,7 @@
 #include "ostree_http_repo.h"
 #include "ostree_ref.h"
 #include "test_utils.h"
+
 std::string port = "2443";
 TemporaryDirectory temp_dir;
 
--- a/src/sota_tools/ostree_http_repo_test.cc
+++ b/src/sota_tools/ostree_http_repo_test.cc
@@ -3,7 +3,9 @@
 #include "deploy.h"
 #include "ostree_http_repo.h"
 #include "ostree_ref.h"
+#include "server_credentials.h"
 #include "test_utils.h"
+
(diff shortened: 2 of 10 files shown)


src/sota_tools/deploy_test.cc
```diff
@@ -5,7 +5,9 @@
 #include "ostree_dir_repo.h"
 #include "ostree_http_repo.h"
 #include "ostree_ref.h"
+#include "server_credentials.h"
```
patriotyk: Looks like we don't need this

src/sota_tools/ostree_http_repo_test.cc
```diff
@@ -3,7 +3,9 @@
 #include "deploy.h"
 #include "ostree_http_repo.h"
 #include "ostree_ref.h"
+#include "server_credentials.h"
```
patriotyk: Do we really need this lnclude?

src/sota_tools/deploy_test.cc
patrickvacek: Probably not, since it is included in `deploy.h`.

src/sota_tools/ostree_http_repo_test.cc
patrickvacek: Probably yes, because I don't see it included in `ostree_http_repo.h`.

PR #968, opened 2018-10-18
Refactor/ota 1331/return values

patriotyk:


src/libaktualizr/primary/aktualizr.cc
```diff
@@ -63,9 +63,7 @@ void Aktualizr::CycleEventHandler::handle(const std::shared_ptr<event::BaseEvent
```
lbonn: The original plan was also to rewrite this loop to use the functions return values instead of this huge event handler. The code would then look sequential: `CheckUpdates(); Download(); Install();`.

Could be done in a separate PR if it's convenient.

src/libaktualizr/primary/reportqueue.h
```diff
@@ -78,8 +79,7 @@ class ReportQueue {
   std::condition_variable cv_;
   std::mutex thread_mutex_;
   std::mutex queue_mutex_;
-  std::queue<std::unique_ptr<ReportEvent>> report_queue_;
-  Json::Value report_array{Json::arrayValue};
+  std::vector<std::pair<std::shared_ptr<std::promise<bool>>, std::unique_ptr<ReportEvent>>> report_queue_;
```
lbonn: Would it be realistic to embed the promise inside `ReportEvent`? It could simplify this class a bit.

src/libaktualizr/utilities/events.h
```diff
@@ -156,23 +148,6 @@ class AllInstallsComplete : public BaseEvent {
...
-};
-
-/**
- * A campaign has been successfully accepted.
- */
-class CampaignAcceptComplete : public BaseEvent {
- public:
-  explicit CampaignAcceptComplete();
-};
-
```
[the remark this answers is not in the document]
lbonn: I don't care too much but we can check how it's used by the library users (if there is any for this feature)

src/libaktualizr/primary/reportqueue.h
patriotyk: Yes I just embeded, but not sure is it better.

src/libaktualizr/primary/reportqueue.cc
```diff
@@ -32,28 +32,41 @@ void ReportQueue::run() {
   }
 }
 
-void ReportQueue::enqueue(std::unique_ptr<ReportEvent> event) {
+std::future<bool> ReportQueue::enqueue(std::unique_ptr<ReportEvent> event) {
+  auto report_promise = std::unique_ptr<std::promise<bool>>(new std::promise<bool>);
```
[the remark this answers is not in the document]
patriotyk: Fixed

patriotyk:
@patrickvacek Ok, go ahead. I just updated with some fixes and will not push here anymore.

src/libaktualizr/utilities/events.h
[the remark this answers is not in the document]
patrickvacek: Actually, since it looks like we want to move towards async API, we probably want to keep the events for now. I'll add them back for now.

patrickvacek:
> I just updated with some fixes and will not push here anymore.

@patriotyk Thanks, sounds good. One question, though: it doesn't look like anything is using the futures in `ReportQueue`. What was the goal there?

src/libaktualizr/primary/reportqueue.cc
```diff
@@ -32,28 +32,40 @@ void ReportQueue::run() {
...
 }
 
-void ReportQueue::enqueue(std::unique_ptr<ReportEvent> event) {
+std::future<bool> ReportQueue::enqueue(std::unique_ptr<ReportEvent> event) {
+  event->promise = std_::make_unique<std::promise<bool>>();
+  auto report_future = event->promise->get_future();
   {
     std::lock_guard<std::mutex> queue_lock(queue_mutex_);
-    report_queue_.push(std::move(event));
+    report_queue_.emplace_back(std::unique_ptr<ReportEvent>(std::move(event)));
```
lbonn: Now, it can just be `report_queue_.push_back(std::move(event));`

src/libaktualizr/primary/reportqueue.h
lbonn: If you don't think it's a good change, we can talk about it that's fine.

It's just that when `enqueue` is called, the ownership of `event` is passed to the queue anyway so it should be able to modify it and use an internal member for bookkeeping.

src/libaktualizr/primary/aktualizr.cc
patrickvacek: Done. How does it look to you? I still need to test it more, though.

patrickvacek:
@lbonn ReportQueue changes removed for now, so this shouldn't conflict with your plans anymore.

patrickvacek:
@lbonn I actually still have more work to do, specifically testing and getting a meaningful return from installation. Do you think we should merge this now and I should carry on in a new PR? That might actually be easier, although there is a risk in the meantime due to lack of testing.

lbonn:
If you want to add something, go ahead. What's your plan on testing? You intended to add new cases related to the refactor or run some more oe-selftests/manual tests?

patrickvacek:
> What's your plan on testing? You intended to add new cases related to the refactor or run some more oe-selftests/manual tests?

1. Manual testing of hmi_stub and aktualizr (the executable). Unfortunately this is generally still something that has to be done manually when we make API changes.
2. Review aktualizr_test to make sure it is actually exercising the return values of the API calls.
3. oe-selftest just to be sure (not usually necessary, but probably useful here).

I also want to work on the install return value and I need to update the docs.

src/libaktualizr/utilities/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ set(SOURCES dequeue_buffer.cc
 set(HEADERS config_utils.h
             dequeue_buffer.h
             exceptions.h
+            results.h
```
lbonn: Minor: it's a bit odd that `results.h` sits in `utilities`. It was already the case before with `events.h`, so it's not a new problem.

lbonn:
You're right, `events.cc` is basically useless now, would be nice to merge it in the header.

src/libaktualizr/utilities/CMakeLists.txt
patrickvacek: Yes, I agree, but I couldn't find a better solution at the moment.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -141,12 +141,15 @@ void SotaUptaneClient::PackageInstallSetResult(const Uptane::Target &target) {
...
   storage->storeInstallationResult(result);
+  return result;
 }
 
 void SotaUptaneClient::reportHwInfo() {
   Json::Value hw_info = Utils::getHardwareInfo();
   if (!hw_info.empty()) {
     http->put(config.tls.server + "/core/system_info", hw_info);
+  } else {
+    LOG_WARNING << "Unable to fetch hardware information from host system.";
```
lbonn: Cool!

patrickvacek: Yeah, that probably should've been there ages ago.

patrickvacek:
> You're right, `events.cc` is basically useless now, would be nice to merge it in the header.

Done.

PR #970, opened 2018-10-19
Get campaign size

eu-smirnov:


lbonn:
Why is it not in metadata like the rest? You've checked with backend?

eu-smirnov:
> Why is it not in metadata like the rest? You've checked with backend?

Ask Vladimir :) It is tested with the backend.

patrickvacek:
@koshelev Can you confirm that this looks correct? If that is what we need, it is fine with me.

koshelev:
@lbonn The metadata is what a user enters while creating a campaign. Size is calculated per vehicle.
@patrickvacek I can confirm it is what we need :)

PR #972, opened 2018-10-23
Added pause/resume support

patriotyk:


config/sql/schema.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE tls_creds(ca_cert BLOB, ca_cert_format TEXT,
                        client_cert BLOB, client_cert_format TEXT,
                        client_pkey BLOB, client_pkey_format TEXT);
 CREATE TABLE meta(meta BLOB NOT NULL, repo INTEGER NOT NULL, meta_type INTEGER NOT NULL, version INTEGER NOT NULL, UNIQUE(repo, meta_type, version));
-CREATE TABLE target_images(filename TEXT UNIQUE, image_data BLOB NOT NULL);
+CREATE TABLE target_images(filename TEXT UNIQUE, image_data BLOB NOT NULL, real_size INTEGER NOT NULL DEFAULT 0);
```
patrickvacek: If you change the SQL schema, you will also need to write a migration script. We've also decided that we need to be able to migrate in reverse as well. We don't need to be able to go back further than where we are now (because it's too late for that to be meaningful), but from this point on, we have to be able to migrate forwards and backwards.

src/hmi_stub/main.cc
```diff
@@ -165,6 +165,21 @@ int main(int argc, char *argv[]) {
...
+              break;
+            } else if (c == 'r') {
+              std::cout << "resuming\n";
+              aktualizr.Resume();
+              break;
+            }
+          }
+        })
+            .detach();
+
```
patrickvacek: This presumably will also need to break once the download is complete, right? Or is that automatic in some way that I'm not seeing?

I'd originally assumed you'd want to base this PR off of the API PR that we've been working on, but this is actually simpler than I expected, so maybe it is fine as it is. That other PR still isn't finished but I hopefully will get it done soon.

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -91,6 +93,8 @@ class SotaUptaneClient {
   void sendImagesToEcus(const std::vector<Uptane::Target> &targets);
   bool hasPendingUpdates(const Json::Value &manifests);
   void sendDownloadReport();
+  void downloadFinished(const unsigned int targets_count);
```
patrickvacek: Where is this function implemented?

config/sql/schema.sql
patriotyk: Yes I know and I did it. I just didn't include it in the commit

src/libaktualizr/primary/sotauptaneclient.h
patriotyk: Sorry this is odd function

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -952,18 +951,40 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
     }
-    auto statement = db_.prepareStatement<std::string, SQLZeroBlob>(
-        "INSERT OR REPLACE INTO target_images (filename, image_data) VALUES (?, ?);", filename_,
-        SQLZeroBlob{expected_size_});
 
-    if (statement.step() != SQLITE_DONE) {
-      LOG_ERROR << "Statement step failure: " << db_.errmsg();
-      throw exc;
+    auto statement = db_.prepareStatement<std::string>(
+        "SELECT rowid, length(real_size) FROM target_images WHERE filename = ?;", filename_);
```
OYTIS: `length(real_size)` looks like a typo.

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -49,32 +49,54 @@ static size_t DownloadHandler(char* contents, size_t size, size_t nmemb, void* u
...
+          if (response.curl_code == CURLE_WRITE_ERROR) {
+            throw OversizedTarget(target.filename());
+          }
+          throw Exception("image", "Could not download file, error: " + response.error_message);
         }
-        throw Exception("image", "Could not download file, error: " + response.error_message);
-      }
+        if (pause_) {
+          pause_mutex_.lock();
+          pause_mutex_.unlock();
```
OYTIS: Why?

src/libaktualizr/storage/sqlstorage.cc
patriotyk: Yes, should be `length(image_data)`

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -49,32 +49,54 @@ static size_t DownloadHandler(char* contents, size_t size, size_t nmemb, void* u
     auto event = std::make_shared<event::DownloadProgressReport>(ds->target, "Downloading", calculated);
     (*(ds->events_channel))(event);
   }
+  if (ds->fetcher->isPaused()) {
+    return written_size + 1;
```
OYTIS: Sorry, could you add clarifying comments here?

src/libaktualizr/uptane/fetcher.cc
patriotyk: We should unlock this mutex after another thread freed it because second time we will get  a dead lock. We don't use this mutex for data synchronization, only for waiting while it is blocked in another thread.

src/libaktualizr/uptane/fetcher.cc
patriotyk: We return incorrect size because we need stop download. libcurl has some magic value we could return to pause download (https://curl.haxx.se/libcurl/c/curl_easy_pause.html) and I tried this approach but I rejected it because it keeps connection alive and after some time server drops this connection with timeout error.

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -49,32 +49,54 @@ static size_t DownloadHandler(char* contents, size_t size, size_t nmemb, void* u
...
+  }
   return written_size;
 }
 
+void Fetcher::setPause(bool pause) {
+  if (pause) {
+    pause_mutex_.lock();
+  } else {
+    pause_mutex_.unlock();
```
OYTIS: Unlocking an unlocked mutex can throw.

patrickvacek:
> what is our desired usage of this API? Run `aktualizr::Download()` in a thread and issue `aktualizr::setPause()` asyncronously?

Currently, yes. This may change once we have actual asynchronous API calls, which is planned for an upcoming sprint.

src/libaktualizr/uptane/fetcher.cc
OYTIS: Yes, looks like the best way to abort the installation. It would be nice to have the comment in the code though.

src/libaktualizr/uptane/fetcher.cc
patriotyk: Good catch. Also we will get deadlock if call setPause(true) twice.

patriotyk:
> This usage of a mutex looks a bit non-idiomatic (i.e. generates wtfs). Could we use something else, like conditional variables?
Yes we can. But mutexes requires less code.

> This mechanism is going to break if the file has been updated between download attempts. Maybe it's enough to say you shouldn't update a file once it is in the repo, but maybe it is also a good idea to use some If-Modified-Since magic.
I think we should add checksum field to the images table(but checksum should be from metadata) and check if checksum is different thats mean it is new file and we will update it. If checksum is the same but length in database if smaller, that means it is not finished image, and if checksum and size is the same we could skip download.

src/libaktualizr/uptane/fetcher.cc
patriotyk: Updated. Now it is impossible

patriotyk:
Looks like it is ready now

patrickvacek:
I think this looks good! The fetcher_test is great. My preference would be to merge this now and then do a follow-up PR to finish things up. In the meantime I can rebase https://github.com/advancedtelematic/aktualizr/pull/968 and get that merged as well. @lbonn @OYTIS any objections?

The things that still need work (that I can see):
1. The comment that Anton requested about aborting the download when it is paused.
2. A higher-level (or "integration") test in `aktualizr_tests.cc` that demonstrates the events are generated as expected. You might be able to reuse and expand an existing test for this.
3. It is still possible to call `Fetcher::setPause(false)` twice at the same time from two threads and thus unlock the same lock twice if we were to get really unlucky. We have to fix that somehow.

patriotyk:
> It is still possible to call Fetcher::setPause(false) twice at the same time from two threads and thus unlock the same lock twice if we were to get really unlucky. We have to fix that somehow.

No, it is not possible. I added `if` statment at the beiginig of setPause fucntion.

> After downloading a file, if I immediately try to download again, it re-downloads the entire file. Shouldn't it recognize that it is already in the storage?

Yes it is current behavior and I didn't change it. To implement this fucntionality we would need to add checksum to the database table.

patrickvacek:
I'll start real testing in just a minute, but based on the code, we still have to do something about `setPause` getting called twice simultaneously. We need a condition variable or some other construct to prevent that scenario.

patrickvacek:
> 9. It does look like downloads can be resumed after aborting aktualizr mid-download. However, when the download finished, this is what I got:
> 
> ```
> Error while downloading a target: The target's calculated hash did not match the hash in the metadata.
> Received Error event
> Only 0 of 1 were successfully downloaded. Report not sent.
> Received Error event
> API Download finished
> ```
> 
> I was still able to install the file afterwards, though.

I'm still seeing this issue as well. To reproduce:
1. Start a download.
2. `pause`
3. `shutdown`
4. Restart `hmi_stub`.
5. `fetchmeta`
6. `download`

patriotyk:
About download again while pause. Yes I can reproduce it, I have not tried this scenario before. But it is not relited to pause/resume, it is also reproducable just when call two download commands simultaneously. So we should check if we in downloading and don't allow second download.

patriotyk:
> Yes, it is possible for two threads to call the function practically simultaneously with the same value of pause and it would cause problems. Unlikely but possible, and we have to watch out for things like that.

Yes you are right, I have not thought simultaneous execution. I have added mutex to this call.

patriotyk:
> 8. After pausing a download, you can't shut down hmi_stub because the other thread is still hung. If you try, it hangs indefinitely.

I cannot reproduce it also. I am able to exit while in pause

patriotyk:
Just found that download restore issue is no to so easy to fix. The problem here is multi part hasher. So we should initialize it to prevouse state to calculate hash correct, I think there is only 2 sollutions how to fix it.
1) Find way how to save in database and then restore partialy downloaded hash state.
2) When restore downloading, just load partialy download data and update hasher to restore it to correct state.

patrickvacek:
> Just found that download restore issue is no to so easy to fix. The problem here is multi part hasher.

Good point, I see the problem. Either solution is fine, but the second option (reload the data from storage and calculate the hash of that) seems safest. That would be my preference.

patriotyk:
Fixed download restore issue

patriotyk:
Yes, you are right. `git diff` showed me that changes. I have pushed those changes together with high level test. But I havn't add this tets to aktualizr_tesr.cc because this requires real http server which we already have in fetcher_test.cc. So, to avoid tests duplication, I just added event check functionality to that test.

patriotyk:
I cannot reproduce issue with  `Aktualizr.FullNoCorrelationId`  failure. Also I connot open Jenkins.

patrickvacek:
> Also I connot open Jenkins.

It looked like this last time Jenkins was just complaining about formatting, so I fixed that. Hopefully both PRs will be green after that.

patriotyk:
Looks like docker servers fails

patrickvacek:
> Looks like docker servers fails

Worked after a restart. CI is happy now and I'm looking into your latest changes.

patrickvacek:
@patriotyk I added just a couple extra checks to the pause/resume test. I think we can safely merge this once CI accepts that change. But then I think you should make a new PR with a couple extra tests:

1. Pause while paused (and confirm the result).
2. Resume while not paused (and confirm the result).
3. Pause while not downloading (and confirm the result).
4. Pause download, shut down SotaUptaneClient, create a new instance, resume, and verify that it picks up the download where it left off. (I'm not quite sure yet how to verify that that actually worked other than that there should not be any errors or problems.)

patrickvacek:
I think we should merge this now. @lbonn @OYTIS any objections?

src/libaktualizr/storage/invstorage.h
```diff
@@ -57,6 +58,9 @@ class StorageTargetWHandle {
     }
     return is;
   }
+
+ protected:
+  size_t written_size_{0};
```
lbonn: I don't think it's a new issue but we should probably make a pass to make sure that we store full sizes of files with 64 bits on all architectures.

lbonn:
I have not re-reviewed the concurrency but if you're happy let's merge it.

PR #973, opened 2018-10-23
Read secondary configs json from one directory

lbonn:
It allows us to treat `secondary_configs_dir` as a regular configuration option, which can be set in the main configuration and not only from the command line.

Will need changes in meta-updater and probably docs. Shout if you don't agree!

### DIFF
--- a/docs/deb-package-install.adoc
+++ b/docs/deb-package-install.adoc
@@ -10,10 +10,10 @@ sudo apt install ./aktualizr.deb
 
 The debian package will install, enable, and start an `aktualizr` systemd service immediately after it's installed. However, there are some configuration steps that should be taken before the service starts. To use aktualizr with a server (i.e. https://github.com/advancedtelematic/ota-community-edition/[OTA Community Edition] or https://docs.atsgarage.com[HERE OTA Connect]), you will need to download the provisioning credentials file provided by the server and place it at `/var/sota/sota_provisioning_credentials.zip`.
 
-Additionally, if you wish to pass extra command line arguments to aktualizr, you can do so by creating `/usr/lib/sota/sota.env` to set the `AKTUALIZR_SECONDARY_CONFIG` environment variable. For example, to enable a secondary ECU with a config file at `/var/sota/sec_ecu.json`, the contents of the file would be:
+Additionally, if you wish to pass extra command line arguments to aktualizr, you can do so by creating `/usr/lib/sota/sota.env` to set the `AKTUALIZR_SECONDARY_CONFIG` environment variable. For example, to set up the log level to debug:
 
 ----
-AKTUALIZR_SECONDARY_CONFIG="--secondary-config /var/sota/sec_ecu.json"
+AKTUALIZR_SECONDARY_CONFIG="--loglevel 0"
 ----
 
 You can pass any other command line arguments in this file, as well.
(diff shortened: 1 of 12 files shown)


docs/deb-package-install.adoc
```diff
@@ -10,10 +10,10 @@ sudo apt install ./aktualizr.deb
...
 
-Additionally, if you wish to pass extra command line arguments to aktualizr, you can do so by creating `/usr/lib/sota/sota.env` to set the `AKTUALIZR_SECONDARY_CONFIG` environment variable. For example, to enable a secondary ECU with a config file at `/var/sota/sec_ecu.json`, the contents of the file would be:
+Additionally, if you wish to pass extra command line arguments to aktualizr, you can do so by creating `/usr/lib/sota/sota.env` to set the `AKTUALIZR_SECONDARY_CONFIG` environment variable. For example, to set up the log level to debug:
 
 ----
-AKTUALIZR_SECONDARY_CONFIG="--secondary-config /var/sota/sec_ecu.json"
+AKTUALIZR_SECONDARY_CONFIG="--loglevel 0"
```
eu-smirnov: Rather a strange name for the variable that adds an arbitrary configuration.

patrickvacek: I assume that is a typo or leftover from debugging?

lbonn: Yes you're right, thanks! In fact in meta-updater, we use `AKTUALIZR_CMDLINE_PARAMETERS`: https://github.com/advancedtelematic/meta-updater/blob/c035200bda65f96880cb1663a5841faeb4470861/recipes-sota/aktualizr/files/aktualizr.service#L10

src/libaktualizr/config/config.cc
```diff
@@ -248,15 +253,19 @@ void Config::updateFromCommandLine(const boost::program_options::variables_map&
...
-  for (auto it = sconfigs.cbegin(); it != sconfigs.cend(); ++it) {
-    uptane.secondary_configs.emplace_back(Uptane::SecondaryConfig(*it));
+void Config::readSecondaryConfigs(const boost::filesystem::path& sconfigs_dir) {
+  if (!boost::filesystem::is_directory(sconfigs_dir)) {
+    LOG_ERROR << "Could not read secondary configs from " << sconfigs_dir << ": not a directory";
+    return;
+  }
+  for (const auto& config_file : Utils::glob((sconfigs_dir / "*.json").string())) {
+    LOG_INFO << "Parsing secondary config: " << config_file;
+    uptane.secondary_configs.emplace_back(config_file);
```
patrickvacek: Maybe we should allow providing a single json file if the user wants to? As in, the target can be a directory or a single file.

docs/deb-package-install.adoc
patrickvacek: Ah, now I see. Can we change that variable safely? Or do we even care, since debian support is not actually essential anymore?

Besides, you removed `AKTUALIZR_SECONDARY_CONFIG` above from the `.service` file (probably rightfully so).

src/libaktualizr/config/config.cc
lbonn: I think it would complicate the docs and the cli a little bit. Because we essentially use this feature for demos, I don't know if we need to be that flexible for the moment.

The primary goal was just to get rid of this ugly CLI-only feature and its associated hacks.

lbonn:
I've used this PR as a pretext for introducing a changelog to the project, with a rough summary of our historical versions. It should be a quick go-to document for aktualizr users to track new features, bug fixes and usage changes.

So, it would not  contain internal changes that only belong to the git log: tests and ci improvements, internal refactoring...

src/libaktualizr/config/config.cc
patrickvacek: That's fine. Not worth investing effort in something that no one needs. I'm happy with this upgrade as it is.

OYTIS:
Oh nice, I've just implemented it in a custom way for Sachertorte. But having it in libaktualizr is of course much better.

lbonn:
@patrickvacek I forgot about your comment on hmi_stub. I'll update it before the merge.

lbonn:
I've tested the last commit on hmi-stub manually, so let's just merge!

PR #978, opened 2018-10-26
Don't create test files in the source tree

patriotyk:


tests/sota_tools/test-garage-deploy-offline-signing
```diff
@@ -18,8 +18,10 @@ export PATH=$PATH:$TEMP_DIR
 
 echo $TREEHUB > $TEMP_DIR/treehub.json
 ./tests/sota_tools/treehub_server.py $PORT &
+PUSH_CREDS="$(pwd)/tests/sota_tools/auth_test_good_offline.zip"
```
lbonn: Is it just a style change or does it fix a specific issue?

Issue #974, opened 2018-10-23
Provisioning docs are outdated


Bubu:
https://github.com/advancedtelematic/aktualizr/blob/master/docs/implicit-provisioning.adoc seems to be outdated.

1) It has a link to https://github.com/advancedtelematic/aktualizr/blob/master/config/sota_implicit_prov.toml which gives `404`
2) The `aktualizr-implicit-prov` method seems to have been removed from the meta-updater repository.

Bubu:
The docs here are outdated as well:https://github.com/advancedtelematic/meta-updater#sota-related-variables-in-localconf

patrickvacek:
@Bubu what did you notice was wrong with the README in meta-updater?

Bubu:
> @Bubu what did you notice was wrong with the README in meta-updater?

Can't find it now. I probably misread something yesterday.

patrickvacek:
Fixed by https://github.com/advancedtelematic/aktualizr/pull/979. Feel free to reopen or open a new issue if you find something is still missing or wrong. Thanks again for pointing out the problem.

PR #975, opened 2018-10-24
Update implicit provisioning docs.

patrickvacek:
Fixes #974. @OYTIS can you confirm this matches the work you've done recently?

PR #979, opened 2018-10-26
Fix/rm implicit writer

patrickvacek:
Currently includes https://github.com/advancedtelematic/aktualizr/pull/975 (after rebasing).


patrickvacek:
Fixes #974.

PR #981, opened 2018-10-26
Expose downloaded binaries via API

OYTIS:


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -813,12 +813,17 @@ UpdateCheckResult SotaUptaneClient::checkUpdates() {
   unsigned int ecus_count = 0;
   if (!uptaneOfflineIteration(&updates, &ecus_count)) {
     LOG_ERROR << "Invalid UPTANE metadata in storage.";
-    result = UpdateCheckResult({}, 0, UpdateStatus::kError, "Invalid UPTANE metadata in storage.");
+    result = UpdateCheckResult({}, 0, UpdateStatus::kError, Json::nullValue, "Invalid UPTANE metadata in storage.");
   } else {
+    std::string director_targets;
+    storage->loadNonRoot(&director_targets, Uptane::RepositoryType::Director, Uptane::Role::Targets());
```
eu-smirnov: Should we check the return value here?

OYTIS: If `loadNonRoot()` returns `false`, `director_targets` will be an empty string and `parseJSON()` will return `Json::Nullvalue` which will be passed as targets metadata. It looks like a reasonable error handling to me. Could we do something better?

src/libaktualizr/primary/aktualizr.h
```diff
@@ -83,6 +83,14 @@ class Aktualizr {
    */
   DownloadResult Download(const std::vector<Uptane::Target>& updates);
 
+  /**
+   * Get target downloaded in Download call. Returned target is guaranteed to be verified and up-to-date
```
lbonn: I think these guarantees are not actually fulfilled in the current implementation (SQLStorage). We even commit changes in partial downloads.

src/libaktualizr/primary/aktualizr.h
```diff
@@ -83,6 +83,14 @@ class Aktualizr {
...
+   * according to the Uptane metadata downloaded in CheckUpdates call.
+   * @param filename Name of the binary in the storage
+   * @return Handle to the stored binary. nullptr if none is found.
+   */
+  std::unique_ptr<StorageTargetRHandle> getStoredTarget(const std::string& filename);
```
lbonn: lint: other api calls start with an upper case letter

src/libaktualizr/primary/sotauptaneclient.cc
eu-smirnov: I'm just wondering if it's still OK to return UpdatesAvailable status in this case.

src/libaktualizr/primary/aktualizr.h
OYTIS: OK, it's not true yet, but in Serhiy's PR read handle is throwing on incomplete file (https://github.com/advancedtelematic/aktualizr/pull/972/files#diff-01a41e87d80cc7d2239a7088d4944b83R1087). I would prefer returning `nullptr`, but that is another story.

lbonn: Ok fair enough.

You can also bundle up a changelog update under an `Added` section if you're at it :).

OYTIS: Another thing that is missing for that to be totally true is that we don't clean-up old target files.

lbonn: Yes. So you'd like to use this api now and plan enhancements to the target files handling as soon as possible?

OYTIS: Yes, I will also change the docstring to mention that the guarantee is not really true yet, but we really want to be there =)

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -587,6 +587,9 @@ TEST(Aktualizr, InstallWithUpdates) {
   EXPECT_EQ(result.reports.size(), 0);
   conf.uptane.running_mode = RunningMode::kDownload;
   aktualizr.UptaneCycle();
+  EXPECT_NE(aktualizr.GetStoredTarget("primary_firmware.txt").get(), nullptr);
+  EXPECT_NE(aktualizr.GetStoredTarget("secondary_firmware.txt").get(), nullptr);
```
patrickvacek: I suggest also doing the negative test, as in, check that these are `nullptr` before running the `UptaneCycle`.

src/libaktualizr/utilities/results.h
```diff
@@ -36,11 +36,16 @@ class UpdateCheckResult {
...
+                    Json::Value targets_meta_in, std::string message_in)
+      : updates(std::move(updates_in)),
+        ecus_count(ecus_count_in),
+        status(status_in),
+        targets_meta(std::move(targets_meta_in)),
+        message(std::move(message_in)) {}
   std::vector<Uptane::Target> updates;
   unsigned int ecus_count{0};
   UpdateStatus status{UpdateStatus::kNoUpdatesAvailable};
+  Json::Value targets_meta;
```
patrickvacek: What is the goal of this `targets_meta`?

OYTIS: The goal is to let custom installer traverse targets metadata and extract relevant information from "custom" fields.

src/libaktualizr/primary/aktualizr_test.cc
OYTIS: Sure, makes sense.

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: It's more a style question to me. If `uptaneOfflineIteration` returned `true`, there should be some targets in the storage, so this check will test what our own code does (which is a questionable style to me). And even if it would be possible that `uptaneOfflineIteration` succeeds, but there is no targets metadata in the storage, I don't think it's a reason to fail, because our default solution ignores targets metadata in `UpdateCheckResults`.

OYTIS:
@patrickvacek Sorry, how does this `system_info_count` work? It fails for me, and only on Jenkins for some reason.

patrickvacek:
> Sorry, how does this `system_info_count` work? It fails for me, and only on Jenkins for some reason.

Not sure why it would fail only on Jenkins, but the point of that test and those checks is to track how many calls are made to each endpoint on the server. We assume that there should be only one call to `/core/system_info` during normal operation.

src/libaktualizr/primary/sotauptaneclient.cc
eu-smirnov: OK, makes sense to me. Thanks for the explanation!

OYTIS:
@patrickvacek Thanks! So getHardwareInfo() is failing somehow with my changes to parseJSON.

patrickvacek:
> So getHardwareInfo() is failing somehow with my changes to parseJSON.

I guess so? Is `reader.parse()` failing silently somehow?

src/libaktualizr/primary/aktualizr.h
patriotyk: I don't throw exception when we open partially downloaded image anymore. It is needed to update multipart hasher.

OYTIS:
@patrickvacek Funny story. Apparently `lshw -json` really outputs a slightly invalid JSON (with a trailing comma in the array). Somehow that doesn't matter unless you explicitly check parsing result, which I decided to do in this PR just because it felt right. So I will revert it now, and everyone is going to be happy.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1108,7 +1108,11 @@ class SQLTargetRHandle : public StorageTargetRHandle {
 };
 
 std::unique_ptr<StorageTargetRHandle> SQLStorage::openTargetFile(const std::string& filename) {
-  return std::unique_ptr<StorageTargetRHandle>(new SQLTargetRHandle(*this, filename));
+  try {
+    return std::unique_ptr<StorageTargetRHandle>(new SQLTargetRHandle(*this, filename));
+  } catch (...) {
```
lbonn: It should at least log some part of the exception.
Also, this method is used internally and in existing code, it doesn't check for `nullptr` (relies on exception).

So we should revisit all calls and make them check for `nullptr`. Alternatively (I would prefer that actually), the exception catching could be in the new `Aktualizr::GetStoredTarget` method.

OYTIS: Yes, I agree it would be better/easier. Thank you!

OYTIS:
@patrickvacek @lbonn 
BTW, you've probably already known, but I've just found that you can provide a custom error message for a failed assertion in gtest like that: `EXPECT_EQ(a, b) << "a and b differ"`. Would be great to use it for new tests we write, because not all of assertions are self-explanatory.

lbonn:
@OYTIS yes, we have some of them sprinkled around, usually with `FAIL()`. Trying to generalize them sounds like a good plan.

OYTIS:
Added more checks for `Aktualizr::GetStoredTarget` test as @patrickvacek has suggested. Let's wait for CI to complete and merge.

PR #982, opened 2018-10-29
Adding "INDEX.md" to be read in by Doxygen's USE_MDFILE_AS_MAINPAGE option

merltron:
Prompted by this case: https://saeljira.it.here.com/browse/OTA-1455

The Doxygen doc is missing content on the main page. The Doxygen option "USE_MDFILE_AS_MAINPAGE" needs an MD file rather than an "*.adoc" file. For now, I created a separate index file for this purpose.

However, if you are fine with this text, we should replace the equivalent text on the README.adoc page. I don't know how helpful it is to explicitly call out the  "GENIVI SOTA" compatibility on the index/front page. I think we could ditch that stuff no?


INDEX.md
```diff
@@ -0,0 +1,16 @@
+Aktualizr is the codename of the HERE OTA Connect client. This client runs on any embedded device and can check for updates periodically or triggered by another system interaction. You can build this client from source or you can integrate it into your firmware with the Yocto toolset.
```
patrickvacek: "Codename" feels like a funny way to describe aktualizr. I'd prefer to just say "Aktualizr is the HERE OTA Connect client."

INDEX.md
```diff
@@ -0,0 +1,16 @@
...
+* Communicating with the HERE OTA Connect server
+* Authenticating using locally available device and user credentials
+* Reporting current software and hardware configuration to the server
+* Checking for any available updates for the device
+* Downloaded any available updates
+* Installing the updates on the system, or notifying other services of the availability of the downloaded file
+* Receiving or generating installation reports (success or failure) for attempts to install received software
+* Submitting installation reports to the server
+
+For all controllers that run aktualizr, you’ll need to implement some form of key provisioning. One option is to use a Hardware Security Module (HSM). The OTA Connect documentation explains in detail [how to set up an HSM](https://docs.atsgarage.com/prod/enable-implicit-provisioning.html) to hold potentially-sensitive provisioning key material.
```
patrickvacek: It seems odd to me to call out the HSM support but not mention the other provisioning options.

merltron: OK, I linked to the "overview of provisioning methods" page instead.

INDEX.md
merltron: OK, reworded.

merltron:
Almighty, responded to comments and moved INDEX.md to "/docs" - might might need to update the doxyfile? I couldn't get it to pick up the MD file properly but I'm not a Doxygen expert.

lbonn:
@merltron I've pushed some changes in `Doxyfile` to pick the .md properly again and make the path handling a bit less confusing there.

Does it look like what you wanted to achieve?

merltron:
@lbonn Thanks, when I pulled your change and tried to build, I still got a blank front page. What build command do you use for to generate the content that shows up here?: https://advancedtelematic.github.io/aktualizr/

I used this command from the root of the repo:
`doxygen ./docs/Doxyfile.in`

patrickvacek:
@merltron It should just be `make docs` from whatever your build directory is.

patrickvacek:
Also, it works for me!

lbonn:
Ok I'll just merge it then.

PR #984, opened 2018-10-29
Return values for pause and resume.

patrickvacek:
Sorry for the confusion of making a PR on a PR, but I realized we'd missed a requirement because of mutual misunderstanding (as in, I didn't realize we needed it until an hour ago) and I knew I could kick it out quickly. All this really does is add return values for pause/resume, since it is possible that they can succeed or fail for various reasons.


patriotyk:
Unfortunatelly I cannot open Jenkins because vpn you have provided me doesn't work anymore.

patrickvacek:
> Unfortunatelly I cannot open Jenkins because vpn you have provided me doesn't work anymore.

Okay, we'll look into that. Do you mind emailing us whatever logs you get from trying to connect?

patriotyk:
Vpn connects successfully, but opening jenkins give me timout error in browser.

PR #985, opened 2018-10-30
Make TestHelperProcess variadic

lbonn:
Avoid having to add a new string argument all the time and do not supply extraneous empty arguments to subprocesses

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -677,23 +677,28 @@ class HttpFakeNoCorrelationId : public HttpFake {
 TEST(Aktualizr, FullNoCorrelationId) {
   TemporaryDirectory temp_dir;
   auto http = std::make_shared<HttpFakeNoCorrelationId>(temp_dir.Path());
-  Config conf = makeTestConfig(temp_dir, http->tls_server);
 
-  auto storage = INvStorage::newStorage(conf.storage);
-  auto sig = std::make_shared<boost::signals2::signal<void(std::shared_ptr<event::BaseEvent>)>>();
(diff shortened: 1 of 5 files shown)


patrickvacek:
Looks promising, I like the idea!

OYTIS:
Looks great!
Could we add some comments to the TestHelperProcess just to make clear what it does and why it's there. The name looks quite generic. Thanks!

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -677,23 +677,25 @@ class HttpFakeNoCorrelationId : public HttpFake {
...
+    DownloadResult download_result = aktualizr.Download(update_result.updates);
+    EXPECT_EQ(download_result.status, DownloadStatus::kSuccess);
 
-  InstallResult install_result = aktualizr.Install(download_result.updates);
-  for (const auto& r : install_result.reports) {
-    EXPECT_EQ(r.status.result_code, data::UpdateResultCode::kOk);
+    InstallResult install_result = aktualizr.Install(download_result.updates);
+    for (const auto& r : install_result.reports) {
+      EXPECT_EQ(r.status.result_code, data::UpdateResultCode::kOk);
+    }
```
patrickvacek: Not that this is bad or something, but why the extra scope enclosure?

lbonn: It's explained in the commit message. I've seen two spurious failures of this test and I think that's because we count the number of events before the queue had the time to be flushed.

lbonn: That said, it probably needs a comment, good point :)

patrickvacek: Oh, perfect, now I get it. Sorry, I do normally read the commits but forgot this time.

PR #986, opened 2018-10-31
WIP: Use checksums to match images in db

patriotyk:


src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1028,67 +1024,97 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
-    if (sqlite3_blob_open(db_.get(), "main", "target_images", "image_data", row_id, 0, &blob_) != SQLITE_OK) {
-      LOG_ERROR << "Could not open blob: " << db_.errmsg();
-      throw exc;
-    }
-    size_ = static_cast<size_t>(sqlite3_blob_bytes(blob_));
+    int statement_state;
+    while ((statement_state = statement.step()) == SQLITE_ROW) {
+      auto sha256 = statement.get_result_col_str(2);
+      auto sha512 = statement.get_result_col_str(3);
+      if (!sha256 && !sha512) {
```
lbonn: `!sha256 || !sha512`?

patriotyk: No, we skip if target don't have both hashes

src/libaktualizr/uptane/fetcher_test.cc
```diff
@@ -56,20 +74,58 @@ TEST(fetcher, fetch_with_pause) {
...
+  
+  Uptane::Fetcher f(config, storage, http);
+  auto result = f.fetchVerifyTarget(target);
+  EXPECT_TRUE(result);
+  auto h2 = storage->openTargetFile(target);
+  EXPECT_NE(h2, nullptr);
+  EXPECT_FALSE(h2->isPartial());
+  EXPECT_EQ(h2->rsize(), 2048);
+}
+
```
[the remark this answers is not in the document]
patriotyk: fetchVerifyTarget checks hashsum and returns true only if it is correct.

src/libaktualizr/storage/sqlstorage.cc
lbonn: Oh ok. But the code that follows is calling `*` on both. So if one is defined and not the other, it will pass the check and then crash when dereferencing the other.

src/libaktualizr/uptane/fetcher_test.cc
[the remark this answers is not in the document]
patrickvacek: But why is it incorrect the first time we check it?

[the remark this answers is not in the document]
patriotyk: Because  first time, server returns only half of file.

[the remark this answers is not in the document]
patrickvacek: Oh, now I see it. Thanks! I missed that before.

src/libaktualizr/storage/sqlstorage.cc
patriotyk: No, because it is NOT NULL column. So it  can be empty string  or hash. But in this case `if (!sha256 && !sha512) {` is incorrect, we should check for empty string. So it should be `if ((*sha256).empty() && (*sha512).empty()) {`

src/libaktualizr/storage/invstorage.h
```diff
@@ -145,11 +149,13 @@ class INvStorage {
   virtual void storeInstallationResult(const data::OperationResult& result) = 0;
   virtual bool loadInstallationResult(data::OperationResult* result) = 0;
   virtual void clearInstallationResult() = 0;
+  virtual boost::optional<std::pair<int64_t, size_t>> isTargetFileExists(const Uptane::Target& target) const = 0;
```
patrickvacek: Not a big deal, but would it be better if this were named something like `checkTargetFile`?

tests/fake_http_server/fake_http_server.py
```diff
@@ -26,7 +26,7 @@ def do_GET(self):
                 if auth_list[0] == 'Bearer' and auth_list[1] == 'token':
                     self.wfile.write(b'{"status": "good"}')
             self.wfile.write(b'{}')
-        elif self.path.endswith('/large_file'):
+        elif self.path.endswith('/large_file') or self.path.endswith('/large_iterupted'):
```
patrickvacek: Very minor, but should be `large_interrupted`.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -933,61 +936,97 @@ void SQLStorage::clearInstallationResult() {
...
+      "SELECT rowid, real_size, sha256, sha512 FROM target_images WHERE filename = ?;", target.filename());
+
+  int statement_state;
+  while ((statement_state = statement.step()) == SQLITE_ROW) {
+    auto sha256 = statement.get_result_col_str(2);
+    auto sha512 = statement.get_result_col_str(3);
+    if ((*sha256).empty() && (*sha512).empty()) {
+      LOG_WARNING << "Image without checksum: " << target.filename();
+      continue;
+    }
```
patrickvacek: This will probably be a problem for anyone who upgrades to this version of aktualizr with a database with targets stored before we added these hashes. We either will have to manually add those hashes while migrating or accept targets without any hashes. Or perhaps some way to indicate a legacy target could work.

patriotyk:
@lbonn I restored storage method to throw exception again

src/libaktualizr/storage/sqlstorage.cc
patrickvacek: After our discussion, I agree that migration compatibility is actually not a problem. The worst that would happen is the user would have to re-download a package, but that's actually quite unlikely to be possible or necessary. Perhaps leaving a comment somewhere to explain that is a good idea, though.

patriotyk:
Have I missed something?

PR #989, opened 2018-11-05
Reference doxygen docs in the main docs and explain how to update them.

patrickvacek:


patrickvacek:
We might want to consider having CI do this automatically at some point.

docs/README.adoc
```diff
@@ -29,3 +29,19 @@ link:./security.adoc[security.adoc] - A brief description of properties a system
...
+
+== Developer documentation
+
+Additional documentation intended for developers that may need to use the libaktualizr API should refer to link:https://advancedtelematic.github.io/aktualizr/index.html[the doxygen output on github].
+
+=== Updating doxygen on github
+
+To update the doxygen documentation on github, you will need to do something like the following:
+
+1. In an aktualizr repo, run `make docs` in the build directory.
```
lbonn: Every item starts with `1. `. Is it intentional?

patrickvacek: Yep, when it's rendered, the numbering is automatic: https://github.com/advancedtelematic/aktualizr/blob/169013c20e14cfb22268b7509ca0676a9ac41b16/docs/README.adoc

lbonn: Ok thanks. Could have looked that up I guess, sorry.

tkfu:
> We might want to consider having CI do this automatically at some point.

I think I've got a ticket for that somewhere--we definitely should, especially because we really ought to have doxygen docs available for each branch. Depending on how much of a backlog I have when I'm back in the office, I'll try and do it next week.

tkfu:
(I clicked the wrong button on mobile and closed by accident, sorry.)

merltron:
Generally looks fine to me. Would be good to know if there's a way to automatically added the applicable tag into the doc. The doc that we have published on github.io is for latest master I assume? Is that what we want? or should we only publishing for content that is tagged "2018.xx?"

merltron:
We should also look at publishing all our doc together rather than having some some doc on "whatever.github.io" and some on "docs.whatever.com". We could use the HERE branding docker image to make the doxygen look more like HERE doc. (they do the same for javadoc: https://developer.here.com/documentation/android-starter/api_reference_java/index.html )

patrickvacek:
@merltron Good question about latest docs versus a specific release. It's probably a good idea to stick to only tagged versions, since that is what we are moving towards in general. We should then list the release version in the doxygen frontpage as well, just so that it is clear. We could then think about automating this so that it only happens when making a new release.

Regarding publishing things with the same theme and in the same place, I have no argument. Perhaps we could leave the aktualizr repo as is but then mirror the contents (or reframe them, or whatever) in some official HERE place where it all fits together in one place with one style.

PR #991, opened 2018-11-05
Disable timeout for download operation 

eu-smirnov:
Timeout for all libcurl operations is set to 60 sec,
which is not suitable for download, as it could take
much longer. To prevent slow retrieval attack we set
CURLOPT_LOW_SPEED_LIMIT and CURLOPT_LOW_SPEED_TIME.

### DIFF
--- a/src/libaktualizr/http/httpclient.cc
+++ b/src/libaktualizr/http/httpclient.cc
@@ -239,6 +239,7 @@ HttpResponse HttpClient::download(const std::string& url, curl_write_callback ca
   curlEasySetoptWrapper(curl_download, CURLOPT_FOLLOWLOCATION, 1L);
   curlEasySetoptWrapper(curl_download, CURLOPT_WRITEFUNCTION, callback);
   curlEasySetoptWrapper(curl_download, CURLOPT_WRITEDATA, userp);
+  curlEasySetoptWrapper(curl_download, CURLOPT_TIMEOUT, 0);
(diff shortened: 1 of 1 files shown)


eu-smirnov:
I don't get why garage-deploy-missing-commit test on jenkins has failed. Could someone have a look, please?

src/libaktualizr/logging/boost_logging.cc
```diff
@@ -18,9 +19,10 @@ int64_t get_curlopt_verbose() { return gLoggingThreshold <= boost::log::trivial:
 
 void logger_init() {
   gLoggingThreshold = boost::log::trivial::info;
-  logging::add_console_log(std::cout, boost::log::keywords::format = "%Message%",
+  logging::add_console_log(std::cout, boost::log::keywords::format = "[%TimeStamp%]: %Message%",
```
lbonn: I'm unsure about this, when logging to stdout it's either handled by systemd which does the timestamping or during debug, when we can use `ts` from [moreutils](https://joeyh.name/code/moreutils/): `aktualizr | ts "%F %H:%M:%.S"`.

If we decide to have more flexible logging builtin, storing timestamps in log files would make sense though.

eu-smirnov: Yes, you are right, will be ugly in systemd case.

eu-smirnov:
Well, looks like the problem with garage-deploy-missing-commit test solved itself, so I'm going to merge this.

PR #995, opened 2018-11-07
Async api with std::future

patriotyk:


src/libaktualizr/primary/aktualizr.cc
```diff
@@ -43,33 +43,54 @@ void Aktualizr::systemSetup() {
...
+      auto task = api_queue_.dequeue();
+      if (shutdown_) {
+        return;
+      }
+      task();
+    }
+  })
+      .detach();
+  uptane_client_->initialize();
+}
```
patrickvacek: We should probably initialize SotaUptaneClient before launching the thread just to be safe.

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -43,33 +43,54 @@ void Aktualizr::systemSetup() {
...
 
 int Aktualizr::RunForever() {
-  SendDeviceData();
+  SendDeviceData().get();
   while (!shutdown_) {
-    UptaneCycle();
+    UptaneCycle().get();
     std::this_thread::sleep_for(std::chrono::seconds(config_.uptane.polling_sec));
   }
   return EXIT_SUCCESS;
```
patrickvacek: While `RunForever` does not need to return a future, it probably should be asynchronous and return immediately.

src/libaktualizr/primary/aktualizr.h
```diff
@@ -13,6 +13,38 @@
...
+  void shutDown() {
+    shutdown_ = true;
+    c_.notify_all();
+  }
+
+ private:
+  std::queue<std::function<void()>> q_;
+  mutable std::mutex m_;
+  std::condition_variable c_;
+  std::atomic_bool shutdown_{false};
```
patrickvacek: Can this reuse the `shutdown_` member of the Aktualizr class? It seems odd to have two variables that seem to do the same thing.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -32,6 +32,7 @@ Config makeTestConfig(const TemporaryDirectory& temp_dir, const std::string& url
 }
 
 void verifyNothingInstalled(const Json::Value& manifest) {
+  std::cout << "MMMMMMMMMMM\n";
```
patrickvacek: lol

src/libaktualizr/primary/aktualizr.h
patriotyk: We can use only shutdown_ from queue in Aktualizr class, but I am not sure it is good idea. What do you think?

patrickvacek: What would be the problem with only using one? I just don't like the redundancy of having two variables that to the same thing. It seems like you basically have to set both to actually get the system to shut down. Or am I missing something? If they really are different, maybe we should rename one of them.

patrickvacek: No, you are probably right. It's fine how it is. It was just my overactive sense for removing redundancy!

src/libaktualizr/primary/aktualizr.cc
patriotyk: Do you mean it should run in separate thread?

patrickvacek: Yeah, mostly likely. Maybe there's a clever way to reuse the API queue thread, maybe not.

patriotyk: Yes, we cannot, because we will get dead lock.

patrickvacek:
Looks pretty good to me, thanks! I like the new tests. A bit silly but I think it works. I've pushed a small update to the changelog, and then I think we can merge when that passes.

lbonn:
Also like it!

There is this `OstreeObject.UpdateSuccess` failure on Jenkins that has been happening quite a lot recently, this test is pretty unstable for some reason.

But, in all likelihood, it's unrelated to your changes so I'd say we are good to go.

patrickvacek:
@lbonn Can you restart the Jenkins test? I can't get to Jenkins, try as I might.

lbonn:
done

lbonn:
@patriotyk looks like the APICheck is not so reliable, there was a failure on the master branch CI.

We should avoid hard-coded waiting time in tests as it is rather brittle (and CI machines are often under load which makes everything slower). So the test should be fixed to use some proper synchronization, possibly with a (large) timeout.

PR #996, opened 2018-11-08
fetcher: urlencode target names

OYTIS:

Needed for Sachertorte, but a good thing to have in general probably.


patrickvacek:
+1 for tests.

PR #998, opened 2018-11-09
Assign explicit num code to UpdateResultCode

lbonn:
This is part of our external API (in the general sense), we should not rely on C's tricks to define this enum.

### DIFF
--- a/src/libaktualizr/utilities/types.h
+++ b/src/libaktualizr/utilities/types.h
@@ -151,45 +151,45 @@ enum class UpdateResultCode {
   /// Operation executed successfully
   kOk = 0,
   /// Operation has already been processed
-  kAlreadyProcessed,
+  kAlreadyProcessed = 1,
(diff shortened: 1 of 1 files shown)


lbonn:
@patrickvacek, these codes are supposed to be stable in time. I once read a nice set of recommendations for writing stable C apis and explicitly defining enum values was part of it (wish I could find this doc again).

It's very easy (dumb I know), to add a code in the middle and break everything.

Plus, backend asked me about these codes and since C is not their go-to language, I had to explain the trick.

patrickvacek:
Fair enough. Sometimes it's worth learning from python. [Explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python).

patrickvacek:
And "Readability counts".

lbonn:
Yes it could apply to `results.h`. I'll leave it for the moment, we can change it next time we touch this file.

PR #1002, opened 2018-11-12
api: add API to send installation manifest manually

OYTIS:
A somewhat naive API to send installation result manually. A better way to solve the problem would be to redesign what Install call does, but it requires a bit more consideration (and we hope to deliver the demo next week).


lbonn:
Any update on this?

lbonn:
Closing for now, the manifest generation part is currently being reworked and should also contain equivalent functionality.

PR #1003, opened 2018-11-12
Tests/ota 1353/actions update

patrickvacek:
Mostly documentation and comments, but a few minor fixes to various tests.


lbonn:
Great work! We still have some of these unstable tests causing trouble unfortunately.

patrickvacek:
Rebased and removed my failed attempt to fix `authenticate_test`'s `good_cert_zip` test case. We'll have to come back to that another time.

patrickvacek:
I think I've finally fixed one of the broken tests (`test-missing-commit`). I broke it further in this PR and fixed both problems with it. I have no idea how this worked before. `set -e` and `set -o pipefail` were both disastrous for that test.

patrickvacek:
@lbonn No, those are valid points. It took seconds to fix them, so I did. I left `exit $?` for now just to be explicit.

PR #1004, opened 2018-11-13
Detect reboot

lbonn:
Not useful by itself, will need to be integrated closely to the ostree package manager.

### DIFF
--- a/src/libaktualizr/bootloader/bootloader.cc
+++ b/src/libaktualizr/bootloader/bootloader.cc
@@ -1,3 +1,8 @@
+#include <fcntl.h>
+#include <sys/stat.h>
+
+#include <boost/filesystem.hpp>
+
 #include "bootloader.h"
 #include "utilities/config_utils.h"
 #include "utilities/exceptions.h"
@@ -41,6 +46,23 @@ void BootloaderConfig::updateFromPropertyTree(const boost::property_tree::ptree&
 
 void BootloaderConfig::writeToStream(std::ostream& out_stream) const {
   writeOption(out_stream, rollback_mode, "rollback_mode");
+}
+
+static const boost::filesystem::path reboot_beacon_dir = "/var/run/aktualizr-session";
+static const boost::filesystem::path reboot_beacon_filename = "need_reboot";
+
+Bootloader::Bootloader(const BootloaderConfig& config, INvStorage& storage) : config_(config), storage_(storage) {
+  if (mkdir(reboot_beacon_dir.c_str(), S_IRWXU) == -1) {
+    struct stat st {};
+    stat(reboot_beacon_dir.c_str(), &st);
+    if (((st.st_mode & S_IFDIR) == 0) || (st.st_mode & (S_IRGRP | S_IROTH | S_IWGRP | S_IWOTH)) != 0) {
+      LOG_WARNING << "Could not create " << reboot_beacon_dir << " securely, reboot detection support disabled";
+      reboot_detect_supported_ = false;
+      return;
+    }
+    // mdkir failed but directory is here with good permissions: someone created it for us
+  }
+  reboot_detect_supported_ = true;
 }
 
 void Bootloader::setBootOK() const {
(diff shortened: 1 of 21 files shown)


src/libaktualizr/bootloader/bootloader.cc
```diff
@@ -37,10 +42,30 @@ inline void CopyFromConfig(RollbackMode& dest, const std::string& option_name, c
...
+    struct stat st {};
+    stat(config_.reboot_sentinel_dir.c_str(), &st);
+    if (((st.st_mode & S_IFDIR) == 0) || (st.st_mode & (S_IRGRP | S_IROTH | S_IWGRP | S_IWOTH)) != 0) {
+      LOG_WARNING << "Could not create " << config_.reboot_sentinel_dir
+                  << " securely, reboot detection support disabled";
+      reboot_detect_supported_ = false;
+      return;
+    }
+  }
+  reboot_detect_supported_ = true;
```
patrickvacek: Why bother with the second check here? If you can't create the directory for any reason, then there's a problem. If the first test succeeds but the second fails, you are in a strange state that probably won't work but `reboot_detect_supported_` is set to true.

src/libaktualizr/bootloader/bootloader.cc
```diff
@@ -94,3 +119,42 @@ void Bootloader::updateNotify() const {
...
+
+  // true if set in storage and no volatile flag
+
+  bool beacon_exists = boost::filesystem::exists(reboot_sentinel_);
+  bool need_reboot = false;
+
+  storage_.loadNeedReboot(&need_reboot);
+
+  return need_reboot && !beacon_exists;
+}
```
patrickvacek: The name here is slightly misleading, although given the context, it's probably fine. The problem is just that this does more than just detecting a reboot -- it's detecting an *expected* reboot. If we rebooted and weren't waiting for it, this would return false, despite that based on the name, you'd expect true. This isn't a real problem unless we reuse this logic for something else later without being careful.

src/libaktualizr/primary/aktualizr.h
```diff
@@ -159,6 +159,13 @@ class Aktualizr {
    */
   void UptaneCycle();
 
+  // reboot handling
+
+  bool RebootDetected();
+
+  void RebootFlagSet();
+  void RebootFlagClear();
+
```
patrickvacek: What are these for? Do they need to be part of the API?

src/libaktualizr/bootloader/bootloader.cc
lbonn: If `mkdir` fails but all the conditions checks out, it means that the directory was already there (I hope, you can double check).

The point is to avoid race conditions (better ask forgiveness...)

I'll add a comment to clarify.

src/libaktualizr/primary/aktualizr.h
lbonn: Remnants of a first design, they should go away (for now at least). Thanks!

src/libaktualizr/bootloader/bootloader.cc
lbonn: Yes, you're probably right. Do you have an alternate naming suggestion or should it just be better documented (well-placed comment?)

lbonn:
Thanks patrick, I've acted on some of your concerns. There remains the question of naming.

I'll update the action lists as well but for now I would prefer keeping it outside of the changelog as it's more like an internal building brick than a full-fledged feature.

src/libaktualizr/bootloader/bootloader.cc
patrickvacek: Perfect, that makes sense.

src/libaktualizr/bootloader/bootloader.cc
patrickvacek: No, I think it's fine for now. I can't think of anything better, so a good comment is probably enough.

PR #1006, opened 2018-11-14
Make test_fetcher more stable

lbonn:
There was a race at the end


lbonn:
@patriotyk ok I see. I think the race is still there, it shouldn't be a problem of 2 or 4 seconds since it seems to be between the last dummy pause and the end of download.

So I'll merge anyway, it shouldn't be too hard to rebase yours.

PR #1008, opened 2018-11-14
Only sleep as long as it takes for the server to be ready.

patrickvacek:
Lots of tests sleep for 3 or 4 seconds in the expectation that the test server will be ready in that time. However, our CI regularly fails when that assumption is not true. So instead of increasing the time, just poll the server for readiness and sleep only as long as necessary.

I'm not sure this is bulletproof yet, and it seems valgrind doesn't play nicely. But I think this is a better solution than just sleeping.

### DIFF
--- a/src/sota_tools/authenticate_test.cc
+++ b/src/sota_tools/authenticate_test.cc
@@ -129,10 +129,11 @@ int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
   TestHelperProcess server_process("tests/fake_http_server/ssl_server.py");
   TestHelperProcess server_noauth_process("tests/fake_http_server/ssl_noauth_server.py");
-  sleep(4);
-  // TODO: these do not work because the server expects auth!
+  // TODO: this do not work because the server expects auth! Let's sleep for now.
+  // (could be replaced by a check with raw tcp)
   // TestUtils::waitForServer("https://localhost:1443/");
-  // TestUtils::waitForServer("https://localhost:2443/");
+  sleep(4);
+  TestUtils::waitForServer("https://localhost:2443/");
   return RUN_ALL_TESTS();
 }
 #endif
--- a/src/sota_tools/authenticate_test.cc
+++ b/src/sota_tools/authenticate_test.cc
@@ -129,8 +129,10 @@ int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
   TestHelperProcess server_process("tests/fake_http_server/ssl_server.py");
   TestHelperProcess server_noauth_process("tests/fake_http_server/ssl_noauth_server.py");
-  TestUtils::waitForServer("http://localhost:1443/");
-  TestUtils::waitForServer("http://localhost:2443/");
+  sleep(4);
+  // TODO: these do not work because the server expects auth!
+  // TestUtils::waitForServer("https://localhost:1443/");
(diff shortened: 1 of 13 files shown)


lbonn:
I like this a lot (if it works fine)! The wait time could be a bit less than 1s, so that we waste even less time. Maybe 200ms?

patrickvacek:
> (if it works fine)

That is the question. I'm struggling to understand why this doesn't work. It worked for me locally yesterday, but CI obviously doesn't like it, and when I run it through ctest locally (i.e. with valgrind) it fails. Now when I try it locally (gdb without valgrind), sometimes it works and sometimes it doesn't.

> The wait time could be a bit less than 1s, so that we waste even less time.

Agreed.

lbonn:
It should probably time out at some point too.

lbonn:
I pushed another version on top with timeout and shorter delays. It looks like it's still failing though

patrickvacek:
> I pushed another version on top with timeout and shorter delays.

Thanks!

> It looks like it's still failing though

I'm wondering if it has something to do with the SSL authentication that some of our test servers require. Some of these reliably succeed and some reliably fail, so there must be something in common.

lbonn:
A handful of special cases were hiding in the test suite. In particular, test of mutual TLS and UDP discovery.

A more robust detection could perhaps be done with raw sockets but I think we can call it a day for now.

src/sota_tools/authenticate_test.cc
```diff
@@ -130,6 +130,9 @@ int main(int argc, char **argv) {
   TestHelperProcess server_process("tests/fake_http_server/ssl_server.py");
   TestHelperProcess server_noauth_process("tests/fake_http_server/ssl_noauth_server.py");
   sleep(4);
+  // TODO: these do not work because the server expects auth!
+  // TestUtils::waitForServer("https://localhost:1443/");
+  // TestUtils::waitForServer("https://localhost:2443/");
```
patrickvacek: Does the noauth server really require auth?

tests/fake_http_server/ssl_server.py
```diff
@@ -1,8 +1,14 @@
...
+import socket
 import ssl
 
-httpd = HTTPServer(('localhost', 1443), SimpleHTTPRequestHandler)
+class ReUseHTTPServer(HTTPServer):
+    def server_bind(self):
+        self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        HTTPServer.server_bind(self)
+
+httpd = ReUseHTTPServer(('localhost', 1443), SimpleHTTPRequestHandler)
```
patrickvacek: I would love to get rid of these hardcoded ports, but that will probably be a job for another day.

tests/test_utils.cc
```diff
@@ -56,6 +89,8 @@ void TestHelperProcess::run(const char *argv0, const char *args[]) {
     prctl(PR_SET_PDEATHSIG, SIGTERM);
 #endif
     execvp(argv0, const_cast<char *const *>(args));
+    std::cout << "Could not execute child " << argv0 << "\n";
+    exit(1);
```
patrickvacek: +1!

tests/fake_http_server/ssl_server.py
lbonn: Yes, I agree...

PR #1009, opened 2018-11-15
Set an `ecu_serial` in example virtualsec.json

lbonn:
It is actually needed.

Thanks @EpicNuts!

### DIFF
--- a/config/secondary/virtualsec.json
+++ b/config/secondary/virtualsec.json
@@ -1,6 +1,7 @@
 {
   "secondary_type" : "virtual",
   "partial_verifying" : "false",
+  "ecu_serial" : "demo-vsec1-serial",
   "ecu_hardware_id" : "demo-virtual",
 
   "full_client_dir" : "storage/demo-vsec1",


lbonn:
Yes but that's probably a bug. I have seen error messages that ends with `error ... with Ecu: `.

If you look in the code, it reads it from the Json and just picks an empty serial if it's not there.

patrickvacek:
> it reads it from the Json and just picks an empty serial if it's not there.

I see, that's probably not ideal. We should complain if someone tries to do that. I was hoping that one was generated for virtual secondaries, but if we don't do that, we shouldn't allow empty serials at all.

lbonn:
Hmm ok I overlooked some stuff...
https://github.com/advancedtelematic/aktualizr/blob/348822d914f422a8d7b34a1e98cfdd1fa3cf816c/src/libaktualizr/config/config_test.cc#L127

https://github.com/advancedtelematic/aktualizr/blob/348822d914f422a8d7b34a1e98cfdd1fa3cf816c/src/libaktualizr/uptane/managedsecondary.h#L29

The message I saw was probably an early error during initialization, maybe we should just fix that.

patrickvacek:
> maybe we should just fix that

Sure. I was pretty sure there were hacks to make it fairly "user-friendly". I think I recall specifically trying to allow not specifying the serial (for virtual secondaries only!) so that we could make testing easier. Otherwise you run into problems with reusing the serial with the same account. Maybe that is less of a problem now, but it used to be worse.

lbonn:
Closing it for now

PR #1011, opened 2018-11-19
Update /docs/README.adoc, move instructions for fake device

tkfu:
Should not be merged until after https://github.com/advancedtelematic/ats-garage-docs/pull/65.

The doc is currently live at http://docs-staging.atsgarage.com.s3-website.eu-central-1.amazonaws.com/quickstarts/install-a-client-locally-with-fake-secondaries.html


patrickvacek:
Would it perhaps be better to link directly from the root README or docs/README to the new guide? The extra layer of indirection seems less than ideal to me.

merltron:
FYI the terminology "primary" and "secondary" is specific to Uptane right? If it's a quick start, might it be possible that I haven't read anything about Uptane yet? Perhaps just say "device" in the title and link to the general Uptane explanation and the differences between "primary and secondary"?

PR #1014, opened 2018-11-23
Fix/hmi stub pause

patrickvacek:
Fix deadlock in hmi-stub caused by checking for updates while paused.

Not perfect (we just ignore the result of the check) but better than freezing the application. We might want to improve this later, because at present waiting on futures while paused is asking for trouble.

### DIFF
--- a/actions.md
+++ b/actions.md
@@ -126,7 +126,7 @@ These are the primary actions that a user of libaktualizr can perform through th
       - [x] Send an event report (see below)
     - [x] Install an update on the primary
       - [ ] Install an OSTree update on the primary
-      - [ ] Notify "reboot needed" after an OStree update
+      - [ ] Notify "reboot needed" after an OSTree update
       - [x] Install a binary update on the primary (uptane_test.cc, aktualizr_test.cc)
     - [x] Store installation result for primary (uptane_test.cc)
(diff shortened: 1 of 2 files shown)


patrickvacek:
> What if we check `isPaused` flag in API calls and just fail when it's true?

Sounds reasonable, but is that actually what we want to do? I'd want to check with the PO/PMs that that would match their expectations.

>  It also seems to me that we might want to have some kind of a `Abort` call complimenting the `Resume`.

Yes, absolutely, but I was expecting there to be a ticket created for that, and I haven't seen it yet.

I will bring these issues up and see about getting the necessary decisions made.

patrickvacek:
> What if we check `isPaused` flag in API calls and just fail when it's true?

Made a ticket after discussing with Xavier: https://saeljira.it.here.com/browse/OTA-1726

> It also seems to me that we might want to have some kind of a `Abort` call complimenting the `Resume`.

Xavier is looking into this. Last he had heard this was not actually a priority. I think it makes sense, but it does require deciding how to handle this and understanding the likely use case.

patrickvacek:
I think we should consider this a quick fix to solve a problem in hmi-stub. Addressing the underlying API issues will require further consideration and work.

PR #1017, opened 2018-11-27
More tests

patriotyk:


src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -295,6 +295,27 @@ TEST(Uptane, UptaneSecondaryAdd) {
...
+  config.provision.primary_ecu_serial = "testecuserial";
+  config.storage.path = temp_dir.Path();
+  config.pacman.type = PackageManager::kNone;
+  UptaneTestCommon::addDefaultSecondary(config, temp_dir, "secondary_ecu_serial", "secondary_hardware");
+  UptaneTestCommon::addDefaultSecondary(config, temp_dir, "secondary_ecu_serial", "secondary_hardware");
+
+  auto storage = INvStorage::newStorage(config.storage);
+  EXPECT_THROW(SotaUptaneClient::newTestClient(config, storage, http), std::runtime_error);
+}
+
```
patrickvacek: Looks good. I have a few suggestions for this one:
1. I think we can remove most of the configuration. Most of it will never get touched because we fail so early.
2. Change the hardware ID of the second secondary.
3. After the failure, change the serial of the second secondary and try again to prove that it should work.

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -316,6 +316,67 @@ TEST(Uptane, UptaneSecondaryAddSameSerial) {
...
+    if (ecus[0].serial.ToString() == "new_secondary_ecu_serial") {
+      EXPECT_EQ(ecus[0].state, EcuState::kNotRegistered);
+      EXPECT_EQ(ecus[1].serial.ToString(), "secondary_ecu_serial");
+      EXPECT_EQ(ecus[1].state, EcuState::kOld);
+    } else if (ecus[0].serial.ToString() == "secondary_ecu_serial") {
+      EXPECT_EQ(ecus[0].state, EcuState::kOld);
+      EXPECT_EQ(ecus[1].serial.ToString(), "new_secondary_ecu_serial");
+      EXPECT_EQ(ecus[1].state, EcuState::kNotRegistered);
+    } else {
+      FAIL();
```
patrickvacek: Can you add an informative error message here? You can do something like `FAIL() << "Unexpected secondary serial in storage " << ecus[0].serial.ToString();`.

src/libaktualizr/uptane/uptane_test.cc
patriotyk: > 3. After the failure, change the serial of the second secondary and try again to prove that it should work.
We have had bug here. We didn't clear misconfigured ecus.

patrickvacek: > We didn't clear misconfigured ecus.

That is true and that's fine for now. But you can clear the misconfiguration table manually, right?

patriotyk: I have fixed it, then commeneted here, but haven't pushed changes.

src/libaktualizr/uptane/fetcher_test.cc
```diff
@@ -81,9 +89,10 @@ void test_pause(const Uptane::Target& target) {
   std::promise<void> end_pausing;
 
   EXPECT_EQ(f.setPause(true), PauseResult::kNotDownloading);
+  EXPECT_EQ(f.setPause(false), PauseResult::kNotPaused);
 
   std::thread([&f, &end_pausing] {
-    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+    std::this_thread::sleep_for(std::chrono::milliseconds(400));  // wait for download start
```
patrickvacek: Could we change this to something that waits only as long as necessary? I think we should move away from waiting an arbitrary number of seconds or milliseconds.

src/libaktualizr/uptane/uptane_test.cc
patrickvacek: I think points 1 and 3 are covered now, but I think point 2 still stands: I recommend changing the hardware ID of the second secondary.

tests/httpfake.h
```diff
@@ -110,11 +110,12 @@ class HttpFake : public HttpInterface {
   }
 
   HttpResponse put(const std::string &url, const Json::Value &data) override {
-    (void)data;
+    if (url.find("putmanifesterror") != std::string::npos) {
+      return HttpResponse(url, 504, CURLE_OK, "");
+    }
```
patrickvacek: Laurent recently did some work to try to remove most of the test-case-specific code out of httpfake.h itself and instead derive a new class that overloads the one thing necessary. Could you do that instead? There should be plenty of examples in uptane_test.cc.

patriotyk:
Now It should be ok

PR #1018, opened 2018-11-27
Feat/ota 1569/manifest after reboot

lbonn:
Based on #1016. Worked as PoC, will need to be more extensively tested.

### DIFF
--- a/actions.md
+++ b/actions.md
@@ -18,10 +18,11 @@ These are the primary actions that a user of libaktualizr can perform through th
   - [x] Initialize device ID
     - [x] Use a provided device ID (OTA-985, uptane_init_test.cc)
     - [x] Generate a random device ID (OTA-986, utils_test.cc, uptane_init_test.cc)
-  - [ ] Finalize a pending update that requires reboot
-    - [ ] Store installation result
-    - [ ] Send manifest
-    - [ ] Update is not in pending state anymore after processing (success or failure)
+  - [x] Finalize a pending update that requires reboot (aktualizr_test.cc)
+    - [x] Store installation result (aktualizr_test.cc)
+    - [x] Send manifest (aktualizr_test.cc)
+    - [x] Update is not in pending state anymore after successful finalization (aktualizr_test.cc)
+    - [ ] Update is not in pending state anymore after failed finalization
   - [x] Provision with the server
     - [x] Automatically provision (OTA-983, uptane_init_test.cc, uptane_ci_test.cc, auto_prov_test.py)
       - [x] Extract credentials from a provided archive (config_test.cc, utils_test.cc)
@@ -131,9 +132,9 @@ These are the primary actions that a user of libaktualizr can perform through th
     - [x] Install an update on the primary
       - [ ] Install an OSTree update on the primary
       - [ ] Notify "reboot needed" after an OSTree update trigger
-      - [ ] Set new version to pending status after an OSTree update trigger
-      - [ ] Send EcuInstallationAppliedReport to server after an OSTree update trigger
-      - [ ] Uptane check for updates and manifest sends are disabled while an installation is pending reboot
+      - [x] Set new version to pending status after an OSTree update trigger (aktualizr_test.cc)
+      - [x] Send EcuInstallationAppliedReport to server after an OSTree update trigger (aktualizr_test.cc)
+      - [x] Uptane check for updates and manifest sends are disabled while an installation is pending reboot (aktualizr_test.cc)
       - [ ] Trigger a system reboot if `reboot_after_install` is set in the configuration
       - [x] Install a binary update on the primary (uptane_test.cc, aktualizr_test.cc)
     - [x] Store installation result for primary (uptane_test.cc)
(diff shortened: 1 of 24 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -245,6 +286,9 @@ void SotaUptaneClient::initialize() {
 
   verifySecondaries();
   LOG_DEBUG << "... provisioned OK";
+
+  // TODO: right place?
+  finalizeAfterReboot();
```
patrickvacek: Yes, I think so. The other option, particularly if we only want to incur network traffic when the user requests it, is in sendDeviceData or checkUpdates, but then we have to have logic to make sure we only send it after a reboot and not every time. This is probably good for now.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -255,16 +255,19 @@ Json::Value SotaUptaneClient::AssembleManifest() {
 }
 
 bool SotaUptaneClient::hasPendingUpdates(const Json::Value &manifests) {
+  // TODO: is this still relevant? Should use new installed_versions?
+  bool has_secondary_in_progress = false;
```
patrickvacek: Pretty sure that your work entirely supersedes what was there before, although that depends on how we set the `pending` flag in the secondary. I'd prefer to unify this if possible and use one canonical method for the pending state, and your new method seems like the better method to me.

lbonn:
- rebased on master
- added some (difficult) testing actions to the list
- reworked some parts around installation_result...

Needs:

- fixing current tests
- probably at least one convincing automatic test (if you have great ideas, please tell)
- (maybe) auto-reboot config flag. In theory I could skip it for now but I think it will probably be needed for oe-selftests...

actions.md
```diff
@@ -126,7 +130,11 @@ These are the primary actions that a user of libaktualizr can perform through th
       - [x] Send an event report (see below)
     - [x] Install an update on the primary
       - [ ] Install an OSTree update on the primary
-      - [ ] Notify "reboot needed" after an OSTree update
+      - [ ] Notify "reboot needed" after an OSTree update trigger
+      - [ ] Set new version to pending status after an OSTree update trigger
+      - [ ] Send EcuInstallationAppliedReport to server after an OSTree update trigger
+      - [ ] Uptane check for updates and manifest sends should be disabled while an installation is pending reboot
+      - [ ] Should trigger a system reboot if `reboot_after_install` is set in the configuration
```
patrickvacek: Minor, but I've preferred stating things definitively instead of using "should", i.e. "Disable checking for updates and sending manifest while an installation is pending reboot" and "Trigger a system reboot..."

lbonn:
Ok the ci should pass right now but I still think it needs more love.

TODO:

* have a version (probably a config flag) of the fake package manager that simulates the two step install process, with `install()` returning pending and needing a `finalize()`
* use this simulator to write realistic tests, approximate the ostree behaviour
* (maybe) mock the hard-to-test part of OstreeManager: `getCurrent()` which uses `ostree_sysroot_get_booted_deployment()` which asks the kernel, for more realistic coverage

Not TODO:

* auto-reboot feature, will be in a separate work when/if needed, maybe when we update e2e

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Good plan. I've debated removing OPC-UA entirely, but for now just leaving comments about things presumed to be broken is a good idea.

patrickvacek:
Agreed with your plan. I'd say this is mergeable once you have the upgraded fake package manager and the corresponding tests. If writing the mocked version of the OstreeManager is not too hard, go for it, but otherwise we can save that for a follow-up PR and test it manually in the meantime.

lbonn:
Ok I've added a test with a simulated OStree update with the fake package manager.

actions.md
```diff
@@ -131,9 +131,9 @@ These are the primary actions that a user of libaktualizr can perform through th
     - [x] Install an update on the primary
       - [ ] Install an OSTree update on the primary
       - [ ] Notify "reboot needed" after an OSTree update trigger
-      - [ ] Set new version to pending status after an OSTree update trigger
-      - [ ] Send EcuInstallationAppliedReport to server after an OSTree update trigger
-      - [ ] Uptane check for updates and manifest sends are disabled while an installation is pending reboot
+      - [x] Set new version to pending status after an OSTree update trigger
+      - [x] Send EcuInstallationAppliedReport to server after an OSTree update trigger
+      - [x] Uptane check for updates and manifest sends are disabled while an installation is pending reboot
```
patrickvacek: Would you mind listing in parentheses the file in which the test that covers these cases can be found, and also copying the same text here as a comment before the test? Minor, but for now that's the best I can do to link things together easily.

lbonn: Is it better now?

patrickvacek: Yes, but I have two further suggestions. First, the ticket number probably isn't necessary for now; I've only included ones that were requirements tickets. I don't know what the status of that system is, but I haven't seen any movement on that for a while. Second, could you add the text of the action to the code? As in, specifically put the items like "Set new version to pending status after an OSTree update trigger" in a comment before the test. It's not a perfect solution, but I just want to make sure that when we say the test is covered in some file, we can find exactly what we were thinking of when we made that claim.

lbonn: Updated again :)

PR #1019, opened 2018-11-30
Fix tests for non ostree build

eu-smirnov:


src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -477,6 +477,7 @@ TEST(Uptane, ProvisionOnServer) {
 TEST(Uptane, FsToSqlFull) {
   TemporaryDirectory temp_dir;
   Utils::copyDir("tests/test_data/prov", temp_dir.Path());
+  chmod(temp_dir.Path().c_str(), S_IRWXU);
```
patrickvacek: Why don't we just do this automatically in `copyDir`? I think it's safe to say that we always want to enforce the same specific permissions.

CMakeLists.txt
```diff
@@ -136,7 +136,7 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Valgrind)
 endif(NOT CMAKE_BUILD_TYPE)
 
-set(CMAKE_CXX_FLAGS_DEBUG "-g")
+set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g")
```
patrickvacek: How about `-0g`? I forget if clang supports that, but g++ definitely does, and it helps keep down the bloat a bit.

src/libaktualizr/uptane/uptane_test.cc
eu-smirnov: I think it's better to do it explicitly. Also, usual `copy` behavior is to copy permissions from the source. So if we replace `Utils::copyDir` with `copy_directory` from boost or C++14 this line will be needed.

CMakeLists.txt
eu-smirnov: Thanks, that will be better. Clang also has it since 4.0.

src/libaktualizr/uptane/uptane_test.cc
patrickvacek: Fair enough, that sounds reasonable.

PR #1020, opened 2018-11-30
Feat/ostree walker

patrickvacek:
I wrote most of this while trying to debug the missing OSTree object errors, and I figured it was worth polishing up a bit and submitting. Note that I've never successfully gotten a walk to finish, so I don't know what happens if it does. I suspect it will block indefinitely, but I'm not sure. Hopefully sooner or later I'll be able to figure that out.


patrickvacek:
By the way, I've solved the halting problem and optimized several parts of the code. I'm looking into  more optimizations for garage-deploy, but this work is ready for review. Note that I still haven't gotten to use the garage-push `--walk-tree` option in full glory yet.

lbonn:
So, did you manage to run a full `garage-check --walk-tree`?

patrickvacek:
> So, did you manage to run a full `garage-check --walk-tree`?

Yes, finally! I've also run a full `garage-push --walk-tree`, but I still haven't seen a case where it successfully found an uploaded an unexpectedly missing object.

patrickvacek:
Added some more work to improve garage-deploy speed.

For a fairly simple change, garage-push (with these changes) took this long (note that this does not include garage-sign, but that is usually trivial), measured with `time`:
```
real  0m9,098s
user  0m0,947s
sys 0m0,317s
```

Using garage-deploy from master to upload the same objects:
```
real  3m13,168s
user  0m27,168s
sys 0m1,309s
```

Using garage-deploy from this branch to upload the same objects:
```
real  1m37,113s
user  0m24,744s
sys 0m1,211s
```

I haven't been able to compare a larger set of changes/objects because of unreliable connectivity. (I actually had one error with the garage-deploy in this case, but it recovered and was trivial enough that it still outperformed the old garage-deploy by quite a bit.)

I'm also still quite impressed with how fast garage-push is compared with garage-deploy. The advantage of reading from disk instead of over a network is obvious, but I'm still wondering if there are further inefficiencies I haven't yet uncovered.

patrickvacek:
I think this should be reviewed and merged now. I've tested that garage-push will successfully push children objects even if the parents exist when running with `-w` and I've proven that garage-deploy is now substantially faster than before.

src/sota_tools/deploy.cc
```diff
@@ -11,8 +11,16 @@
 #include "treehub_server.h"
 #include "utilities/utils.h"
 
+bool CheckPoolState(const OSTreeObject::ptr &root_object, const RequestPool &request_pool) {
+  if (request_pool.run_mode() == RunMode::kWalkTree || request_pool.run_mode() == RunMode::kPushTree) {
+    return !request_pool.is_idle() && !request_pool.is_stopped();
+  } else {
```
lbonn: Clarity: having an exhaustive switch/case on all existing running modes would make this part easier to follow as it wouldn't need to know about the exact enum cases.

src/sota_tools/ostree_object_test.cc
```diff
@@ -12,7 +12,9 @@
 
 std::string port;
 
-/* Verify that constructor does not accept a nonexistent repo. */
+/* Verify that constructor does not accept a nonexistent repo.
+ *
+ * Note that this will not fail on release builds! */
```
lbonn: Good catch, but isn't that a bit bad, for example if we want to run our test suite on qemu? This could be a FIXME/TODO.

patrickvacek: Yeah, I'm actually working on fixing that right now! My current solution is to wrap it around `NDEBUG` so that it doesn't run on Release builds.

lbonn: Ok but what does it test then? It's not the actual behaviour of the class.

patrickvacek: Currently it fails in debug builds because of an assert that checks that the repo is valid. Would you prefer that instead threw an exception so it failed in all cases?

src/sota_tools/ostree_repo.cc
```diff
@@ -0,0 +1,65 @@
...
+}
+
+OSTreeObject::ptr OSTreeRepo::GetObject(const OSTreeHash hash, const OstreeObjectType type) const {
+  otable::const_iterator it;
+  it = ObjectTable.find(hash);
+  if (it != ObjectTable.end()) {
+    return it->second;
+  }
+
+  const std::string exts[] = {".filez", ".dirtree", ".dirmeta", ".commit"};
```
lbonn: `std::map exts{{OstreeObjectType::OSTREE_OBJECT_TYPE_FILE, "filez"}, ...}`?

src/sota_tools/ostree_object_test.cc
lbonn: Either that or just remove this test. If we just test that the constructor has an assert, I don't think it brings much value.

src/sota_tools/request_pool.cc
```diff
@@ -80,22 +80,35 @@ void RequestPool::LoopListen() {
...
-    select(maxfd + 1, &fdread, &fdwrite, &fdexcept, timeoutms == -1 ? nullptr : &timeout);
+    select(maxfd + 1, &fdread, &fdwrite, &fdexcept, &timeout);
   } else {
-    LOG_DEBUG << "Waiting 100ms for curl";
-    // If maxfd == -1, then wait 100ms. See:
-    // https://curl.haxx.se/libcurl/c/curl_multi_timeout.html
-    timeout.tv_sec = 0;
-    timeout.tv_usec = 100 * 1000;
+    // If maxfd == -1, then wait the lesser of timeoutms and 100ms. See:
+    // https://curl.haxx.se/libcurl/c/curl_multi_fdset.html
```
lbonn: No if?

```
long nofd_timeoutms = std::min(timeoutms, 100);
LOG_DEBUG << "Waiting " << nofd_timeoutms << " ms for curl";
timeout.tv_sec = nofd_timeoutms / 1000;
timeout.tv_usec = 1000 * (nofd_timeoutms % 1000);
```

src/sota_tools/ostree_object.cc
```diff
@@ -241,7 +247,24 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
   request_start_time_ = std::chrono::steady_clock::now();
 }
 
-void OSTreeObject::PresenceUnknown(RequestPool &pool, const int64_t rescode) {
+void OSTreeObject::CheckChildren(RequestPool &pool, const long rescode) {  // NOLINT
```
lbonn: This clang-tidy lint is a bit annoying https://clang.llvm.org/extra/clang-tidy/checks/google-runtime-int.html. Even if I agree with the intent, it is a bit strict.

Maybe we could disable it. Otherwise, maybe use the specific lint id: `// NOLINT(google-runtime-int)`

patrickvacek: Generally, I like it, it's just because of the curl API that we have the problem. You are right about specifying the lint ID, though.

PR #1022, opened 2018-12-04
Change message about no new updates to debug level.

### LINKED ISSUE #1021: Improvements to logging levels
When I look at the logs of devices I manage they are basically a never ending stream of:
~~~
2018-11-30 16:41.04 hikey aktualizr[483] : got UpdateCheckComplete event
2018-11-30 16:41.15 hikey aktualizr[483] : No new updates found in Uptane metadata.
~~~

I think at a minimum it would be a good idea to change the "no new updates found" message to be DEBUG and not INFO.

However, I'm starting to think it might make sense to make the `LOG_INFO << "got " << event->variant << " event";` a LOG_DEBUG. Perhaps its worth keeping at LOG_INFO for certain types of events != UpdateCheckComplete?

patrickvacek:
If there are updates, still print at info level. I also silenced the log message in aktualizr-primary and hmi-stub that we got the event; it was redundant.

Fixes #1021.

### DIFF
--- a/src/aktualizr_primary/main.cc
+++ b/src/aktualizr_primary/main.cc
@@ -85,7 +85,11 @@ bpo::variables_map parse_options(int argc, char *argv[]) {
 }
 
 void process_event(const std::shared_ptr<event::BaseEvent> &event) {
-  if (event->variant != "DownloadProgressReport") {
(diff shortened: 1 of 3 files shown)


doanac:
thanks!

PR #1023, opened 2018-12-05
Cache writes to sqlstorage

eu-smirnov:


src/libaktualizr/uptane/fetcher.cc
```diff
@@ -22,24 +22,18 @@ static size_t DownloadHandler(char* contents, size_t size, size_t nmemb, void* u
...
-    if (ds->downloaded_length > 0) {
-      std::cout << "\r";
-    }
-  }
   ds->downloaded_length += downloaded;
-  calculated = static_cast<unsigned int>((ds->downloaded_length * 100) / expected);
+  auto calculated = static_cast<unsigned int>((ds->downloaded_length * 100) / expected);
   if (loggerGetSeverity() <= boost::log::trivial::severity_level::trace) {
-    std::cout << "Downloading: " << calculated << "%";
+    std::cout << "\rDownloading: " << calculated << "%";
```
patrickvacek: It might be time to get rid of this `cout` business entirely. It's basically useless. If you are using trace debugging, there is way too much information to catch these, and the `\r` means if you go back looking for it, you only see `Downloading: 100%`. hmi-stub already prints out download progress; aktualizr probably should at debug or even info level in the same fashion as hmi-stub. On the other hand, I'm not sure why I didn't see the `Download status` bit that you added below.

eu-smirnov: I agree, printing the events should be enough. Forgot `\n` in `Download status`, maybe that's the reason.

eu-smirnov:
@patrickvacek Yes, I think so. I'll add it to the next PR.

PR #1024, opened 2018-12-06
Small changes for Angel Delight project

OYTIS:
For `copyCertsToCurl` it's just nice to be able to apply it to bare `HttpInterface` instances.

For `downloadAsync` it can be necessary to access curl handler from the caller. In the proxy we use it to set response code to what we've got from the server before we have returned from `curl_easy_perform`, which happens only after the whole file has been downloaded.

### DIFF
--- a/src/libaktualizr/crypto/keymanager.cc
+++ b/src/libaktualizr/crypto/keymanager.cc
@@ -189,13 +189,13 @@ std::string KeyManager::getCN() const {
   return cn;
 }
 
-void KeyManager::copyCertsToCurl(const std::shared_ptr<HttpInterface> &http) {
+void KeyManager::copyCertsToCurl(HttpInterface &http) {
   std::string pkey = getPkey();
   std::string cert = getCert();
   std::string ca = getCa();
 
   if ((pkey.size() != 0u) && (cert.size() != 0u) && (ca.size() != 0u)) {
-    http->setCerts(ca, config_.tls_ca_source, cert, config_.tls_cert_source, pkey, config_.tls_pkey_source);
+    http.setCerts(ca, config_.tls_ca_source, cert, config_.tls_cert_source, pkey, config_.tls_pkey_source);
   }
 }
 
--- a/src/libaktualizr/http/httpclient.cc
+++ b/src/libaktualizr/http/httpclient.cc
@@ -50,7 +50,6 @@ HttpClient::HttpClient() : user_agent(std::string("Aktualizr/") + AKTUALIZR_VERS
     throw std::runtime_error("Could not initialize curl");
   }
   headers = nullptr;
-  http_code = 0;
 
   curlEasySetoptWrapper(curl, CURLOPT_NOSIGNAL, 1L);
   curlEasySetoptWrapper(curl, CURLOPT_TIMEOUT, 60L);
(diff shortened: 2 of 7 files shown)


src/libaktualizr/crypto/keymanager.cc
```diff
@@ -189,13 +189,13 @@ std::string KeyManager::getCN() const {
...
 
-void KeyManager::copyCertsToCurl(const std::shared_ptr<HttpInterface> &http) {
+void KeyManager::copyCertsToCurl(HttpInterface &http) {
   std::string pkey = getPkey();
   std::string cert = getCert();
   std::string ca = getCa();
 
   if ((pkey.size() != 0u) && (cert.size() != 0u) && (ca.size() != 0u)) {
-    http->setCerts(ca, config_.tls_ca_source, cert, config_.tls_cert_source, pkey, config_.tls_pkey_source);
+    http.setCerts(ca, config_.tls_ca_source, cert, config_.tls_cert_source, pkey, config_.tls_pkey_source);
```
patrickvacek: I'm not sure I understand the benefit here of the reference instead of the shared_ptr. How is that actually more flexible?

OYTIS: In that I (the caller) can have plain HttpInterface pointer and still use this interface.

OYTIS: There seems to be no reason to use <shared_ptr> here, because `copyCertsToCurl` doesn't share or store this pointer anywhere, it just uses it and forgets about it after it returns.

lbonn: Yes, there is no dealing with ownership here, a reference is the obvious choice.

See this nice blogpost (again, sorry :)): https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

OYTIS:
What I particularly don't like here is that we add curl-specific data here. But I can't see how it can be worked around.

src/libaktualizr/crypto/keymanager.cc
patrickvacek: Ah, I see now. Yes, agreed. And the blog post is excellent, I wish that type of logic was more generally observed and enforced. (I mean in general; I think we mostly do it right.)

src/libaktualizr/http/httpclient.cc
```diff
@@ -226,11 +236,16 @@ HttpResponse HttpClient::download(const std::string& url, curl_write_callback ca
...
+  auto promise = std::make_shared<std::promise<HttpResponse>>();
+  std::thread([curlp, promise]() {
+    CURLcode result = curl_easy_perform(curlp.get());
+    long http_code;
+    curl_easy_getinfo(curlp.get(), CURLINFO_RESPONSE_CODE, &http_code);
+    HttpResponse response("", http_code, result, (result != CURLE_OK) ? curl_easy_strerror(result) : "");
+    promise->set_value(response);
+  })
+      .detach();
+  return promise->get_future();
```
eu-smirnov: It's more like a style preference, but wouldn't it be cleaner to get future first and then move the promise?

OYTIS: Hm... I don't know, it's copied not moved after all. Does it look confusing?

eu-smirnov: Yes, a bit. The whole point of creating a shared pointer is only to do promise->get_future() after the lambda statement.

OYTIS: Got it, sorry. Fixed.

PR #1025, opened 2018-12-06
Fix races in aktualizr with running modes

lbonn:
We did not wait for the result of the futures after issuing commands like "CheckUpdates".

This broke the existing behaviour of `aktualizr check`.

Bug introduced in 79283a3c77a584d9a7f5e48f83297c4201e394fe


lbonn:
@patrickvacek yes, codecov sees clear through our game: `codecov/patch — 0% of diff hit (target 82.35%)`

PR #1026, opened 2018-12-07
Improve locking on the SQLite database

OYTIS:
- one mutex per path instead of a global mutex
- automatically lock when two threads try to acquite database for
writing

### DIFF
--- a/src/libaktualizr/storage/CMakeLists.txt
+++ b/src/libaktualizr/storage/CMakeLists.txt
@@ -4,7 +4,7 @@ add_custom_command(OUTPUT sql_schemas.cc sql_schemas_target
 )
 
 if(STORAGE_TYPE STREQUAL "sqlite")
-  set(SOURCES sqlstorage.cc)
+  set(SOURCES sqlstorage.cc sql_utils.cc)
   set(HEADERS sqlstorage.h sql_utils.h)
 elseif(STORAGE_TYPE STREQUAL "android")
   set(SOURCES androidstorage.cc)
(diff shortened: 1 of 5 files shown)


OYTIS:
This contraption seems to work, all style/performance improvements are welcome.

OYTIS:
Wow, I've really made a deadlock. Will take a look on Monday, should be pretty trivial.

lbonn:
@eu-smirnov It's used in aktualizr-info: https://github.com/advancedtelematic/aktualizr/blob/328417faa25bd9bf8a7365182e020d2e73d645bd/src/aktualizr_info/main.cc#L44

We used to have this conflict between the two tools trying to fight for the database during tests.

OYTIS:
No, connection granularity is too coarse for our applications, need to implement locking on per-transaction basis.

OYTIS:
Good news: LGTM works. The new solution, that is locking on transaction, seems to work as well.
Bad news: now we have to start a transaction whenever we want to write to the database. This makes code in `sqlstorage.cc` more verbose and can also have performance implications, although I doubt they will be significant.

eu-smirnov:
> This makes code in `sqlstorage.cc` more verbose and can also have performance implications, although I doubt they will be significant.
I'm not so sure about that, tbh. I think we might want to do some benchmarks.

OYTIS:
@eu-smirnov Would running `aktualizr_test` with and without this PR several times give a good metric in your opinion?

OYTIS:
Abandoned, will get rid of parallel downloads instead.

PR #1027, opened 2018-12-07
Fix/more with installed versions

lbonn:
Part of the bigger installation result story.

I plan to move most of manifest generation out of the SotaUptaneClient class, as it can mostly be done from storage right now.


patrickvacek:
Looks promising!

PR #1030, opened 2018-12-10
Document how to compile the SOTA tools on the Mac

jochenschneider:


jochenschneider:
It doesn't find `libintl`. Which is also not really needed. I tried to find a nicer way, but couldn't come up with one and then saw this: https://github.com/EOSIO/eos/issues/2028. I can live with it since it is mostly for developers.

lbonn:
Thanks!

PR #1031, opened 2018-12-11
Serialize downloads

OYTIS:


patrickvacek:
Before we merge this, we should double-check that no one is relying on something that we will break. I'll look into that. Conceptually, though, I have been convinced that this is the right way forward.

OYTIS:
@patrickvacek What do you mean? Right now we only have sachertorte project as an external user, and I'm pretty sure it's fine.

patrickvacek:
@OYTIS I want to make sure that no one else was expecting this mis-feature. :)

Also, could you try adding the feature to implicitly pause when calling shutdown via the API?

patrickvacek:
Do you think it is reasonable to actually measure the download times with and without this change? I believe this should be an improvement, and I also know it will make things easier to maintain, but it's worth knowing what effects this has in practice.

eu-smirnov:
@patrickvacek Yes, makes sense to measure. We know the current speed already, it's about 64 kbit/s. When this PR is merged I'll create another one with updated fetcher test, which measures the speed.

lbonn:
Some tests are failing in the debian package manager case (which is pretty obsolete, but still...)

OYTIS:
@lbonn  Yes, it uses WHandle directly, fixed that.

OYTIS:
@eu-smirnov Thank you, I was wondering how this should be tested.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -422,11 +422,15 @@ void process_events_FullMultipleSecondaries(const std::shared_ptr<event::BaseEve
...
     manifest_FullMultipleSecondaries = true;
+  } else if (event->variant == "DownloadPaused") {
+    // phony event on shutdown, ignore
+    return;
   }
   // It is possible for the PutManifestComplete to come before we get the
   // InstallTargetComplete depending on the threading, so check for both.
   if (allcomplete_FullMultipleSecondaries && complete_FullMultipleSecondaries == 2 &&
       manifest_FullMultipleSecondaries) {
+    LOG_ERROR << "SETVALUE";
```
patrickvacek: Is this leftover from testing or intentional?

OYTIS: @patrickvacek That's  debug output, thanks, I'll remove it.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1127,23 +1122,20 @@ class SQLTargetWHandle : public StorageTargetWHandle {
     }
 
     row_id_ = sqlite3_last_insert_rowid(db_.get());
-    db_.commitTransaction();
   }
 
   ~SQLTargetWHandle() override {
     if (!closed_) {
       LOG_WARNING << "Handle for file " << target_.filename() << " has not been committed or aborted, forcing abort";
+      db_.commitTransaction();
```
eu-smirnov: Didn't notice it before, but why do we call commit here? We want to abort the transaction, right?

OYTIS: Not quite sure actually. My idea was that if we don't abort explicitly then we probably want to save what we've already downloaded.
By the way, maybe we don't need explicit `pause` from `Aktualizr::Shutdown` in the end, @patrickvacek's commit confused me somewhat. Just need to add the test that the file really gets committed.

patrickvacek: @OYTIS good call, actually! I overlooked the commit in the destructor, which is actually an even better solution than the implicit call to pause.

eu-smirnov:
I locally rebased my changes to fetcher/fetcher test to measure the download speed. It certainly got much better, up to 4 Mbit/s from previous 64 kbit/s.
For some reason, speed drops significantly after pause/resume cycle. I'll look into that.

OYTIS:
@eu-smirnov Could you please share the rebased code on some branch? I might try to debug that as well.

eu-smirnov:
@OYTIS Sure, pushed it to `fix/fetcher-test`.

OYTIS:
@eu-smirnov Thanks!

OYTIS:
OK, as we've found, transactions around `sqlite_blob_*` API calls are not needed at all. As @eu-smirnov suggested, we'll rewrite it to keep the blob open over `wfeed` operations, but the database schema needs to be reworked for that.

This PR already improves the speed significantly, but we can do better by keeping the blob open.

patrickvacek:
This looks fine to me, but I'd prefer someone with better SQL knowledge to sign off if possible.

> I want to make sure that no one else was expecting this mis-feature.

It doesn't sound like anyone will be upset if we remove it, especially since the numbers clearly demonstrate that this is a major improvement. (Or more accurately, this retracts what we now know to have been a downgrade, and it appears to moderately improve upon what we had before that.)

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1158,28 +1152,27 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
+    if (blob_ != nullptr) {
+      sqlite3_blob_close(blob_);
+      blob_ = nullptr;
+      auto statement = db_.prepareStatement<int64_t, int64_t>("update target_images SET real_size = ? where rowid = ?;",
+                                                              static_cast<int64_t>(written_size_), row_id_);
+
+      int err = statement.step();
+      if (err != SQLITE_DONE) {
+        LOG_ERROR << "Could not save size in db: " << db_.errmsg();
+        throw StorageTargetWHandle::WriteError("could not update sise of " + target_.filename() + " in sql storage");
```
lbonn: typo: "size"

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1158,28 +1152,27 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
-    wcommit();
-    db_.commitTransaction();
     return size;
   }
 
   void wcommit() override {
+    if (blob_ != nullptr) {
+      sqlite3_blob_close(blob_);
+      blob_ = nullptr;
+      auto statement = db_.prepareStatement<int64_t, int64_t>("update target_images SET real_size = ? where rowid = ?;",
```
lbonn: really small, but we upper case sql operations everywhere else, would be nice to keep a consistent style for easy grepping etc

lbonn:
Looks good. I don't know if I really have more knowledge of sql but it seems to be consistent with the approached we agreed with.

eu-smirnov:
Awesome, now it takes 7 seconds to download a 100 MB file, which is around 120 Mbit/s or ~2000 times faster. Let's merge it, guys :)

patrickvacek:
>  For some reason, speed drops significantly after pause/resume cycle. I'll look into that.

@eu-smirnov Is this still happening? My guess is no, but just wanted to be sure...

eu-smirnov:
@patrickvacek nope, that one is also seems to be fixed

src/libaktualizr/storage/sqlstorage.cc
OYTIS: Thanks.

src/libaktualizr/storage/sqlstorage.cc
OYTIS: Sure, will fix.

PR #1032, opened 2018-12-11
Fixes for fetcher

eu-smirnov:
There are several problems with pause-resume fetcher API and the corresponding fetcher test that I see at the moment.
1. Fetcher test currently uses `large_interrupted` file for pause/resume. With `large_interrupted` server aborts after transferring half of a file (1000 bytes). This should lead to the `The target's calculated hash did not match the hash in the metadata` error and, after that, the upper layer may choose to retry. However, currently, the test succeeds, because while committing the first half to the db, fetcher gets `pause` event, which makes it retry silently after `resume` is called. This behavior itself is not a problem, the problem is that it is also correct for the test to fail, if `pause` comes after the first chunk is committed.
2. Race condition between `fetchVerifyTarget()` or `DownloadHandler()` and `setPause()`:
   ```
   http->download()
                                               setPause(true) 
   DownloadHandler() {
   ...
   return written_size + 1
                                               setPause(false)
   http->download() == CURLE_WRITE_ERROR && 
   pause_ == false
   throw OversizedTarget
   ```
3. Using a mutex to pause the fetcher is not only non-idiomatic, but it is also an undefined behavior, as the thread which called `pause` may terminate and/or resume might get called from another thread. See [https://en.cppreference.com/w/cpp/thread/mutex] and [https://en.cppreference.com/w/cpp/thread/mutex/unlock]. It seems to work now, but it can easily break anytime.
4. The `downloading_` guard in fetcher is pretty useless IMO. If I understand correctly its main purpose is to return `kNotDownloading`, if the `Download` API wasn't called yet. But the `downloading_` flag
   doesn't guarantee anything. The high-level `Download` API call is asynchronous and the execution of underlying `Fetcher::fetchVerifyTarget` will be performed at some point in the future on the command queue. `Pause`, on the other hand, is synchronous, so it's entirely possible that caller gets a `kNotDownloading` even after he has called `Download`. That means that caller should track the downloading state himself and provide some retry logic for the aforementioned case.

   So why bother with the `kNotDownloading` at all?
   Cleaner approach IMO would be to set `pause` no matter if the `Download` was called or not. In case `Download` called after a `Pause` it should just block until the `Resume` is called.

This PR attempts to solve issues 1-3, issue 4 is out of scope currently, I just want to start a discussion on it.

### DIFF
--- a/src/libaktualizr/http/httpclient.cc
+++ b/src/libaktualizr/http/httpclient.cc
@@ -213,12 +213,12 @@ HttpResponse HttpClient::perform(CURL* curl_handler, int retry_times, int64_t si
   return response;
 }
 
-HttpResponse HttpClient::download(const std::string& url, curl_write_callback callback, void* userp, size_t from) {
+HttpResponse HttpClient::download(const std::string& url, curl_write_callback callback, void* userp, curl_off_t from) {
   return downloadAsync(url, callback, userp, from, nullptr).get();
 }
 
 std::future<HttpResponse> HttpClient::downloadAsync(const std::string& url, curl_write_callback callback, void* userp,
-                                                    size_t from, CurlHandler* easyp) {
+                                                    curl_off_t from, CurlHandler* easyp) {
   CURL* curl_download = Utils::curlDupHandleWrapper(curl, pkcs11_key);
   CurlHandler curlp = CurlHandler(curl_download, curl_easy_cleanup);
 
@@ -234,7 +234,7 @@ std::future<HttpResponse> HttpClient::downloadAsync(const std::string& url, curl
   curlEasySetoptWrapper(curl_download, CURLOPT_TIMEOUT, 0);
   curlEasySetoptWrapper(curl_download, CURLOPT_LOW_SPEED_TIME, speed_limit_time_interval_);
   curlEasySetoptWrapper(curl_download, CURLOPT_LOW_SPEED_LIMIT, speed_limit_bytes_per_sec_);
-  curlEasySetoptWrapper(curl_download, CURLOPT_RESUME_FROM, from);
+  curlEasySetoptWrapper(curl_download, CURLOPT_RESUME_FROM_LARGE, from);
 
   std::promise<HttpResponse> resp_promise;
   auto resp_future = resp_promise.get_future();
--- a/src/libaktualizr/http/httpinterface.h
+++ b/src/libaktualizr/http/httpinterface.h
@@ -41,9 +41,9 @@ class HttpInterface {
   virtual HttpResponse post(const std::string &url, const Json::Value &data) = 0;
   virtual HttpResponse put(const std::string &url, const Json::Value &data) = 0;
 
-  virtual HttpResponse download(const std::string &url, curl_write_callback callback, void *userp, size_t from) = 0;
+  virtual HttpResponse download(const std::string &url, curl_write_callback callback, void *userp, curl_off_t from) = 0;
   virtual std::future<HttpResponse> downloadAsync(const std::string &url, curl_write_callback callback, void *userp,
-                                                  size_t from, CurlHandler *easyp) = 0;
+                                                  curl_off_t from, CurlHandler *easyp) = 0;
   virtual void setCerts(const std::string &ca, CryptoSource ca_source, const std::string &cert,
                         CryptoSource cert_source, const std::string &pkey, CryptoSource pkey_source) = 0;
   static constexpr int64_t kNoLimit = 0;  // no limit the size of downloaded data
--- a/src/libaktualizr/http/httpinterface.h
+++ b/src/libaktualizr/http/httpinterface.h
@@ -14,6 +14,7 @@
 using CurlHandler = std::shared_ptr<CURL>;
 
 struct HttpResponse {
+  HttpResponse() = default;
   HttpResponse(std::string body_in, const long http_status_code_in, CURLcode curl_code_in,  // NOLINT
                std::string error_message_in)
       : body(std::move(body_in)),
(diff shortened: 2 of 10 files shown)


src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -18,13 +18,7 @@
...
 static void aktualizr_progress_cb(OstreeAsyncProgress *progress, gpointer data) {
   auto *mt = static_cast<PullMetaStruct *>(data);
-  bool has_been_paused = false;
-  if (!mt->pause_mutex->try_lock()) {  // If can't aquire the mutex thats mean
-    mt->pause_mutex->lock();           // mutex is locked by pause thread, and
-    has_been_paused = true;            // we will wait, until it is locked.
-  } else {
-    mt->pause_mutex->unlock();  // Unlock locked by try_lock() mutex.
-  }
+  mt->pause();
```
lbonn: Can we add a comment here, to document that we expect to wait for the download to not be paused anymore?

This callback is passed through several method calls and constructors and the intent is a bit obfuscated here, IMO.

eu-smirnov: Agree, also the name is a bit confusing, as it will not pause on every call, but I wasn't able to come up with something better. I'll add a comment.

src/libaktualizr/http/httpinterface.h
```diff
@@ -14,15 +14,16 @@
...
   HttpResponse(std::string body_in, const long http_status_code_in, CURLcode curl_code_in,  // NOLINT
                std::string error_message_in)
       : body(std::move(body_in)),
         http_status_code(http_status_code_in),
         curl_code(curl_code_in),
         error_message(std::move(error_message_in)) {}
   std::string body;
-  long http_status_code;  // NOLINT
-  CURLcode curl_code;
+  int64_t http_status_code{0};
```
OYTIS: We were using `long`, because it is the type in the signature of `curl_easy_getinfo`.

eu-smirnov: OK, makes sense to change it back then.

src/libaktualizr/package_manager/ostreemanager.cc
patrickvacek: +1 for a comment, but even just `checkPause` would be a better name.

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -119,23 +109,19 @@ bool Fetcher::fetchVerifyTarget(const Target& target) {
...
+        response = http->download(config.uptane.repo_server + "/targets/" + Utils::urlEncode(target.filename()),
+                                  DownloadHandler, &ds, ds.downloaded_length);
+      } while (retry_);
+      if (!response.isOk()) {
+        if (response.curl_code == CURLE_WRITE_ERROR) {
+          throw OversizedTarget(target.filename());
         }
-      } while (retry);
+        throw Exception("image", "Could not download file, error: " + response.error_message);
+      }
```
patrickvacek: This cleanup is definitely an improvement, but I think we are losing our retry functionality. We currently retry downloads at a high level if they fail for basically any reason. I think that's worth keeping. Originally, I think we gave it three tries before failing.

src/libaktualizr/uptane/fetcher_test.cc
```diff
@@ -11,13 +11,32 @@
 #include "test_utils.h"
 #include "tuf.h"
 
+static const int pause_after = 10;        // percent
+static const int pause_duration = 10;     // seconds
+static const int download_timeout = 200;  // seconds.
```
patrickvacek: This looks like a major improvement to this test. But do we really have to wait 10 seconds? Maybe we do, but it seems like a lot of time to spend just waiting.

src/libaktualizr/uptane/fetcher.cc
eu-smirnov: Thanks! I forgot to look into sotauptaneclient.cc. The retry indeed will not work. But does it work now? The code on master throws in exactly the same way.

src/libaktualizr/uptane/fetcher_test.cc
eu-smirnov: True, that's a debug leftover. I'll reduce it to 1 sec.

src/libaktualizr/uptane/fetcher.cc
patrickvacek: My bad, the actual retry logic is in `SotaUptaneClient::downloadImage`. All is well, sorry!

src/libaktualizr/package_manager/ostreemanager.cc
eu-smirnov: True, I'll rename it.

lbonn: Sorry, a bit late again. But it probably wouldn't hurt to have a null check as well before making the call.

eu-smirnov: @lbonn I don't know, std:function will throw in case it's empty. Do we want to proceed if pause callback isn't provided?

lbonn: Just having:

```
if (checkPause != nullptr) {
  checkPause()
}
```

Then, users of the class could just pass a null pointer if they don't need the pause functionality. It's only an added functionality, not a part of the core duty of the class.

src/libaktualizr/http/httpclient.cc
```diff
@@ -234,7 +234,7 @@ std::future<HttpResponse> HttpClient::downloadAsync(const std::string& url, curl
   curlEasySetoptWrapper(curl_download, CURLOPT_TIMEOUT, 0);
   curlEasySetoptWrapper(curl_download, CURLOPT_LOW_SPEED_TIME, speed_limit_time_interval_);
   curlEasySetoptWrapper(curl_download, CURLOPT_LOW_SPEED_LIMIT, speed_limit_bytes_per_sec_);
-  curlEasySetoptWrapper(curl_download, CURLOPT_RESUME_FROM, from);
+  curlEasySetoptWrapper(curl_download, CURLOPT_RESUME_FROM_LARGE, from);
```
OYTIS: Would it make sense then to change type of `from` to `curl_off_t`?

src/libaktualizr/package_manager/ostreemanager.cc
eu-smirnov: Agreed, that would make sense.

src/libaktualizr/http/httpclient.cc
eu-smirnov: Yes, absolutely. Thanks!

PR #1033, opened 2018-12-12
Refactor/events api

lbonn:
Trying to move stuff at their right place again. Also planning to put the data types in `results.h` inside a namespace, some of them have really confusing names.

### DIFF
--- a/src/libaktualizr/uptane/fetcher.cc
+++ b/src/libaktualizr/uptane/fetcher.cc
@@ -34,10 +34,9 @@ static size_t DownloadHandler(char* contents, size_t size, size_t nmemb, void* u
 
   ds->downloaded_length += downloaded;
   auto progress = static_cast<unsigned int>((ds->downloaded_length * 100) / expected);
-  if (ds->events_channel && progress > ds->last_progress) {
+  if (ds->progress_cb && progress > ds->last_progress) {
     ds->last_progress = progress;
-    auto event = std::make_shared<event::DownloadProgressReport>(ds->target, "Downloading", progress);
-    (*(ds->events_channel))(event);
+    ds->progress_cb(ds->target, "Downloading", progress);
   }
   if (ds->fetcher->isPaused()) {
     ds->fetcher->setRetry(true);
@@ -53,18 +52,15 @@ PauseResult Fetcher::setPause(bool pause) {
   if (pause_ == pause) {
     if (pause) {
       LOG_INFO << "Download is already paused.";
-      sendEvent<event::DownloadPaused>(PauseResult::kAlreadyPaused);
       return PauseResult::kAlreadyPaused;
     } else {
       LOG_INFO << "Download is not paused, can't resume.";
-      sendEvent<event::DownloadResumed>(PauseResult::kNotPaused);
       return PauseResult::kNotPaused;
     }
   }
 
   if (pause && downloading_ == 0u) {
     LOG_INFO << "No download in progress, can't pause.";
-    sendEvent<event::DownloadPaused>(PauseResult::kNotDownloading);
     return PauseResult::kNotDownloading;
   }
 
@@ -72,10 +68,8 @@ PauseResult Fetcher::setPause(bool pause) {
   cv_.notify_all();
 
   if (pause) {
-    sendEvent<event::DownloadPaused>(PauseResult::kPaused);
     return PauseResult::kPaused;
   } else {
-    sendEvent<event::DownloadResumed>(PauseResult::kResumed);
     return PauseResult::kResumed;
   }
 }
@@ -98,7 +92,7 @@ bool Fetcher::fetchVerifyTarget(const Target& target) {
         LOG_INFO << "Image already downloaded skipping download";
         return true;
       }
-      DownloadMetaStruct ds(target, events_channel);
+      DownloadMetaStruct ds(target, progress_cb);
       ds.fetcher = this;
       if (!target_exists) {
         ds.fhandle = storage->allocateTargetFile(false, target);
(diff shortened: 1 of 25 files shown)


lbonn:
Hm sorry, putting back the not-ready flag because I also wanted to namespace some of the data structures.

src/libaktualizr/uptane/fetcher_test.cc
```diff
@@ -15,55 +16,6 @@ static std::string server = "http://127.0.0.1:";
...
-      EXPECT_EQ(resume_event->result, PauseResult::kNotPaused);
-      break;
-    }
-    default:
-      std::cout << "event #" << num_events_DownloadPause << " is: " << event->variant << "\n";
-      EXPECT_EQ(event->variant, "");
-  };
-  ++num_events_DownloadPause;
-}
-
```
patrickvacek: I understand why this is removed, but then we don't test that those events are actually created. I note that the events aren't actually mentioned in `actions.md`, and perhaps it isn't actually important, but is there a way we can still test these events easily?

src/libaktualizr/uptane/managedsecondary.cc
```diff
@@ -1,11 +1,14 @@
-#include "uptane/managedsecondary.h"
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
 
 #include <boost/algorithm/hex.hpp>
 #include <boost/filesystem.hpp>
 
+#include "uptane/managedsecondary.h"
+
```
patrickvacek: Normally I think we prefer a `.cc`'s `.h` file to be the first header file listed. I think that's what the Google Style Guide recommends, too.

src/libaktualizr/uptane/fetcher_test.cc
lbonn: They'd have to be tested on `Aktualizr`'s side which might be hard...

Maybe we don't have to do it for every logic case, like it's done here, because the translation from pause/resume returns to an event is actually straightforward there.

patrickvacek: Yeah, and to be honest, emitting events for pause/resume was only done to be consistent. They aren't very useful since they just duplicate the function return values, and the function should return quickly.

lbonn: Ah indeed... But it makes some kind of sense if you want to observe (log) the systems through all these callbacks. I can try to come up with a quick test and if it's too convoluted, we could leave it like that?

src/libaktualizr/uptane/fetcher.h
```diff
@@ -6,8 +6,24 @@
...
+  kResumed,
+  /* Download was already paused, so there is nothing to do. */
+  kAlreadyPaused,
+  /* Download has already completed, so there is nothing to do. */
+  kAlreadyComplete,
+  /* No download is in progress, so there is nothing to do. */
+  kNotDownloading,
+  /* Download was not paused, so there is nothing to do. */
+  kNotPaused,
+};
```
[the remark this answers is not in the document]
patrickvacek: > Or we could rename the enum here and have a `using PauseResult = FetcherPauseRet` in aktualizr.h?

I don't think that actually helps if the definition is still somewhere else.

Also, shouldn't aktualizr.h include fetcher.h? That would at least make it a bit clearer where the definition _might_ come from.

src/libaktualizr/uptane/fetcher_test.cc
patrickvacek: If you've got an idea, go for it! (And if it does result in new test coverage, might as well add the items to actions.md.)

src/libaktualizr/uptane/fetcher.h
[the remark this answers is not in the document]
lbonn: Ah yes that's true. I've looked again and the plan would be to have `#include "fetcher.h"` in results.h + the using directive.

The api is not sitting in one file anyway, we already have `aktualizr.h`, `results.h` so for now it doesn't seem too ugly to me. Maybe we'll want to isolate completely aktualizr.h from the rest of the code-base at some point, but that would be a bigger effort.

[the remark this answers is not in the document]
patrickvacek: > I've looked again and the plan would be to have `#include "fetcher.h"` in results.h + the using directive.

That would be fine, but I think how it is now in your PR is also okay.

> The api is not sitting in one file anyway

Yes, it's true. I've tried to limit the number of files an external developer has to dig through, but at least doxygen is smart enough to pick up on where these things are defined, so it shouldn't really be a big deal in the end.

I think you can leave this issue as it is for now and we can revisit if we have good reason to some other time.

lbonn:
Ok, I've tried to take the comments into account. Left to do:

- namespacing
- (maybe) test Aktualizr class' pause/resume events

lbonn:
Ok should be good now!

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -139,9 +133,9 @@ bool Fetcher::fetchVerifyTarget(const Target& target) {
 #ifdef BUILD_OSTREE
       KeyManager keys(storage, config.keymanagerConfig());
       keys.loadKeys();
-      std::function<void()> pause = std::bind(&Fetcher::checkPause, this);
-      data::InstallOutcome outcome =
-          OstreeManager::pull(config.pacman.sysroot, config.pacman.ostree_server, keys, target, pause, events_channel);
+      std::function<void()> check_pause = std::bind(&Fetcher::checkPause, this);
+      data::InstallOutcome outcome = OstreeManager::pull(config.pacman.sysroot, config.pacman.ostree_server, keys,
```
eu-smirnov: Was it compiling here after you removed `pause`? Didn't know that it's ok to pass `int (*)(void)` instead of `std::function<void()>`.

lbonn: Yes that's here, I didn't know either...

Would have liked to find the relevant rule in the standard but I failed to do so.

PR #1034, opened 2018-12-14
Android support

riqkum:
* build script update required by atkualizr-android
* package manager for android
* log system integration

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,15 @@ set(BOOST_COMPONENTS log_setup log filesystem program_options)
 set(Boost_USE_STATIC_LIBS OFF)
 add_definitions(-DBOOST_LOG_DYN_LINK)
 
+if(ANDROID)
+    set(BOOST_ROOT ${PREBUILD_BOOST_ROOT})
+    set(Boost_INCLUDE_DIR ${BOOST_ROOT}/include)
+    set(Boost_LIBRARY_DIR ${BOOST_ROOT}/libs/llvm/${ANDROID_ABI})
+    if(Boost_FOUND)
+        include_directories(${Boost_INCLUDE_DIRS})
+    endif()
+endif(ANDROID)
+
 if(BUILD_OPCUA)
     list(APPEND BOOST_COMPONENTS serialization iostreams)
 endif(BUILD_OPCUA)
@@ -87,6 +96,7 @@ find_package(sodium REQUIRED)
 find_package(SQLite3 REQUIRED)
 find_package(Git)
 find_package(Asn1c REQUIRED)
+find_package(ZLIB)
 
 if(NOT AKTUALIZR_VERSION)
     if(GIT_EXECUTABLE)
@@ -297,6 +307,7 @@ include_directories(${SQLITE3_INCLUDE_DIRS})
 include_directories(${LIBP11_INCLUDE_DIR})
 include_directories(${sodium_INCLUDE_DIR})
 include_directories(${OPENSSL_INCLUDE_DIR})
+include_directories(${CURL_INCLUDE_DIR})
 include_directories(${LibArchive_INCLUDE_DIR})
 
 set_source_files_properties(third_party/jsoncpp/jsoncpp.cpp PROPERTIES COMPILE_FLAGS -w)
@@ -373,7 +384,12 @@ set (AKTUALIZR_EXTERNAL_LIBS
     ${LIBP11_LIBRARIES}
     ${LIBDPKG_LIBRARIES}
     ${GLIB2_LIBRARIES}
-    ${SYSTEMD_LIBRARY})
+    ${SYSTEMD_LIBRARY}
+    ${ZLIB_LIBRARY})
+
+if(ANDROID)
+    list(APPEND AKTUALIZR_EXTERNAL_LIBS liblog.so)
+endif()
 
 get_directory_property(hasParent PARENT_DIRECTORY)
 if(hasParent)
@@ -387,7 +403,6 @@ if(BUILD_WITH_CODE_COVERAGE)
     set(COVERAGE_COMPILER_FLAGS "--coverage -fprofile-arcs -ftest-coverage" CACHE INTERNAL "")
     list(APPEND TEST_LIBS gcov)
 endif(BUILD_WITH_CODE_COVERAGE)
-
 
 include(CTest)
 ENABLE_TESTING()
--- a/src/libaktualizr/config/config.cc
+++ b/src/libaktualizr/config/config.cc
@@ -141,9 +141,7 @@ KeyManagerConfig Config::keymanagerConfig() const {
 void Config::postUpdateValues() {
   logger_set_threshold(logger);
 
-  if (provision.provision_path.empty()) {
-    provision.mode = ProvisionMode::kImplicit;
-  }
+  provision.mode = provision.provision_path.empty() ? ProvisionMode::kImplicit : ProvisionMode::kAutomatic;
 
   if (tls.server.empty()) {
     if (!tls.server_url_path.empty()) {
@@ -255,7 +253,7 @@ void Config::readSecondaryConfigs(const boost::filesystem::path& sconfigs_dir) {
     LOG_ERROR << "Could not read secondary configs from " << sconfigs_dir << ": not a directory";
     return;
   }
-  for (const auto& config_file : Utils::glob((sconfigs_dir / "*.json").string())) {
+  for (const auto& config_file : Utils::getDirEntriesByExt(sconfigs_dir, ".json")) {
     LOG_INFO << "Parsing secondary config: " << config_file;
     uptane.secondary_configs.emplace_back(config_file);
   }
--- a/src/libaktualizr/package_manager/androidmanager.cc
+++ b/src/libaktualizr/package_manager/androidmanager.cc
@@ -1,0 +1,105 @@
+#include <regex>
+
+#include "androidmanager.h"
+
+#include "utilities/utils.h"
+
+Json::Value AndroidManager::getInstalledPackages() const {
+  Json::Value packages(Json::arrayValue);
+  Json::Value package;
+  package["name"] = "aktualizr";
+  package["version"] = "n/a";
+  packages.append(package);
+  return packages;
+}
+
+Uptane::Target AndroidManager::getCurrent() const {
+  std::vector<Uptane::Target> installed_versions;
+  std::string current_hash;
+  size_t current_k = SIZE_MAX;
+  storage_->loadPrimaryInstalledVersions(&installed_versions, &current_k, nullptr);
+  if (current_k != SIZE_MAX) {
+    current_hash = installed_versions[current_k].sha256Hash();
+  }
+  AndroidInstallationState installation_state;
+  if (AndroidInstallationDispatcher::GetState().state_ == AndroidInstallationState::STATE_INSTALLED) {
+    installation_state = AndroidInstallationDispatcher::Dispatch();
+    if (installation_state.state_ == AndroidInstallationState::STATE_UPDATE) {
+      current_hash = installation_state.payload_;
+    }
+  }
+  std::vector<Uptane::Target>::iterator it;
+  for (it = installed_versions.begin(); it != installed_versions.end(); it++) {
+    if (it->sha256Hash() == current_hash) {
+      if (installation_state.state_ == AndroidInstallationState::STATE_UPDATE) {
+        storage_->savePrimaryInstalledVersion(*it, InstalledVersionUpdateMode::kCurrent);
+        auto outcome = data::InstallOutcome(data::UpdateResultCode::kOk, "package successfully intalled");
+        storage_->storeInstallationResult(data::OperationResult(it->filename(), outcome));
+      }
+      return *it;
+    }
+  }
+  return getUnknown();
+}
+
+data::InstallOutcome AndroidManager::install(const Uptane::Target& target) const {
+  if (AndroidInstallationDispatcher::GetState().state_ == AndroidInstallationState::STATE_NOP) {
+    LOG_INFO << "Begin Android package installation";
+    storage_->savePrimaryInstalledVersion(target, InstalledVersionUpdateMode::kPending);
+    auto package_filename = (Utils::getStorageRootPath() / target.filename()).string() + "." + target.sha256Hash();
+    std::ofstream package_file(package_filename.c_str());
+    if (!package_file.good()) {
+      throw std::runtime_error(std::string("Error opening file ") + package_filename);
+    }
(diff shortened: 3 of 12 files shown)


src/libaktualizr/logging/boost_logging.cc
```diff
@@ -1,11 +1,36 @@
...
 #include <boost/log/utility/setup/console.hpp>
+#pragma GCC diagnostic pop
+#else
+#include <boost/log/utility/setup/console.hpp>
+#endif
 
 namespace logging = boost::log;
 using boost::log::trivial::severity_level;
 
+#if defined(ANDROID)
```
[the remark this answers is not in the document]
riqkum: It is possible, but aktualizr's logging implementation is need to be changed first. While I just add another boost/log sink for android, splitting to a separate source file will lead to undesired code duplication.

patriotyk:
Please update your commit using following command: 'git commit --amend -s', then push with force flag.

src/libaktualizr/config/config.cc
```diff
@@ -258,10 +256,21 @@ void Config::readSecondaryConfigs(const boost::filesystem::path& sconfigs_dir) {
     LOG_ERROR << "Could not read secondary configs from " << sconfigs_dir << ": not a directory";
     return;
   }
+#if (!defined(ANDROID) || __ANDROID_API__ >= 28)
   for (const auto& config_file : Utils::glob((sconfigs_dir / "*.json").string())) {
```
OYTIS: Can we use the more generic `directory_iterator`-based implementation for both cases and get rid of external compilation here?

src/libaktualizr/utilities/config_utils.h
```diff
@@ -152,9 +152,19 @@ class BaseConfig {
         continue;
       }
       if (boost::filesystem::is_directory(config)) {
+#if (!defined(ANDROID) || __ANDROID_API__ >= 28)
         for (const auto& config_file : Utils::glob((config / "*.toml").string())) {
```
OYTIS: Same here, let's just use generic version in both cases.

src/libaktualizr/utilities/utils.cc
```diff
@@ -642,6 +643,7 @@ boost::filesystem::path Utils::absolutePath(const boost::filesystem::path &root,
   return (root / file);
 }
 
+#if (!defined(ANDROID) || __ANDROID_API__ >= 28)
 std::vector<boost::filesystem::path> Utils::glob(const std::string &pat) {
```
OYTIS: `Utils::glob` can be deleted or replaced with cross-platform implementation.

src/libaktualizr/utilities/utils.cc
```diff
@@ -731,6 +741,21 @@ class SafeTempRoot {
   boost::filesystem::path path;
 };
 
+#if defined(ANDROID)
+std::mutex Utils::storage_root_path_mutex_;
+std::string Utils::storage_root_path_;
+
+void Utils::setStorageRootPath(const std::string &storage_root_path) {
```
OYTIS: Where/how is it used?

src/libaktualizr/logging/boost_logging.cc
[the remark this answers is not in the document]
OYTIS: I'd also prefer to get rid of ifdefs wherever possible, so let's restructure logging if that helps.

CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
     message(FATAL_ERROR "Aktualizr does not support building in the source tree. Please remove CMakeCache.txt and the CMakeFiles/ directory, then create a subdirectory to build in: mkdir build; cd build; cmake ..")
 endif()
 
-set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake-modules)
+set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake-modules;${PROJECT_SOURCE_DIR}/aktualizr/cmake-modules")
```
patrickvacek: Should this extra bit get set in aktualizr-android? `${PROJECT_SOURCE_DIR}/aktualizr` doesn't exist in this repo, so it seems quite odd.

src/libaktualizr/package_manager/androidmanager.cc
```diff
@@ -0,0 +1,102 @@
...
+    if (boost::filesystem::is_directory(*entryIt)) {
+      continue;
+    }
+    auto package_file_path = entry_path.string();
+    auto ext = entry_path.extension().string();
+    if (ext == ".installed") {
+      return {AndroidInstallationState::STATE_INSTALLED, package_file_path};
+    } else if (ext == ".inprogress") {
+      return {AndroidInstallationState::STATE_IN_PROGRESS, package_file_path};
+    } else if (true /* check ext is actually a hash */) {
```
patrickvacek: Is this still a TODO? What's the challenge there?

src/libaktualizr/package_manager/androidmanager.h
```diff
@@ -0,0 +1,43 @@
...
+  Json::Value getInstalledPackages() const override;
+
+  Uptane::Target getCurrent() const override;
+  bool imageUpdated() override { return true; };
+
+  data::InstallOutcome install(const Uptane::Target &target) const override;
+  data::InstallOutcome finalizeInstall(const Uptane::Target &target) const override {
+    (void)target;
+    throw std::runtime_error("Unimplemented");
+  }
```
patrickvacek: We might want to just return a positive result instead of throwing if there really is nothing to do here. This function exists for the case when installation requires reboot or some external action before we can really call it complete. Does that apply to the android case?

src/libaktualizr/logging/boost_logging.cc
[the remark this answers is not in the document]
riqkum: I'd like to address that with a separate PR.

src/libaktualizr/config/config.cc
riqkum: I was unsure about Utils::glob purpose. It could be definitely changed to iterator-based implementation if its only aimed to select entries by extension.

src/libaktualizr/utilities/utils.cc
riqkum: It is also could be allowed once we upgrade API from 27 to 28.

src/libaktualizr/utilities/utils.cc
riqkum: I use it to point aktualizr to an allowed private storage on Android.

CMakeLists.txt
riqkum: Yes, you're right. This shouldn't be there.

src/libaktualizr/package_manager/androidmanager.cc
riqkum: No challenge, I'll restore the implementation soon.

src/libaktualizr/package_manager/androidmanager.h
riqkum: I did android package manager when finalizeInstall call was not yet introduced. It may leads some changes in the current implementation, I need to check.

src/libaktualizr/logging/boost_logging.cc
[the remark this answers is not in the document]
riqkum: Sounds reasonable. I think, I can do it.

src/libaktualizr/utilities/config_utils.h
```diff
@@ -144,9 +144,8 @@ class BaseConfig {
         continue;
       }
       if (boost::filesystem::is_directory(config)) {
-        for (const auto& config_file : Utils::glob((config / "*.toml").string())) {
+        for (const auto& config_file : Utils::getDirEntriesByExt(config, ".toml"))
```
lbonn: Flagged by clang-tidy: we do not allow braceless for loops.

src/libaktualizr/utilities/utils.h
```diff
@@ -44,10 +43,19 @@ struct Utils {
...
-  static std::vector<boost::filesystem::path> glob(const std::string &pat);
   static void createDirectories(const boost::filesystem::path &path, mode_t mode);
   static std::string urlEncode(const std::string &input);
   static CURL *curlDupHandleWrapper(CURL *curl_in, bool using_pkcs11);
+  static std::vector<boost::filesystem::path> getDirEntriesByExt(const boost::filesystem::path &dir_path,
+                                                                 const std::string &ext);
+#if defined(ANDROID)
+  static void setStorageRootPath(const std::string &storage_root_path);
+  static boost::filesystem::path getStorageRootPath();
+  static std::mutex storage_root_path_mutex_;
```
lbonn: Does it really need a mutex? I would have thought that `setStorageRootPath()` would be called once in early initialization.

I also think we could forego the `#if defined` here and just make this api available in all cases, with a suitable comment (like "used for Android support").
The code in `SafeTempRoot` could default to use this path if not empty and attempt to use a temporary directory otherwise.

So, `if (!Utils::getStorageRootPath())` instead of `#if defined(ANDROID)`.

src/libaktualizr/package_manager/androidmanager.h
```diff
@@ -0,0 +1,39 @@
...
+
+  data::InstallOutcome install(const Uptane::Target &target) const override;
+  data::InstallOutcome finalizeInstall(const Uptane::Target &target) const override;
+
+ private:
+  std::shared_ptr<INvStorage> storage_;
+};
+
+struct AndroidInstallationState {
+  enum State { STATE_NOP, STATE_READY, STATE_IN_PROGRESS, STATE_INSTALLED, STATE_UPDATE };
```
lbonn: Another style consideration: we now prefer using `enum class` and name enum variants according to google's c++ style.

e.g.: https://github.com/advancedtelematic/aktualizr/blob/97214dcab4acc2284d4169503ae09fabefe6af34/src/libaktualizr/primary/initializer.h#L11

src/libaktualizr/package_manager/androidmanager.cc
```diff
@@ -0,0 +1,105 @@
...
+  Json::Value packages(Json::arrayValue);
+  Json::Value package;
+  package["name"] = "aktualizr";
+  package["version"] = "n/a";
+  packages.append(package);
+  return packages;
+}
+
+Uptane::Target AndroidManager::getCurrent() const {
+  std::vector<Uptane::Target> installed_versions;
```
lbonn: Looks like there is a lot going on behind the scene between this and the aktualizr-android code. I don't object merging soon since it's platform-specific code but it would be nice to have quick summaries:

- what are exactly the roles of the different components and how they interact with the state machine?
- do you think the need to structure the code like that came from limitations of our current package manager api (not flexible enough)? If so, do you have suggestions to change it to handle this scenario more nicely?

src/libaktualizr/utilities/utils.h
riqkum: Yes, that's a better idea.

src/libaktualizr/package_manager/androidmanager.cc
riqkum: Existing package manager interface doesn't fit well while aktualizr is in role of a slave within aktualizr-android application. Extra entity (dispatcher) that I've introduced is required in that case as a sync. point between the sides. I did so to minimize interaction complexity and decrease amount of data to be exposed. And it brings me faster to the point I could play with an update mechanism on Android side.  

However I have the same doubts about clearness of the current solution. I'll look what I can do to simplify it.

While entire update approach on Android is not well-established yet, I have no precise suggestions about changing package manager api.

lbonn: Thanks for your summary!

If you don't manage to simplify the integration dramatically, I think it could wait. But some documentation about the complete installation process with details about each actors' interactions would be greatly appreciated (when you think it's ready). Either an external doc or helpful code comments would work.

src/libaktualizr/package_manager/androidmanager.cc
```diff
@@ -0,0 +1,127 @@
...
+#include <forward_list>
+
+#include "androidmanager.h"
+
+#include "utilities/utils.h"
+
+#include <boost/phoenix/stl/algorithm/transformation.hpp>
+#include <boost/spirit/include/phoenix_container.hpp>
+#include <boost/spirit/include/phoenix_core.hpp>
+#include <boost/spirit/include/qi.hpp>
```
patrickvacek: What was it that required using these boost libraries? I see the code below, but I'm curious while normal C++ string parsing was inadequate.

src/libaktualizr/utilities/utils.cc
```diff
@@ -736,6 +740,12 @@ class SafeTempRoot {
   boost::filesystem::path path;
 };
 
+std::string Utils::storage_root_path_;
+
+void Utils::setStorageRootPath(const std::string &storage_root_path) { storage_root_path_ = storage_root_path; }
```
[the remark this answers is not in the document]
patrickvacek: Never mind, just realized it's called by aktualizr-android.

src/libaktualizr/package_manager/androidmanager.cc
```diff
@@ -0,0 +1,127 @@
...
+  return packages;
+}
+
+Uptane::Target AndroidManager::getCurrent() const {
+  using boost::phoenix::push_front;
+  using boost::spirit::ascii::xdigit;
+  using qi::_1;
+
+  std::string getprop_output;
+  if (0 == Utils::shell("getprop ota.last_installed_package_file", &getprop_output)) {
```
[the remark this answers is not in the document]
riqkum: See [explanation](https://github.com/advancedtelematic/aktualizr-android/tree/feat/ota-enabled-android-instructions#installed-target-hash-property) in aktualizr-android.

lbonn:
I think this is also good to go when we'll have resolved your comments @patrickvacek. The boost spirit added dependency doesn't worry me too much if it stays in the android part of code.

src/libaktualizr/package_manager/androidmanager.cc
riqkum: I'm not aware of all the approaches you're trying to follow in the project, so be specific. Standard library does not offer comparable functionality, and hand-written find/extract would be less expressive than using boost/spirit/Qi.

src/libaktualizr/package_manager/androidmanager.cc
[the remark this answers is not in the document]
patrickvacek: Great, thanks!

src/libaktualizr/package_manager/androidmanager.cc
patrickvacek: No problem, I think I've got a good enough handle on it.

PR #1035, opened 2018-12-17
Add image without actual file

patriotyk:


patriotyk:
I don't understand two other tasks:
'setting director's targets to be empty.' and 'setting staging (unsigned) director targets to be the same as current director targets.'

patrickvacek:
@OYTIS can probably explain better than I can!

OYTIS:
@patriotyk 

Current workflow of scheduling an update to director in aktualizr-repo is as following:

1. Add several targets to director with `aktualizr-repo --command addtarget`. The targets are added to the staged version of diretor/targets.json that is not accessible to remote clients.
2. Sign all the added targets scheduling and update `aktualizr-repo --command signtargets`

So there are two problems here:
1. We can't clear targets.json. The only way to create a new staging targets.json is to issue `addtarget` command, after which it will obviously be non-empty.
2. We're starting from scratch every time. It would be convenient to be able to take the current (signed) targets.json and use it as a base.

So I imagined two new commands:

- `emptytargets` will just make staging targets empty, if the user signs it immediately with `signtargets`, it will schedule an empty update (what our backend currently does when no update is scheduled).
- `oldtargets` will populate staging targets with what is currently signed (stripping the signature of course). User can then use `addtarget` command to add new targets to it. I think we'll need `deltarget` as well, but that belongs to another task.

src/aktualizr_repo/main.cc
```diff
@@ -24,7 +24,10 @@ int main(int argc, char **argv) {
     ("keyname", po::value<std::string>(), "name of key's role")
     ("repotype", po::value<std::string>(), "director|image")
     ("correlationid", po::value<std::string>()->default_value(""), "correlation id")
-    ("keytype", po::value<std::string>()->default_value("RSA2048"), "UPTANE key type");
+    ("keytype", po::value<std::string>()->default_value("RSA2048"), "UPTANE key type")
+    ("targetname", po::value<std::string>(), "target's name")
+    ("targethash", po::value<std::string>(), "target's hash")
+    ("targetlength", po::value<uint64_t>(), "target's length");
```
patrickvacek: These three new options should have some explanation here that they are intended for a special purpose. Something like "target name (for adding metadata without an actual file)".

src/aktualizr_repo/image_repo.cc
```diff
@@ -44,4 +50,17 @@ void ImageRepo::addImage(const boost::filesystem::path &image_path) {
   timestamp["meta"]["snapshot.json"]["version"] = snapshot["version"].asUInt();
   Utils::writeFile(repo_dir / "timestamp.json",
                    Utils::jsonToCanonicalStr(signTuf(Uptane::Role::Timestamp(), timestamp)));
+}
+
+void ImageRepo::addImage(const std::string &name, const std::string &hash, uint64_t length) {
```
patrickvacek: Not big, but looks like `length` can be `const`.

patriotyk:
@OYTIS Thanks, now I understand.

patriotyk:
Looks like I have added all commands

patrickvacek:
@patriotyk if Travis is giving you trouble, rebase so that you get Laurent's fix to speed up the aktualizr-repo test.

patriotyk:
Are there is something to fix?

src/aktualizr_repo/image_repo.h
```diff
@@ -8,6 +8,10 @@ class ImageRepo : public Repo {
   ImageRepo(boost::filesystem::path path, const std::string &expires, std::string correlation_id)
       : Repo(Uptane::RepositoryType::Image(), std::move(path), expires, std::move(correlation_id)) {}
   void addImage(const boost::filesystem::path &image_path);
```
lbonn: It feels like too much overloading of this `addImage`.

Could be helpful to give them different names.

src/aktualizr_repo/image_repo.cc
```diff
@@ -44,4 +50,17 @@ void ImageRepo::addImage(const boost::filesystem::path &image_path) {
   timestamp["meta"]["snapshot.json"]["version"] = snapshot["version"].asUInt();
   Utils::writeFile(repo_dir / "timestamp.json",
                    Utils::jsonToCanonicalStr(signTuf(Uptane::Role::Timestamp(), timestamp)));
+}
+
+void ImageRepo::addImage(const std::string &name, const std::string &hash, const uint64_t length) {
+  Json::Value target;
+  target["length"] = Json::UInt(length);
+  if (hash.size() == 64) {
```
lbonn: Not really elegant... Could we use `Uptane::Hash` and have too separate arguments on the commandline, like `targetsha256` and `targetsha512`, or does it make cli handling too complicated?

lbonn:
Looks fine to me, except for the two small comments.

If you make changes again, I can take a quick look on Friday.

PR #1036, opened 2018-12-17
Disconnect all slots on Aktualizr destructor

patriotyk:
This PR fixes APICheck test fails. 
But the problem was in Aktualizr class where we didn't disconnects from connected slots at destruction.

Signed-off-by: Serhiy Stetskovych <patriotyk@gmail.com>

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -909,6 +909,7 @@ TEST(Aktualizr, APICheck) {
     aktualizr.Shutdown();
   }
   EXPECT_LT(num_events_UpdateCheck, 100);
+  EXPECT_GT(num_events_UpdateCheck, 0);
 }
 
 class HttpPutManifestFail : public HttpFake {
@@ -920,6 +921,15 @@ class HttpPutManifestFail : public HttpFake {
   }
 };
 
+int num_events_PutManifestError = 0;
+void process_events_PutManifestError(const std::shared_ptr<event::BaseEvent>& event) {
+  std::cout << event->variant << "\n";
+  if (event->variant == "UpdateCheckComplete") {
+    EXPECT_EQ(std::static_pointer_cast<event::UpdateCheckComplete>(event)->result.status, result::UpdateStatus::kError);
+    num_events_PutManifestError++;
+  }
+}
+
 /* Send UpdateCheckComplete event after failure */
 TEST(Aktualizr, PutManifestError) {
   TemporaryDirectory temp_dir;
@@ -928,15 +938,15 @@ TEST(Aktualizr, PutManifestError) {
   Config conf = makeTestConfig(temp_dir, "http://putmanifesterror");
 
   auto storage = INvStorage::newStorage(conf.storage);
-  std::function<void(std::shared_ptr<event::BaseEvent> event)> f_cb = process_events_UpdateCheck;
+  std::function<void(std::shared_ptr<event::BaseEvent> event)> f_cb = process_events_PutManifestError;
 
-  num_events_UpdateCheck = 0;
+  num_events_PutManifestError = 0;
   Aktualizr aktualizr(conf, storage, http);
   boost::signals2::connection conn = aktualizr.SetSignalHandler(f_cb);
   aktualizr.Initialize();
   auto result = aktualizr.CheckUpdates().get();
   EXPECT_EQ(result.status, result::UpdateStatus::kError);
-  EXPECT_EQ(num_events_UpdateCheck, 1);
+  EXPECT_EQ(num_events_PutManifestError, 1);
 }
 
 /* Test that Aktualizr retransmits DownloadPaused and DownloadResumed events */


src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -926,13 +926,14 @@ TEST(Aktualizr, APICheck) {
     Aktualizr aktualizr(conf, storage, up, sig);
     boost::signals2::connection conn = aktualizr.SetSignalHandler(f_cb);
     aktualizr.Initialize();
-    for (int i = 0; i < 100; ++i) {
+    for (int i = 0; i < 101; ++i) {
```
OYTIS: Why?

patriotyk: To make 100 cycles, otherwise we make only 99

OYTIS:
Solves the problem for me. I'm not completely comfortable with the solution though. We should clearly document that our users are never supposed to run multiple instances of Aktualizr. Or maybe `shared_ptr` is just wrong abstraction here.
Nevertheless, I believe it should be merged.

src/libaktualizr/primary/aktualizr_test.cc
OYTIS: Um... no.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -944,6 +945,14 @@ class HttpPutManifestFail : public HttpFake {
   }
 };
 
+int num_events_PutManifestError = 0;
+void process_events_PutManifestError(const std::shared_ptr<event::BaseEvent>& event) {
+  std::cout << event->variant << "\n";
+  if (event->variant == "UpdateCheckComplete") {
+    num_events_PutManifestError++;
```
eu-smirnov: We should check here, that it's really an error.

patrickvacek:
> We should clearly document that our users are never supposed to run multiple instances of Aktualizr at least.

You are probably right that we should make that clear. At first glance, it seems obvious to me, but perhaps it isn't. Even better: is there a way we could detect if aktualizr is already running and abort?

> Or maybe `shared_ptr` is just wrong abstraction here.

For what exactly?

OYTIS:
@patrickvacek 
> For what exactly?

For sharing an event channel. Channel is obviously something asymmetric, while `shared_ptr` is symmetric. Moreover, `shared_ptr` makes an impression that it can be shared between multiple instances of `Aktualizr`, which it even technically can.

OYTIS:
Maybe `events::Channel` should not be really exposed to Aktualizr user. It was convenient when a channel was a queue of events, but now that we are using `boost::signal2` with callbacks, I see little value and a decent amount of confusion here.

OYTIS:
Yes, you're right, the public constructor doesn't allow the user to do it. It's worth fixing the test then to not abuse the private constructor (like, keep the scope of `event::Channel` the same as the scope of `Aktualizr`). Would be better than `disconnect_all_slots` IMO.

OYTIS:
This is a more pleasant solution IMO: https://github.com/advancedtelematic/aktualizr/pull/1037

src/libaktualizr/primary/aktualizr_test.cc
patriotyk: I can change it to 'i <=100 '

patriotyk:
Oh I just only noticed that sig_ comes from outside of Aktualizr instance and dissconnecting slots inside is wrong. So I agree with Anton's #1037 PR. But it complements my PR, not replaces.

src/libaktualizr/primary/aktualizr_test.cc
eu-smirnov: There are 100 numbers in the range [0, 99]. `i < 100` is exactly what we want.

src/libaktualizr/primary/aktualizr_test.cc
patriotyk: We do that in 947 line

src/libaktualizr/primary/aktualizr_test.cc
eu-smirnov: @patriotyk happens to me as well sometimes :)

src/libaktualizr/primary/aktualizr_test.cc
eu-smirnov: Well, there we check a result, not an event. I think it won't hurt to check if event status matches the result.

patriotyk: OK, Updated

patrickvacek:
> dissconnecting slots inside is wrong.

This PR is still disconnecting the slots, though. Is that unintentional now? I think I agree with Anton that i'd rather not mess with the slots internally. It seems too likely to be hiding other problems.

patriotyk:
It disconnect slots in destructor but for now member is incapsulated and never exposed. So there is no any drawback to dissconnect slots from that member which will be destroyed immediatelyy after Aktualizr destructor finished.

patriotyk:
But If we destroy this object with destructor we should not worry about disconnection and all slots will be disconnected implicitly. It is maybe beter to remove this line of code. So I have updated PR and removed it.

PR #1038, opened 2018-12-18
Only run aktualizr-repo tests with ed25519

lbonn:
They used to be run with all key types, including big RSA.

This was our longest running test, despite testing one of our less critical piece of code. Let's just use the fastest key type we have.

### DIFF
--- a/src/aktualizr_repo/repo_test.cc
+++ b/src/aktualizr_repo/repo_test.cc
@@ -10,7 +10,7 @@
 #include "test_utils.h"
 #include "uptane_repo.h"
 
-KeyType key_type = KeyType::kUnknown;
+KeyType key_type = KeyType::kED25519;
 std::string generate_repo_exec;
 
 TEST(aktualizr_repo, generate_repo) {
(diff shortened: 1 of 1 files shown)


src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -1016,6 +1016,8 @@ int main(int argc, char** argv) {
   logger_init();
   logger_set_threshold(boost::log::trivial::trace);
 
+  // note: we used to run tests for all key types, which was pretty slow.
+  // Now, they only run with ed25519
   return RUN_ALL_TESTS();
```
patrickvacek: Did you mean for this comment to be here, or should it be in `repo_test.cc`? Otherwise, I'm in favor!

patrickvacek:
Nice:
` 87/112 Test  #91: test_aktualizr_repo .................................   Passed    7.77 sec`

src/libaktualizr/primary/aktualizr_test.cc
lbonn: Wow, what a mix-up... Thanks, I'll fix that

lbonn:
@patriotyk no problem, feel free to submit enhancements in other PRs if we can have better tests with a reasonable running time.

I went for the easiest way because the running time really got out of hand and was causing timeouts on Travis.

PR #1039, opened 2018-12-18
Get rid of running modes.

patrickvacek:
aktualizr still has an input parameter (--run-mode) that does the same thing, but this is no longer wired directly into libaktualizr; it exists purely in the wrapper outside it. This should be much cleaner and simpler to use.

Note that once upon a time I advocated for all aktualizr commandline options to be also represented in the config. I no longer believe this to be a good idea for things that only matter outside of libaktualizr. An internal "run mode" for libaktualizr does not make sense. It should be driven by the code outside of the API.

### DIFF
--- a/actions.md
+++ b/actions.md
@@ -256,11 +256,11 @@ These are internal requirements that are relatively opaque to the user and/or co
 
 This is just the list of sequences currently covered. It is likely that there are more worth testing, but these tests are expensive.
 
-- [x] Initialize -> CheckUpdates -> no updates -> no further action or events (aktualizr_test.cc)
+- [x] Initialize -> UptaneCycle -> no updates -> no further action or events (aktualizr_test.cc)
 - [x] Initialize -> UptaneCycle -> updates downloaded and installed for primary and secondary (aktualizr_test.cc)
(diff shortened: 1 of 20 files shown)


lbonn:
Great, thanks! Some instances left (bad doc and failing tests):

```
$ git grep running-mode
docs/selectively-triggering-aktualizr.adoc:It is also possible to trigger the update cycle, or individual parts of the update cycle, manually or programmatically. This is done by using aktualizr's `--running-mode` option, or just supplying the operation name as a subcommand.
docs/selectively-triggering-aktualizr.adoc:Installation reports are sent when aktualizr polls the server for updates, so the `check` running-mode should be used again after installing:
scripts/test_aktualizr_repo.sh:aktualizr --config "${TMPDIR}/sota.toml" --running-mode once
scripts/test_install_aktualizr_and_update.sh:aktualizr -c /persistent/selfupdate.toml -c "$TEMP_DIR/conf.toml" --running-mode=once
```

patrickvacek:
> Some instances left (bad doc and failing tests)

On it, I noticed at the same time you did!

PR #1041, opened 2018-12-18
Rework installation results in manifest

lbonn:
It now includes individual results for all ecus + a global device result.

The device result is now a simple AND on all ecu results.

Not ready, I need to add correlation ids as well. And should probably wait until January anyway :)


lbonn:
Just rebased it on master

src/libaktualizr/storage/storage_common_test.cc
```diff
@@ -384,24 +384,45 @@ TEST(storage, load_store_installed_versions) {
   }
 }
 
-/* Load and store an installation result. */
-TEST(storage, load_store_installation_result) {
+/* Load and store a installation results (ecu and device). */
+TEST(storage, load_store_installation_results) {
```
patrickvacek: If you change this line, you should probably change the corresponding line in `actions.md`. And actually, I think you've added some new actions that are worth specifying. (And very minor, but you can drop the "a".)

lbonn:
Backend seems to agree with the format but they'll need to make the corresponding changes there. So we might have to put this PR on hold until it's ready there.

I'll update the actions list in the meantime.

lbonn:
Small fix in the last commit to not include an installation report with empty content when not needed.

src/libaktualizr/storage/storage_common_test.cc
patrickvacek: Sorry to be stubborn, but in lieu of a better way to link the test case to the action, I think the comment here should match the actions exactly, e.g.:
```
/*
 * Load and store an ecu installation result in an SQL database
 * Load and store a device installation result in an SQL database
 */
```

lbonn: Sure thing. Must have thought that the "SQL database" part is too specific, but I'll change it for consistency.

patrickvacek: Fair point. The specificity in a place that should be generic is annoying, but I think the need for linkage outweighs that at least for now.

lbonn:
Ok, got the green light. Should be ready now if you like it.

src/libaktualizr/utilities/types_test.cc
```diff
@@ -31,6 +31,20 @@ TEST(TimeStamp, Now) {
...
+  EXPECT_EQ(ok_res.num_code, data::ResultCode::Numeric::kOk);
+  EXPECT_EQ(ok_res.toString(), "OK");
+  std::string repr = ok_res.toRepr();
+  EXPECT_EQ(repr, "OK:0");
+  EXPECT_EQ(data::ResultCode::fromRepr(repr), ok_res);
+
+  // !
+  EXPECT_NE(ok_res, data::ResultCode(data::ResultCode::Numeric::kOk, "OK2"));
+  EXPECT_NE(ok_res, data::ResultCode(data::ResultCode::Numeric::kGeneralError, "OK"));
+  EXPECT_EQ(data::ResultCode::fromRepr("OK"), data::ResultCode(data::ResultCode::Numeric::kUnknown, "OK"));
```
patrickvacek: What's with the comment? Is there something unexpected here?

lbonn:
@patrickvacek I've broken the migration of installation_result (and not installed_versions!), from fs but also from previous sql because we've changed the format quite a bit.

It means that updating to this version from an older aktualizr would probably display a failed or pending update on the ui! We can discuss if you think it's breaking too much. In any case, it warrants at least a manual test (will do).

src/libaktualizr/utilities/types_test.cc
lbonn: The comparison operator could appear surprising. For example, if there is no `:k` part, the numeric code is unknown. These are three instances of gotcha comparison, but the comment could be more verbose I admit.

patrickvacek: It looks fine to me, I was just alarmed by the exclamation point. A better comment may be helpful, but it's fine as it is.

patrickvacek:
> I've broken the migration of installation_result (and not installed_versions!)

Oops, sorry, that was a typo on my part.

> It means that updating to this version from an older aktualizr would probably display a failed or pending update on the ui! We can discuss if you think it's breaking too much. In any case, it warrants at least a manual test (will do).

I think we agreed previously that this is fine, since the worst that will happen is we will need to re-start the update process, but we shouldn't get stuck. But yes, a manual test is a great idea.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -940,48 +1005,53 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
...
+                               install_res);
     // TODO: other updates for primary
   } else {
     LOG_INFO << "No update to install on primary";
   }
 
   auto sec_reports = sendImagesToEcus(updates);
-  result.reports.insert(result.reports.end(), sec_reports.begin(), sec_reports.end());
+  result.ecu_reports.insert(result.ecu_reports.end(), sec_reports.begin(), sec_reports.end());
+  computeDeviceInstallationResult(&result.dev_report, correlation_id);
```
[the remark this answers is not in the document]
lbonn: So it's a bit tricky. At install(), a package manager can return:

- kOk. If all ecus are ok, device installation result will be ok and the manifest will be sent immediately (since there is no pending update)
- kNeedCompletion. If one ecu needs completion, the manifest won't be sent until the flag is unset which will happen at next reboot. Then device result will be ok or error depending on the final states of all ecus
- an error. If one ecu has an error, device installation result will not be ok. But one ecu needing completion will still block manifest emission.

Don't know if it's really clear, maybe it would need a state machine drawing. Also, kNeedCompletion will only be supported for primaries right now.

lbonn:
For the other points:

1. Yes `OperationResult` is gone but has another shape now :). The hierarchy I've made might be a bit convoluted,
2. Yes raw_report is a free-form string per device. It's partially supported now, in the internal api but, right now, the device installation result is computed by SotaUptaneClient which uses an empty raw report (see `computeDeviceInstallationResult`). We'll need to think of an entry point of injection in the api.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -940,48 +1005,53 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
...
+      if (install_res.result_code.num_code == data::ResultCode::Numeric::kNeedCompletion) {
         // update needs a reboot, send distinct EcuInstallationApplied event
-        report_queue->enqueue(std_::make_unique<EcuInstallationAppliedReport>(ecu_serial, correlation_id));
-        sendEvent<event::InstallTargetComplete>(ecu_serial, true);  // TODO: distinguish from success here?
-      } else if (status.result_code == data::UpdateResultCode::kOk) {
-        report_queue->enqueue(std_::make_unique<EcuInstallationCompletedReport>(ecu_serial, correlation_id, true));
-        sendEvent<event::InstallTargetComplete>(ecu_serial, true);
+        // note: installation result is not save here, will be done after finalization
+        report_queue->enqueue(std_::make_unique<EcuInstallationAppliedReport>(primary_ecu_serial, correlation_id));
+        sendEvent<event::InstallTargetComplete>(primary_ecu_serial, true);  // TODO: distinguish from success here?
```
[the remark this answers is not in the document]
lbonn: It's a bit false/misleading now, good catch. I'll delete it.

lbonn:
> 
> > It means that updating to this version from an older aktualizr would probably display a failed or pending update on the ui! We can discuss if you think it's breaking too much. In any case, it warrants at least a manual test (will do).
> 
> I think we agreed previously that this is fine, since the worst that will happen is we will need to re-start the update process, but we shouldn't get stuck. But yes, a manual test is a great idea.

After a test with new backend, it actually seems to be fine from the user's perspective: they accept an empty installation_report as success if aktualizr reports the new version as installed.

lbonn:
I've added two unrelated commits, after [this comment](https://github.com/advancedtelematic/aktualizr/pull/1072#issuecomment-459014525) + some internal discussions.

patrickvacek:
The new commits look promising and I'd like to get this in soon so I can rebase the delegations PR on it.

lbonn:
@patrickvacek the last one breaks the test suite, I'll spin it off to another PR so that it doesn't block the flow.

PR #1043, opened 2018-12-28
More tests

patriotyk:


src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -329,6 +331,91 @@ TEST(Uptane, InstallFake) {
...
+
+  Uptane::RawMetaPack meta;
+  storage->loadLatestRoot(&meta.director_root, Uptane::RepositoryType::Director());
+  storage->loadNonRoot(&meta.director_targets, Uptane::RepositoryType::Director(), Uptane::Role::Targets());
+  storage->loadLatestRoot(&meta.image_root, Uptane::RepositoryType::Image());
+  storage->loadNonRoot(&meta.image_timestamp, Uptane::RepositoryType::Image(), Uptane::Role::Timestamp());
+  storage->loadNonRoot(&meta.image_snapshot, Uptane::RepositoryType::Image(), Uptane::Role::Snapshot());
+  storage->loadNonRoot(&meta.image_targets, Uptane::RepositoryType::Image(), Uptane::Role::Targets());
+
+  EXPECT_CALL(*sec, putMetadata(matchMeta(meta)));
```
lbonn: cool!

codecov.yml
```diff
@@ -1,4 +1,4 @@
 ignore:
   - "third_party"
   - "tests"
-  - "*_test.cc"
+  - "**/*_test.cc"
```
lbonn: Good catch!
For lcov reports, this list is used https://github.com/advancedtelematic/aktualizr/blob/19c6f433c7c9ad217c37b25f551613082a05d228/CMakeLists.txt#L341. It has a different format...

Now, the two coverage figures on this PR are closer but still different:
- 77.6% according to lcov (on Jenkins)
- 75.86% according to codecov

But it should not cause too much worry, not a big problem.

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -227,13 +229,66 @@ TEST(Uptane, PutManifest) {
...
+
+  auto storage = INvStorage::newStorage(conf.storage);
+  auto events_channel = std::make_shared<event::Channel>();
+  std::function<void(std::shared_ptr<event::BaseEvent> event)> f_cb = process_events_PutManifestError;
+  events_channel->connect(f_cb);
+  num_events_PutManifestError = 0;
+  auto sota_client = SotaUptaneClient::newTestClient(conf, storage, http, events_channel);
+  EXPECT_NO_THROW(sota_client->initialize());
+  auto result = sota_client->putManifest();
+  EXPECT_FALSE(result);
```
patrickvacek: Why do we expect this to fail? Is it just because of an invalid URL? It's worth making it explicit why this works as it does.

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -918,6 +1090,36 @@ TEST(Uptane, offlineIteration) {
   EXPECT_TRUE(sota_client->uptaneOfflineIteration(&targets_offline, nullptr));
   EXPECT_EQ(targets_online, targets_offline);
 }
+/*
+ Ignore updates for unrecognized ECUs.
+*/
+TEST(Uptane, DISABLED_IgnoreUnknownUpdate) {
+  // FIXME This test is disabled because it conflicts with some tests in uptane_vector_tests.cc
```
patrickvacek: What's the problem here? I don't understand how it conflicts.

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -918,6 +1090,36 @@ TEST(Uptane, offlineIteration) {
...
+
+  EXPECT_NO_THROW(sota_client->initialize());
+  sota_client->AssembleManifest();
+
+  std::vector<Uptane::Target> targets;
+  EXPECT_TRUE(sota_client->uptaneIteration());
+  EXPECT_TRUE(sota_client->getNewTargets(&targets));
+
+  EXPECT_EQ(targets.size(), 1);
+  EXPECT_EQ(targets[0].filename(), "secondary_firmware.txt");
```
patrickvacek: We need to do just a bit more here to prove that nothing was actually updated and the targets were ignored.

patrickvacek: Also, there are two places that we need to verify that unknown targets are ignored: during the fetch and during the actual check for updates. Seems a bit redundant at a high level, but worth checking both, and since most of the work is already done here, it seems like it should be easy to check both here.

patrickvacek: Actually, we can probably cover the `Reject targets which do not match a known ECU` case here as well. Just try installing the bogus target and verify that it is rejected.

src/libaktualizr/uptane/uptane_test.cc
patriotyk: Oh yes, I forgot to tell you about this test. uptane_vector_tests expects exception if we get unknown update. But this test checks, do we ignore unknown update. So we should decide, should we throw exception or ignore.

src/libaktualizr/uptane/uptane_test.cc
patriotyk: This test checks if we get event 'PutManifestComplete' even when putManifest fails. It fails by returning non 200 return code.

I think the name of the test speaks for itself.

patrickvacek: It's true, the name does make it clear that the putManifest fails, but what is not immediately clear to me is WHY it fails.

src/libaktualizr/uptane/uptane_test.cc
patrickvacek: I don't see the conflict. In `SotaUptaneClient::getNewTargets`, it looks to me like unrecognized serials or HW IDs set `last_exception` but then return false. `uptane_vector_tests` just checks for the expected value in `last_exception`. I don't like this whole `last_exception` business, but it isn't a problem here.

src/libaktualizr/uptane/uptane_test.cc
patrickvacek: Ah, I missed that 504 when I was scanning the code before. My bad.

src/libaktualizr/uptane/uptane_test.cc
patriotyk: I am not sure it is good idea. Returning false, means some error happens and we won't be able to install good update, so that's mean we dont ignore bad updates. So we should return true even when wrong ecu update happens.

patriotyk: In my opinion, if we are going to ignore updates with wrong ecu, we should enable this test and remove appropriate tests in uptane_vector_tests, or just change them to check if we really ignore bad ecu updates. But it is not so trivial, so my suggestion is to remove them.

src/libaktualizr/uptane/uptane_test.cc
patrickvacek: @patriotyk Exactly! They are the same. Perhaps using the word "ignore" was confusing. But for now the correct thing to do is to reject and complain.

patriotyk: Ok now I understand, it is the same but first one for installing second one for downloading

patrickvacek: Close. `Reject targets which do not match a known ECU` is part of `Install updates`, but `Ignore updates for unrecognized ECUs` is part of both `Fetch metadata from server` and `Check for updates`. Currently we have a lot of redundancy between those last two steps but we haven't yet come to a conclusion on what the right thing to with all that is. For now, we should cover all of those cases.

src/libaktualizr/uptane/uptane_test.cc
patriotyk: I have enabled this test with some fixes.

src/libaktualizr/uptane/uptane_test.cc
patriotyk: I have added "Reject targets which do not match a known ECU" test case

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -1113,12 +1113,18 @@ TEST(Uptane, DISABLED_IgnoreUnknownUpdate) {
...
+  EXPECT_STREQ(sota_client->getLastException().what(),
+               "The target had an ECU ID that did not match the client's configured ECU id.");
+  sota_client->last_exception = Uptane::Exception{"", ""};
+  result = sota_client->checkUpdates();
+  EXPECT_EQ(result.status, result::UpdateStatus::kError);
+  EXPECT_STREQ(sota_client->getLastException().what(),
+               "The target had an ECU ID that did not match the client's configured ECU id.");
+  std::vector<Uptane::Target> packages_to_install = UptaneTestCommon::makePackage("testecuserial", "testecuhwid");
+  auto report = sota_client->uptaneInstall(packages_to_install);
+  EXPECT_EQ(report.reports.size(), 0);
```
patrickvacek: Nice, this is great. It looks we don't actually reject unknown targets during installation, we just silently ignore them, which perhaps isn't ideal. But given how the code works now, this is what we need. Thanks!

PR #1045, opened 2019-01-03
Avoid races in parallel tests that count events

cajun-rat:
Without this I get random failures for the test_aktualizr test because two
tests can cause UpdateCheckComplete events to arrive in parallel.

Signed-off-by: Phil Wise <phil@phil-wise.com>

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -856,13 +856,34 @@ TEST(Aktualizr, FullNoCorrelationId) {
   EXPECT_EQ(http->events_seen, 8);
 }
 
-int num_events_UpdateCheck = 0;
-void process_events_UpdateCheck(const std::shared_ptr<event::BaseEvent>& event) {
-  std::cout << event->variant << "\n";
-  if (event->variant == "UpdateCheckComplete") {
-    num_events_UpdateCheck++;
+class CountUpdateCheckEvents {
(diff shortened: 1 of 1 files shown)


patrickvacek:
Was this not already addressed by https://github.com/advancedtelematic/aktualizr/pull/1036? The name of that PR is deceiving; the final version was just a simple fix for the same issue that I think this is intended to fix.

cajun-rat:
Erm, possibly. I rebased this patch just before opening the PR, so it may not be needed any more.  I'm naturally suspicious of any global state like `num_events_UpdateCheck`--I wonder if a mock object like CountUpdateCheckEvents makes for a cleaner solution?

cajun-rat:
> Eventually we might want to move towards making CountUpdateCheckEvents a bit more generic and having the various tests that use this style use subclasses as necessary.

Yup, certainly. I found the `std::function<void(std::shared_ptr<event::BaseEvent>)> signal_{
      [this](std::shared_ptr<event::BaseEvent> e) { this->count(e); }};` really ugly, but I couldn't find @lbonn to ask what the correct solution was.  There _must_ be an incantation of `operator()` that is acceptable to boost::signals2, but I couldn't find it :)

lbonn:
I'm back :).

Yes, working with closures like that is often a pain... I wouldn't sweat it at the moment since it's only testing code.

We'll see when we'll revisit this code again.

PR #1046, opened 2019-01-03
RFC: Docker compose

cajun-rat:
I wanted to get some comments on this before pushing much further. The goal here is to have secondary whose 'firmware' is a docker-compose yaml file.

Pushing a docker-compose.yml via HERE OTA Connect basically works, but I haven't done a proper set of testing on this.

### DIFF
--- a/config/docker-compose-secondaries/example-redis.yml
+++ b/config/docker-compose-secondaries/example-redis.yml
@@ -1,0 +1,5 @@
+version: '3'
+services:
+  redis:
+    image: "redis:alpine"
+    restart: always
--- a/src/libaktualizr/utilities/utils.cc
+++ b/src/libaktualizr/utilities/utils.cc
@@ -313,7 +313,8 @@ static ssize_t read_cb(struct archive *a, void *client_data, const void **buffer
   return s->is.gcount();
 }
 
-void Utils::writeFile(const boost::filesystem::path &filename, const std::string &content, bool create_directories) {
+void Utils::writeFile(const boost::filesystem::path &filename, const std::string &content, bool create_directories,
+                      bool atomic) {
   // also replace the target file atomically by creating filename.new and
   // renaming it to the target file name
   boost::filesystem::path tmpFilename = filename;
@@ -327,13 +328,19 @@ void Utils::writeFile(const boost::filesystem::path &filename, const std::string
     throw std::runtime_error(std::string("Error opening file ") + tmpFilename.string());
   }
   file << content;
+  if (atomic) {
+    sync();
+  }
   file.close();
-
   boost::filesystem::rename(tmpFilename, filename);
+  if (atomic) {
+    sync();
+  }
 }
 
-void Utils::writeFile(const boost::filesystem::path &filename, const Json::Value &content, bool create_directories) {
-  Utils::writeFile(filename, jsonToStr(content), create_directories);
+void Utils::writeFile(const boost::filesystem::path &filename, const Json::Value &content, bool create_directories,
+                      bool atomic) {
+  Utils::writeFile(filename, jsonToStr(content), create_directories, atomic);
 }
 
 std::string Utils::jsonToStr(const Json::Value &json) {
--- a/config/docker-compose-secondaries/docker-compose.json
+++ b/config/docker-compose-secondaries/docker-compose.json
@@ -1,0 +1,11 @@
+{
+  "secondary_type" : "docker_compose",
+  "partial_verifying" : "false",
+  "ecu_hardware_id" : "docker-compose",
+  "full_client_dir" : "storage/demo-vsec1",
+  "ecu_private_key" : "sec.private",
+  "ecu_public_key" : "sec.public",
+  "firmware_path" : "storage/demo-vsec1/docker.yml",
+  "target_name_path" : "storage/demo-vsec1/target_name",
+  "metadata_path" : "storage/demo-vsec1/metadata"
+}
--- a/src/libaktualizr/uptane/CMakeLists.txt
+++ b/src/libaktualizr/uptane/CMakeLists.txt
@@ -1,4 +1,5 @@
-set(SOURCES fetcher.cc
+set(SOURCES dockersecondary.cc
+    fetcher.cc
     ipsecondarydiscovery.cc
     ipuptanesecondary.cc
     managedsecondary.cc
@@ -12,7 +13,8 @@ set(SOURCES fetcher.cc
     imagesrepository.cc
     virtualsecondary.cc)
 
-set(HEADERS exceptions.h
+set(HEADERS dockersecondary.h
+    exceptions.h
     fetcher.h
     ipsecondarydiscovery.h
     ipuptanesecondary.h
--- a/src/libaktualizr/uptane/CMakeLists.txt
+++ b/src/libaktualizr/uptane/CMakeLists.txt
@@ -5,6 +5,7 @@ set(SOURCES fetcher.cc
     partialverificationsecondary.cc
     role.cc
     root.cc
+    secondaryconfig.cc
     tuf.cc
     uptanerepository.cc
     directorrepository.cc
--- a/src/libaktualizr/uptane/dockersecondary.cc
+++ b/src/libaktualizr/uptane/dockersecondary.cc
@@ -1,0 +1,38 @@
+#include "uptane/dockersecondary.h"
+
+#include "logging/logging.h"
+
+#include <sstream>
+
+using std::stringstream;
+
+namespace Uptane {
+DockerComposeSecondary::DockerComposeSecondary(const SecondaryConfig& sconfig_in) : ManagedSecondary(sconfig_in) {}
+
+bool DockerComposeSecondary::storeFirmware(const std::string& target_name, const std::string& content) {
+  Utils::writeFile(sconfig.target_name_path, target_name, true);
+  Utils::writeFile(sconfig.firmware_path, content, true);
+  stringstream cmd_line;
+  cmd_line << "docker-compose -f " << sconfig.firmware_path << " up -d";
+  LOG_INFO << " Running " << cmd_line.str();
+  int rescode = system(cmd_line.str().c_str());
+  return rescode == 0;
+}
+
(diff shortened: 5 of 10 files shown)


doanac:
This is really complimentary to the stuff we do on the Linux microPlatform. My comment would be that we routinely find that just a single compose file isn't enough. We need a "project", ie a directory with docker-compose.yml and some other files that get mounted into different containers. We also sometimes need to choose a non-default compose file (eg docker-compose-foo.yml).

So maybe rather than just passing a single file, the code could also take a tarball?

agners:
@cajun-rat this looks quite good to me. I am not sure about the exact docker compose command line arguments, can we make this a configuration option?

It would be nice to see this upstream, but then the whole ecosystem might evolve quite a bit in the next month and we might want break it/create something different. Can we mark it as experimental or similar?

@doanac I remember we were discussion that option too. I think a multiple files makes it quite a bit more complex, e.g. it would make `getFirmwareInfo` cumbersome.

I considered a single file to be an acceptable limitation since most embedded use cases are "single application". 

Can we have multiple secondaries of the same type?

cajun-rat:
Yes, multiple secondaries of  the same type should work fine. We would need to associate them with a docker-compose 'project' in order that docker-compose treats them as independent targets, but that should be easy.

@doanac Can you describe the tarball use-case a bit more?  My understanding was that (from a Docker point of view) docker-compose.yml was meant to be most-top-level thing in a deployment.

@agners Good point on the experimental-nature of this, will put a note is saying so.

doanac:
The case I run into is when you have a service like:
~~~
nginx-coap-proxy:
    image: ${REGISTRY:-hub.foundries.io}/nginx:${TAG:-latest}
    volumes:
     - ./nginx/nginx-lwm2m.conf:/etc/nginx/nginx.conf
~~~
This is using a file from a directory relative to the docker-compose. You could argue it could be done with a custom container that has that file, but that starts to become a maintenance pain.

Its possible i'm an edge case.

doanac:
@cajun-rat - that looks really interesting. i'd have to read a little closer, but I'm not sure it deals with volumes like in the issue I've presented. But maybe that's something that could be worked on upstream?

cajun-rat:
@doanac I haven't looked into what it does beyond reading the a page of the docs, so maybe it doesn't do what I'd hoped for :) What would be _nice_ is a way to turn the outcome of docker-compose into a stand-alone file that can be freely shipped around.

agners:
Docker compose bundle seems to be a nice approach to generate a single Docker compose file, will look in that too.

I guess what @doanac to also ship arbitrary files with Docker compose, e.g. like the `nginx.conf` file to parameterize containers. So this would need support for multiple files, or even complete file system trees. It almost seems that using a second OSTree would be a possible solution?

We are looking at ways to customize the base system (/etc) without rebuilding the image using OSTree union functionality. That scheme could also be used to update the Docker compose file and arbitrary additional files. However, we dislike that idea to update the compose file since it requires a reboot (to move to the new OSTree).

Can we maybe solve that by using another container? E.g. create a config container which contains the applications specific configuration files for all the containers?

doanac:
> Can we maybe solve that by using another container? E.g. create a config container which contains the applications specific configuration files for all the containers?

totally - it adds a little more process, and gives you more work. you have to make sure you are rebuilding your container when the base changes. but its certainly an option.

You could probably even pass the config files as b64 encoded data to the environment. So don't let my comment derail things too much.

cajun-rat:
If we want to distribute something looking like a directory tree or file system image, then OSTree does seem like a good solution: The delta story is already worked out, and we already have support for it throughout the server and client

agners:
@cajun-rat what are your thoughts wrt docker compose cmd line?

cajun-rat:
@agners Do you mean setting the docker-compose command line in the secondary config file? That makes sense to me. Maybe we provide a 'sane' default and allow the user to override it?  On the other hand, if there is only one 'correct' answer, then we might as well just hard-code that in...

agners:
There are quite some options, and some might be application specific I guess?
https://docs.docker.com/compose/reference/up/

Sane default with an option to override the cmd sounds good.

If you prefer we also can keep what we have now and add if we really see the need, that is fine for me too.

ricardosalveti:
> @doanac I haven't looked into what it does beyond reading the a page of the docs, so maybe it doesn't do what I'd hoped for :) What would be nice is a way to turn the outcome of docker-compose into a stand-alone file that can be freely shipped around.

Docker apps (https://github.com/docker/app) might be the supported path later on, but at the moment there are still quite a few things missing there.

LukaPivk:
> @agners Do you mean setting the docker-compose command line in the secondary config file? That makes sense to me. Maybe we provide a 'sane' default and allow the user to override it? On the other hand, if there is only one 'correct' answer, then we might as well just hard-code that in...

This sounds ok for me. I did some testing today and this is exactly what I would need :)

src/libaktualizr/utilities/utils.cc
```diff
@@ -327,13 +328,19 @@ void Utils::writeFile(const boost::filesystem::path &filename, const std::string
     throw std::runtime_error(std::string("Error opening file ") + tmpFilename.string());
   }
   file << content;
+  if (atomic) {
```
lbonn: Why do we need sync calls to make it atomic?

config/docker-compose-secondaries/example-redis.yml
```diff
@@ -0,0 +1,5 @@
+version: '3'
+services:
+  redis:
+    image: "redis:alpine"
+    restart: always
```
patrickvacek: Do we need this example file? It doesn't seem relevant to our/your needs.

cajun-rat: It is the example 'secondary image' that I uploaded to OTA Connect. No strict need for it to be present, but it was useful to have something for testing.

patrickvacek: Oh, I see! That's simpler than I expected, sorry.

cajun-rat: np. I'll get some documentation in place before merging, but I'd like to make sure this is on the right path first.

src/libaktualizr/utilities/utils.cc
lbonn: Ok the first sync operation is probably needed to be safe. However:

- that function was written to be atomic in the first place, so I don't think we need the boolean flag
- `sync()` is expensive if we call it every time we create a file. Calling `fsync()` would be ideal but looks difficult/impossible on a standard ofstream. Maybe we'd have to resort to boost (https://stackoverflow.com/questions/676787/how-to-do-fsync-on-an-ofstream) or POSIX C
- renaming a file on the same filesystem should be atomic, so I don't see the point of the `sync()` after rename

OYTIS:
I feel like it asks for being generalized into a "shell-out secondary" =)

OYTIS:
Thank you all for your input.
Preliminarily I would say that it would be a cleaner solution to have multiple targets instead of a tarball containing all of them. As the use case has been described, I'm not also sure that the target should be executed as a part of installation process or the user/some other process would choose a personality at a later point in time.
We're looking forward to seeing more details from Foundries.io.

doanac:
I haven't really thought much on this. However, I did want to add a pointer to a newer tool I think we should all evaluate, Docker App:

 https://blog.docker.com/2018/12/docker-app-and-cnab/
 https://github.com/docker/app

Reading the blog and the project's README, it seems to address the concerns I've had. Another nice feature to this is that it integrates with docker's Notary (a TUF server like ota-tuf).

I guess in this model you could just send the secondary a simple dictionary like:
~~~
{
  "app": "<your-namespace>/monitoring:0.1.0",
  "params": {
    " ports.alertmanager": "9095"
  }
}
~~~
which we'd turn into a command like: `docker-app install <your-namespace>/monitoring:0.1.0 --set ports.alertmanager=9095`

doanac:
This might be worth me creating a GitHub Issue to discuss, but I'm starting to form some more ideas around this. Full disclosure: This is mostly motivated by my interest in the aktualizr-lite work, but I also think we might be missing something if the idea of containers doesn't complement both.

The first thing I'm starting to think is that docker-app is the way to go. It address some of our questions around managing the update payload *and* it includes a "sign apps with notary" TODO item which might compliment this work.

The second thing I'm starting to think (but don't have enough experience in the code base): A docker-app or docker-compose PackageManager implementation feels like it makes sense.

Combining all that you could have a single targets.json (or maybe use the new delegation logic) that includes your base OS targets *and* docker-app targets. eg:
~~~
        "dragonboard-410c-lmp-osf-0.32" : {
            "custom" : {
               "version" : "osf-0.32",
               "targetFormat" : "OSTREE",
               "createdAt" : "2018-10-30T00:18:03Z",
               "updatedAt" : "2018-10-30T00:18:03Z",
               "uri" : "https://app.foundries.io/mp/lmp/455/",
               "name" : "dragonboard-410c-lmp",
               "hardwareIds" : [
                  "dragonboard-410c"
               ]
            },
            "length" : 0,
            "hashes" : {
               "sha256" : "f063554b57ccef4b0203135d2ffeeae3f1b013e546acfb72a6fbe8ccac81e6a3"
            }
         },
        "wordpress-osf-0.32" : {
            "custom" : {
               "version" : "osf-0.32",
               "targetFormat" : "DOCKER_APP",
               "createdAt" : "2018-10-30T00:18:03Z",
               "updatedAt" : "2018-10-30T00:18:03Z",
               "uri" : "hub.docker.io/foundriesio/wordpress",
               "name" : "wordpress",
               "docker-app-params": {
                   " ports.alertmanager": "9095"
                }
            },
            "length" : 0,
            "hashes" : {
               "sha256" : "sha-of-the-app-docker-hub"
            }
         },
~~~

I *think* in this model, you could push updates to your device that were either OS or docker-app, and the sotauptaneclient could execute the proper code path by simply looking at the "targetFormat" attribute.

I *think* aktualizr-lite could also do something similar as well, eg it could include a new "--targetFormat" flag.

patrickvacek:
@cajun-rat Is this still something you are interested in, or has this been sufficiently superseded by aktualizr-lite?

lbonn:
This PR hasn't moved in a while and the core idea seems to have been addressed.

I'll close for now, feel free to reopen if needed.

PR #1050, opened 2019-01-09
Update changelog with recent work.

patrickvacek:
Anything missing? Or is this too much? I'd like to make a release very soon, but realized this doc was out of date.


lbonn:
Found one possible additional entry:

Pause/resume was supposed to be added in last release but only works with OStree in the future one.

patrickvacek:
Looks like codecov probably shouldn't track documentation files.

patrickvacek:
Finally got the test fixed, so this should be ready to go.

PR #1055, opened 2019-01-11
Test: fetcher death

eu-smirnov:
A test scenario for the use-case when aktualizr process gets killed while being in a `pause` state. When the fetcher is called again, it's expected to start the download from the point it was paused on.
Currently, a delay of 15 seconds between `pause` call and killing the process is needed, as we have no reliable way to know when all background operations are completed, so we can resume later.
The PR also includes a fix for a stack overflow issue, that is triggered by this test on `resume`.


patrickvacek:
Looks good to me!

PR #1057, opened 2019-01-16
Show primary ecu versions in aktualizr-info

lbonn:
The initial motivation was for enhancing e2e tests support.


src/aktualizr_info/main.cc
```diff
@@ -142,6 +143,22 @@ int main(int argc, char **argv) {
       std::cout << "Private key:" << std::endl << priv << std::endl;
     }
 
+    auto pacman = PackageManagerFactory::makePackageManager(config.pacman, storage, nullptr);
+
+    Uptane::Target current_target = pacman->getCurrent();
+
+    if (current_target.IsValid()) {
+      std::cout << "Current primary ecu running version: " << current_target.sha256Hash() << std::endl;
+    }
```
patrickvacek: Should we print something if the target is not valid?

src/libaktualizr/uptane/tuf.h
```diff
@@ -253,6 +257,7 @@ class Target {
   friend std::ostream &operator<<(std::ostream &os, const Target &t);
 
  private:
+  bool valid{true};
```
patrickvacek: `valid` as a boolean is slightly weird, but I think it's harmless for now and will be easy to change if we later need to.

lbonn: It's a bit weird to have inside the type I agree. Did you have a specific improvement idea?

src/aktualizr_info/main.cc
lbonn: Good suggestion!

src/libaktualizr/uptane/tuf.h
patrickvacek: Not yet. I think it's fine for now.

PR #1058, opened 2019-01-16
primary: fix bug/typo in resume()

OYTIS:
Not sure how it could work, probably some tests are also missing.


eu-smirnov:
Lol. Yes, it looks like we need to rework our API level test to something more sensible.

Issue #1056, opened 2019-01-16
Aktualizr "Anonymous" Mode


doanac:
### Context
Aktualizr is awesome, but the effort to get things going is large. Part of that complexity is because of Uptane. There are many users who want the TUF security aspects of Aktualizr and don’t need to Director portions.

Additionally, Foundries.io has an open platform where people are free to install their OS. It would be great if anonymous users were able to use the TUF/OSTree logic in Aktualizr for updates and then have the ability to convert device(s) into a full-blow OTA Connect + Uptane setup.

To help prototype possibilities, Foundries.io has created a simple TUF/OSTree [project](https://github.com/foundriesio/tuftree/). The idea presented here, is how to get aktualizr to work in a similar fashion.

### Requirements
I'd like to propose a new "anonymous" feature for aktualizr with a minimal set of requirements:
* Backend
  *  Can be deployed with docker-compose
  * TUF
    * Docker Notary or OTA-Tuf
  * OSTree
    * Vanilla HTTP Server or TreeHub
  * Doesn’t require authentication (client just uses HTTPS anonymously)
* Client
  * User can run an “update” command pulls latest OSTree and personality and deploy
  * User can list available updates

### Changes to Aktualizr
This needs real experts, but here is doanac’s 20,000 ft level view:

It's probably a new binary that uses libaktualizr. Maybe like hmi_stub is done in-tree now. Then we look at pieces that can operate “anonymously”. By anonymous we mean:

* Doesn’t talk to device-gateway (so clients don’t need pre-provisioned keys)
* Only talks to OStree and TUF

**Pieces a client would need:**
Validating TUF meta-data: The “image repository” seems like the right place and it looks to be stand-alone already. However, the code to use it properly seems to be in the “sotaupdateclient” (updateImagesMeta()). I’m not sure if it would be a good idea to use that class directly since its got Uptane logic (and device-registry) in it. So there might be some work untangling that into a  common helper/utility function/method.

OSTree Pull and Install: The “ostree manager” handles this, and seems to have no uptane dependencies.

**Nice-to-haves:**
Rather than pinning the client’s configuration to a specific OSTree server, it would be nice if the server was an optional field in the TUF “custom” data. Everything else in targets.json should still work just fine.

### Proposed Client CLI
Not sure what to call it. I’ve used tuftree in the past, but maybe this is aktualizr-lite or something along those lines. I’ll call it $TOOL below:
~~~
$TOOL initialize --help
[...]

PR #1059, opened 2019-01-16
sotauptaneclient: Provide access to image-repository targets

doanac:
This is related to issue:
  https://github.com/advancedtelematic/aktualizr/issues/1056

In order for an application to find the latest target, it will need
access to the list of targets found in the image repository.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -1250,3 +1250,10 @@ std::string SotaUptaneClient::secondaryTreehubCredentials() const {
     return "";
   }
 }
+
+const std::vector<Uptane::Target> SotaUptaneClient::GetRepoTargets() {
+  if (!updateImagesMeta()) {
+    LOG_ERROR << "Unable to get latest repo data";
+  }
+  return images_repo.getTargets();
+}
--- a/src/libaktualizr/primary/sotauptaneclient.h
+++ b/src/libaktualizr/primary/sotauptaneclient.h
@@ -55,6 +55,8 @@ class SotaUptaneClient {
   result::CampaignCheck campaignCheck();
   void campaignAccept(const std::string &campaign_id);
 
+  const std::vector<Uptane::Target> GetRepoTargets();
+
  private:
   FRIEND_TEST(Aktualizr, FullNoUpdates);
   FRIEND_TEST(Aktualizr, FullMultipleSecondaries);
--- a/src/libaktualizr/uptane/imagesrepository.h
+++ b/src/libaktualizr/uptane/imagesrepository.h
@@ -23,6 +23,8 @@ class ImagesRepository : public RepositoryCommon {
   bool snapshotExpired() { return snapshot.isExpired(TimeStamp::Now()); }
   int64_t snapshotSize() { return timestamp.snapshot_size(); }
 
+  const std::vector<Target> getTargets() { return targets.targets; }
+
   Exception getLastException() const { return last_exception; }
 
  private:


doanac:
This is sort of an RFC. I intentionally chose the easiest way to expose Targets to a sotauptaneclient caller. Another approach approach could be creating a new TufClient class to handle this specific stuff and then base the sotauptaneclient on it. That's a lot more moving code around, so I wanted to get opinions before going too far with this.

doanac:
With these changes, you can create code like:
~~~
static void foo() {
    Config config;
    config.uptane.repo_server = "https://api.foundries.io/lmp/repo/release/api/v1/user_repo/";
    config.storage.path = "/tmp/sota";
    config.storage.uptane_metadata_path = BasedPath(config.storage.path / "metadata");
    config.pacman.type = PackageManager::kNone;

    auto storage = INvStorage::newStorage(config.storage);
    auto client = SotaUptaneClient::newDefaultClient(config, storage);
    client->GetRepoTargets();
    for(auto x : client->GetRepoTargets()) {
      LOG_INFO << x;
    }
}
~~~

patrickvacek:
As a way to punch through Uptane and get straight to TUF, this does work. However, it implies that there will be separate paths for TUF-only and full Uptane workflows in the API, which seems less than ideal. I'm wondering if we should have a configuration setting for "director-less" mode, in which you would still make the same API calls and the Uptane vs. TUF-only verification would be the only part that is different.

Coincidentally, I'm in the middle of working through delegations, which will also change the way we store and process targets in the images repository. It doesn't really change how this PR would work, but just letting you know.

doanac:
It might be possible to make an uptane-less mode. I didn't look too much into the Uptane code flows, but I was a little worried it might turn into a lot of "if uptane" blocks that might be hard to maintain. However, the Uptane path seems to primarily just want a "Targets". Although that normally comes from Director, I *think* it could be a Targets object I got from TUF and it might just work. At this point, you could probably check "if config.uptane == NOT_CONFIGURED then dont send results to server"

If we go this route, I'd probably like to defer to you guys for implementation, because the complexity would probably lead to me missing edge cases I don't fully understand.

doanac:
@OYTIS - thanks for the input. My original idea was some sort of method that returned an iterator. I'm not sure the most C++ idiomatic way to do this. Would something like a liboost [filter iterator](https://www.boost.org/doc/libs/1_66_0/libs/iterator/doc/html/iterator/specialized/filter.html) be appropriate here? Maybe something like this make sense?
~~~
  typedef boost::filter_iterator<...> TargetFilter

  TargetFilter filterByMachineId(string &machine_id);
~~~

OYTIS:
@doanac 

> I've based by thinking around the Notary API which supports delegation (its API just gives an array of targets back to the caller). And also the fact that the OTA Connect Web App needs to be able to list applicable targets. It feels weird to build this "anonymous" tool w/o being able to provide "list targets" type command.

I believe you're right, but we still want to prevent user from shooting the foot. Could you please point me to the relevant place in the Notary API?

> 
> I have two thoughts:
> 
> 1. Maybe we should table this PR until the delegation implementation is submitted. Then there's something concrete to work against.

Absoulutely. Delegations should be ready soon.

> 2. The nature of what you are describing feels a bit like something that might fit well with a callback function. One place this is used effectively is in GLIB for things like lists. This would allow the implmentation to manage the memory in whatever way it chooses (and change w/o fear), while also allowing consumers to see what's going on.

Yes, callbacks will work as well. I was rather thinking of a [forward iterator](http://www.cplusplus.com/reference/iterator/ForwardIterator/), but let's go for whatever is easier to use of course.

doanac:
> I don't have the exact place they handle delegations, but here's an example "client" that lists targets:

https://github.com/theupdateframework/notary/blob/master/client/example_client_test.go

I don't really know best practices in big C++ code bases anymore, so I'll just trust you guys on things like iterators, callback funcs, etc.

doanac:
I think this can be closed for now. The real fix looks like its going to involve building some kind of iterator out of the new getTargets logic: https://github.com/advancedtelematic/aktualizr/pull/1074/files#diff-55cff2ab63b7a704e8b46b05a052f4b6R126

PR #1060, opened 2019-01-16
Logging implementation revised

riqkum:
Signed-off-by: Maksim Beznos <maximbns@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,6 @@ set(SOTA_PACKED_CREDENTIALS "" CACHE STRING "Credentials.zip for tests involving
 set(TESTSUITE_ONLY "" CACHE STRING "Only run tests matching this list of labels")
 set(TESTSUITE_EXCLUDE "" CACHE STRING "Exclude tests matching this list of labels")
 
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -285,6 +284,11 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
 
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "4.9" OR CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.9")
         add_definitions(-Wshadow)
+        if (CMAKE_SYSTEM_NAME MATCHES "Android")
+            # treat boost headers as system to ignore warnings,
+            # this is alternative to in-code diagnostics disabling
+            add_definitions(--system-header-prefix=boost/)
```
lbonn: It looks like it's a clang-only option whereas the condition above also covers gcc 4.9+.

If we expect Android to always be compiled with clang, the second condition could be out of the first `if` (current version would be misleading).
Otherwise it would be best to have a clang-only check (current version would be incorrect).

PR #1061, opened 2019-01-17
Add delegate functionality for aktualizr-repo

patriotyk:


src/aktualizr_repo/director_repo.cc
```diff
@@ -35,7 +35,7 @@ void DirectorRepo::signTargets() {
   }
 
   Utils::writeFile(path_ / "repo/director/targets.json",
-                   Utils::jsonToCanonicalStr(signTuf(Uptane::Role::Targets(), targets_unsigned)));
+                   Utils::jsonToCanonicalStr(signTuf("targets", targets_unsigned)));
```
patrickvacek: I would prefer not to eliminate the specialized type usage. I've expanded the `Role` enum class to support delegations. In a minute I'll refactor my code to pull out a commit with just that change and I'll put that on a branch that you can then rebase on top of.

patrickvacek: Here's the branch with the single useful commit: https://github.com/advancedtelematic/aktualizr/tree/feat/delegation-roles. If you rebase on that, then you can keep the `Uptane::Role` objects instead of using strings.

src/aktualizr_repo/repo.h
```diff
@@ -14,24 +14,62 @@ struct KeyPair {
...
+    }
+
+    if (pattern.empty()) {
+      throw std::runtime_error("Could not found delegation role in targets.json");
+    }
+  }
+  bool isMatched(const boost::filesystem::path &image_path) {
+    (void)image_path;
+    glob_t globbuf;
+    glob(pattern.c_str(), 0, nullptr, &globbuf);
```
patrickvacek: Three things:
1. What pulls in `glob_t`? I think a header file is missing.
2. I think we were trying to remove globbing from aktualizr because we didn't really need it, it is platform-specific code, and it wasn't available for Android. That said, aktualizr-repo is a host-side tool, so we are probably free to use it here.
3. Why the `(void)image_path` line?

src/aktualizr_repo/repo_test.cc
```diff
@@ -98,6 +98,48 @@ TEST(aktualizr_repo, copy_image) {
...
+  }
+  Uptane::Root root(Uptane::RepositoryType::Image(), Utils::parseJSONFile(temp_dir.Path() / "repo/image/root.json"));
+  auto test_delegate = Utils::parseJSONFile(temp_dir.Path() / "repo/image/test_delegate.json");
+  Uptane::Targets delegate_targets(Uptane::RepositoryType::Image(), test_delegate, root);
+  EXPECT_EQ(delegate_targets.targets.size(), 1);
+  EXPECT_EQ(delegate_targets.targets[0].filename(), "tests/CMakeLists.txt");
+  EXPECT_EQ(delegate_targets.targets[0].length(), 8370);
+  EXPECT_EQ(delegate_targets.targets[0].sha256Hash(),
+            "9cb7c207dfda2cbdc2378c32ee4f77260fadd5f3f2671af2985074f66cde0b5a");
+}
```
patrickvacek: While I applaud using what is already available, we will have to edit this test every time a change is made to `tests/CMakeLists.txt`. Perhaps it would be better to reuse a different test data file we already have in the repo?

patriotyk:
@patrickvacek  About `terminating` as I understand we don't support nesting so this option is not useful. And is it enought to setup `terminating` value when we create delegation? Maybe we should be able to modify it? About `--dname`, I have addressed it, I have added `isBadName` method in repo.h which I call when create delegation and when add image.

patrickvacek:
>  About `terminating` as I understand we don't support nesting so this option is not useful. And is it enought to setup `terminating` value when we create delegation? Maybe we should be able to modify it?

For now, since we aren't initially supporting nested delegations, we should default `terminating` to `true`. Later, when we do support nesting, it should be an option, and we can consider allowing changes to the delegated roles at that time.

> About `--dname`, I have addressed it, I have added `isBadName` method

Sorry, I missed that! Thanks.

src/aktualizr_repo/repo.cc
```diff
@@ -164,7 +164,7 @@ void Repo::generateRepo(KeyType key_type) {
   timestamp["meta"]["snapshot.json"]["length"] = static_cast<Json::UInt>(signed_snapshot.length());
   timestamp["meta"]["snapshot.json"]["version"] = 1;
   Utils::writeFile(repo_dir / "timestamp.json",
-                   Utils::jsonToCanonicalStr(signTuf(Uptane::Role::Snapshot(), timestamp)));
+                   Utils::jsonToCanonicalStr(signTuf(Uptane::Role::Timestamp(), timestamp)));
```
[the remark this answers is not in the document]
patriotyk: We didn't cover it in our tests

src/aktualizr_repo/main.cc
```diff
@@ -68,7 +71,14 @@ int main(int argc, char **argv) {
...
         if (vm.count("filename") > 0) {
-          repo.addImage(vm["filename"].as<std::string>());
+          Delegation delegation;
+          if (vm.count("dname") != 0) {
+            delegation = Delegation(vm["path"].as<boost::filesystem::path>(), vm["dname"].as<std::string>());
+            if (!delegation.isMatched(vm["filename"].as<boost::filesystem::path>())) {
+              throw std::runtime_error("Image path doesn't match delegation!");
+            }
+          }
+          repo.addImage(vm["filename"].as<boost::filesystem::path>(), delegation);
```
patrickvacek: For now, we need to set `terminating` to `true`. That should be the default, but it would also be acceptable to explicitly set it to `true` here.

patrickvacek:
I see another problem. I added a delegation with `--dpattern abc`. Then I tried to add an image with something like `aktualizr-repo image --path temp-repo/ --filename tests/test_data/firmware.txt --dname first`. aktualizr-repo complained, and rightfully so, that the name didn't match. But it strips the path from `--filename` and only tried to match `firmware.txt`. We don't have a way to add images with paths. Perhaps the `image` command needs to also optionally accept `--targetname` as a way to override the parsing of `--filename`. Then I could add `--targetname abc/firmware.txt` and it would work. Make sense?

patrickvacek:
And another issue. I tried adding an image with `--targetname abc/firmware.txt --targetsha256 qwer --targetlength 5` and without `--filename` to circumvent the above issue. However, the delegation logic wasn't applied and it was instead added to `targets.json`.

Also, in case you didn't see, I made a [PR against this PR](https://github.com/advancedtelematic/aktualizr/pull/1065) with some documentation and output improvements. Feel free to merge it in if you don't object to it.

patrickvacek:
Issue 3: I finally got a delegation to work by using a pattern like `temp*`. That's cool. But after adding an image, I see that the delegated_targets.json was signed by the same key that signed `targets.json` instead of the key listed in `target.json`'s `keys` and `roles` sections.

patrickvacek:
Issue 4: Just for fun, I tried the `addtarget` command on the delegated image I'd added. It failed with `No such temp_firmware.txt target in the image repository`. Is `addtarget` not checking for delegated targets, or is this somehow also caused by the signature problem?

patriotyk:
I think should be ok for now

src/aktualizr_repo/repo_test.cc
```diff
@@ -123,6 +123,7 @@ TEST(aktualizr_repo, delegation) {
   EXPECT_EQ(targets["signed"]["delegations"]["roles"][0]["name"].asString(), "test_delegate");
   EXPECT_EQ(targets["signed"]["delegations"]["roles"][0]["paths"][0].asString(), "tests/test_data/*.txt");
 
+  std::cout << "KEYTYPE: " << keytype_stream.str() << "\n";
```
patrickvacek: Did you mean to leave this line in? It looks like this commit was duplicated during a merge or rebase, but this version of it only serves to add this one line. Ideally this bogus commit should probably be eliminated.

patrickvacek:
> when I added a delegation with `--dpattern abc/`, it got turned into `abc/**`. What's up with that?

By the way, I think I was wrong that Uptane requires `abc/` to match anything in a subdirectory of that. I only can find references to wildcard expansion, and our server implementation [also appears to work that way](https://github.com/advancedtelematic/ota-tuf/blob/cc39b7b38c79c084fdd4c3e336a3963ccd8ede12/libtuf/src/test/scala/com/advancedtelematic/libtuf/data/TufClientCodecsSpec.scala#L39). So while I still don't understand why two wildcards were added, it looks like we can actually simplify that logic a bit. I guess directories aren't special after all.

patrickvacek:
It looks like you force-pushed over the last commit I added (`3570a9a0a72facb6fc1dcca6fc84abc5d82d2217`). I can re-add it or put it in the branch I'm working on if that's easier.

patriotyk:
Please readd it.

patrickvacek:
> Please readd it.

Done. But CI is complaining about DCO signoff and clang-tidy static checks.

patrickvacek:
Minor annoyance: `addtarget` requires `--filename` but it is really using `--targetname`. Maybe we should be able to accept either version.

src/aktualizr_repo/repo.cc
```diff
@@ -184,7 +196,7 @@ void Repo::readKeys() {
     KeyType key_type;
     key_type_str >> key_type;
     std::string private_key_string(Utils::readFile(p / "private.key"));
-    keys_[Uptane::Role(p.path().filename().string())] =
+    keys_[Uptane::Role(p.path().filename().string(), true)] =
```
patrickvacek: This is wrong. `delegations` should only be set to `true` if it really is a delegation. I added some logic in my branch to `role.cc` to prevent creating delegations with reserved names, and after rebasing on top of this branch, my check breaks because of this line.

patriotyk:
> Minor annoyance: addtarget requires --filename but it is really using --targetname. Maybe we should be able to accept either version.

I don't see where any refferencies to --targetname:
screenshot_20190128_130453

src/aktualizr_repo/repo.cc
patriotyk: Done

patrickvacek:
> I don't see any refferencies to the --targetname:

Exactly. When you add an image with `--command image`, what gets stored internally is the `--targetname`, not the `--filename`. So when you use `--comand addtarget`, the value you pass to `--filename` is whatever was previously used in `--filename`. That's confusing. For consistency, it really should be `--targetname`.

patrickvacek:
I'm trying to run a generated repo with a delegation through the version of aktualizr I have that has WIP delegation support. Aktualizr complains that in the image repository, the hash for `targets.json` in `snapshot.json` doesn't match the actual hash of `targets.json`. Can you confirm this problem? Could it be that the `snapshot.json` hash isn't correctly updated after adding the delegation?

patrickvacek:
After further investigation, I have confirmed that the hash stated in `snapshot.json` does not match the real hash of `targets.json`. Furthermore, the length stated in `snapshot.json` also does not match the real length of `targets.json`.

patriotyk:
> After further investigation, I have confirmed that the hash stated in snapshot.json does not match the real hash of targets.json. Furthermore, the length stated in snapshot.json also does not match the real length of targets.json.

Added fix. Waiting CI

patrickvacek:
> Added fix.

Thanks. Could we write a test to catch this?

patriotyk:
Covered with tests.

PR #1063, opened 2019-01-18
Fixes for ReportQueue

eu-smirnov:
I've played a bit with a [ThreadSanitizer](https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual). There is a considerable amount of false positives, as ThreadSanitizer requires all libraries to be rebuild, and I didn't do that. But it still was able to find some interesting stuff. 
There are two issues with the current syncronization scheme of ReportQueue:
Member `shutdown_` is modified and checked out of the critical section, which may lead to various race conditions, e.g.:
```
      Run() checks that shutdown is false
      ~ReportQueue() sets shutdown_ to true
      ~ReportQueue() locks mutex and does cv.notify_all()
      Run() locks mutex and starts to wait on the CV
```
`flushQueue()` checks `report_queue_.empty()` out of the critical section and uses a different mutex, not the one than is used with CV. Consider:
```
      flushQueue() checks that report_queue_.empty() is true
      enqueue() locks, pushes and notifies
      flushQueue() starts to to wait on the CV
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,8 +137,8 @@ endif(FAULT_INJECTION)
 #add_definitions(-DBOOST_LOG_DYN_LINK)
 
 if(NOT CMAKE_BUILD_TYPE)
-    message(STATUS "No CMAKE_BUILD_TYPE specified, defaulting to Valgrind")
-    set(CMAKE_BUILD_TYPE Valgrind)
+    set(CMAKE_BUILD_TYPE "Release")
+    message(STATUS "No CMAKE_BUILD_TYPE specified, defaulting to ${CMAKE_BUILD_TYPE}")
 endif(NOT CMAKE_BUILD_TYPE)
 
 set(CMAKE_CXX_FLAGS_DEBUG "-Og -g")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,14 @@ set(CMAKE_CXX_FLAGS_DEBUG "-Og -g")
 SET(CMAKE_CXX_FLAGS_VALGRIND "-O1 -g")
 SET(CMAKE_C_FLAGS_VALGRIND "-O1 -g")
 
+SET(CMAKE_CXX_FLAGS_ASAN "-O1 -g -fsanitize=address -fsanitize=leak -fsanitize=undefined -fno-omit-frame-pointer")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -137,15 +137,22 @@ endif(FAULT_INJECTION)
 #add_definitions(-DBOOST_LOG_DYN_LINK)
 
 if(NOT CMAKE_BUILD_TYPE)
-    message(STATUS "No CMAKE_BUILD_TYPE specified, defaulting to Valgrind")
-    set(CMAKE_BUILD_TYPE Valgrind)
+    message(STATUS "No CMAKE_BUILD_TYPE specified, defaulting to ${CMAKE_BUILD_TYPE}")
```
lbonn: Should we default to something else then? Or just print a warning, possibly with the available build types?

In any case, this message looks funny as `${CMAKE_BUILD_TYPE}` should be empty there.

eu-smirnov: But it's not empty, it's `Release`.

src/libaktualizr/primary/reportqueue.h
```diff
@@ -76,11 +76,10 @@ class ReportQueue {
   std::shared_ptr<HttpInterface> http;
   std::thread thread_;
   std::condition_variable cv_;
-  std::mutex thread_mutex_;
-  std::mutex queue_mutex_;
+  std::mutex m_;
   std::queue<std::unique_ptr<ReportEvent>> report_queue_;
   Json::Value report_array{Json::arrayValue};
-  bool shutdown_;
+  bool shutdown_{false};
```
lbonn: That's fine but I'd suggest removing it from the initializer list of the class' constructor, so that we keep one source of truth

CMakeLists.txt
eu-smirnov: oops, looks like I was doing something wrong, it's empty indeed.

lbonn:
Really cool! So what was the process exactly? Compiling with `-DCMAKE_BUILD_TYPE=Tsan` and running the test suite?

eu-smirnov:
@lbonn Yes, as simple as that. I also did the same with AddressSanitizer, which apparently can find some stuff that Valgrind cannot (see https://github.com/google/sanitizers/wiki/AddressSanitizerComparisonOfMemoryTools), but it didn't find anything. As I understand, together with MemorySanitizer they can replace Valgrind, while being 10 times faster. The only downside is that to avoid false positives one needs to rebuild ALL libraries with the same flags.

PR #1064, opened 2019-01-18
Feat/parametrize faults

lbonn:


patrickvacek:
Seems cool!

lbonn:
Thanks patrick!

I also looked at the `failinfo` parameter supported by libfiu. When injecting faults through the api, it can be any value (void*). I initially thought it couldn't be used through `fiu-run` but it's in fact mentioned in the man pages. However it's poorly documented and by reading the code, it only supports passing unsigned long through the command line (see https://blitiri.com.ar/git/r/libfiu/b/master/t/libfiu/f=fiu-rc.c.html):

    fiu-run -c 'enable name=no1,failinfo=10' ./print1

So, it can be a bit awkward to maintain a list of numeric failure causes but its usage is a bit more elegant.

eu-smirnov:
Not sure, why do we need this fail cause stuff? To me, it looks more like a test for libfiu itself.
And I'm also a bit reluctant about mixing instrumentation and production code. Well, maybe, fake package manager is not exactly production code, but still. The better solution IMO would be to redefine existing functions in a header, something like:
```
int _do_some_stuff(int a);
#ifdef FIU
int do_some_stuff(int a) {
  if (fiu_fail()) {
    return ERR;
  } else {
    return _do_some_stuff(a);
  }
}
#else 
int do_some_stuff(int a) {
  return _do_some_stuff(a);
}
#endif
'''
What do you think?

**Comment by lbonn** [comment]
@eu-smirnov what's the actual concern here?

In production code, `fiu_fail` is defined to 0 at the pre-processor level, so everything inside the condition will be optimized away. Your solution should be basically equivalent except that it introduces more code duplication.

The need right now is having different kinds of installation failures injected into the system through simulated devices.

**Comment by eu-smirnov** [comment]
@lbonn Sure, there will be no difference in the compiled code, and maybe it's fine for this particular case. I'm just thinking that in general, if we define more fail injection points it would be great to have them somehow localized and not just randomly scattered over the source tree. And you are right, that was a duplication, simply having
```
int _do_some_stuff(int a);
int do_some_stuff(int a) {
  if (fiu_fail()) {
    return ERR;
  } else {
    return _do_some_stuff(a);
  }
}
```
would be enough.

**Comment by lbonn** [comment]
We sure have to be careful with these. IMO, in this example, the code inside the `if ()` is on the limit of being an obstacle to readability.

One way to mitigate that could be to have the failure simulation logic in a separate function:

```
if (fiu_fail(...)) {
  simulate_xxx_failure()
}
// regular code
```

which should be marginally better.

Let's exercise good judgement for future use cases.

PR #1065, opened 2019-01-21
Doc/aktualizr repo delegations etc

patrickvacek:
Just trying to clean up some stuff and make things clearer. I had to take some time to understand these things myself, so I might as well record what I found to make it easier for myself and other in the future.

### DIFF
--- a/docs/aktualizr-repo.adoc
+++ b/docs/aktualizr-repo.adoc
@@ -1,10 +1,10 @@
 = Aktualizr repo
 
-aktualizr-repo director contains a basic implementation of Uptane server. It is comprised of three tools:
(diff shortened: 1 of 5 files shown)


docs/aktualizr-repo.adoc
```diff
@@ -1,38 +1,99 @@
...
+2. Add a target to the images metadata (this can be done as many times as necessary):
++
+```
+aktualizr-repo --path <repo path> --command image --filename <image name>
+```
+
+3. Prepare director targets metadata for a given device (this can be done as many times as necessary):
++
+```
+aktaulizr-repo --path <repo path> --command addtarget --filename <image name> --hwid <hardware ID> --serial <ECU serial>
```
lbonn: "aktaulizr-repo" here

PR #1068, opened 2019-01-23
Remove instances of popen_subprocess in prov tests

lbonn:
There is no need to launch `aktualizr once` as a background process before checking for provisioning as it should terminate by itself.

I'm not sure of the original intent but it should hopefully work this way.

### DIFF
--- a/tests/auto_prov_test.py
+++ b/tests/auto_prov_test.py
@@ -7,7 +7,7 @@ import tempfile
 
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


patrickvacek:
> I'm not sure of the original intent but it should hopefully work this way.

Good call! I overlooked this last time I worked on this. I think the tests were written before we had `once` and so we had to wait on aktualizr-info. You are correct to suspect that that is entirely unnecessary now.

PR #1072, opened 2019-01-28
Feat/backward migrations

patriotyk:


src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1375,19 +1372,50 @@ bool SQLStorage::dbMigrate() {
...
-    LOG_ERROR << "Only forward migrations are supported. You cannot migrate to an older schema.";
-    return false;
-  }
-
   SQLite3Guard db = dbConnection();
-  for (int32_t k = schema_num_version + 1; k <= current_schema_version; k++) {
-    if (db.exec(schema_migrations.at(static_cast<size_t>(k)), nullptr, nullptr) != SQLITE_OK) {
-      LOG_ERROR << "Can't migrate db from version " << (k - 1) << " to version " << k << ": " << db.errmsg();
-      return false;
+  if (schema_version != DbVersion::kEmpty && schema_num_version > current_schema_version) {
```
lbonn: I'd suggest making two sub-methods for backward and forward migration, it's a bit heavy on flow-control right now.

config/sql/migration/migrate.15.sql
```diff
@@ -0,0 +1,7 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+BEGIN TRANSACTION;
+
+CREATE TABLE migrations(migration TEXT NOT NULL, version INT NOT NULL);
```
lbonn: `migrations` is not enough self-explanatory IMO, as it's only storing backward migrations which are a special case. `version` could also be `version_from` or `version_to` depending on which is it.

We have to be careful about the naming here because we probably won't be able to rename this table later.

lbonn: Also, the version field should probably be the primary key

config/sql/migration/migrate.15.sql
```diff
@@ -0,0 +1,7 @@
...
+DROP TABLE version;
```
lbonn: Do we really need to stop having the version directly in the db and compute it from the content of `migrations`? It's complicating the logic significantly.

src/libaktualizr/storage/sqlstorage_base.cc
```diff
@@ -61,6 +63,48 @@ std::string SQLStorageBase::getTableSchemaFromDb(const std::string& tablename) {
   return schema.value() + ";";
 }
 
+bool SQLStorageBase::dbMigrateForward(int version_from) {
+  SQLite3Guard db = dbConnection();
+  for (int32_t k = version_from + 1; k <= current_schema_version_; k++) {
+    if (db.exec(schema_migrations_.at(static_cast<size_t>(k)), nullptr, nullptr) != SQLITE_OK) {
```
lbonn: Thoughts:

- for safety, first insert the rollback migrations, then execute the forward migration. This way, there is no way to lose the backward migrations if the migration process was interrupted. Even cleaner would be to use sqlite save points (https://www.sqlite.org/lang_savepoint.html) to nest all this migration process in a big transaction. Not sure if we can still do that and keep compatibility. It might be fine though, need to think a bit more about it
- use `INSERT OR REPLACE`, as we'll probably want to have the more recent versions of these

config/sql/migration/migrate.15.sql
```diff
@@ -0,0 +1,9 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+BEGIN TRANSACTION;
+
+CREATE TABLE migrations(version_from INT PRIMARY KEY, migration TEXT NOT NULL);
```
lbonn: Can we call this table `rollback_migrations` to be more explicit?

src/libaktualizr/storage/sqlstorage_base.cc
```diff
@@ -61,6 +63,48 @@ std::string SQLStorageBase::getTableSchemaFromDb(const std::string& tablename) {
...
+      LOG_ERROR << "Can't extract migration script: " << db.errmsg();
+      return false;
+    }
+    auto migration = statement.get_result_col_str(0);
+    statement.step();
+    if (db.exec(*migration, nullptr, nullptr) != SQLITE_OK) {
+      LOG_ERROR << "Can't migrate db from version " << (ver) << " to version " << ver - 1 << ": " << db.errmsg();
+      return false;
+    }
+    if (db.exec(std::string("DELETE FROM migrations WHERE version_from=") + std::to_string(ver) + ";", nullptr,
```
lbonn: Even though it shouldn't hurt, do you think it's actually needed?

src/libaktualizr/storage/sqlstorage_test.cc
```diff
@@ -139,12 +139,49 @@ TEST(sqlstorage, migrate) {
 
   SQLStorage storage(config, false);
   boost::filesystem::remove_all(config.sqldb_path.get(config.path));
-
   EXPECT_FALSE(dbSchemaCheck(storage));
   EXPECT_TRUE(storage.dbMigrate());
   EXPECT_TRUE(dbSchemaCheck(storage));
 }
 
+/* Migrate backward through SQL schemas. */
```
lbonn: We'll have to update actions.md and changelog.

src/libaktualizr/storage/sqlstorage_test.cc
```diff
@@ -139,12 +139,49 @@ TEST(sqlstorage, migrate) {
...
+    DROP TABLE test_table; \
+    DELETE FROM version;\
+    INSERT INTO version VALUES(" +
+      std::to_string(static_cast<int>(ver)) + ");";
+
+  auto statement =
+      db.prepareStatement("insert into migrations VALUES (?,?);", static_cast<int>(ver) + 1, back_migration_script);
+  statement.step();
+
+  EXPECT_EQ(static_cast<int>(storage.getVersion()), 16);
```
lbonn: 16 is hardcoded here, which will cause breakage. `ver+1`?

lbonn:
Thanks for the update, this is looking good. Sorry for all the nitpick comments, but they should be easy fixes.

src/libaktualizr/storage/embed_schemas.py
```diff
@@ -54,6 +57,18 @@ def apend_migration(migration_path, header_file):
             header_file.write("\",\n")
         apend_migration(os.path.join(migration_folder, migration_list[-1]), header_file)
         header_file.write("\"\n};\n")
+
+        header_file.write("extern const std::map<int, std::string> {}_schema_rollback_migrations = {{".format(prefix))
```
lbonn: We could get away with a `std::vector` where the missing elements are empty strings.
For consistency with the other migration list, also I'm wary of what a static std::map compiles to.

src/libaktualizr/storage/sqlstorage_base.cc
patriotyk: Why not? We don't need this anymore so this data is garbage.

src/libaktualizr/storage/sqlstorage_base.cc
patriotyk: I tried to swap, but it doesn't work. Doesn't work because first we need to migrate to create backward_migrations table, then we can insert data in to it. I also tried  to add save point but PrepareStatement always fails when I add it.

lbonn:
I've added a commit to use save points for forward migration. It changes our existing migrations scripts which could be controversial but I don't think it has a negative impact.

src/libaktualizr/storage/sqlstorage_base.cc
lbonn: Good point for swapping, I missed that detail!

I took the liberty to push a version which should work with save points on top of your PR. Please check if it doesn't break your work.

src/libaktualizr/storage/sqlstorage_base.cc
lbonn: Ok fair enough, I won't fight on this.

patriotyk:
> I've added a commit to use save points for forward migration. It changes our existing migrations scripts which could be controversial but I don't think it has a negative impact.

Thank you, I was trying to accomplish this without migration scripts modification.

patrickvacek:
Looks fine to me after a precursory glance, and after discussion with @lbonn, I think it's fine if both of you are happy. It sounds like Laurent wants to do a manual test first, though.

src/libaktualizr/storage/sqlstorage_base.cc
```diff
@@ -61,6 +63,59 @@ std::string SQLStorageBase::getTableSchemaFromDb(const std::string& tablename) {
...
+      return false;
+    }
+    auto migration = statement.get_result_col_str(0);
+    statement.step();
+    if (db.exec(*migration, nullptr, nullptr) != SQLITE_OK) {
+      LOG_ERROR << "Can't migrate db from version " << (ver) << " to version " << ver - 1 << ": " << db.errmsg();
+      return false;
+    }
+    if (db.exec(std::string("DELETE FROM rollback_migrations WHERE version_from=") + std::to_string(ver) + ";", nullptr,
+                nullptr) != SQLITE_OK) {
```
patrickvacek: `DELETE` probably won't hurt, but shouldn't be necessary if we add with `INSERT OR REPLACE`.

lbonn: Here is the original discussion: https://github.com/advancedtelematic/aktualizr/pull/1072#discussion_r251880431. Now I don't feel too strongly about it.

patrickvacek: Sorry, missed that conversation. And I agree, I don't feel strongly, I just wanted to leave the thought.

lbonn:
Ok I've added the test I wanted in #1041 (hard to do it here in a convincing fashion) and it passes.

So let's merge!

PR #1074, opened 2019-01-29
Basic first-order delegation support

patrickvacek:
Yes, it's really here! However, there are still open questions and more work to be done:

1. Nested delegations are not yet supported. It's probably not that much more work, but that wasn't my goal at this time.
2. This has been tested with aktualizr-repo but not with the real ota-tuf on the server.
3. tuf-test-vectors should really be updated with delegation cases.
4. I'm open to more unit test suggestions.
5. The `delegation_basic` test I wrote is mostly a copy of the `full_no_correlation_id`. Should these be combined?
6. There is some inconsistency in using `Role` objects vs. just their names. This could probably be improved.

Anything else missing or wrong?


config/sql/migration/migrate.15.sql
```diff
@@ -0,0 +1,9 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+BEGIN TRANSACTION;
+
+CREATE TABLE delegations(meta BLOB NOT NULL, role_name TEXT NOT NULL, UNIQUE(role_name));
```
OYTIS: Why not a new type in `meta` table?

src/libaktualizr/storage/invstorage.h
```diff
@@ -131,13 +131,16 @@ class INvStorage {
 
   virtual void storeRoot(const std::string& data, Uptane::RepositoryType repo, Uptane::Version version) = 0;
   virtual bool loadRoot(std::string* data, Uptane::RepositoryType repo, Uptane::Version version) = 0;
-  bool loadLatestRoot(std::string* data, Uptane::RepositoryType repo) {
+  bool loadLatestRoot(std::string* data, const Uptane::RepositoryType repo) {
```
OYTIS: Why? `const` when passing by value is not forbidden, and actually makes some sense, just looks a bit weird to me. I think if we decide it's a good style, it's better apply it consistently (like in the other methods of INvStorage).

config/sql/migration/migrate.15.sql
patrickvacek: I went back and forth on this, but the main problem is that the `meta` table expects a `meta_type` integer, which we don't have for delegations. They can come and go over the lifespan of a device, so assigning them arbitrary numbers seems like unnecessary work. We also don't track versions on non-root metadata in the `meta` table as it is, so it seemed like more waste in the delegation case.

Furthermore, the `meta` table is fairly rigidly structured. We only replace items in it. It seemed like delegations were different enough to warrant a separate table with less stringent rules about the contents.

src/libaktualizr/storage/invstorage.h
lbonn: I'm not a fan of these to be honest, I think the readability cost out-weights the benefits.

patrickvacek: I prefer using `const` when possible to make the intention of immutability clear. The reason why this one stands out is that it is the only function definition in the header file. All of the other definitions in the .cc files have `const`, but clang-tidy complains about using `const` for declarations in header files because it has no effect. The only part where it matters is the definition.

lbonn: The problem is that const for pass-by-value is not a meaningful info for the caller, so it's not relevant in the prototype. It's only useful for the inside of the function, where it's equivalent to a const variable.

But it's bad style (imo) to modify a pass-by-value parameter anyway (in that case, define another intermediary variable), so this const doesn't bring much info.

And when looking at a prototype, the brain usually expects a reference or pointer after a const declaration. It breaks the conventional style of most of C and C++ libraries (including the standard ones).

Plus in this case, it breaks uniformity, so it pops out and slows down readability.

patrickvacek: This case is already an outlier since it's a complete function definition in the header file instead of just a virtual function declaration. However, I appear to be in the minority here, and my opinion is not strong, so I'll undo it. Should I also undo the other `const RepositoryType` instances in the .cc files?

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -513,6 +513,61 @@ void SQLStorage::clearMetadata() {
...
 
+void SQLStorage::storeDelegation(const std::string& data, const Uptane::Role role) {
+  SQLite3Guard db = dbConnection();
+
+  if (!db.beginTransaction()) {
+    LOG_ERROR << "Can't start transaction: " << db.errmsg();
+    return;
+  }
+
+  auto del_statement = db.prepareStatement<std::string>("DELETE FROM delegations WHERE role_name=?;", role.ToString());
```
lbonn: I think it could be replace by a single "INSERT OR REPLACE"

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -513,6 +513,61 @@ void SQLStorage::clearMetadata() {
...
+  }
+
+  db.commitTransaction();
+}
+
+bool SQLStorage::loadDelegation(std::string* data, const Uptane::Role role) {
+  SQLite3Guard db = dbConnection();
+
+  auto statement = db.prepareStatement<std::string>(
+      "SELECT meta FROM delegations WHERE role_name=? ORDER BY role_name DESC LIMIT 1;", role.ToString());
```
lbonn: order by role_name when role_name is fixed?

Also, role_name has a unique constraint, so there should only be one, so no need to `LIMIT 1`.

src/libaktualizr/uptane/imagesrepository.cc
```diff
@@ -102,8 +124,32 @@ bool ImagesRepository::verifyTargets(const std::string& targets_raw) {
 }
 
 std::unique_ptr<Uptane::Target> ImagesRepository::getTarget(const Uptane::Target& director_target) {
-  auto it = std::find(targets.targets.cbegin(), targets.targets.cend(), director_target);
-  if (it == targets.targets.cend()) {
+  // Search for the target in the top-level targets metadata.
+  Uptane::Targets toplevel = targets["targets"];
+  auto it = std::find(toplevel.targets.cbegin(), toplevel.targets.cend(), director_target);
+  if (it == toplevel.targets.cend()) {
```
lbonn: To reduce nesting, I think it could benefit from early exit.

e.g:

    if (it != toplevel.targets.cend()) {
      return std_::make_unique<Uptane::Target>(*it);
    }
    ...

Also:

    if (!fnmatch(pattern ...)) {
      continue;
     }

And:

    if (d_it !=delegate.targets.cend()) {
        return std_::make_unique<Uptane::Target>(*d_it);
    }
    ...

src/libaktualizr/storage/sqlstorage.cc
patrickvacek: Fair point. I copied the style of the `NonRoot` functions. Why don't those do the same?

src/libaktualizr/storage/sqlstorage.cc
OYTIS: I just normally put `LIMIT 1` everywhere to make it clear that we expect exactly one value (the constraints are not visible in the code). Do you think it's a bad style or can have performance implications?

src/libaktualizr/uptane/imagesrepository.cc
patrickvacek: Done. Good call, it looks much better.

src/libaktualizr/storage/invstorage.h
patrickvacek: I undid all of the unnecessary `const` pass-by-value instances.

src/libaktualizr/storage/sqlstorage.cc
lbonn: Yes ok that's a fair point

patrickvacek: Well, I changed it according to @lbonn 's initial comment before I saw @OYTIS 's. Should I restore the `LIMIT 1`? IIRC you can't have that without the `ORDER BY`, though. Alternatively, we could programmatically verify that the statement only returned one row.

src/libaktualizr/storage/sqlstorage.cc
lbonn: `meta` has a `UNIQUE(repo, meta_type, version)` but the delete is `DELETE FROM meta WHERE (repo=? AND meta_type=?);`.

With the current schema, `INSERT OR REPLACE` would allow multiple versions to accumulate while the current `storeNonRoot` only keep the last updated version. To be honest, I'm not sure which would be the preferred semantics. @OYTIS, any input?

src/libaktualizr/uptane/imagesrepository.cc
```diff
@@ -102,12 +124,36 @@ bool ImagesRepository::verifyTargets(const std::string& targets_raw) {
 }
 
 std::unique_ptr<Uptane::Target> ImagesRepository::getTarget(const Uptane::Target& director_target) {
-  auto it = std::find(targets.targets.cbegin(), targets.targets.cend(), director_target);
-  if (it == targets.targets.cend()) {
-    return std::unique_ptr<Uptane::Target>(nullptr);
-  } else {
+  // Search for the target in the top-level targets metadata.
```
doanac: It might be nice (but I can try after this is merged) to turn this logic into an iterator. Then I think the logic of this function would become a std::find call to the new iterator

patrickvacek: Sounds feasible. But just so you know, I'm expecting this particular function to get reworked again when we actually get around to supporting nested delegations. It may also get more complex if/when we support prioritization and/or [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md) (multi-role delegations).

config/sql/migration/migrate.15.sql
```diff
@@ -1,3 +1,4 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
```
lbonn: Thanks!

src/libaktualizr/storage/sqlstorage.cc
patrickvacek: I've restored `LIMIT 1` without the `ORDER BY` bit.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -552,38 +552,76 @@ bool SotaUptaneClient::updateImagesMeta() {
...
 
-    if (images_repo.targetsExpired()) {
+    if (images_repo.targetsExpired("targets")) {
       last_exception = Uptane::ExpiredMetadata("repo", "targets");
       return false;
     }
   }
+
+  // Update Images delegated Targets Metadata
+  for (const std::string &delegation : images_repo.delegations("targets")) {
```
OYTIS: IIRC @heartsucker had an idea that the delegation tree can be huge and we only want to download what is relevant on this particular update. Also one of the drivers to add support for delegations now was the idea to save space/bandwidth on potentially huge `targets.json` for maps by splitting them by region.

So something like an optional (to also support tuf-only scenario) parameter specifying list of targets from director would be useful. Or as an alternative (and probably closer to the spirit of TUF spec) solution, we could only download delegations when intending to download a target or are explicitly asked to.

It may better be left for a next PR though.

patrickvacek: I recall the same conversations and had similar concerns. For the moment, we've "avoided" the problem by only processing first-order delegations. On one hand, I'm not sure how likely it really is that we will have more than a couple layers of delegation in the automotive world, but on the other hand I think it makes sense to be flexible with our design goals and thus careful and efficient in our implementation.

I agree that we probably shouldn't download all delegations and instead only get what we need. However, that would require a fairly substantial refactoring and I wasn't willing to do it for this PR. Perhaps that should be a requirement for the nested delegation support, though.

patrickvacek:
These Travis errors are pretty annoying. It's failing to download an Ubuntu package in Docker. Not related to my PR; every Travis run is currently failing in a similar manner.

By the way, I've gotten as far as testing this PR against the server to the point that aktualizr goes looking for the delegated_targets.json, but it gets a 404.

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: @patrickvacek 
I think these issues somewhat correlate, but we can also run into bandwidth problems with simple non-nested delegations. For the aforementioned example of region-based delegation for maps, these don't necessarily need to be nested.

src/libaktualizr/storage/sqlstorage.cc
OYTIS: @lbonn Yes, the preferred semantics was to only keep the last version. Again `ORDER BY` and `LIMIT` are just for precaution, also known as defensive programming, which is not always considered a good style. We can delete them if you feel like it.

lbonn: I'm not necessarily against defensive programming. Thing is, in this case the schema doesn't protect against multiple version. So should the unique constraint be only `UNIQUE(repo, meta_type)`? I find it weird that we keep this constraint in code only.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Fair point. Sounds like a higher-priority chunk of work than nesting, then. It shouldn't be too hard, but since I'm unlikely to be able to get it done in the next week or so, would it be possible to merge this as it is and then do that as a follow-up? Others are welcome to contribute; I'm just a bit loathe to let this languish for a week while I'm away.

In other news, it sounds like the backend puts delegated_targets.json into a special `delegations/` subdirectory in the images repo.

PR #1075, opened 2019-01-29
Increase scope of pause api

lbonn:
Now succeeds at all time: pauses current download as well as the api queue.

I still need to write some tests


lbonn:
Significant changes since last review.

I also removed DownloadPaused and DownloadResumed events.

PR #1079, opened 2019-01-31
Fix/embed schemas rollback

lbonn:


lbonn:
Travis is still choking with our docker image but this is ready and probably not controversial. There was a missing dependency on rollback scripts for generating sql_schemas.cc.

PR #1080, opened 2019-01-31
Add support for GitLab Integration resolves OTA-1937

kmanikandan-kmk:
Signed-off-by: Kandasamy <manikandan.kandasamy@here.com>


lbonn:
We normally don't require opening a fresh PR to clean things up, but that's also good :).

PR #1083, opened 2019-02-04
Feat/ota 678/delegation revoke tests

patriotyk:
@patrickvacek  Those tests fails because there is some issues in delegation implementation I think.
So there is 3 tests, and 2 of them fails. The problem is when we check updates with delegation then we remove delegation and fetch again and we get an error "Hash verification for targets metadata failed". The problem is the same if we download targets before revoking delegation. But if we install delegated updates before revoke, the everything is ok.

I also tried to reproduce this issue by commenting the first block of code where we run aktualizr with delegation, to check does aktualizr can work with repository with revoked delegations, and it works well. So revoke functionality doesn't break the repo.

### DIFF
--- a/src/aktualizr_repo/image_repo.cc
+++ b/src/aktualizr_repo/image_repo.cc
@@ -73,7 +73,7 @@ void ImageRepo::addDelegation(const Uptane::Role &name, const std::string &path,
   updateRepo();
 }
 
-void ImageRepo::revokeDelegtion(const Uptane::Role &name) {
+void ImageRepo::revokeDelegation(const Uptane::Role &name) {
   if (keys_.count(name) == 0) {
     throw std::runtime_error("Delegation does not exist.");
   }
--- a/src/aktualizr_repo/image_repo.cc
+++ b/src/aktualizr_repo/image_repo.cc
@@ -73,6 +73,43 @@ void ImageRepo::addDelegation(const Uptane::Role &name, const std::string &path,
   updateRepo();
 }
 
+void ImageRepo::revokeDelegtion(const Uptane::Role &name) {
+  if (keys_.count(name) == 0) {
+    throw std::runtime_error("Delegation does not exist.");
+  }
+  if (Uptane::Role::IsReserved(name.ToString())) {
+    throw std::runtime_error("Delegation name " + name.ToString() + " is reserved.");
+  }
+
+  boost::filesystem::path keys_dir = path_ / ("keys/image/" + name.ToString());
+  boost::filesystem::remove_all(keys_dir);
+
+  boost::filesystem::path repo_dir(path_ / "repo/image");
+  boost::filesystem::remove(repo_dir / (name.ToString() + ".json"));
+  Json::Value targets = Utils::parseJSONFile(repo_dir / "targets.json")["signed"];
+  auto keypair = keys_[name];
+  targets["delegations"]["keys"].removeMember(keypair.public_key.KeyId());
+  Json::Value new_roles(Json::arrayValue);
+  for (const auto &role : targets["delegations"]["roles"]) {
+    if (role["name"].asString() != name.ToString()) {
+      new_roles.append(role);
+    }
+  }
+  targets["delegations"]["roles"] = new_roles;
+  Utils::writeFile(repo_dir / "targets.json", Utils::jsonToCanonicalStr(signTuf(Uptane::Role::Targets(), targets)));
+  updateRepo();
+}
+
+std::vector<std::string> ImageRepo::getDelegationTargets(const Uptane::Role &name) {
+  std::vector<std::string> result;
+  boost::filesystem::path repo_dir(path_ / "repo/image");
+  auto targets = Utils::parseJSONFile(repo_dir / (name.ToString() + ".json"))["signed"]["targets"];
+  for (Json::ValueIterator it = targets.begin(); it != targets.end(); ++it) {
+    result.push_back(it.key().asString());
+  }
+  return std::move(result);
+}
+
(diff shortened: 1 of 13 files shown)


src/aktualizr_repo/image_repo.h
```diff
@@ -11,6 +11,8 @@ class ImageRepo : public Repo {
                 const Delegation &delegation = {});
   void addCustomImage(const std::string &name, const Uptane::Hash &hash, uint64_t length, const Delegation &delegation);
   void addDelegation(const Uptane::Role &name, const std::string &path, KeyType key_type, bool terminating);
+  void revokeDelegtion(const Uptane::Role &name);
```
lbonn: There is a "a" missing in the method's name.

patriotyk: Thanks

src/libaktualizr/uptane/CMakeLists.txt
```diff
@@ -64,8 +64,7 @@ add_aktualizr_test(NAME uptane SOURCES uptane_test.cc PROJECT_WORKING_DIRECTORY)
 set_tests_properties(test_uptane PROPERTIES LABELS "crypto")
 
 add_aktualizr_test(NAME uptane_delegation SOURCES uptane_delegation_test.cc PROJECT_WORKING_DIRECTORY
-                   ARGS ${PROJECT_BINARY_DIR}/uptane_repos)
-add_dependencies(t_uptane_delegation uptane_repo_delegation_basic)
```
lbonn: Too bad that we can't generate these through CMake anymore, as it allows an easy inspection. Could we still do it for the basic test without revocation?

In any case, there should be a line `add_dependencies(t_uptane_delegation aktualizr-repo)` if we call the executable directly.

src/aktualizr_repo/image_repo.cc
```diff
@@ -73,6 +73,43 @@ void ImageRepo::addDelegation(const Uptane::Role &name, const std::string &path,
...
+  }
+  if (Uptane::Role::IsReserved(name.ToString())) {
+    throw std::runtime_error("Delegation name " + name.ToString() + " is reserved.");
+  }
+
+  boost::filesystem::path keys_dir = path_ / ("keys/image/" + name.ToString());
+  boost::filesystem::remove_all(keys_dir);
+
+  boost::filesystem::path repo_dir(path_ / "repo/image");
+  boost::filesystem::remove(repo_dir / (name.ToString() + ".json"));
```
lbonn: Should revoke delete the delegated repository? In the classic scenario, revoked repos should be ignored even if still reachable via network, by signing new delegating metadata without it.

lbonn:
The source of the issue seems to be here:

https://github.com/advancedtelematic/aktualizr/blob/595356011bb27a452c942033dfd02b8c461b8ef8/src/libaktualizr/primary/sotauptaneclient.cc#L577

When stepping with gdb, `local_version` and `remote_version` actually have the same value, so nothing is updated. Later, the old images targets repo is used which contains delegations not present anymore.

Is it a problem of aktualizr-repo not updating the version when revoking?

lbonn:
The tests pass when incrementing the version in `DirectorRepo::revokeTargets` and `ImageRepo::revokeDelegation`. I think there is still a dangerous bit in the aktualizr.

The image targets version should also increase when adding a delegation I guess?

src/aktualizr_repo/image_repo.cc
patriotyk: Yeah, but why it's needed? We couldn't restore it anymore, and it will be overwritten if new delegation with the same name is created.

src/libaktualizr/uptane/CMakeLists.txt
patriotyk: With `add_dependencies(t_uptane_delegation aktualizr-repo)` agree. But about `delegation_basic.sh`, I think it is not a problem to run it manually when needed. Otherwise we will have different ways to generate delegations. What do you think?

patriotyk:
@lbonn, you was right, after updating version it wors.

src/libaktualizr/uptane/CMakeLists.txt
lbonn: Ok let's keep that for now, thanks for fixing the dependency

tests/uptane_repo_generation/delegation_basic.sh
```diff
@@ -27,11 +27,15 @@ PRIMARY_FIRMWARE="$IMAGES/primary.txt"
...
 echo "secondary" > "$SECONDARY_FIRMWARE"
-
-akrepo --command generate --expires 2021-07-04T16:33:27Z
-akrepo --command adddelegation --dname new-role --dpattern "abc/*"
-akrepo --command image --filename "$PRIMARY_FIRMWARE" --targetname primary.txt
-akrepo --command image --filename "$SECONDARY_FIRMWARE" --targetname "abc/secondary.txt" --dname new-role
-akrepo --command addtarget --hwid primary_hw --serial CA:FE:A6:D2:84:9D --targetname primary.txt
-akrepo --command addtarget --hwid secondary_hw --serial secondary_ecu_serial --targetname "abc/secondary.txt"
-akrepo --command signtargets
+echo "oooo: $REVOKE"
```
lbonn: debug `echo` still here

src/aktualizr_repo/image_repo.cc
lbonn: I believe tests would be more robust if we actually check that delegations are ignored even if still reachable. But that's fine, we can still change that later if needed.

PR #1084, opened 2019-02-04
Do not add source in testing docker images

lbonn:
Mess with caching as we build images for different PRs all the time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,18 @@ branches:
     - master
     - /^\d\d\d\d\.\d+$/
 env:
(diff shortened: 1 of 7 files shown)


lbonn:
Ended up changing Travis as well, but the idea was to speed up Gitlab and Jenkins docker setup.

First useless work is the `ADD .` we had in our recipes, while we can just use a volume to share code.
Second, gitlab docker caching seemed a bit suspicious. I tried to make reality match with the expected intent:

- if a cache exists for the current branch, then use it
- otherwise, cache from last master build

Then, our caches should stay pretty warm most of the time.

PR #1089, opened 2019-02-08
Refactor delegations to hopefuly be more secure

OYTIS:
Also, this enables nested delegations

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -555,7 +555,10 @@ bool SotaUptaneClient::updateImagesMeta() {
     std::string images_targets;
     Uptane::Role targets_role = Uptane::Role::Targets();
 
-    int64_t targets_size = (images_repo.targetsSize() > 0) ? images_repo.targetsSize() : Uptane::kMaxImagesTargetsSize;
+    auto targets_size = images_repo.getRoleSize(Uptane::Role::Targets());
+    if (targets_size <= 0) {
+      targets_size = Uptane::kMaxImagesTargetsSize;
+    }
     if (!uptane_fetcher->fetchLatestRole(&images_targets, targets_size, Uptane::RepositoryType::Image(),
                                          targets_role)) {
       return false;
@@ -823,29 +826,44 @@ std::unique_ptr<Uptane::Target> SotaUptaneClient::findTargetHelper(const Uptane:
     // Target name matches one of the patterns
 
     std::string delegation_meta;
-    std::shared_ptr<Uptane::Targets> delegation;
+
+    auto version_in_snapshot = images_repo.getRoleVersion(delegate_role);
 
     if (storage->loadDelegation(&delegation_meta, delegate_role)) {
-      delegation = Uptane::ImagesRepository::verifyDelegation(delegation_meta, delegate_role, cur_targets);
-      if (delegation == nullptr) {
-        storage->deleteDelegation(delegate_role);
+      auto version = Uptane::extractVersionUntrusted(delegation_meta);
+
+      if (version > version_in_snapshot) {
+        throw Uptane::SecurityException("images", "Rollback attempt on delegated targets");
+      } else if (version < version_in_snapshot) {
         delegation_meta.clear();
+        storage->deleteDelegation(delegate_role);
       }
     }
 
-    if (delegation == nullptr) {
+    bool delegation_remote = delegation_meta.empty();
+    if (delegation_remote) {
       if (!uptane_fetcher->fetchLatestRole(&delegation_meta, Uptane::kMaxImagesTargetsSize,
                                            Uptane::RepositoryType::Image(), delegate_role)) {
         throw std::runtime_error(std::string("Couldn't fetch ") +
                                  delegate_role.ToString());  // TODO: connectivity exception
       }
+    }
 
-      delegation = Uptane::ImagesRepository::verifyDelegation(delegation_meta, delegate_role, cur_targets);
+    if (!images_repo.verifyRoleHashes(delegation_meta, delegate_role)) {
+      throw Uptane::SecurityException("images", "Delegation hash verification failed");
+    }
 
-      if (delegation == nullptr) {
-        throw Uptane::SecurityException("images", "Delegation verification failed");
+    auto delegation = Uptane::ImagesRepository::verifyDelegation(delegation_meta, delegate_role, cur_targets);
+
+    if (delegation == nullptr) {
+      throw Uptane::SecurityException("images", "Delegation verification failed");
+    }
+
+    if (delegation_remote) {
+      if (delegation->version() != version_in_snapshot) {
+        throw Uptane::SecurityException(
+            "images", std::string("Delegated role version didn't match the entry in Snapshot metadata"));
       }
-
       storage->storeDelegation(delegation_meta, delegate_role);
     }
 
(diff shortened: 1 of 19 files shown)


src/libaktualizr/uptane/imagesrepository.h
```diff
@@ -29,9 +29,13 @@ class ImagesRepository : public RepositoryCommon {
 
   Exception getLastException() const { return last_exception; }
 
+  static std::shared_ptr<Uptane::Targets> verifyDelegation(const std::string& delegation_raw, const Uptane::Role& role,
+                                                           const Targets& parent_target);
+  std::shared_ptr<Uptane::Targets> getTrustedTargets() { return targets; }
+
  private:
   // Map from role name -> metadata ("targets" for top-level):
```
patrickvacek: This comment is no longer relevant.

src/libaktualizr/uptane/imagesrepository.h
```diff
@@ -29,9 +29,13 @@ class ImagesRepository : public RepositoryCommon {
 
   Exception getLastException() const { return last_exception; }
 
+  static std::shared_ptr<Uptane::Targets> verifyDelegation(const std::string& delegation_raw, const Uptane::Role& role,
+                                                           const Targets& parent_target);
+  std::shared_ptr<Uptane::Targets> getTrustedTargets() { return targets; }
```
patrickvacek: "Trusted targets" is an interesting name. Based on the line `auto toplevel_targets = images_repo.getTrustedTargets();` it appears to be just the toplevel targets; is that accurate? Maybe a comment in the header to explain what is meant by "trusted" here would be helpful.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -841,14 +787,106 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets, u
...
+    // Target name matches one of the patterns
+
+    std::string delegation_meta;
+    std::shared_ptr<Uptane::Targets> delegation;
+
+    if (storage->loadDelegation(&delegation_meta, delegate_role)) {
+      delegation = Uptane::ImagesRepository::verifyDelegation(delegation_meta, delegate_role, cur_targets);
+      if (delegation == nullptr) {
+        storage->deleteDelegation(delegate_role);
+        delegation_meta.clear();
```
patrickvacek: Just to make sure I have this straight: we only delete a delegation if verification fails? So if a delegation is removed and later re-added, it wouldn't get deleted after removal, but hopefully verification will fail and then it will get deleted. Right?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -841,14 +787,106 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets, u
...
+      }
+    }
+
+    if (delegation == nullptr) {
+      if (!uptane_fetcher->fetchLatestRole(&delegation_meta, Uptane::kMaxImagesTargetsSize,
+                                           Uptane::RepositoryType::Image(), delegate_role)) {
+        throw std::runtime_error(std::string("Couldn't fetch ") +
+                                 delegate_role.ToString());  // TODO: connectivity exception
```
patrickvacek: Don't we still need to fetch even if we have a stored version in case the version has incremented? Maybe I'm confusing something with versions, but this seems like it won't ever fetch new versions.

OYTIS: Oh wait, you're right. So storing delegations in a persistent storage doesn't make any sense at all, we don't know what's there until we go to the server. That's annoying.

OYTIS: I can propose one of the two procedures then.

OYTIS: Either:

1. At the beginning of update cycle clear all delegations.
2. Cache delegations during the update cycle.
3. Use cached version when it's available, download otherwise.

This way we can avoid downloading the same delegation over for different targets in the same update cycle, but still can't share them between the update cycles.

OYTIS: Or:

1. Store delegations permanently.
2. Use the same procedure that is implemented now on the first pass.
3. If some a target has not been found, update delegations for it on a second pass.

This way we are more bandwith-effective, but there is a potential security problem in not being able to revoke trust to targets in delegated roles in a timely fashion.

patrickvacek: Second option doesn't seem safe enough to me, but the first option is interesting. Does it make since to do any sort of comparison between old and new versions? If we have both available anyway, I wonder if there's anything intelligent we can do with them or if we really just have to ignore the old one and throw it away.

OYTIS: That's what's in docker notary documentation by the way:

> The snapshot key signs the snapshot metadata file, which enumerates the filenames, sizes, and hashes of the root, targets, and delegation metadata files for the collection.

Makes a lot of sense to me and resolves some of the points of confusion.

OYTIS: This way we can cache the delegated meta and only update it if the version in snapshot.json has been incremented. And versioning for delegated targets suddenly makes sense again.

patrickvacek: That's fine, but is that what the backend is doing? I also want to make sure TUF/Uptane are friendly to that notion, but I think they should be. To make sure I understand correctly: this seems like the second option you mentioned, but more robust, because we don't store delegations "permanently", just as long as the version specified in snapshot.json hasn't changed.

OYTIS: Yes. TUF spec says

> The snapshot.json file is signed by the snapshot role. It lists the version numbers of all metadata on the repository, excluding timestamp.json and mirrors.json. For the root role, the hash(es), size, and version number are listed.

which doesn't contradict to what's in Notary docs, just does not name delegations explicitly. I believe that if backend doesn't add delegated roles to snapshot.json yet, it should do so, because otherwise we have weaker security guarantees for delegated targets than for top-level target, and I can't see a reason why it should be the case.

src/libaktualizr/uptane/imagesrepository.cc
```diff
@@ -65,54 +63,55 @@ bool ImagesRepository::verifySnapshot(const std::string& snapshot_raw) {
...
+        break;
+      case Hash::Type::kSha512:
+        if (Hash(Hash::Type::kSha512, boost::algorithm::hex(Crypto::sha512digest(canonical))) != it) {
+          LOG_ERROR << "Hash verification for " << role.ToString() << " metadata failed";
+          return false;
+        }
+        hash_exists = true;
+        break;
+      default:
+        break;
```
patrickvacek: Looks like my comment may have disappeared. Is it possible for the other `verify` functions to reuse `verifyRoleHashes` as well? If it's too complicated, don't worry about it.

OYTIS: `verifyTargets` is already using it. Other roles are not listed in `snapshot.json` (well, `root.json` is, but it makes no sense to verify it).

patrickvacek: I meant specifically the hashing logic in `verifySnapshot`. But after looking more closely, it doesn't isn't worth trying to combine those.

OYTIS:
@patrickvacek Yes, that breaks compatibility with the backend as implemented now. But we have already agreed, it should be fixed there, and no-one uses it yet I believe.

src/libaktualizr/uptane/imagesrepository.cc
OYTIS: @patrickvacek Yes, maybe the logic of iterating over a list of hashes could be abstracted out a bit, but right now it's not too bad I think, it's just two places.

PR #1091, opened 2019-02-11
Store image files on the filesystem

patriotyk:
Signed-off-by: Serhiy Stetskovych <patriotyk@gmail.com>

### DIFF
--- a/src/libaktualizr/storage/sqlstorage.cc
+++ b/src/libaktualizr/storage/sqlstorage.cc
@@ -1190,7 +1190,7 @@ boost::optional<std::pair<size_t, std::string>> SQLStorage::checkTargetFile(cons
     }
     if (((*sha256).empty() || sha256_match) && ((*sha512).empty() || sha512_match)) {
       if (boost::filesystem::exists(images_path_ / *statement.get_result_col_str(3))) {
-        return {std::make_pair(static_cast<size_t>(statement.get_result_col_int(0)), *statement.get_result_col_str(3))};
+        return {{static_cast<size_t>(statement.get_result_col_int(0)), *statement.get_result_col_str(3)}};
       } else {
         return boost::none;
       }
@@ -1242,8 +1242,6 @@ class SQLTargetWHandle : public StorageTargetWHandle {
   ~SQLTargetWHandle() override { SQLTargetWHandle::wcommit(); }
 
   size_t wfeed(const uint8_t* buf, size_t size) override {
-    StorageTargetWHandle::WriteError exc("could not save file " + target_.filename() + " to sql storage");
-
     stream_.write(reinterpret_cast<const char*>(buf), static_cast<std::streamsize>(size));
     written_size_ += size;
 
--- a/src/libaktualizr/storage/sqlstorage.cc
+++ b/src/libaktualizr/storage/sqlstorage.cc
@@ -1160,11 +1160,11 @@ void SQLStorage::clearInstallationResults() {
   db.commitTransaction();
 }
 
-boost::optional<size_t> SQLStorage::checkTargetFile(const Uptane::Target& target) const {
+boost::optional<std::pair<size_t, std::string>> SQLStorage::checkTargetFile(const Uptane::Target& target) const {
   SQLite3Guard db = dbConnection();
 
   auto statement = db.prepareStatement<std::string>(
-      "SELECT real_size, sha256, sha512 FROM target_images WHERE filename = ?;", target.filename());
+      "SELECT real_size, sha256, sha512, filename FROM target_images WHERE targetname = ?;", target.filename());
 
   int statement_state;
   while ((statement_state = statement.step()) == SQLITE_ROW) {
@@ -1189,7 +1189,11 @@ boost::optional<size_t> SQLStorage::checkTargetFile(const Uptane::Target& target
       }
     }
     if (((*sha256).empty() || sha256_match) && ((*sha512).empty() || sha512_match)) {
-      return {static_cast<size_t>(statement.get_result_col_int(0))};
+      if (boost::filesystem::exists(images_path_ / *statement.get_result_col_str(3))) {
+        return {std::make_pair(static_cast<size_t>(statement.get_result_col_int(0)), *statement.get_result_col_str(3))};
+      } else {
+        return boost::none;
+      }
     }
   }
 
@@ -1206,11 +1210,8 @@ boost::optional<size_t> SQLStorage::checkTargetFile(const Uptane::Target& target
 class SQLTargetWHandle : public StorageTargetWHandle {
  public:
   SQLTargetWHandle(const SQLStorage& storage, Uptane::Target target)
-      : db_(storage.dbPath()), target_(std::move(target)), closed_(false), row_id_(0) {
-    StorageTargetWHandle::WriteError exc("could not save file " + target_.filename() + " to sql storage");
-    if (!db_.beginTransaction()) {
-      throw exc;
-    }
+      : db_(storage.dbPath()), target_(std::move(target)) {
+    StorageTargetWHandle::WriteError exc("could not save file " + target_.filename() + " to the filesystem");
 
     std::string sha256Hash;
     std::string sha512Hash;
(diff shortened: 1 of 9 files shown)


src/libaktualizr/storage/sqlstorage_base.cc
```diff
@@ -77,6 +77,42 @@ bool SQLStorageBase::dbMigrateForward(int version_from, int version_to) {
...
+        }
+        unsigned int buff_size = 1024;
+        char buf[buff_size];
+        unsigned int total_red_bytes = 0;
+        unsigned int blob_size = sqlite3_blob_bytes(blob);
+        boost::filesystem::create_directories(images_path_);
+
+        std::ofstream image_file((images_path_ / (*statement.get_result_col_str(1))).string());
+        if (!image_file.good()) {
+          throw std::runtime_error(std::string("Error opening image file for writting"));
```
patrickvacek: typo: "writing".

src/libaktualizr/storage/sqlstorage_base.cc
```diff
@@ -77,6 +77,42 @@ bool SQLStorageBase::dbMigrateForward(int version_from, int version_to) {
   }
 
   for (int32_t k = version_from + 1; k <= version_to; k++) {
+    if (k == 18) {  // Hack, because we can't write to the file from sql scripts.
```
patrickvacek: Unfortunately we'll also need a backward migration hack as well, right?

patriotyk: Unfortunatelly it is the biggest problem. The old aktualizr will not have this code and it is not possible to backmigrate this migration. We should store this migration in the database but we could not do this using sql.

lbonn: I think we should not migrate these files, the primary will download them again if needed anyway (or it should!)

patrickvacek: Fair points. Perhaps we should state that in the code as a comment, though.

eu-smirnov:
It looks to me that we will enter unrecoverable state if the binary file disappears from the filesystem:
 `checkTargetFile()` will return `true` as it finds the database entry, but subsequent `SQLTargetRHandle` constructor will throw.

eu-smirnov:
I'm also not so sure about storing the files under the user-provided filename. As there can be additional directories in the filename, we either risk to end up with hundreds of empty dirs, or need to implement some directory cleanup logic.
What do you think about having just one folder for all binaries and using a hash as a filename?

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1207,25 +1207,20 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
-        SQLZeroBlob{target_.length()});
-
-    if (statement.step() != SQLITE_DONE) {
-      LOG_ERROR << "Statement step failure: " << db_.errmsg();
-      throw exc;
-    }
-
-    row_id_ = sqlite3_last_insert_rowid(db_.get());
-    statement = db_.prepareStatement<std::string, std::string, std::string>(
+    auto statement = db_.prepareStatement<std::string, std::string, std::string>(
```
eu-smirnov: If we handle the absence of a binary file correctly, we also can get rid of `db_.beginTransaction()` and `db_.commitTransaction()` here. This statement can act as an implicit transaction if I understand sqlite correctly (just how it works in `commit()`).

patrickvacek:
>  What do you think about having just one folder for all binaries and using hash as a filename?

I think that is not a bad idea. However, TUF seems to disagree. See [section 5.5.2](https://github.com/theupdateframework/specification/blob/2b4e18472fe25d5b57f36f6fa50104967c8faeaa/tuf-spec.md#5-detailed-workflows). [Uptane](https://uptane.github.io/uptane-standard/uptane-standard.html#download_image) seems to be on the same page.

eu-smirnov:
@patrickvacek 
> the client MUST write the file to non-volatile storage as FILENAME.EXT

Well, if we say that this requirement is about the name in a filesystem, then storing binary blobs in sqlite db is also wrong :)
I guess it's more about to be able to associate FILENAME.EXT with the corresponding binary data, and the rest is implementation details.

patrickvacek:
> Well, if we say that this requirement is about the name in a filesystem, then storing binary blobs in sqlite db is also wrong :)

Indeed!

>  I guess it's more about to be able to associate FILENAME.EXT with the corresponding binary data, and the rest is an implementation detail.

Yes, I agree. I can't think of a good reason why storing the files by their hash would be a bad thing. I can see some benefit in using the names as provided for visibility and maybe maintaining a "nice" directory structure, but that could be as much of a curse as it could be a blessing, and dealing with special characters could be annoying, and no one is really supposed to be looking at the hard drive anyway, so the perceived benefit would be minimal at best.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1234,35 +1229,24 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
   ~SQLTargetWHandle() override {
     if (!closed_) {
+      std::cout << "commiting\n";
       SQLTargetWHandle::wcommit();
     }
+    stream_.close();
   }
 
   size_t wfeed(const uint8_t* buf, size_t size) override {
     StorageTargetWHandle::WriteError exc("could not save file " + target_.filename() + " to sql storage");
```
eu-smirnov: Not used.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1317,81 +1304,56 @@ std::unique_ptr<StorageTargetWHandle> SQLStorage::allocateTargetFile(bool from_d
...
-      size = size_ - read_size_;
-    }
-    if (size == 0) {
-      return 0;
-    }
-
-    if (sqlite3_blob_read(blob_, buf, static_cast<int>(size), static_cast<int>(read_size_)) != SQLITE_OK) {
-      LOG_ERROR << "Could not read from blob: " << db_.errmsg();
-      return 0;
+    auto red_bytes = stream_.readsome(reinterpret_cast<char*>(buf), static_cast<std::streamsize>(size));
```
eu-smirnov: Why not `read()`?

> `readsome()`: 
The behavior of this function is highly implementation-specific. For example, when used with std::ifstream, some library implementations fill the underlying filebuf with data as soon as the file is opened (and readsome() on such implementations reads data, potentially, but not necessarily, the entire file), while other implementations only read from file when an actual input operation is requested (and readsome() issued after file opening never extracts any characters). Likewise, a call to std::cin.readsome() may return all pending unprocessed console input, or may always return zero and extract no characters.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1317,81 +1304,56 @@ std::unique_ptr<StorageTargetWHandle> SQLStorage::allocateTargetFile(bool from_d
...
+    if (!stream_.good()) {
```
eu-smirnov: This check looks superfluous to me, `read()` will throw an exception if something went wrong.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1317,81 +1304,56 @@ std::unique_ptr<StorageTargetWHandle> SQLStorage::allocateTargetFile(bool from_d
...
+      LOG_ERROR << "Could not read from image file";
+      throw StorageTargetRHandle::ReadError("Could not read from image file");
     }
-    read_size_ += size;
-
-    return size;
+    return static_cast<size_t>(red_bytes);
```
eu-smirnov: Should be `bytes_read`, I guess.

patriotyk:
Please check, should be fine now.

src/libaktualizr/storage/sqlstorage.h
```diff
@@ -79,13 +80,14 @@ class SQLStorage : public SQLStorageBase, public INvStorage {
...
   std::unique_ptr<StorageTargetRHandle> openTargetFile(const Uptane::Target& target) override;
-  boost::optional<size_t> checkTargetFile(const Uptane::Target& target) const override;
+  boost::optional<std::pair<size_t, std::string>> checkTargetFile(const Uptane::Target& target) const override;
   void removeTargetFile(const std::string& filename) override;
   void cleanUp() override;
   StorageType type() override { return StorageType::kSqlite; };
 
  private:
   void cleanMetaVersion(Uptane::RepositoryType repo, const Uptane::Role& role);
+  friend SQLTargetRHandle;
```
lbonn: It's already declared line 21. Didn't it work already?

eu-smirnov:
Thanks, Serhiy! Looks fine to me.

lbonn:
I've rebased the PR and added a small tweaks commit for the minor style issues we've found.

PR #1098, opened 2019-02-13
OTA-2135: Optional reboot to apply pending updates

mike-sul:


src/libaktualizr/primary/aktualizr.cc
```diff
@@ -68,6 +68,15 @@ void Aktualizr::UptaneCycle() {
 
   uptane_client_->uptaneInstall(download_result.updates);
   uptane_client_->putManifest();
+  if (uptane_client_->hasPendingUpdates()) {
+    // UptaneCycle() won't do anything useful if there is any pending updates that effectively require either reboot
+    // (ostree) or aktualizr restart (fake pack mngr)
+    LOG_INFO << "About to exit aktualizr since there are pending updates that require either reboot or aktualizr "
+                "restart, no sense to loop Uptane Cycle";
```
patrickvacek: For the record, merely restarting aktualizr won't do anything interesting, it has to be a full reboot. I think logging this a good idea, but you can probably leave out the "no sense to loop Uptane Cycle" bit at the end.

src/libaktualizr/primary/aktualizr.h
```diff
@@ -71,6 +74,15 @@ class Aktualizr {
...
     }
+    if (config_.pacman.force_reboot) {
+      // TODO: put rebooting code in a correct place. I suggest to introduce Aktualizr::Deinitialize as symetric
+      // operations to Initialize() that actually initializes the api thread
+      // This code can go down to the specific package manager so it decides actually what exactly to do, reboot or just
+      // remove the sentinel file as in case of a fake manager
+      LOG_INFO << "About to reboot the system...";
+      sync();
+      reboot(LINUX_REBOOT_CMD_RESTART);
+    }
```
patrickvacek: I think this code actually belongs in the specific package managers, and it should get called through the package manger interface automatically at the end of installation, presumably after the installation report is sent.

mike-sul: I feel like this code should go to a specific package manager too, will do it in the next commit.
In terms of where the code should be called from I am not sure what is the right place. Based on my understanding of the source code/design I think Actualizr dtor is the right place since here we can be sure that other threads are terminated ( like the API thread or the ReportQueue thread ) and don't do anything that should not be terminated by a reboot call (e.g. rebooting in the middle of writing to DB)

src/libaktualizr/primary/aktualizr.cc
mike-sul: OK. (aktualizr restart helps in case of the fake package manager if fake_need_reboot is on and the sentinel file is removed).

src/libaktualizr/bootloader/bootloader.cc
```diff
@@ -157,5 +160,16 @@ void Bootloader::rebootFlagClear() {
   // clear in storage + volatile flag
 
   storage_.clearNeedReboot();
+  // What's the point to remove the sentinel file here if rebootFlagClear is called
```
lbonn: It's an API, so `rebootFlagClear()` could in theory be called just after `rebootFlagSet()`. It could be expressed in term of invariants.

src/libaktualizr/package_manager/debianmanager.cc
```diff
@@ -52,6 +52,10 @@ data::InstallationResult DebianManager::install(const Uptane::Target &target) co
   return data::InstallationResult(data::ResultCode::Numeric::kInstallFailed, output);
 }
 
+void DebianManager:: completeInstall() const {
+    LOG_ERROR << "Complete installation should not happen for the debian package manager as its updates are applied immidiatelly";
```
lbonn: This will probably stay dead code because `hasPendingUpdate()` will always be false for the debian case.

It could be a default implementation in PackageManagerInterface which throws an exception ("unimplemented") as it should be a programming error. Then, the debian case will be taken care of by itself.

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -204,6 +204,13 @@ data::InstallationResult OstreeManager::install(const Uptane::Target &target) co
   return data::InstallationResult(data::ResultCode::Numeric::kNeedCompletion, "Application successful, need reboot");
 }
 
+void OstreeManager::completeInstall() const {
+    if (config.force_reboot) {
+      LOG_INFO << "About to reboot the system in order to aplly pending updates...";
```
lbonn: "apply" instead of "aplly"

src/libaktualizr/package_manager/packagemanagerconfig.h
```diff
@@ -18,6 +18,7 @@ struct PackageConfig {
 
   // Options for simulation (to be used with kNone)
   bool fake_need_reboot{false};
+  bool force_reboot{false};
```
lbonn: Grouped like this, it looks like it's only for simulation (kNone which is the fake package manager). Maybe put it before, just after packages_file and keep this order in updateFromPropertyTree and writeToStream.

src/libaktualizr/package_manager/packagemanagerfake.cc
```diff
@@ -49,6 +49,13 @@ data::InstallationResult PackageManagerFake::install(const Uptane::Target &targe
   return data::InstallationResult(data::ResultCode::Numeric::kOk, "Installing fake package was successful");
 }
 
+void PackageManagerFake::completeInstall() const {
+    if (config.force_reboot) {
+      LOG_INFO << "Emulate the system reboot in case of the fake package manager";
```
lbonn: That's nice that we also have this logic in the fake package manager!

Maybe the log message could be simpler: "Emulating a system reboot"

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -83,7 +83,12 @@ SotaUptaneClient::SotaUptaneClient(Config &config_in, std::shared_ptr<INvStorage
   }
 }
 
-SotaUptaneClient::~SotaUptaneClient() { conn.disconnect(); }
+SotaUptaneClient::~SotaUptaneClient() {
+    conn.disconnect();
```
lbonn: If this is triggered by SotaUptaneClient's destructor, it will happen even when using the api manually.

During our discussion, the consensus was to only trigger it in Aktualizr's "automatic" mode, i.e. from `Aktualizr::RunForever` but it was probably not made very clear. @patrickvacek is that what you understood?

But I see that doing it from this destructor is convenient since it happens after flushing the report queue. Maybe a bit ugly but this code could be moved after `while (!shutdown_) {}` in RunForever and somehow do a manual flush before that?

patrickvacek: Yes, I agree about only doing this in the "automatic" mode, that was also my understanding.

I'd originally thought that we would trigger the reboot in a `postInstall` function in the package managers, but since that has the problem of flushing the queue, perhaps doing it in the Aktualizr class is better. (And I apologize @mike-sul that that is closer to where you originally had it!)

lbonn:
Ah since we forgot last time, it would be nice to update `docs/configuration.adoc`. And possibly the changelog and actions.md (specifies all features and how we test them).

src/libaktualizr/bootloader/bootloader.cc
mike-sul: OK. Will remove the comment.

src/libaktualizr/package_manager/debianmanager.cc
mike-sul: OK. Agree. Just think that assert(0) also could be an option here, but it really depends on an approach to assert usage in the project.

src/libaktualizr/package_manager/ostreemanager.cc
mike-sul: OK.

src/libaktualizr/package_manager/packagemanagerconfig.h
mike-sul: OK.

src/libaktualizr/package_manager/packagemanagerfake.cc
mike-sul: OK. Will make it simpler.

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: Yes, the original consensus was to trigger it at the end of UptaneCycle within Aktualizr::RunForever. 
But, we also realized/agreed that it won't solve the sync issue, i.e. rebooting a device while two other active threads (the API and Report threads) might do some actions interruption of which can cause a problem during the following aktualizr run/start.  e.g. rebooting during writing to the DB or some report event wasn't sent to the backend so the local DB and the backend's DN are not in sync. I am speculating here of course since I haven't learned the code & design very well yet, so you should know better if my statement is correct.
My thought process around the sync issue resolution was the following:

1) Graceful reboot (no any thread & action interruption) can be done only after graceful exit from the three active/looping threads running withing the aktualizr process;

2) the UptainCycle thread can be exited gracefully by setting the shutdown flag hence I added Shutdown call;

3) the API thread is shutdown-ed by the aforementioned Aktualizr::Shutdown() call but the thread is "joined" at the Aktualizr dtor so we cannot reboot before (if we wanna clean/graceful reboot)

4) the Report thread is shutdown-ed and "joined" at the ReportQueue object dtor lifetime of which is owned by SotaUptaneClient object lifetime of which in turn is owned by  Aktualizr object created on a stack of the main thread and being destroyed only when the main thread goes out of the main block.

Therefore, within the current implementation the correct place/moment for the clean/graceful reboot call is at the ReportQueue dtor.

If there is no need in the clean/graceful reboot in this case then we can trigger the reboot from whatever place. If it's important to have the graceful reboot then it should be done at the ReportQueue dtor or somewhere down of a control flow. Also, there is an option to refactor the code slightly and trigger the graceful reboot in more appropriate place.

But, again, I am a new guy here and haven't learned the code very well, so I can be incorrect in my statements :).

mike-sul:
I agree with all you comments and will update the code accordingly. The only things that remains is to agree on a place from which to trigger the reboot.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1134,6 +1129,17 @@ void SotaUptaneClient::campaignAccept(const std::string &campaign_id) {
   report_queue->enqueue(std_::make_unique<CampaignAcceptedReport>(campaign_id));
 }
 
+bool SotaUptaneClient::isInstallCompletionRequired() {
+  bool force_install_completion = (hasPendingUpdates() and config.uptane.force_install_completion);
```
lbonn: We usually avoid these synonym boolean operators as they are quite uncommon in C and C++ contexts. Would prefer `&&` here.

CHANGELOG.md
```diff
@@ -122,3 +122,6 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 ## [2018.2] - 2018-02-16
 
 ## [2018.1] - 2018-02-05
+
+### Added
+- An uptane configuration parameter (force_install_completion) that triggers a system reboot if there is any pending updates that can be applied only after a reboot (the ostree package manager). Emulates a reboot in case of the fake package manager
```
lbonn: You've added this part in the wrong section. It should be at the top of the file in the `[??? (unreleased)]` section.

Also, we keep these fairly high level, so I would drop the part which mentions the fake package manager.

Would tweak it like that:

> A new uptane configuration parameter `force_install_completion` that triggers a system reboot at the end of the installation process for update mechanisms that need one to complete (e.g. OSTree package manager)

actions.md
```diff
@@ -143,7 +143,7 @@ These are the primary actions that a user of libaktualizr can perform through th
       - [x] Set new version to pending status after an OSTree update trigger (aktualizr_test.cc)
       - [x] Send EcuInstallationAppliedReport to server after an OSTree update trigger (aktualizr_test.cc)
       - [x] Uptane check for updates and manifest sends are disabled while an installation is pending reboot (aktualizr_test.cc)
-      - [ ] Trigger a system reboot if `reboot_after_install` is set in the configuration
+      - [ ] Trigger a system reboot if `force_install_completion` is set in the package manager configuration. Emulates a reboot in case of the fake package manager.
```
lbonn: I would put two entries, one which explicitly mentions OSTree, and the other the fake package manager.

It should also be possible to write a test in `aktualizr_test.cc` which uses the fake package manager implementation and then check this box.

lbonn:
See my new (fairly minor) comments. It's looking good but I think it would be nice to have an unit test, probably in `aktualizr_test.cc` which uses the fake package manager.

Would also like to test it with a full yocto built qemu image before merging here, when you'll have some working recipes changes.

CHANGELOG.md
msul: OK. Never used it before.Read this https://keepachangelog.com/en/1.0.0/ and try follow it further.

actions.md
msul: ok

src/libaktualizr/primary/sotauptaneclient.cc
msul: Yeah, I've done too much python coding recently :)

msul:
I've updated PR according to the comment also fixed two bugs in one code line here Bootloader::reboot(), one bug is very stupid (should use ::reboot instead of just reboot) and the second one is subtle related to the reboot() system call...

actions.md
```diff
@@ -143,7 +143,8 @@ These are the primary actions that a user of libaktualizr can perform through th
       - [x] Set new version to pending status after an OSTree update trigger (aktualizr_test.cc)
       - [x] Send EcuInstallationAppliedReport to server after an OSTree update trigger (aktualizr_test.cc)
       - [x] Uptane check for updates and manifest sends are disabled while an installation is pending reboot (aktualizr_test.cc)
-      - [ ] Trigger a system reboot if `reboot_after_install` is set in the configuration
+      - [ ] Trigger a system reboot at the end of the installation process for an ostree-based update mechanisms if `force_install_completion` is set in an uptane configuration
+      - [ ] Exit aktualizr and emulates a reboot in case of the fake package manager if `force_install_completion` is set in an uptane configuration
```
patrickvacek: Just a couple small suggestions here:
* "Trigger a system reboot at the end of an OSTree update if `force_install_completion` is set in the configuration"
* "Emulate a reboot for the fake package manager if `force_install_completion` is set in the configuration (aktualizr_test.cc)"

This is assuming that I understand correctly that the second item is covered by a test now. The same text should be in the comment for the test itself so there is an easy way to grep for it. (That's the best we have for now; hopefully we'll get to spend some time improving that mechanism sooner or later.)

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -1163,6 +1220,7 @@ int main(int argc, char** argv) {
     std::cerr << "Error: " << argv[0] << " requires the path to the base directory of uptane repos.\n";
     return EXIT_FAILURE;
   }
+  std::cout << "MIKESUL:  repos" << argv[1] << std::endl;
```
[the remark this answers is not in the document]
mike-sul: good catch, could'n figured out what the correct path to the repo so I can run & debug a specific test (t_aktualizr)

actions.md
mike-sul: OK. I think, I got your idea and it makes sense.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -441,6 +441,63 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+  conf.bootloader.reboot_sentinel_dir = temp_dir.Path();
+
+  {
+    // first run: do the install, exit UptaneCycle and emulate reboot since force_install_completion is set
+    auto storage = INvStorage::newStorage(conf.storage);
+    Aktualizr aktualizr(conf, storage, http);
+
+    aktualizr.Initialize();
+    auto aktualizr_cycle_thread = aktualizr.RunForever();
+    auto aktualizr_cycle_thread_status = aktualizr_cycle_thread.wait_for(std::chrono::seconds(4));
```
lbonn: So the 4 seconds is a timeout and shouldn't be reached, I think?

Maybe set a larger value so that it won't fail even on loaded CI machines, maybe 20 seconds?

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -441,6 +441,63 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    EXPECT_EQ(update_res.status, result::UpdateStatus::kNoUpdatesAvailable);
+
+    // primary is installed, nothing pending
+    size_t current_target = SIZE_MAX;
+    size_t pending_target = SIZE_MAX;
+    std::vector<Uptane::Target> targets;
+    storage->loadPrimaryInstalledVersions(&targets, &current_target, &pending_target);
+    EXPECT_LT(current_target, targets.size());
+    EXPECT_EQ(pending_target, SIZE_MAX);
+  }
```
lbonn: Could have `EXPECT_EQ(http->manifest_sends, xxx);` for good measure

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -441,6 +441,63 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+ * Verifies an auto reboot and pending updates 'finalization' after a reboot
+ *
+ * Checks actions:
+ *
+ * - [x] Auto reboot triggering just after installation if `force_install_completion` is set
+ *   - [x] Update download and initial installation result storing
+ *   - [x] Exit from Uptane Cycle just after installation
+ *   - [x] Reboot just after installation and exit form the uptane cycle
+ *
+ *  - [x] Pending updates are installed on primary after an auto-rebbot
```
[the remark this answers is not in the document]
lbonn: For 1., I think the duplicated part is quite small so it's fine from right now my point of view. We can still refactor later. Or maybe add a comment referencing the two tests for each others.

actions.md
```diff
@@ -143,7 +143,8 @@ These are the primary actions that a user of libaktualizr can perform through th
       - [x] Set new version to pending status after an OSTree update trigger (aktualizr_test.cc)
       - [x] Send EcuInstallationAppliedReport to server after an OSTree update trigger (aktualizr_test.cc)
       - [x] Uptane check for updates and manifest sends are disabled while an installation is pending reboot (aktualizr_test.cc)
-      - [ ] Trigger a system reboot if `reboot_after_install` is set in the configuration
+      - [ ] Trigger a system reboot at the end of the installation process in case of the ostree package manager usage (OTA-2135)
+      - [x] Emulate a reboot at the end of the installation process in case of the fake package manager usage (OTA-2135, aktualizr_test.cc)
```
patrickvacek: BTW In the past I only added ticket numbers for Requirement tickets. It seems like no one is very interested in those, though, and I'm hesitant to advocate for more Jira bureaucracy, so that hasn't gone anywhere in a while. There's no harm in keeping the ticket number for the task here, but generally you probably don't need to worry about that.

src/libaktualizr/primary/aktualizr_test.cc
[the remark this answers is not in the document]
lbonn: What you were looking for is here I believe: https://github.com/advancedtelematic/aktualizr/blob/a4d3c4a44e5ea41721a1192d9cc6f02a83a03b72/src/libaktualizr/primary/CMakeLists.txt#L16

Or run the test with `-V`: `ctest -V '^test_aktualizr$'.

But in some cases, it's useful to inspect temporary directories. In that case, I use gdb breakpoints to pause it so that they are not deleted yet.

lbonn:
I've tested it with a qemu image, seems to run fine.

Also looking good apart from the few minor comments!
And a rebase on last master would be nice :).

patrickvacek:
The Travis failure seems unrelated; don't worry about that for now. I think the only thing left to do here is address Laurent's comment about `EXPECT_EQ(http->manifest_sends, xxx);`.

mike-sul:
Added `EXPECT_EQ(http->manifest_sends, xxx);` as Laurent suggested. For some reason missed it in the first place.

PR #1099, opened 2019-02-15
Fix two log messsages

lbonn:
- the report in "Report not sent" does not exist anymore
- the putManifest disabled message sounds alarming, move it to DEBUG

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -924,8 +924,7 @@ result::Download SotaUptaneClient::downloadImages(const std::vector<Uptane::Targ
     if (targets.size() == downloaded_targets.size()) {
       result = result::Download(downloaded_targets, result::DownloadStatus::kSuccess, "");
     } else {
-      LOG_ERROR << "Only " << downloaded_targets.size() << " of " << targets.size()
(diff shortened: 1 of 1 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1189,7 +1188,7 @@ bool SotaUptaneClient::putManifestSimple() {
   // does not send event, so it can be used as a subset of other steps
   auto manifest = AssembleManifest();
   if (hasPendingUpdates()) {
-    LOG_INFO << "An update is pending, putManifest is disabled";
+    LOG_DEBUG << "An update is pending, putManifest is disabled";
```
doanac: thanks, that scared me for a while until i read the code :)

lbonn: haha, yes I've got that feedback a lot. Will try to think more about logging as an user interface :).

patrickvacek: This is fine, but is it worth expanding the message more, perhaps? "disabled" sounds off to me; perhaps "An update is pending. Skipping manifest upload until installation is complete." would be better?

lbonn: ok, reworked it again

PR #1100, opened 2019-02-15
Fix/ota 2071/implicit prov docs

patrickvacek:
See also https://github.com/advancedtelematic/ats-garage-docs/pull/76.


patrickvacek:
> Gateway url needs to be provided for HSM implicit provisioning

Done, but it only affects the official docs, so no changes to this PR.

PR #1081, opened 2019-02-02
Add a targets iterator

doanac:
This is related to issue:
  https://github.com/advancedtelematic/aktualizr/issues/1056
    
In order for an application to find the latest target, it will need
access to the list of targets found in the image repository.

I considered using a C++ iterator for this and before the delegation
merge it wasn't too bad. However, doing this as an iterator was
requiring keeping state of 3 iterators (the toplevel, the delegated roles,
and targets for a delegated role) and the resulting code seemed a
bit awkward compared to this.

I'm totally open to alternatives from you guys, but I thought I'd try 
and see if this might be okay. Feel free to just cherry-pick the
unit-test if you want that and nothing else :)

PR #1104, opened 2019-02-19
Add iterator to loop over all targets in delegation tree

OYTIS:
It is not a part of API yet, and probably shouldn't be
part of Uptane API ever. It is possible to introduce TUF
API in future.

Suggested-by: Andy Doan <doanac@beadoan.com>

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -790,6 +790,55 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets, u
   return true;
 }
 
+Uptane::Targets SotaUptaneClient::getTrustedDelegation(const Uptane::Role &delegate_role,
+                                                       const Uptane::Targets &parent_targets,
+                                                       const Uptane::ImagesRepository &images_repo, INvStorage &storage,
+                                                       Uptane::Fetcher &fetcher) {
+  std::string delegation_meta;
+
+  auto version_in_snapshot = images_repo.getRoleVersion(delegate_role);
+
+  if (storage.loadDelegation(&delegation_meta, delegate_role)) {
+    auto version = Uptane::extractVersionUntrusted(delegation_meta);
+
+    if (version > version_in_snapshot) {
+      throw Uptane::SecurityException("images", "Rollback attempt on delegated targets");
+    } else if (version < version_in_snapshot) {
+      delegation_meta.clear();
+      storage.deleteDelegation(delegate_role);
+    }
+  }
+
+  bool delegation_remote = delegation_meta.empty();
+  if (delegation_remote) {
+    if (!fetcher.fetchLatestRole(&delegation_meta, Uptane::kMaxImagesTargetsSize, Uptane::RepositoryType::Image(),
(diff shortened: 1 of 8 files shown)


OYTIS:
Not sure if this solution is better than one proposed by @doanac (https://github.com/advancedtelematic/aktualizr/pull/1081), but it's an iterator, and can be used as such.

The overhead is the necessity to store the whole delegation tree (without the actual contents of delegated targets metadata, just the tree structure), and to stick to secure metadata update procedure, going up and down the tree generated additional and probably not strictly necessary signature checks. Refetching of once verified and stored metadata should never happen.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1431,3 +1440,142 @@ std::string SotaUptaneClient::secondaryTreehubCredentials() const {
     return "";
   }
 }
+
+LazyTargetsList SotaUptaneClient::allTargets() { return LazyTargetsList(images_repo, storage, uptane_fetcher); }
```
doanac: I think a call to `updateImagesMeta();` would be really handy. Otherwise, the only code path to trigger that logic is via Uptane specific logic. Alternatively, you could make updateImagesMeta a public method, but I think just call updateImagesMeta might make things easier.

With that fix, things are working just fine for my aktualizr-lite logic

OYTIS:
I wonder what causes the selfupdate failure on Travis. I can see the same problem in @mike-sul's https://github.com/advancedtelematic/aktualizr/pull/1098 and I wasn't able to reproduce it locally.

src/libaktualizr/primary/sotauptaneclient.cc
OYTIS: @doanac I think for now we'll have to just make it public to keep the interface of `SotaUptaneClient` consistent in that each call does a simple contained thing. What we could do to make it user-friendly is to add TUF API pretty much as `class Aktualizr` represents our Uptane API. That could wrap two calls into one.

OYTIS:
@patrickvacek I believe LazyTargetsList can be a separate module. It would require `SotaUptaneClinet::getTrustedDelegations` to be public though, which doesn't improve encapsulation. Would be cool to have a SotaTufClient class which would take some burden from SotaUptaneClient.

patrickvacek:
@OYTIS Yes, my concerns aren't enough to hold up this PR, but I want to keep the refactor in mind. Maybe we need a generic SotaClient at a lower level and then have UptaneClient and TufClient as a middle ground between the Aktualizr (and AktualizrTuf?) class and SotaClient.

BTW, "SotaUptaneClient" is a bit redundant, right? Uptane is only relevant for SOTA.

OYTIS:
@patrickvacek Just checked the history. Originally there were two clients `SotaHttpClient` for RVI protocol and `SotaUptaneClient` for Uptane. The first couldn't obviously be named `HttpClient`, and the second was named after the first =)

doanac:
@patrickvacek - nope LGTM. thanks guys. I started down the iterator road almost exactly like this PR, and then got lazy and did that callback hack. This is nicer.

OYTIS:
@doanac Thank you! Merging then.

PR #1106, opened 2019-02-20
Update changelog for a new release.

patrickvacek:
I'd like to make a new release for @stevana. Any objections?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,19 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 
 ## [??? (unreleased)]
 
+## [2019.2] - 2019-02-21
+
 ### Added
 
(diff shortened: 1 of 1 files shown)


lbonn:
Should we wait for the auto-reboot to land, or does it answer an urgent concern?

patrickvacek:
> Should we wait for the auto-reboot to land, or does it answer an urgent concern?

Sure, I think we can wait for that. Shouldn't be too long now, anyway!

patrickvacek:
Rebased.

patrickvacek:
Jenkins is finally happy. The Travis error is bogus. Anyone want to take a look while the date in the text is still accurate?

patrickvacek:
Merging since this is docs-only, small, presumably uncontroversial, and the date is embedded in the commit.

PR #1107, opened 2019-02-20
Create new "aktualizr-lite" command

doanac:
This PR is meant to address the functionality discussed in issue #1056. I'm not super happy about the lack of any new test coverage, but I was a little concerned it would require so much mocking that it might not provide much value. However, I'd be more than happy to work on that if you guys think its needed.

I've tested this against two different services:

1) Foundries.io's OTA Connect Deployment
   https://api.foundries.io/lmp/repo/release/api/v1/user_repo/targets.json

2) An "aktualizr-lite" only docker-compose setup
  https://github.com/doanac/ota-compose

### DIFF
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -6,10 +6,7 @@ install(TARGETS aktualizr-lite RUNTIME DESTINATION bin COMPONENT aktualizr-lite)
 
 set(TEST_SOURCES test_lite.sh)
 
-add_custom_target(t_aktualizr-lite)
-add_dependencies(t_aktualizr-lite aktualizr-repo)
-add_dependencies(t_aktualizr-lite aktualizr-lite)
-add_dependencies(build_tests t_aktualizr-lite)
+add_dependencies(build_tests aktualizr-lite)
 
 add_test(test_aktualizr-lite
     ${CMAKE_CURRENT_SOURCE_DIR}/test_lite.sh
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -1,0 +1,9 @@
+if(BUILD_OSTREE)
+add_executable(aktualizr-lite main.cc)
+target_link_libraries(aktualizr-lite aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
+
+install(TARGETS aktualizr-lite RUNTIME DESTINATION bin COMPONENT aktualizr-lite)
+
+endif(BUILD_OSTREE)
+aktualizr_source_file_checks(main.cc)
+# vim: set tabstop=4 shiftwidth=4 expandtab:
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -4,6 +4,22 @@ target_link_libraries(aktualizr-lite aktualizr_static_lib ${AKTUALIZR_EXTERNAL_L
 
 install(TARGETS aktualizr-lite RUNTIME DESTINATION bin COMPONENT aktualizr-lite)
 
+set(TEST_SOURCES test_lite.sh)
+
+add_custom_target(t_aktualizr-lite)
+add_dependencies(t_aktualizr-lite aktualizr-repo)
+add_dependencies(t_aktualizr-lite aktualizr-lite)
+add_dependencies(build_tests t_aktualizr-lite)
+
+add_test(test_aktualizr-lite
+    ${CMAKE_CURRENT_SOURCE_DIR}/test_lite.sh
+        ${CMAKE_BINARY_DIR}/src/aktualizr_lite/aktualizr-lite
+        ${CMAKE_BINARY_DIR}/src/aktualizr_repo/aktualizr-repo
+        ${PROJECT_SOURCE_DIR}/tests
+        ${RUN_VALGRIND}
+)
+
 endif(BUILD_OSTREE)
+
 aktualizr_source_file_checks(main.cc)
 # vim: set tabstop=4 shiftwidth=4 expandtab:
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -49,7 +49,7 @@ static int list_main(Config &config, const bpo::variables_map &unused) {
   return 0;
 }
 
-static std::unique_ptr<Uptane::Target> find_target(std::shared_ptr<SotaUptaneClient> client,
+static std::unique_ptr<Uptane::Target> find_target(const std::shared_ptr<SotaUptaneClient> &client,
                                                    Uptane::HardwareIdentifier &hwid, const std::string &version) {
   std::unique_ptr<Uptane::Target> rv;
   if (!client->updateImagesMeta()) {
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -1,0 +1,123 @@
+#include <iostream>
+
+#include <openssl/ssl.h>
+#include <boost/filesystem.hpp>
+#include <boost/program_options.hpp>
+
+#include "config/config.h"
+#include "package_manager/ostreemanager.h"
+#include "primary/sotauptaneclient.h"
+
+namespace bpo = boost::program_options;
+
+struct SubCommand {
+  const char *name;
+  int (*main)(Config &, const bpo::variables_map &);
+};
+static SubCommand commands[] = {};
+
+void check_info_options(const bpo::options_description &description, const bpo::variables_map &vm) {
+  if (vm.count("help") != 0 || vm.count("command") == 0) {
+    std::cout << description << '\n';
+    exit(EXIT_SUCCESS);
+  }
+  if (vm.count("version") != 0) {
+    std::cout << "Current aktualizr version is: " << AKTUALIZR_VERSION << "\n";
+    exit(EXIT_SUCCESS);
+  }
+}
+
+bpo::variables_map parse_options(int argc, char *argv[]) {
+  std::string subs("Command to execute: ");
+  for (size_t i = 0; i < sizeof(commands) / sizeof(SubCommand); i++) {
+    if (i != 0) {
+      subs += ", ";
+    }
+    subs += commands[i].name;
+  }
+  bpo::options_description description("aktualizr-lite command line options");
+  // clang-format off
+  // Try to keep these options in the same order as Config::updateFromCommandLine().
+  // The first three are commandline only.
+  description.add_options()
+      ("help,h", "print usage")
+      ("version,v", "Current aktualizr version")
+      ("config,c", bpo::value<std::vector<boost::filesystem::path> >()->composing(), "configuration file or directory")
+      ("loglevel", bpo::value<int>(), "set log level 0-5 (trace, debug, info, warning, error, fatal)")
+      ("repo-server", bpo::value<std::string>(), "url of the uptane repo repository")
+      ("ostree-server", bpo::value<std::string>(), "url of the ostree repository")
+      ("primary-ecu-hardware-id", bpo::value<std::string>(), "hardware ID of primary ecu")
+      ("command", bpo::value<std::string>(), subs.c_str());
+  // clang-format on
+
+  // consider the first positional argument as the aktualizr run mode
+  bpo::positional_options_description pos;
+  pos.add("command", 1);
+
+  bpo::variables_map vm;
+  std::vector<std::string> unregistered_options;
+  try {
+    bpo::basic_parsed_options<char> parsed_options =
+        bpo::command_line_parser(argc, argv).options(description).positional(pos).allow_unregistered().run();
+    bpo::store(parsed_options, vm);
+    check_info_options(description, vm);
+    bpo::notify(vm);
+    unregistered_options = bpo::collect_unrecognized(parsed_options.options, bpo::exclude_positional);
+    if (vm.count("help") == 0 && !unregistered_options.empty()) {
+      std::cout << description << "\n";
+      exit(EXIT_FAILURE);
+    }
+  } catch (const bpo::required_option &ex) {
+    // print the error and append the default commandline option description
+    std::cout << ex.what() << std::endl << description;
+    exit(EXIT_FAILURE);
+  } catch (const bpo::error &ex) {
+    check_info_options(description, vm);
+
+    // log boost error
+    LOG_ERROR << "boost command line option error: " << ex.what();
+
+    // print the error message to the standard output too, as the user provided
+    // a non-supported commandline option
+    std::cout << ex.what() << '\n';
+
+    // set the returnValue, thereby ctest will recognize
+    // that something went wrong
+    exit(EXIT_FAILURE);
+  }
+
+  return vm;
+}
+
+int main(int argc, char *argv[]) {
+  logger_init();
+  logger_set_threshold(boost::log::trivial::info);
+
+  bpo::variables_map commandline_map = parse_options(argc, argv);
+
+  int r = -1;
+  try {
+    if (geteuid() != 0) {
+      LOG_WARNING << "\033[31mRunning as non-root and may not work as expected!\033[0m\n";
+    }
+
+    Config config(commandline_map);
+    config.storage.uptane_metadata_path = BasedPath(config.storage.path / "metadata");
+    if (config.logger.loglevel <= boost::log::trivial::debug) {
+      SSL_load_error_strings();
+    }
+    LOG_DEBUG << "Current directory: " << boost::filesystem::current_path().string();
+
+    std::string cmd = commandline_map["command"].as<std::string>();
+    for (size_t i = 0; i < sizeof(commands) / sizeof(SubCommand); i++) {
+      if (cmd == commands[i].name) {
+        return commands[i].main(config, commandline_map);
+      }
+    }
+    throw bpo::invalid_option_value(cmd);
+  } catch (const std::exception &ex) {
+    LOG_ERROR << ex.what();
+    r = -1;
+  }
+  return r;
+}
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -49,8 +49,8 @@ static int list_main(Config &config, const bpo::variables_map &unused) {
   return 0;
 }
 
-static std::unique_ptr<Uptane::Target> latest_version(std::shared_ptr<SotaUptaneClient> client,
-                                                      Uptane::HardwareIdentifier &hwid) {
+static std::unique_ptr<Uptane::Target> find_target(std::shared_ptr<SotaUptaneClient> client,
+                                                   Uptane::HardwareIdentifier &hwid const std::string &version) {
   std::unique_ptr<Uptane::Target> rv;
   if (!client->updateImagesMeta()) {
     LOG_ERROR << "Unable to update latest metadata, using local copy";
@@ -58,21 +58,26 @@ static std::unique_ptr<Uptane::Target> latest_version(std::shared_ptr<SotaUptane
   for (auto &t : client->allTargets()) {
     for (auto const &it : t.hardwareIds()) {
       if (it == hwid) {
-        return std_::make_unique<Uptane::Target>(t);
+        if (version == "latest" || version == t.filename() || version == t.custom_version()) {
+          return std_::make_unique<Uptane::Target>(t);
+        }
       }
     }
   }
   throw std::runtime_error("Unable to find update");
 }
 
-static int update_main(Config &config, const bpo::variables_map &unused) {
-  (void)unused;
+static int update_main(Config &config, const bpo::variables_map &variables_map) {
   auto storage = INvStorage::newStorage(config.storage);
   auto client = SotaUptaneClient::newDefaultClient(config, storage);
   Uptane::HardwareIdentifier hwid(config.provision.primary_ecu_hardware_id);
 
-  LOG_INFO << "Finding latest version to update to...";
-  auto target = latest_version(client, hwid);
+  std::string version("latest");
+  if (variables_map.count("update-name") > 0) {
+    version = variables_map["update-name"].as<std::string>();
+  }
+  LOG_INFO << "Finding " << version << " to update to...";
+  auto target = find_target(client, hwid, version);
   LOG_INFO << "Updating to: " << *target;
 
   std::vector<Uptane::Target> targets{*target};
@@ -135,6 +140,7 @@ bpo::variables_map parse_options(int argc, char *argv[]) {
       ("repo-server", bpo::value<std::string>(), "url of the uptane repo repository")
       ("ostree-server", bpo::value<std::string>(), "url of the ostree repository")
       ("primary-ecu-hardware-id", bpo::value<std::string>(), "hardware ID of primary ecu")
+      ("update-name", bpo::value<std::string>(), "optional name of the update when running \"update\". default=latest")
       ("command", bpo::value<std::string>(), subs.c_str());
   // clang-format on
 
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -10,11 +10,25 @@
 
 namespace bpo = boost::program_options;
 
+static int status_main(Config &config, const bpo::variables_map &unused) {
+  (void)unused;
+  GObjectUniquePtr<OstreeSysroot> sysroot_smart = OstreeManager::LoadSysroot(config.pacman.sysroot);
+  OstreeDeployment *deployment = ostree_sysroot_get_booted_deployment(sysroot_smart.get());
+  if (deployment == nullptr) {
+    LOG_INFO << "No active deployment found";
+  } else {
+    LOG_INFO << "Active image is: " << ostree_deployment_get_csum(deployment);
+  }
+  return 0;
+}
+
 struct SubCommand {
   const char *name;
   int (*main)(Config &, const bpo::variables_map &);
 };
-static SubCommand commands[] = {};
+static SubCommand commands[] = {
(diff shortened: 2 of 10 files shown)


doanac:
I see some code formatting failures. i'll get them fixed

OYTIS:
Hi Andy,

the easiest way to fix the formatting errors is to run `make format`. Beware of outdated `clang-format` versions though. 6.0.0 and 7.0.0 should work just fine.

More importantly: what do you thing the CI strategy for this code might be? I would imagine adding `ota-compose` as a submodule and writing a script to test at least some basic scenario (see, for instance https://github.com/advancedtelematic/aktualizr/blob/master/scripts/test_install_aktualizr_and_update.sh). Could we make it work?

patrickvacek:
It looks like CI is complaining:
```
CMake Error at CMakeLists.txt:419 (message):
  src/aktualizr_lite/main.cc not checked
```
I think this is because we expect every .cc/.h file to get added to our source control checks, which normally is the case, but not if `BUILD_OSTREE` is off. That might be the problem. It might be as easy as moving the `aktualizr_source_file_checks` line outside of the `BUILD_OSTREE` guard.

We were also wondering if there was a way we could test the code in CI. @OYTIS noticed https://github.com/doanac/ota-compose and we are wondering if you might be able to integrate some or all of that as a way to run a test.

doanac:
> the easiest way to fix the formatting errors is to run `make format`

That seems to be doing nothing for me. I've got to poke around and see why.

doanac:
As per the ota-compose idea. I'm not sure that would work well because it requires docker compose. I'm not sure travis ci type stuff is going to want to do docker-in-docker stuff. However, you guys already have the aktualizr-repo stuff in your tests and all this code needs is the TUF repo it makes.

I think there are 2 ways we could test this. I'm not sure which you prefer. Both ways start with populating some data using aktualizr-repo. Then we could:

1) Have a test *script* that calls aktualizr-lite with things like "status" and "list" and greps the STDOUT to see if its operating correctly. I'm not sure if that type approach works well with your CI system.

2) We could make some of the helper functions in my main.cc be non-static and then build a more typical C++ test case that calls the helpers. Maybe we could even mock/replace std::cout so that we could validate what it prints?

patrickvacek:
> I think there are 2 ways we could test this. I'm not sure which you prefer. Both ways start with populating some data using aktualizr-repo.

I agree that generating metadata with aktualizr-repo is a good plan. We have some examples of scripts that do this dynamically for some of the advanced tests (particularly delegations). Either of the two options for evaluating the aktualizr-lite functions is fine by me; I don't have a strong preference.

doanac:
I'll start up on the unit tests today. Could you guys give a cursory glance at the code in the meantime? I'm worried it may have some fundamental issue you don't like and I'm going to wind up writing tests for something that will have to dramatically change.

src/aktualizr_lite/main.cc
```diff
@@ -0,0 +1,211 @@
+#include <iostream>
+
+#include <openssl/ssl.h>
+#include <boost/filesystem.hpp>
+#include <boost/program_options.hpp>
```
patrickvacek: Normally our linter complains about alphabetizing includes within a category, but sometimes it has unexpected preferences that deviate from that. However, I'd expect you'll need to put openssl after boost to make it happy.

src/aktualizr_lite/main.cc
```diff
@@ -0,0 +1,211 @@
...
+      SSL_load_error_strings();
+    }
+    LOG_DEBUG << "Current directory: " << boost::filesystem::current_path().string();
+
+    std::string cmd = commandline_map["command"].as<std::string>();
+    for (size_t i = 0; i < sizeof(commands) / sizeof(SubCommand); i++) {
+      if (cmd == commands[i].name) {
+        return commands[i].main(config, commandline_map);
+      }
+    }
```
patrickvacek: This seems like an obscure way to call the subcommands, and if you expect the number of commands to increase and the input parameters to vary between them, this could get more unwieldy. That said, I see no real problem with it as is, and we can change it in the future if we really want or need to.

src/aktualizr_lite/main.cc
```diff
@@ -0,0 +1,211 @@
...
+      result.status != result::DownloadStatus::kNothingToDownload) {
+    LOG_ERROR << "Unable to download update: " + result.message;
+    return 1;
+  }
+  auto iresult = client->PackageInstall(*target);
+  if (iresult.result_code.num_code != data::ResultCode::Numeric::kOk &&
+      iresult.result_code.num_code != data::ResultCode::Numeric::kNeedCompletion) {
+    LOG_ERROR << "Unable to install update: " << iresult.description;
+    return 1;
+  }
```
patrickvacek: Do you want to say anything about requiring a reboot if the result code is `kNeedCompletion`? We recently changed the code to fix a longstanding issue that an OSTree installation wasn't really complete until the system had restarted. We are now trying to make this much more clear, and you might want to do the same here.

patrickvacek:
Oh, Travis is still complaining about `src/aktualizr_lite/main.cc not checked` when `BUILD_OSTREE` is off as mentioned before.

doanac:
This has been rebased on the latest master. I've added a couple of changes suggested by @patrickvacek . I then attempted a commit to add unit testing. I'm pretty sure the cmake changes for the unit test might not be quite right, but you can run `make t_aktualizr-lite-list` and it seems to do the right thing (builds aktualizr-lite, aktualizr-repo, and runs the test). I'm totally open to suggestions for the testing logic.

src/aktualizr_lite/test_list.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
+set -e
+
+[ -z $DEBUG ] || exec >/tmp/tmp.txt 2>&1
+port="${PORT-9876}"
```
patrickvacek: We have a script for finding an open port: `tests/get_open_port.py`.

src/aktualizr_lite/test_list.sh
```diff
@@ -0,0 +1,85 @@
...
+ostree admin init-fs sysroot
+
+sota_dir=$dest_dir/sota
+mkdir $sota_dir
+cat >$sota_dir/sota.toml <<EOF
+[uptane]
+repo_server = "http://localhost:$port/repo/image"
+
+[provision]
+primary_ecu_hardware_id = "raspberrypi3-64"
```
patrickvacek: I might suggest using an obviously test-only hardware ID just to make it unambiguous that this isn't actually intended to run on an rpi.

doanac:
> /aktualizr/src/aktualizr_repo/main.cc: In function 'int main(int, char**)':
> /aktualizr/src/aktualizr_repo/main.cc:115:39: error: variable 'std::ifstream custom_file' has initializer but incomplete type
>              std::ifstream custom_file(vm["targetcustom"].as<boost::filesystem::

I think this is fixed by adding and "#include <fstream>", but I'm not sure exactly how to re-create this build failure at home?

doanac:
I've got a few new commits to address some things @patrickvacek. Hopefully these make sense. The valgrind testing feels a bit different, but running valrind on test_list.sh seemed wrong. If these changes look okay, I'll squash the commits to make a clean PR.

I'm also going to start looking at trying to make a test for the "update" subcommand.

doanac:
Interesting failure here. Seems like a leak in ostree:
~~~
==2606== Command: /aktualizr/build-test/src/aktualizr_lite/aktualizr-lite --loglevel 1 -c /tmp/tmp.pe06usgsEF/sota/sota.toml update --update-name dummy-os
==2606== 
(process:2606): OSTree-CRITICAL **: 23:13:13.246: ostree_sysroot_query_deployments_for: assertion 'osname != NULL || self->booted_deployment != NULL' failed
127.0.0.1 - - [25/Feb/2019 23:13:16] "GET /repo/image/root.json HTTP/1.1" 200 -
127.0.0.1 - - [25/Feb/2019 23:13:16] "GET /repo/image/timestamp.json HTTP/1.1" 200 -
Scanning dependencies of target t_crypto
[ 46%] Building CXX object src/libaktualizr/crypto/CMakeFiles/t_crypto.dir/crypto_test.cc.o
127.0.0.1 - - [25/Feb/2019 23:13:17] "GET /repo/image/snapshot.json HTTP/1.1" 200 -
127.0.0.1 - - [25/Feb/2019 23:13:17] "GET /repo/image/targets.json HTTP/1.1" 200 -
(process:2606): OSTree-CRITICAL **: 23:13:18.126: ostree_sysroot_get_merge_deployment: assertion 'osname != NULL || self->booted_deployment != NULL' failed
==2606== 
==2606== HEAP SUMMARY:
==2606==     in use at exit: 65,392 bytes in 688 blocks
==2606==   total heap usage: 22,179 allocs, 21,491 frees, 4,552,495 bytes allocated
==2606== 
==2606== 65 bytes in 1 blocks are definitely lost in loss record 463 of 680
==2606==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==2606==    by 0x736FA28: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==2606==    by 0x73893AE: g_strdup (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==2606==    by 0x6AB44E0: _ostree_repo_resolve_rev_internal (in /usr/lib/libostree-1.so.1.0.0)
==2606==    by 0x6AB47CF: ostree_repo_resolve_rev (in /usr/lib/libostree-1.so.1.0.0)
==2606==    by 0x1D043D: OstreeManager::install(Uptane::Target const&) const (ostreemanager.cc:145)
==2606==    by 0x1DDB29: SotaUptaneClient::PackageInstall(Uptane::Target const&) (sotauptaneclient.cc:134)
==2606==    by 0x196F0E: update_main(Config&, boost::program_options::variables_map const&) (main.cc:90)
==2606==    by 0x19A628: main (main.cc:208)
==2606== 
==2606== LEAK SUMMARY:
==2606==    definitely lost: 65 bytes in 1 blocks
...
~~~

Not sure I understand why this one is happening:
~~~
 89/118 Test  #99: t_aktualizr-lite-list ...............................***Failed    0.51 sec
/aktualizr/src/aktualizr_lite/test_list.sh: line 8: /tests/get_open_port.py: No such file or directory
~~~

patrickvacek:
@doanac Nice! This is looking really good. The valgrind bit is perhaps awkward, but it is useful. We do suppress some valgrind reports from libostree, but apparently this is new. @lbonn does that look familiar to you? Hard to say if it is legitimate or not, but it's worth investigating.

For the update, is there anything else you can grep for or look for to verify that the expected version is being updated? Matching that seems worthwhile. What you've already got it quite good, though.

> /tests/get_open_port.py: No such file or directory

Does this need a more complete or at least relative path? The leading `/tests` makes me skeptical that that is correct.

lbonn:
About the valgrind leak

Ok, a huge part of the OstreeManager::Install() wasn't covered by our tests: everything from line 152, which is the nominal case... And we had a leak!

The `char * revision` needs to be freed with `gfree()` I suppose or needs to be tracked by some smart pointer (C++ or glib).

src/aktualizr_lite/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+if(BUILD_OSTREE)
+add_executable(aktualizr-lite main.cc)
+target_link_libraries(aktualizr-lite aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
+
+install(TARGETS aktualizr-lite RUNTIME DESTINATION bin COMPONENT aktualizr-lite)
+
+set(TEST_SOURCES test_list.sh)
+
+if((CMAKE_BUILD_TYPE MATCHES "Valgrind") AND (NOT AKTUALIZR_TEST_NO_VALGRIND))
```
lbonn: The `NO_VALGRIND` is an argument that can be given to `add_aktualizr_test`, for a given test.

It doesn't apply in this case, the condition should just be `if(CMAKE_BUILD_TYPE MATCHES "Valgrind")`

src/aktualizr_lite/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+    add_custom_target(t_aktualizr-lite-list
+        COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_list.sh
+        $<TARGET_FILE:aktualizr-repo> ${PROJECT_SOURCE_DIR})
+endif()
+
+add_dependencies(t_aktualizr-lite-list aktualizr-repo)
+add_dependencies(t_aktualizr-lite-list aktualizr-lite)
+add_dependencies(build_tests t_aktualizr-lite-list)
+
+add_test(NAME t_aktualizr-lite-list
```
lbonn: Did you really intend to keep the `t_aktualizr-lite-list` custom target alongside the `t_aktualizr-lite-list` test run by ctest? The fact that they share the same name is a bit confusing and they run the same command with diferrent arguments.

I would expect that you'd only need two `add_test`, one with and without valgrind?

src/aktualizr_lite/test_list.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
+set -e
+
+build_dir=$(pwd)
```
lbonn: This is a bit brittle. Usually ctest runs tests from the root project directory, so it doesn't look so true.
Because you only use `build_dir` to  access `aktualizr-lite`, it could as well be directly given as an argument like `akrepo_bin`.

lbonn:
Thanks for making our test coverage better :).

Here is a PR to fix the ostree leak: https://github.com/advancedtelematic/aktualizr/pull/1114. I've tested that it fixes it, on the top of your branch.

So we could either cherry-pick it here (and I'll close the PR) or wait for a merge and rebase on top.

doanac:
@lbonn 
> So we could either cherry-pick it here (and I'll close the PR) or wait for a merge and rebase on top.

I've got a few commits I'm going to be squashing to clean things up, so I'll just do a rebase once its merged and build up a clean commit history

doanac:
@patrickvacek 
> For the update, is there anything else you can grep for or look for to verify that the expected version is > being updated? Matching that seems worthwhile. What you've already got it quite good, though.

There's nothing really logged to stdout/stderr to grep for. However, I did discover that by simply setting pacman.os in the sota.toml, that the update actually exits successfully. So I think that's a nice improvement. I'll be pushing that change here later today.

doanac:
@lbonn - i thinkt this https://github.com/advancedtelematic/aktualizr/pull/1107/commits/51dfd4cbd9b1856e78fbe23edaa67b23b0d90666 will address the CMake issues you noted.

doanac:
I think we also have a couple more memory leaks from valgrind now that the full update is completing.
~~~
==14366== 980 (96 direct, 884 indirect) bytes in 1 blocks are definitely lost in loss record 1,040 of 1,053
==14366==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==14366==    by 0x736FA28: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==14366==    by 0x73878E5: g_slice_alloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==14366==    by 0x7387D98: g_slice_alloc0 (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==14366==    by 0x70FE924: g_type_create_instance (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x70DF747: ??? (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x70E0EE4: g_object_new_with_properties (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x70E1960: g_object_new (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x6AD0B3C: ostree_deployment_new (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x6ACC5C9: sysroot_initialize_deployment (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x6ACCCAB: ostree_sysroot_deploy_tree (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x1563BD: OstreeManager::install(Uptane::Target const&) const (ostreemanager.cc:180)
==14366==    by 0x16120C: SotaUptaneClient::PackageInstall(Uptane::Target const&) (sotauptaneclient.cc:134)
==14366==    by 0x12D312: update_main(Config&, boost::program_options::variables_map const&) (main.cc:90)
==14366==    by 0x12F78E: main (main.cc:208)
==14366== 
==14366== 2,950 (96 direct, 2,854 indirect) bytes in 1 blocks are definitely lost in loss record 1,050 of 1,053
==14366==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==14366==    by 0x736FA28: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==14366==    by 0x73878E5: g_slice_alloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==14366==    by 0x7387D98: g_slice_alloc0 (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.3)
==14366==    by 0x70FE924: g_type_create_instance (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x70DF747: ??? (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x70E0EE4: g_object_new_with_properties (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x70E1960: g_object_new (in /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0.5600.3)
==14366==    by 0x6AD0B3C: ostree_deployment_new (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x6ABF5ED: parse_deployment (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x6ABF9A3: list_deployments_process_one_boot_entry (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x6AC04AF: ostree_sysroot_load_if_changed (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x6ABFB05: ostree_sysroot_load (in /usr/lib/libostree-1.so.1.0.0)
==14366==    by 0x14D9F1: OstreeManager::LoadSysroot(boost::filesystem::path const&) (ostreemanager.cc:330)
==14366==    by 0x1550B4: OstreeManager::install(Uptane::Target const&) const (ostreemanager.cc:134)
==14366==    by 0x16120C: SotaUptaneClient::PackageInstall(Uptane::Target const&) (sotauptaneclient.cc:134)
==14366==    by 0x12D312: update_main(Config&, boost::program_options::variables_map const&) (main.cc:90)
==14366==    by 0x12F78E: main (main.cc:208)

~~~

doanac:
one other leak:
~~~
==14901== 32 bytes in 1 blocks are definitely lost in loss record 1,471 of 2,335
==14901==    at 0x4C3017F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==14901==    by 0x21FAE9: _S_make_state<std::thread::_Invoker<std::tuple<std::_Bind<void (ReportQueue::*(ReportQueue*))()> > > > (thread:197)
==14901==    by 0x21FAE9: thread<std::_Bind<void (ReportQueue::*(ReportQueue*))()> > (thread:126)
==14901==    by 0x21FAE9: ReportQueue::ReportQueue(Config const&, std::shared_ptr<HttpInterface>) (reportqueue.cc:9)
==14901==    by 0x16058B: construct<ReportQueue, Config&, std::shared_ptr<HttpClient>&> (new_allocator.h:136)
==14901==    by 0x16058B: construct<ReportQueue, Config&, std::shared_ptr<HttpClient>&> (alloc_traits.h:475)
==14901==    by 0x16058B: _Sp_counted_ptr_inplace<Config&, std::shared_ptr<HttpClient>&> (shared_ptr_base.h:526)
==14901==    by 0x16058B: __shared_count<ReportQueue, std::allocator<ReportQueue>, Config&, std::shared_ptr<HttpClient>&> (shared_ptr_base.h:637)
==14901==    by 0x16058B: __shared_ptr<std::allocator<ReportQueue>, Config&, std::shared_ptr<HttpClient>&> (shared_ptr_base.h:1295)
==14901==    by 0x16058B: shared_ptr<std::allocator<ReportQueue>, Config&, std::shared_ptr<HttpClient>&> (shared_ptr.h:344)
==14901==    by 0x16058B: allocate_shared<ReportQueue, std::allocator<ReportQueue>, Config&, std::shared_ptr<HttpClient>&> (shared_ptr.h:691)
==14901==    by 0x16058B: make_shared<ReportQueue, Config&, std::shared_ptr<HttpClient>&> (shared_ptr.h:707)
==14901==    by 0x16058B: SotaUptaneClient::newDefaultClient(Config&, std::shared_ptr<INvStorage>, std::shared_ptr<boost::signals2::signal<void (std::shared_ptr<event::BaseEvent>), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (std::shared_ptr<event::BaseEvent>)>, boost::function<void (boost::signals2::connection const&, std::shared_ptr<event::BaseEvent>)>, boost::signals2::mutex> >) (sotauptaneclient.cc:32)
==14901==    by 0x12C973: update_main(Config&, boost::program_options::variables_map const&) (main.cc:72)
==14901==    by 0x12F78E: main (main.cc:208)
~~~

Also - i'm still seeing the leak lbonn had fixed when i applied that patch. I'm wondering if running valgrind on ostree might produce the same leak. eg- its a bug there?

patrickvacek:
The `ostree_sysroot_deploy_tree` was real and I've got a fix for it. The others I'm still looking into. Unfortunately it's been a while since we've run that part of the code under Valgrind, so if we can't figure all of these out right now, we may want to turn off Valgrind for your test and work on taking care of the leaks more at a later time. These leaks don't appear to have anything to do with your PR!

Also, while it looks like the issue with the CMake test vs target has been a bit cleared up, now the test gets run as a dependency of the `build_tests` target, which is not ideal. It should instead be a test that I can run through ctest. Since you don't have a separate source code file for this test, `build_tests` should be unaffected.

patrickvacek:
> i'm still seeing the leak lbonn had fixed when i applied that patch.

I also can't reproduce the leak that Laurent fixed anymore. I'm still seeing other leaks in OstreeManager, but the issue with `revision` appears to be fixed, although your branch doesn't include it.

I'm still looking into the other issues. The `g_file_new_for_path` one looks like a false positive, though.

doanac:
@patrickvacek - thanks for the cmake point. I was actually thinking about this last night and was worried it was wrong. I think I've finally got it done correctly with:
~~~
iadd_custom_target(t_aktualizr-lite)
add_dependencies(t_aktualizr-lite aktualizr-repo)
add_dependencies(t_aktualizr-lite aktualizr-lite)
add_dependencies(build_tests t_aktualizr-lite)

add_test(test_aktualizr-lite
    ${CMAKE_CURRENT_SOURCE_DIR}/test_list.sh
        ${CMAKE_BINARY_DIR}/src/aktualizr_lite/aktualizr-lite
        ${CMAKE_BINARY_DIR}/src/aktualizr_repo/aktualizr-repo
        ${PROJECT_SOURCE_DIR}/tests
        ${RUN_VALGRIND}
)
~~~

NOTE: I was originally copying this logic: https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr/package_manager/CMakeLists.txt#L33 which i now think might be wrong as well.

I'm going to clean this branch up in a moment and rebase it on the current master. And then we can see the results.

Disabling the valgrind logic would be simple, but i'll keep it for now.

patrickvacek:
@doanac Just finished my investigation: https://github.com/advancedtelematic/aktualizr/pull/1120. Hopefully if you use that you will not see any more leaks.

src/aktualizr_lite/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+target_link_libraries(aktualizr-lite aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
+
+install(TARGETS aktualizr-lite RUNTIME DESTINATION bin COMPONENT aktualizr-lite)
+
+set(TEST_SOURCES test_lite.sh)
+
+add_custom_target(t_aktualizr-lite)
+add_dependencies(t_aktualizr-lite aktualizr-repo)
+add_dependencies(t_aktualizr-lite aktualizr-lite)
+add_dependencies(build_tests t_aktualizr-lite)
```
patrickvacek: This does appear to work now. However, I'm not sure that you need a custom target at all. You should just need `add_dependencies(build_tests aktualizr-lite)`.

patrickvacek:
Travis is complaining about a lack of const:
```
/aktualizr/src/aktualizr_lite/main.cc:52:86: error: the parameter 'client' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
static std::unique_ptr<Uptane::Target> find_target(std::shared_ptr<SotaUptaneClient> client,
                                                   ~~~                               ^
                                                   const                            &
```

src/aktualizr_lite/CMakeLists.txt
doanac: Good point. I was trying to make it really easy to just run "make t_aktualizr-lite" and generate just what you need for unit testing. However, "make build_tests" is fine. Fixed now.

patrickvacek:
Looks pretty good to me! @lbonn @eu-smirnov any other thoughts from your side?

PR #1108, opened 2019-02-22
Dockerfile: Include clang-format

doanac:
"make [check-]format" does not work in this project's default container
because clang-format isn't installed. CMake logs a message about this,
but its hard to notice.


doanac:
I figured out why "make check-format" returned no errors for me as noted here:
  https://github.com/advancedtelematic/aktualizr/pull/1107#issuecomment-465912474

PR #1109, opened 2019-02-22
OTA-1941: Tests: Hostname as hwd ID, Download progress reporting

mike-sul:

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -44,7 +44,7 @@ int num_events_FullNoUpdates = 0;
 std::future<void> future_FullNoUpdates{};
 std::promise<void> promise_FullNoUpdates{};
 void process_events_FullNoUpdates(const std::shared_ptr<event::BaseEvent>& event) {
-  if (event->variant == "DownloadProgressReport") {
+  if (event->isTypeOf(event::DownloadProgressReport::TypeName)) {
     return;
   }
   LOG_INFO << "Got " << event->variant;
@@ -183,7 +183,7 @@ int num_events_FullWithUpdates = 0;
 std::future<void> future_FullWithUpdates{};
 std::promise<void> promise_FullWithUpdates{};
 void process_events_FullWithUpdates(const std::shared_ptr<event::BaseEvent>& event) {
-  if (event->variant == "DownloadProgressReport") {
+  if (event->isTypeOf(event::DownloadProgressReport::TypeName)) {
     return;
   }
   LOG_INFO << "Got " << event->variant;
@@ -581,7 +581,7 @@ int num_events_CheckNoUpdates = 0;
 std::future<void> future_CheckNoUpdates{};
 std::promise<void> promise_CheckNoUpdates{};
 void process_events_CheckNoUpdates(const std::shared_ptr<event::BaseEvent>& event) {
-  if (event->variant == "DownloadProgressReport") {
+  if (event->isTypeOf(event::DownloadProgressReport::TypeName)) {
     return;
   }
   LOG_INFO << "Got " << event->variant;
@@ -651,7 +651,7 @@ int num_events_DownloadWithUpdates = 0;
 std::future<void> future_DownloadWithUpdates{};
 std::promise<void> promise_DownloadWithUpdates{};
 void process_events_DownloadWithUpdates(const std::shared_ptr<event::BaseEvent>& event) {
-  if (event->variant == "DownloadProgressReport") {
+  if (event->isTypeOf(event::DownloadProgressReport::TypeName)) {
     return;
   }
   LOG_INFO << "Got " << event->variant;
@@ -744,7 +744,7 @@ std::promise<void> promise_InstallWithUpdates{};
 void process_events_InstallWithUpdates(const std::shared_ptr<event::BaseEvent>& event) {
   // Note that we do not expect a PutManifestComplete since we don't call
   // UptaneCycle() and that's the only function that generates that.
-  if (event->variant == "DownloadProgressReport") {
+  if (event->isTypeOf(event::DownloadProgressReport::TypeName)) {
     return;
   }
   LOG_INFO << "Got " << event->variant;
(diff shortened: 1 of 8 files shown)


patrickvacek:
Not sure why Jenkins failed. I couldn't find the problem, so I restarted it.

src/libaktualizr/uptane/uptane_init_test.cc
```diff
@@ -261,6 +261,44 @@ TEST(Uptane, InitializeFail) {
...
+ */
+TEST(Uptane, HostnameAsHardwareID) {
+  TemporaryDirectory temp_dir;
+  Config conf("tests/config/basic.toml");
+  conf.storage.path = temp_dir.Path();
+
+  boost::filesystem::copy_file("tests/test_data/cred.zip", temp_dir.Path() / "cred.zip");
+  conf.provision.provision_path = temp_dir.Path() / "cred.zip";
+
+  {
```
patrickvacek: We should explicitly check after constructing the config but before initialization that the config's `provision.primary_ecu_hardware_id` field is empty.

src/libaktualizr/uptane/uptane_init_test.cc
```diff
@@ -261,6 +261,44 @@ TEST(Uptane, InitializeFail) {
...
+
+    EcuSerials ecu_serials;
+    EXPECT_TRUE(storage->loadEcuSerials(&ecu_serials));
+    EXPECT_GE(ecu_serials.size(), 1);
+
+    // A second element of the first tuple in ECU Serials tuple array is a primary hardware ID.
+    // Each client of the storage class needs to know this information.
+    // If it changes then corresponding changes should be done in each storage client.
+    // perhaps it makes sense to introduce get/setPrimaryHardwareID method and incapsulate
+    // this tech info within storage (or maybe some other entity)
```
patrickvacek: I grant that this isn't well-documented, but it is guaranteed in the current implementation that the first entry returned from `loadEcuSerials` is the primary. It's our convention, but also ensured by the SQL command: `SELECT serial, hardware_id FROM ecu_serials ORDER BY is_primary DESC;`. I'm not opposed to a shortcut getter as a convenience function for getting the primary hardware ID, but a setter seems unnecessary for now.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -895,6 +895,53 @@ TEST(Aktualizr, InstallWithUpdates) {
...
+/**
+ * Verifies reporting of update download progress
+ *
+ * Checks actions:
+ *
+ * - [x] Report download progress
+ */
+TEST(Aktualizr, ReportDownloadProgress) {
+  // The test initialization part is repeated in many tests so maybe it makes sense
+  // to define a fixture and move this common initialization part into the fixture's SetUp() method
```
patrickvacek: Yes, I've considered that but never got around to it for some reason. I did unify a lot of the configuration into `makeTestConfig`, and we've done a lot with `HttpFake` reuse, but that's as far as we've gotten for now.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -895,6 +895,53 @@ TEST(Aktualizr, InstallWithUpdates) {
...
+  TemporaryDirectory temp_dir;
+  auto http = std::make_shared<HttpFake>(temp_dir.Path(), "hasupdates");
+  Config conf = UptaneTestCommon::makeTestConfig(temp_dir, http->tls_server);
+  auto storage = INvStorage::newStorage(conf.storage);
+  Aktualizr aktualizr(conf, storage, http);
+
+  /* perhaps should be static const member of DownloadProgressReport */
```
patrickvacek: Static? That would mean that all instances of the class would have the same progress number, which is not desirable.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -895,6 +895,53 @@ TEST(Aktualizr, InstallWithUpdates) {
...
+      [&](const std::shared_ptr<event::BaseEvent>& event) {
+        /* perhaps should be static const member of DownloadProgressReport */
+        if (event->variant != "DownloadProgressReport") {
+          return;
+        }
+
+        auto download_progress_event = std::dynamic_pointer_cast<event::DownloadProgressReport>(event);
+        ASSERT_NE(download_progress_event, nullptr);
+        ASSERT_NE(download_progress_event.get(), nullptr);
+        EXPECT_GT(download_progress_event->progress, download_progress);
```
patrickvacek: Should probably be `EXPECT_GE` in case the percentage hasn't actually increased.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -895,6 +895,53 @@ TEST(Aktualizr, InstallWithUpdates) {
...
+
+  result::UpdateCheck update_result = aktualizr.CheckUpdates().get();
+  ASSERT_EQ(update_result.status, result::UpdateStatus::kUpdatesAvailable);
+  // The test mocks are tailored to emulate a device with a primary ECU and one secondary ECU
+  // for sake of the download progress report testing it's suffice to test it agains just one of the ECUs
+  update_result.updates.pop_back();
+
+  result::Download download_result = aktualizr.Download(update_result.updates).get();
+  EXPECT_EQ(download_result.status, result::DownloadStatus::kSuccess);
+  EXPECT_EQ(download_progress, download_complete_progress_value);
```
patrickvacek: I think there is one thing missing here: a guarantee that at least two progress reports have actually been seen. As currently written, this test will pass even if none are actually generated, which is what will probably happen if the download is small and fast.

src/libaktualizr/uptane/uptane_init_test.cc
mike-sul: Agree

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: Yes, it's a number which represents that download was successful, so clients know criterion of download success.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: ok

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: No, the test will not pass if none progress reports are generated since download_progress will be equal to zero while download_complete_progress_value is equal to 100.

mike-sul: Although, it does not guarantee that at least two reports are received, since just one report with 100 progress value can be sent. The question is if it's an error case ?

src/libaktualizr/uptane/uptane_init_test.cc
mike-sul: ok

src/libaktualizr/primary/aktualizr_test.cc
patrickvacek: Oh sorry, I misunderstood what you meant! Yes, you are probably correct. Feel free to fix it properly.

src/libaktualizr/primary/aktualizr_test.cc
patrickvacek: That's not an error, but it doesn't prove that we are actually sending progress reports if we only get one and it says 100%. We have to see at least two to be able to demonstrate actual progress.

PR #1111, opened 2019-02-25
Don't send manifest & check for updates if there are pending updates

mike-sul:

### DIFF
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -67,12 +67,18 @@ void Aktualizr::UptaneCycle() {
   }
 
   uptane_client_->uptaneInstall(download_result.updates);
-  uptane_client_->putManifest();
+
   if (uptane_client_->isInstallCompletionRequired()) {
     // If there are some pending updates then effectively either reboot (ostree) or aktualizr restart (fake pack mngr)
     // is required to apply the update(s)
     LOG_INFO << "About to exit aktualizr so the pending updates can be applied after reboot";
     Shutdown();
+  }
+
+  if (!uptane_client_->hasPendingUpdates()) {
+    // If updates were applied and no any reboot/finalization is required then send/put manifest
+    // as soon as possible, don't wait for config_.uptane.polling_sec
+    uptane_client_->putManifest();
   }
 }
 
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -324,7 +324,6 @@ void SotaUptaneClient::initialize() {
 }
 
 bool SotaUptaneClient::updateMeta() {
-  reportNetworkInfo();
   // Uptane step 1 (build the vehicle version manifest):
   if (!putManifestSimple()) {
     LOG_ERROR << "could not put manifest";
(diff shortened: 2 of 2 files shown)


src/libaktualizr/primary/aktualizr.cc
```diff
@@ -67,13 +67,18 @@ void Aktualizr::UptaneCycle() {
...
+
   if (uptane_client_->isInstallCompletionRequired()) {
     // If there are some pending updates then effectively either reboot (ostree) or aktualizr restart (fake pack mngr)
     // is required to apply the update(s)
     LOG_INFO << "About to exit aktualizr so the pending updates can be applied after reboot";
     Shutdown();
   }
+
+  if (!uptane_client_->hasPendingUpdates()) {
+    // Why do we need to call putManifest here if it's called implicitly/indirectly at CheckUpdates().get() ?
```
patrickvacek: This is something of a legacy concept. We used to make a more explicit distinction between sending the manifest and checking for updates. Maybe we'll want to re-emphasize that distinction again some day. However, it is still relevant here when running in "automatic" mode with `UptaneCycle`, particularly if `config_.uptane.polling_sec` is set to a large value. Sending the manifest is necessary to tell the server that we've completed an update, and if it's going to be many minutes or hours before we next check for updates, we need to explicitly send the manifest here. Of course, with the recent reboot logic changes, maybe we should reconsider how we do that.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1061,12 +1060,24 @@ void SotaUptaneClient::sendDeviceData() {
...
+
+  if (hasPendingUpdates()) {
+    // no need in update checking if there are some pending updates
+    return result::UpdateCheck({}, 0, result::UpdateStatus::kError, Json::nullValue,
+                               "There are pending updates, no new updates are checked");
+  }
+
   if (updateMeta()) {
+    // is updateMeta() part of the update checking procedure ?
+    // if no then the sendEvent should be at the end of this block
```
patrickvacek: Yes, `updateMeta` is an essential part of the update checking process. In fact, arguably `fetchMeta` is a bit of a misnomer. The issue is another legacy concept: we used to distinguish between fetching the metadata and then actually checking it for updates. Now we generally do it all at once. I think the program flow you have here is probably exactly what is desired.

src/libaktualizr/primary/aktualizr.cc
mike-sul: If there is no any update the control flow never reaches this code.
If there is update(s) but it's pending and waiting for a reboot we don't wont to send any manifest anyway.
Just after reboot the manifest is sent just after update is completed from finalizeAfterReboot() at the initialization state.
The only case when it makes sense is when an update installation doesn't require any reboot (no pending updates) .

patrickvacek: That matches my understanding. Note that this may be more common than expected because of updates to secondaries that will not require rebooting the primary.

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: ok

src/libaktualizr/primary/aktualizr.cc
mike-sul: ok

mike-sul:
Just updated PR, removed the comment-question and update another comment.

patrickvacek:
Confirmed that this fixes the issue by testing with qemu.

PR #1119, opened 2019-02-28
Test/unit with ptest

lbonn:
Some changes to make our test easier to run with ptest. I think everything is here but I might push some more bits.

Also see the forthcoming PR on meta-updater.


src/aktualizr_secondary/CMakeLists.txt
```diff
@@ -112,40 +109,6 @@ add_test(NAME aktualizr_secondary_no_config_check_message
...
-add_test(NAME aktualizr_secondary_log_invalid
-         COMMAND aktualizr-secondary -c ${PROJECT_SOURCE_DIR}/tests/config/aktualizr_secondary.toml --loglevel=6)
-set_tests_properties(aktualizr_secondary_log_invalid
-                     PROPERTIES PASS_REGULAR_EXPRESSION "Invalid log level")
-add_test(NAME aktualizr_secondary_log_negative
-         COMMAND aktualizr-secondary -c ${PROJECT_SOURCE_DIR}/tests/config/aktualizr_secondary.toml --loglevel=-1)
-set_tests_properties(aktualizr_secondary_log_negative
-                     PROPERTIES PASS_REGULAR_EXPRESSION "Invalid log level")
-
-
```
patrickvacek: I think it might be premature to remove all of these tests. Can we just mark them as `noptest` for now?

If we do remove them, we should also remove the checkmarks for these items in `actions.md`. I know that code is all a bit of a mess, but I'm not sure I'm ready to axe it all.

src/libaktualizr/package_manager/ostreemanager_test.cc
```diff
@@ -212,7 +212,15 @@ int main(int argc, char **argv) {
     std::cerr << "Error: " << argv[0] << " requires the path to an OSTree sysroot as an input argument.\n";
     return EXIT_FAILURE;
   }
-  test_sysroot = argv[1];
+
+  TemporaryDirectory temp_sysroot;
+  test_sysroot = temp_sysroot / "sysroot";
+  // uses cp, as boost doesn't like to copy bad symlinks
```
patrickvacek: What's wrong with the symlink? Why is it bad?

src/aktualizr_secondary/CMakeLists.txt
lbonn: Ok fair enough, this was a bit brutal. I had the idea of the "noptest" label after that, so that's why I've started removing stuff I did not like :).

src/libaktualizr/package_manager/ostreemanager_test.cc
lbonn: To be specific, in `ostree_repo/ostree/deploy/dummy-os/var`, there is a `run` symlinking to `../run` which doesn't exist. That's the internal to ostree so it seems to be working fine but when using `Utils::copyDir`, the boost function that we use fails.

I believe Serhiy had a similar issue in the past and used the same solution.

patrickvacek: Hmm. Ugly. Oh well, do what you gotta do.

src/aktualizr_secondary/CMakeLists.txt
lbonn: Now changed to use a softer method.

Still removed `aktualizr_secondary_cmdline_empty` which has the potential to run forever and is not really meaningful IMO.

lbonn:
Hm I've taken another look and I'm not sure the ioctl error is exactly the same. Though it's also probably a false positive.

Do you have the full valgrind log?

lbonn:
Added a cleaner way to handle running aktualizr-repo at configure time when cross-compiling.

PR #1122, opened 2019-03-01
Implement Abort() API call

eu-smirnov:
The main idea of this PR is to provide an ability to abort ongoing downloads and reset the queue to an empty state. It also removes Shutdown() call, which was bringing an Aktualizr object to an unusable state and fixes the issue with a possible race condition between Pause()/Resume() calls.
The PR still lacks tests for the Abort() functionality and more proper handling of an abort event in fetcher and uptane client.
At least it doesn't seem to break existing functionality and I feel that it's already getting quite big,
so it would be great if you could have a look at it next week.


src/hmi_stub/main.cc
```diff
@@ -147,7 +147,6 @@ int main(int argc, char *argv[]) {
     while (std::getline(std::cin, buffer)) {
       boost::algorithm::to_lower(buffer);
       if (buffer == "shutdown") {
-        aktualizr.Shutdown();
         break;
```
patrickvacek: We should add an abort command to hmi_stub so we can try this out more easily.

tests/sota_tools/treehub_server.py
```diff
@@ -22,18 +23,17 @@ def do_GET(self):
             return
         print("Processing GET request %s" % self.path)
         path = os.path.join(repo_path, self.path[1:])
-        if(os.path.exists(path)):
+        if os.path.exists(path):
+            time.sleep(0.5)
```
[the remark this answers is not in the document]
eu-smirnov: Yes, makes sense. The problem here was that if download happens too fast, the OSTree progress callback doesn't get called at all.

tests/sota_tools/repo/refs/heads/master
```diff
@@ -1 +1 @@
-16ef2f2629dc9263fdf3c0f032563a2d757623bbc11cf99df25c3c3f258dccbe
+7f9237e1048fa97b204a2c42028e03eec27d891b5dfab0589589a38987d76199
```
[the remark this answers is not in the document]
eu-smirnov: Absolutely, I'm also not a fan of keeping binary files (especially ones that are needed only for tests) in the repo.
The change was needed to get a progress status from OSTree, with only one object it obviously doesn't work. But that was rather a quick hack to see how it normally works.

src/libaktualizr/utilities/apiqueue.cc
```diff
@@ -0,0 +1,112 @@
...
+    }
+    cv_.notify_all();
+    if (thread_.joinable()) {
+      thread_.join();
+    }
+    {
+      std::lock_guard<std::mutex> g(m_);
+      std::queue<std::packaged_task<void()>>().swap(queue_);
+      token_.reset();
+      shutdown_ = false;
```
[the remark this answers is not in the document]
eu-smirnov: That's right, I'll add a comment.

src/libaktualizr/primary/aktualizr.h
```diff
@@ -166,11 +112,22 @@ class Aktualizr {
...
+  /**
+   * Requests the currently running command to abort and flushes the command
+   * queue.
+   * The `Abort()` function will block until the command queue is empty and
+   * all currently executing commands have stopped. You can also call Abort()
+   * on a previously paused class instance, this will clean the command queue,
+   * but the Aktualizr will remain in the paused state. To continue execution
+   * at some later point one needs to call Resume().
+   */
+  void Abort();
```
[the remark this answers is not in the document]
eu-smirnov: I'm not so sure if it's worth mentioning here. It's unlikely, but what if we decide to abort e.g. CheckForUpdates at some later point?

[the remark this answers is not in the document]
patrickvacek: It's not strictly necessary, but when I read "Requests the currently running command to abort", that leads me to assume that all commands can be aborted while in progress, and that isn't true. If it later becomes true, or even if just one more command is abortable, we can change the doc here again to reflect that as necessary.

eu-smirnov:
Thanks for the review, @patrickvacek! I'll fix the mentioned issues soon.
The idea of the token is to make pause/cancellation more generic, so we can reuse it in other tasks if we want to, and, more important, to control the execution flow from one place, so we avoid race conditions.

patrickvacek:
>  The idea of the token is to make pause/cancellation more generic, so we can reuse it in other tasks if we want to, and, more important, to control the execution flow from one place, so we avoid race conditions.

Great, that was my understanding. That is perfect. The abstraction was confusing to me at first, but it is obviously superior to me now that I understand it better.

src/libaktualizr/primary/aktualizr.h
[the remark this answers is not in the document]
eu-smirnov: Agreed, sounds confusing. I'll rephrase it.

tests/sota_tools/treehub_server.py
```diff
@@ -1,85 +1,155 @@
...
+        if self.drop_check():
+            print("Dropping HEAD request %s" % self.path)
+            return
+        print("Processing HEAD request %s" % self.path)
+        path = os.path.join(repo_path, self.path[1:])
+        if os.path.exists(path):
+            self.send_response_only(200)
+        else:
+            self.send_response_only(404)
+        self.end_headers()
```
patrickvacek: `do_HEAD` looks to be duplicated.

tests/sota_tools/test-garage-deploy-dry-run
```diff
@@ -12,9 +12,9 @@ TREEHUB="{\
 }"
 
 echo $TREEHUB > $TEMP_DIR/treehub.json
-./tests/sota_tools/treehub_server.py $PORT &
-sleep 1
+./tests/sota_tools/treehub_server.py -p${PORT} -c &
```
patrickvacek: Do you not need a space after the `-p`? If not, no problem.

tests/sota_tools/treehub_server.py
```diff
@@ -1,85 +1,155 @@
...
+                    help='sleep for n.n seconds for every GET request')
+parser.add_argument('-t', '--tls', action='store_true',
+                    help='require TLS from clients')
+args = parser.parse_args()
+
+signal.signal(signal.SIGTERM, sig_handler)
 try:
-    httpd.serve_forever()
-except KeyboardInterrupt as k:
+    with ExitStack() as stack:
```
patrickvacek: What benefit does the ExitStack and sig_handler provide? Just curious, not doubting. :)

tests/sota_tools/treehub_server.py
eu-smirnov: oops, thanks!

tests/sota_tools/test-garage-deploy-dry-run
eu-smirnov: no, should work either way

tests/sota_tools/treehub_server.py
eu-smirnov: sys.exit(0) in signal handler will raise a SystemExit exception, which will unwind the ExitStack cleaning up the tmp directory

PR #1123, opened 2019-03-04
aktualizr-info tests

mike-sul:
- Moved the existing (the shell script-based) tests into the gtest-based;
- Added a few more tests.


mike-sul:
I see that the debian build fails, looking into it.

mike-sul:
The failure with building on debian was fixed.

src/aktualizr_info/aktualizr_info_test.cc
```diff
@@ -0,0 +1,478 @@
...
+ */
+TEST_F(AktualizrInfoTest, PrintDirectorTargetsMetadata) {
+  db_storage_->storeEcuSerials({{Uptane::EcuSerial(primary_ecu_serial), Uptane::HardwareIdentifier(primary_ecu_id)}});
+  db_storage_->storeEcuRegistered();
+
+  Json::Value director_root_json;
+  director_root_json["key-002"] = "value-003";
+
+  std::string images_root = Utils::jsonToStr(director_root_json);
+  db_storage_->storeRoot(images_root, Uptane::RepositoryType::Director(), Uptane::Version(1));
```
patrickvacek: Why store the `images_root` here?

src/aktualizr_info/aktualizr_info_test.cc
```diff
@@ -0,0 +1,478 @@
...
+ *  - [x] Print whether metadata has been fetched from the server, if they were fetched
+ */
+TEST_F(AktualizrInfoTest, PrintPrimaryAndSecondaryInfo) {
+  const std::string secondary_ecu_serial = "c6998d3e-2a68-4ac2-817e-4ea6ef87d21f";
+  const std::string secondary_ecu_id = "secondary-hdwr-af250269-bd6f-4148-9426-4101df7f613a";
+  const std::string provisioning_status = "Provisioned on server: yes";
+  const std::string fetched_metadata = "Fetched metadata: yes";
+
+  Json::Value meta_root;
+  std::string director_root = Utils::jsonToStr(meta_root);
```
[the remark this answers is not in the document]
mike-sul: They are composing director's root metadata string which is stored in the DB at line 117. Aktualizr-info will think that the metadata are not fetched by Aktualizr if there is no any jsonified string stored in the DB ('meta' table )

src/aktualizr_info/aktualizr_info_test.cc
mike-sul: Because, the aktualizr-info won't output target metadata from a director repository if its root metadata are not stored in DB. Although, the variable name 'images_root' is incorrect so I am renaming it into 'director_root'.

src/aktualizr_info/aktualizr_info_test.cc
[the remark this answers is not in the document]
patrickvacek: Oops, my bad, I see it now. Thanks for pointing that out!

src/aktualizr_info/aktualizr_info_test.cc
patrickvacek: Great. now I understand. The name threw me off, so thanks for fixing it.

patrickvacek:
Looks like the Jenkins failures are related to a problem on the backend. I restarted them and hopefully all will be well this time.

patrickvacek:
Update: I can reproduce these errors locally on master with both production and SIT credentials. I'm not quite sure what is going wrong, but this isn't unique to your branch.

Issue #1124, opened 2019-03-06
unbound variable DOCKER_OPTS[@] when running scripts/run_docker_test.sh


bclouser:
Am I the only one seeing this error?

`./scripts/run_docker_test.sh: line 49: DOCKER_OPTS[@]: unbound variable`

patrickvacek:
> Am I the only one seeing this error?

Possibly. I've never seen it. Thank you for the PR, though! I'll take a look soon.

patrickvacek:
Closing because this is fixed.

PR #1125, opened 2019-03-06
Fix unbound variable DOCKER_OPTS[@] in run_docker_test.sh

bclouser:
Resolve: './scripts/run_docker_test.sh: line 49: DOCKER_OPTS[@]: unbound variable'

Signed-off-by: Ben Clouser <ben.clouser@toradex.com>

for issue:
https://github.com/advancedtelematic/aktualizr/issues/1124

### DIFF
--- a/scripts/run_docker_test.sh
+++ b/scripts/run_docker_test.sh
@@ -45,5 +45,10 @@ IMG_TAG=dev-$(cat /proc/sys/kernel/random/uuid)
 
 docker build -t "${IMG_TAG}" -f "$DOCKERFILE" .
 
+
+# Prevent DOCKER_OPTS[@]: unbound variable
+# From SO: https://stackoverflow.com/a/34361807/6096518
+OPTS_STR=${DOCKER_OPTS[@]+"${DOCKER_OPTS[@]}"}
+
 # run under current user, mounting current directory at the same location in the container
-docker run -u "$(id -u):$(id -g)" -v "$PWD:$PWD" -w "$PWD" --rm "${DOCKER_OPTS[@]}" -it "${IMG_TAG}" "$@"
+docker run -u "$(id -u):$(id -g)" -v "$PWD:$PWD" -w "$PWD" --rm $OPTS_STR -it "${IMG_TAG}" "$@"


lbonn:
That's intriguing that we haven't noticed that before. What is your bash version?

Also, from the [original StackOverflow question](https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u/34361807):

> A common solution to this is to use ${arr[@]-} instead, thus substituting an empty string instead of the ("undefined") empty array. However this is not a good solution, since now you can't discern between an array with a single empty string in it and an empty array.

Looks like it should not really a problem in our case since we don't need to distinguish between the twos. Have you also tried it?

patrickvacek:
It works both with and without this change on my machine with bash version 4.4.19. If this change makes it work for you, then I think it's fine. But I'm still curious why it works for me and not for you without this fix.

bclouser:
Heh heh, so it is just me!
Ok, so looks like it changed as of version 4.4 of bash

My dev machine is still ubuntu 16 which has 4.3.48 which has this issue:
```
bensThinkPad:scripts $ set -u
bensThinkPad:scripts $ someArray=()
bensThinkPad:scripts $ echo ${someArray[@]}
bash: someArray[@]: unbound variable
```

but in docker with latest ubuntu and bash 4.4.19
```
root@3e98ce94a78d:/# set -u
root@3e98ce94a78d:/# someArray=()
root@3e98ce94a78d:/# echo ${someArray[@]}

root@3e98ce94a78d:/# 
```

Digging further. In the changelog for bash 4.4
```
subst.c
+       - parameter_brace_expand: if ${array[@]} or ${array[*]} is supplied
+         with an unset (or empty) array, and -u is enabled, don't make it
+         an unbound variable error, like $@ or $* when there are no
+         positional parameters.  Based on report from Ian Allen
+         <idallen@idallen-fibe.dyndns.org>
```
So its a real bug (for us older bash users :smile: )

and in response to @lbonn, yes i did try the ${arr[@]-} and it does work, but I figured If i was going through the trouble of submitting a PR I would submit the most complete fix, lest it create some subtle bug in the future where someone was **actually** trying to discern between empty array and just empty string... however unlikely.

lbonn:
Thanks for the investigation! :)

I would vote for the simplest sufficient solution but won't argue too much to defend it. Thoughts @patrickvacek?

In any case, noting that it's a bash < 4.4 issue in the comments, would be great.

patrickvacek:
> I would vote for the simplest sufficient solution but won't argue too much to defend it.

I think it's fine as it is. I also like simpler, but since it is functionally different, it's probably better to do it the right way. Having the comment with a link to the SO page is good for clarifying why we are doing it that way. And I just now realized that further down the comments someone mentions that it changed in bash 4.4, so I think we've even got that part covered. :)

lbonn:
Well, "the right way" is not so clear cut :).

shellcheck now reports warning and `./run_docker_test Dockerfile -e VAR="a b"` does word splitting and gets converted to `docker run -u ... -e VAR=a b ...`. The `${DOCKER_OPTS[@]:-}` does not work in the empty case though.

But this is just a helper script, so let's not worry too much

PR #1129, opened 2019-03-07
Add a valgrind bypass option in run-valgrind

lbonn:
To make it configurable after configure and compile time

### DIFF
--- a/scripts/run-valgrind.in
+++ b/scripts/run-valgrind.in
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+if [ "$NO_VALGRIND" = 1 ]; then
+    "$@"
+    exit $?
+fi
+
 @VALGRIND@ --error-exitcode=1 \
     --leak-check=yes \
     --num-callers=25 \


patrickvacek:
Why exactly? I'm a bit worried about letting users think they are running with valgrind only to actually be running without.

lbonn:
Now, it's a compile time / build time configuration for running tests, which is mostly fine. The problem is that it prepends this run-valgrind script to inside the files used by ctest as instructions.

We'll want to run the suite with valgrind on ptest most of the time but it would also be handy to disable it dynamically to make it run faster and get around false positives. Modifying the test sequence is tricky there because we are out of the configure and build steps and we deploy the build directory on a read-only filesystem.

It seems to me pretty unlikely that users would disable it by mistake, maybe you find the variable name too generic?

patrickvacek:
> It seems to me pretty unlikely that users would disable it by mistake

You're probably right, I'm being too careful. It's probably fine as you have it.

PR #1130, opened 2019-03-11
Use `bare-user-only` for ostree tests

lbonn:
So they can run on tmpfs


lbonn:
Sorry, just pushed another hilarious fix

lbonn:
I'll merge it's really small and passes the tests + fix one ptest case.

PR #1131, opened 2019-03-11
OTA-1685: aktualizr-cert-provider tests

mike-sul:

### DIFF
--- a/actions.md
+++ b/actions.md
@@ -363,36 +363,35 @@ These tools all link with libaktualizr, although they do not necessary use the A
 
 `aktualizr-cert-provider` assists with generating credentials and uploading them to a device for implicit provisioning.
 
-- [ ] Use file paths from config if provided
-- [ ] Use autoprovisioning credentials if fleet CA and private key are not provided
+- [x] Use file paths from config if provided (cert_provider_test.cc)
+- [x] Use autoprovisioning credentials if fleet CA and private key are not provided (cert_provider_autoprov_test.cc)
   - [x] Generate a random device ID (OTA-986, utils_test.cc, uptane_init_test.cc)
-  - [x] Automatically provision (see above)
-- [x] Use fleet credentials if provided (run_certprovider_test.sh)
-  - [ ] Abort if fleet CA is provided without fleet private key
-  - [ ] Abort if fleet private key is provided without fleet CA
-  - [ ] Specify RSA bit length
-  - [ ] Specify device certificate expiration date
-  - [ ] Specify device certificate country code
-  - [ ] Specify device certificate state abbreviation
-  - [ ] Specify device certificate organization name
-  - [ ] Specify device certificate common name
-    - [ ] Generate a random device ID if not specified
-  - [ ] Read fleet CA certificate
-  - [ ] Read fleet private key
-  - [x] Create device certificate (run_certprovider_test.sh)
-  - [ ] Create device keys
-  - [ ] Set public key for the device certificate
-  - [x] Sign device certificate with fleet private key (run_certprovider_test.sh)
-  - [ ] Serialize device private key to a string
-  - [ ] Serialize device certificate to a string
+  - [x] Automatically provision (cert_provider_autoprov_test.cc)
+- [x] Use fleet credentials if provided (cert_provider_test.cc)
+  - [x] Abort if fleet CA is provided without fleet private key (cert_provider_test.cc)
+  - [x] Abort if fleet private key is provided without fleet CA (cert_provider_test.cc)
+  - [x] Specify RSA bit length (cert_provider_test.cc)
+  - [x] Specify device certificate expiration date (cert_provider_test.cc)
+  - [x] Specify device certificate country code (cert_provider_test.cc)
+  - [x] Specify device certificate state abbreviation (cert_provider_test.cc)
+  - [x] Specify device certificate organization name (cert_provider_test.cc)
+  - [x] Specify device certificate common name (cert_provider_test.cc)
+  - [x] Read fleet CA certificate (cert_provider_test.cc)
+  - [x] Read fleet private key (cert_provider_test.cc)
+  - [x] Create device certificate (cert_provider_test.cc)
+  - [x] Create device keys (cert_provider_test.cc)
+  - [x] Set public key for the device certificate (cert_provider_test.cc)
+  - [x] Sign device certificate with fleet private key (cert_provider_test.cc)
+  - [x] Serialize device private key to a string (cert_provider_test.cc)
+  - [x] Serialize device certificate to a string (cert_provider_test.cc)
 - [ ] Read server root CA from credentials archive
   - [ ] Read server root CA from server_ca.pem if present (to support community edition use case)
-  - [ ] Read server root CA from p12 (default case)
-- [ ] Write credentials to a local directory if requested
-  - [ ] Provide device private key
-  - [x] Provide device certificate (run_certprovider_test.sh)
-  - [ ] Provide root CA if requested
-  - [ ] Provide server URL if requested
+  - [x] Read server root CA from p12 (cert_provider_autoprov_test.cc)
+- [x] Write credentials to a local directory if requested (cert_provider_test.cc)
+  - [x] Provide device private key (cert_provider_test.cc)
+  - [x] Provide device certificate (cert_provider_test.cc)
+  - [x] Provide root CA if requested (cert_provider_autoprov_test.cc)
+  - [x] Provide server URL if requested (cert_provider_autoprov_test.cc)
 - [ ] Copy credentials to a device with ssh
   - [ ] Create parent directories
   - [ ] Provide device private key
--- a/src/cert_provider/CMakeLists.txt
+++ b/src/cert_provider/CMakeLists.txt
@@ -11,8 +11,21 @@ add_dependencies(build_tests aktualizr-cert-provider)
 
 install(TARGETS aktualizr-cert-provider RUNTIME DESTINATION bin)
 
-add_test(NAME certprovider-ca COMMAND ${PROJECT_SOURCE_DIR}/tests/run_certprovider_test.sh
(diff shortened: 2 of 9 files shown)


src/cert_provider/CMakeLists.txt
```diff
@@ -11,8 +11,21 @@ add_dependencies(build_tests aktualizr-cert-provider)
...
-        ${PROJECT_SOURCE_DIR}/tests/test_data/CAcert.pem ${PROJECT_SOURCE_DIR}/tests/test_data/CApkey.pem)
+set(CREDS ${SOTA_PACKED_CREDENTIALS})
+
+if (TESTSUITE_EXCLUDE)
+  foreach(label ${TESTSUITE_EXCLUDE})
+    if (${label} STREQUAL "credentials")
+      set(CREDS "")
+    endif()
+  endforeach()
+endif(TESTSUITE_EXCLUDE)
```
patrickvacek: I think this can be done a lot more easily by setting the right LABEL for this test target, for example: `set_tests_properties(test_aktualizr_cert_provider PROPERTIES LABELS "credentials")`. You can see an example of this in `src/libaktualizr/uptane/CMakeLists.txt`. If you do that, then the Travis script that runs the tests will automatically exclude it because of the `-eTEST_TESTSUITE_EXCLUDE=credentials` parameter given to cmake.

actions.md
```diff
@@ -362,36 +362,35 @@ These tools all link with libaktualizr, although they do not necessary use the A
...
-  - [ ] Provide device private key
-  - [x] Provide device certificate (run_certprovider_test.sh)
-  - [ ] Provide root CA if requested
-  - [ ] Provide server URL if requested
+  - [x] Read server root CA from p12 (cert_provider_test.cc)
+- [x] Write credentials to a local directory if requested (cert_provider_test.cc)
+  - [x] Provide device private key
+  - [x] Provide device certificate
+  - [x] Provide root CA if requested
+  - [x] Provide server URL if requested
```
patrickvacek: Would you mind adding `(cert_provider_test.cc)` to each of these lines to keep track of where the tests that cover these actions live?

src/cert_provider/main.cc
```diff
@@ -383,157 +384,196 @@ int main(int argc, char* argv[]) {
...
+    credentials_path = commandline_map["credentials"].as<boost::filesystem::path>();
+  }
 
-    if (!config.import.tls_pkey_path.empty()) {
-      pkey_file = config.import.tls_pkey_path;
-    } else {
-      pkey_file = config.storage.tls_pkey_path;
-    }
+  if (local_dir.empty() && target.empty()) {
+    std::cerr << "Both a local dir and a target are not specified, thus the resultant cert and key will be lost"
```
patrickvacek: Good thing to check for. However, I think it's fine to just say "Please provide a local directory and/or target to output the generated files to."

src/cert_provider/cert_provider_test.cc
```diff
@@ -0,0 +1,503 @@
...
+
+#include <boost/asio/io_service.hpp>
+#include <boost/format.hpp>
+#include <boost/process.hpp>
+
+#include "config/config.h"
+#include "utilities/utils.h"
+
+static const char* CERT_PROVIDER_PATH = nullptr;
+static const char* CREDENTIALS_PATH = nullptr;
```
patrickvacek: We have a fondness for boost::filesystem::path (and descriptive types in general).

src/cert_provider/cert_provider_test.cc
```diff
@@ -0,0 +1,503 @@
...
+  config.import.tls_pkey_path = BasedPath(private_key_file);
+  config.import.tls_clientcert_path = BasedPath(cert_file);
+
+  auto test_conf_file = tmp_dir_ / "conf.toml";
+  boost::filesystem::ofstream conf_file(test_conf_file);
+  config.writeToStream(conf_file);
+  conf_file.close();
+
+  // for some reason the cert-provider uses the config's base path as suffix (the same as the 'directory' param) of the
+  // local dir
```
patrickvacek: This is for when we use it inside bitbake. According to the config, the import dir will be `/var/sota/import`, but when we prepare the image, it will have a really long prefix.

src/cert_provider/CMakeLists.txt
mike-sul: Yes, but it will require splitting the tests into two files, since the aforementioned mechanism works per test/file bases. I'll split the tests into two parts as we discussed offline and apply the described approach.

actions.md
mike-sul: yes, sure.

patrickvacek:
This should also be tested to make sure it doesn't break bitbake's expectations. Have you tried that yet?

src/cert_provider/main.cc
mike-sul: ok.

src/cert_provider/cert_provider_test.cc
mike-sul: ok, will update it.

src/cert_provider/CMakeLists.txt
mike-sul: Actually, splitting tests into two files look more complicated compared to what is implemented in this CMakeLists.

patrickvacek: Why? I expect that some code will need to be put in a common location (possibly the already available TestUtils), but what else has to be taken care of?

lbonn: Also not a big fan but it should at least have an explanatory comment in the CMake definition.

Other points:

- use https://cmake.org/cmake/help/latest/command/list.html#search maybe?
- the skipped test silently succeeding with empty credentials is very misleading

mike-sul: The list search doesn't work for a variable that contains just a single value for some reason.

Yeah, anyway, I agree that it's better actually split these tests into two parts, it just look more readable and better/easier extendable compared to what is now.

mike-sul: Moreover, just setting the label to a test does not help to exclude the test. TESTSUITE_EXCLUDE definition is used/applied solely to tests directory and corresponding ctest parameters are applied only to 'check' target.

lbonn: What do you mean "it does not help to exclude the test"?

It's directly supported by the `check` target, as you've said, which is the recommended way to run the suite (and what is used by CI) and it also works for tests defined outside of the `tests` directory, as you can see on CI.

The reason we have this surprising setup is that we had the problem of tests requiring credentials only being compiled when `SOTA_PACKED_CREDENTIALS` was defined. As we used to not have credentials on our only CI at the time (Travis), and as developers tended to not run the testsuite with the variable either, these tests often weren't compiled at all and broke often. Also in theory, they'd need to be compiled to be checked by clang-tidy (but we disabled the checks for the unit tests IIRC).

So, the solution was to define bogus credentials in Travis, to compile the test but not running them.

actions.md
```diff
@@ -362,36 +362,35 @@ These tools all link with libaktualizr, although they do not necessary use the A
 
 `aktualizr-cert-provider` assists with generating credentials and uploading them to a device for implicit provisioning.
 
-- [ ] Use file paths from config if provided
-- [ ] Use autoprovisioning credentials if fleet CA and private key are not provided
+- [x] Use file paths from config if provided (cert_provider_test.cc)
+- [x] Use autoprovisioning credentials if fleet CA and private key are not provided (cert_provider_test.cc)
```
patrickvacek: Some of these are now in cert_provider_autoprov_test.cc now.

mike-sul:
It looks like I've fixed the issue with a child process exit code.
The travis job still fails but for another reason "The job exceeded the maximum log length, and has been terminated." not sure if know what to do with it.

actions.md
mike-sul: done

patrickvacek:
> It looks like I've fixed the issue with a child process exit code.

Good catch tracking that down. I just realized the `Process::spawn` code is fairly similar to `Utils::shell` in `src/libaktualizr/utilities/utils.cc`. Do you think it makes sense to merge them? Your work appears to me to be more thorough.

patrickvacek:
> The travis job still fails but for another reason "The job exceeded the maximum log length, and has been terminated." not sure if know what to do with it (e.g. limit or get rid of the gcov's output).

That is annoying, sorry that you encountered that. Limiting or even silencing gcov's output is a great idea. We rarely want to see it in the Travis CI logs anyway; it's much easier to just go to the codecov website and review it there. If we need the follow text logs, we can always run it manually, or perhaps we could leave the full output in the Jenkins/gitlab builds.

lbonn:
:+1: for silencing the gcov output, since it's very low value. Might redirect it to a log file just in case but I'm not even sure we need it.

mike-sul:
> > It looks like I've fixed the issue with a child process exit code.
> 
> Good catch tracking that down. I just realized the `Process::spawn` code is fairly similar to `Utils::shell` in `src/libaktualizr/utilities/utils.cc`. Do you think it makes sense to merge them? Your work appears to me to be more thorough.

Well, `Utils::shell` is used not just for tests but also for the aktualizr's bootloader and debian package manager that might be 'sensitive' to  its binary size and additional dependencies. Most likely, the boost::process usage will increase aktualizr's binary size, footprint and introduce another dependency (libboost*.so) for it what might be not optimal in the embedded world. Although, it already uses boost and has dependencies on five boost libs. IMHO, at high level, a decision on it depends on what kind of platforms we are targeting.

Also, test_utils.h::TestHelperProcess can be replaced by boost::process::system, and this tool is used solely for testing purposes so the aforementioned concerns are not relevant in this case.

patrickvacek:
> Well, `Utils::shell` is used not just for tests but also for the aktualizr's bootloader and debian package manager that might be 'sensitive' to its binary size and additional dependencies.

Good point. You're right, let's leave `Utils::shell` as it is.

PR #1133, opened 2019-03-11
Suppress a boost 1.69.0 valgrind error

lbonn:
Could be serious? Maybe we should fix it.

### DIFF
--- a/tests/aktualizr.supp
+++ b/tests/aktualizr.supp
@@ -32,6 +32,12 @@
    ...
 }
 {
+   boost 1.69 directory_iterator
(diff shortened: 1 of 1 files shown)


eu-smirnov:
Looks like it's this one: https://github.com/boostorg/filesystem/commit/bbe9d1771e5d679b3f10c42a58fc81f7e8c024a9

lbonn:
Yes also found that :). I first tried to patch our boost yocto recipe but it diverged way too much between 1.69 and the parent of this commit.

Issue #1118, opened 2019-02-27
Feature Proposal: Docker "packages"


doanac:
This is closely related to PR #1046 by @cajun-rat. I've been playing around some in code trying to get a feel for things and think I may have a more holistic approach for this. From a high-level UX it could simply be a user in app.atsgarage.com doing an "add package" with a targetFormat=DOCKER_APP. _And wow - tufreposerver can store files ... I had no idea._ I think for my production usage, I'd probably use my credentials.zip to push things to TUF by hand that I've "released", but the GUI option seems cool.

The package could then be added to a device(s). Currently it will fail because SotaUptaneClient will say "Cannot install a non-OSTree package on an OSTree system". At this point I *think* we could implement a new DockerApp package manager. I'm not sure how we'd wire things up in code, ie do we create something like a CompositePackageManager that has pointers to both the OstreeManager and the DockerAppManager or do we conditionally compile in some Docker logic into the OstreeManager, or maybe SotaUptaneClient could change to construct the proper PackageManager implementation on-the-fly based on the targetFormat? Either way, you'd probably want a new value for `config.pacman.type = kOstreeAndDocker`.

I'm totally open to trying to make proof-of-concept if this isn't crazy. However, I'm worried about the answer to question 1 below.

## Open Questions/Concerns
1. Would having 2 or really $N "packages" (you could choose multiple docker-apps) on target work with Uptane? 
PackageManagerInterface::getManifest might not be capable of expressing this? However, maybe its fine to just report the installed image?

2. Docker-App is still pretty experimental
Seems like their [DockerCon Preview](https://github.com/docker/app/releases/tag/cnab-dockercon-preview) is enough for us. We coud download the app bundle via the ota-reposerver and then do something like `docker-app-linux render helloworld.dockerapp  | docker-compose -p <package name> -f -  up`

patrickvacek:
I'm just a bit confused about the big picture here. You want to have both docker apps that you can update as well as an OSTree system? These will all live on the same device? Where will the docker apps live such that an OSTree update won't wipe them away?

You are correct that if you want to use docker as a package manager you will probably need to create a new class that inherits from PackageManagerInterface. That shouldn't be too hard. I don't understand how it should interface with OSTree, but in theory you could still just make calls into the OSTree package manager from there if you really wanted to. That might be tricky, but seems possible.

>     1. Would having 2 or really $N "packages" (you could choose multiple docker-apps) on target work with Uptane?

[...]

PR #1135, opened 2019-03-12
RFC: Remove circular dependency between fetcher(uptane) and packagemanager

doanac:
This is a very lightly tested attempt at cleaning up the relationship between the fetcher, sotauptaneclient, and packagemanager.

Its dependent on PR #1122. Its kind of a prereq issue #1118 so that the "app download" logic won't make the existing fetcher logic even messier.


src/libaktualizr/package_manager/fetcher_test.cc
```diff
@@ -56,7 +64,10 @@ void test_pause(const Uptane::Target& target) {
 
   std::shared_ptr<INvStorage> storage(new SQLStorage(config.storage, false));
   auto http = std::make_shared<HttpClient>();
-  Uptane::Fetcher f(config, storage, http, progress_cb);
+
+  auto pacman = std::make_shared<PKGMGR_TYPE>(config.pacman, storage, nullptr, http);
```
lbonn: We try to avoid having tests that run differently depending on CMake flags, so that we can be confident we test everything when the test suite runs with everything `=on`. The CMake flags are only here to add functionality, not to configure it.

In this case, `test_pause()` could have the package manager type as a template parameter, or could take a `std::function` factory.

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -222,14 +222,22 @@ data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &ta
...
   if (sysroot_smart == nullptr) {
     throw std::runtime_error("Could not find OSTree sysroot at: " + config.sysroot.string());
   }
 }
 
+bool OstreeManager::fetchTarget(const Uptane::Target &target, const std::string &repo_server, const KeyManager &keys,
+                                FetcherProgressCb progress_cb, const api::FlowControlToken *token) {
+  if (!target.IsOstree()) {
+    return PackageManagerBase::fetchTarget(target, repo_server, keys, progress_cb, token);
+  }
```
patrickvacek: I was going to say this should be an error (installing a binary file on an OSTree filesystem) but apparently our code allows it at present. I'm not sure that's right, but that's also not something you should worry about for now unless you are depending upon that logic.

src/libaktualizr/package_manager/fetcher_test.cc
doanac: good idea. thanks.

src/libaktualizr/package_manager/ostreemanager.cc
doanac: Correct. I was really trying to make this function *exactly* the same as the old code. However, I think that even though the download passes in the current code base, the install step will fail. So its not a huge difference and something that would be nice to clean up at some point.

It would also probably be smart to update the Web UI to not lead users to think they can install binary files on an OSTree based device.

TBH - I'm a bit confused what the point of installing a file is at all. I guess maybe its applicable to the DebianPackageManager?

That also got me wondering about the DebianPackageManager. Is this really used by anyone? It looks like its only capable of installing a single package, so I was confused how someone might use that?

doanac:
>What is the point of distinguishing between PackageManagerBase and PackageManagerInterface? 

I could easily merge this. My first attempt did, and then I split it out. I prefer the simplicity of just having one class.

> I recommend in any event waiting for #1122 to get merged first and then rebasing this on master.

fantastic. I'll table this for now and we can pick this back up when its merged. FYI - you may want to get that PR rebased on master. I think it may have a merge conflict which would be nice to resolve while its still being tested.

src/libaktualizr/package_manager/ostreemanager.cc
patrickvacek: > However, I think that even though the download passes in the current code base, the install step will fail. So its not a huge difference and something that would be nice to clean up at some point.

That matches my expectations. Not a high priority right now, but I agree we should clean that up (and on the web UI if possible, too).

> I'm a bit confused what the point of installing a file is at all.

There are at least three uses that I am aware of at present:
1. Non-OSTree primaries that use an external installer. This comes up in customer projects. However, all of the examples of that are currently closed-source.
2. Secondaries. Most examples we've seen do not use OSTree and just flash a whole filesystem image.
3. The fake package manager, which is great for testing.

> That also got me wondering about the DebianPackageManager. Is this really used by anyone?

In short, no. It was a PoC that we got working at a basic level and then it never went anywhere. It is mostly dead weight, but it is kind of interesting, so we've kept it around in case it ever becomes useful again.

patrickvacek:
> FYI - you may want to get that PR rebased on master. I think it may have a merge conflict which would be nice to resolve while its still being tested.

You are probably correct. @eu-smirnov is still working on that PR and may already be aware of the conflict, but now he's been tagged and informed if he wasn't already!

patrickvacek:
@doanac The new PR looks very promising, and if this is just the draft version, we can close this now, right?

PR #1136, opened 2019-03-14
Set the required permission for the DB/storage folder

mike-sul:

Set the required permission (700) for the DB/storage folder (<tmp-dir>/sota) in test_lite.sh.
Set '1' as a default and error exit code instead of -1.

### DIFF
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -189,7 +189,7 @@ int main(int argc, char *argv[]) {
 
   bpo::variables_map commandline_map = parse_options(argc, argv);
 
-  int r = -1;
+  int r = 0;
   try {
     if (geteuid() != 0) {
       LOG_WARNING << "\033[31mRunning as non-root and may not work as expected!\033[0m\n";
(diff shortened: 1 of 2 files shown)


patrickvacek:
FYI @doanac. Does this look fine to you?

mike-sul:
@doanac Please, review these tiny corrections.

src/aktualizr_lite/main.cc
```diff
@@ -189,7 +189,7 @@ int main(int argc, char *argv[]) {
 
   bpo::variables_map commandline_map = parse_options(argc, argv);
 
-  int r = -1;
+  int r = 0;
```
doanac: This is fine by me. However, the logic was copied from:
 https://github.com/advancedtelematic/aktualizr/blob/master/src/aktualizr_primary/main.cc#L105

You might want to fix that in the other binaries as well so that things stay consistent (in code and for user error codes)

mike-sul: Yes, I agree, thx for pointing to it.

patrickvacek:
@mike-sul We can merge this now if you'd like and do further fixes for other tools in a separate PR if you'd like.

mike-sul:
> @mike-sul We can merge this now if you'd like and do further fixes for other tools in a separate PR if you'd like.

OK.

PR #1137, opened 2019-03-14
Better "could not put manifest" error

lbonn:
- stop logging it when the manifest put is inhibited by a pending install
- more detailed output when it couldn't be sent to the server


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1227,6 +1226,8 @@ bool SotaUptaneClient::putManifestSimple() {
     storage->clearInstallationResults();
     return true;
   }
+
```
mike-sul:
```suggestion
At first glance it looks like the manifest assembling is redundant if there are pending updates, so I'd move line #1217 just after `if (hasPendingUpdates())` statement.
```

lbonn: Good point, but I'd need to double check. I have vague memories about this function having or used to having side-effects.

PR #1138, opened 2019-03-14
OTA-2320: Add an option to aktualizr-info to print delegations metadata

mike-sul:

1. Added an option `--delegate` to aktualizr-info aimed to print delegations metadata;
2. Added corresponding tests.

### DIFF
--- a/src/aktualizr_info/aktualizr_info_test.cc
+++ b/src/aktualizr_info/aktualizr_info_test.cc
@@ -37,30 +37,33 @@ class AktualizrInfoTest : public ::testing::Test {
 
   virtual void TearDown() {}
 
+  // TODO: replace it with TestUtil::Process once the later is merged into master
   void SpawnProcess(const std::vector<std::string> args = std::vector<std::string>{}) {
     std::future<std::string> output;
     std::future<std::string> err_output;
+    std::future<int> child_process_exit_code;
     boost::asio::io_service io_service;
-    int child_process_exit_code = -1;
 
+    executable_args = default_executable_args;
     executable_args.insert(std::end(executable_args), std::begin(args), std::end(args));
     aktualizr_info_output.clear();
 
     try {
       boost::process::child aktualizr_info_child_process(
           boost::process::exe = executable_to_run, boost::process::args = executable_args,
-          boost::process::std_out > output, boost::process::std_err > err_output, io_service);
+          boost::process::std_out > output, boost::process::std_err > err_output,
+          boost::process::on_exit = child_process_exit_code, io_service);
 
       io_service.run();
 
       // To get the child process exit code we need to wait even in the async case (specifics of boost::process::child)
       ASSERT_TRUE(aktualizr_info_child_process.wait_for(std::chrono::seconds(20)));
-      child_process_exit_code = aktualizr_info_child_process.exit_code();
+
     } catch (const std::exception& exc) {
       FAIL() << "Failed to spawn process " << executable_to_run << " exited with an error: " << exc.what();
     }
 
-    ASSERT_EQ(child_process_exit_code, 0)
+    ASSERT_EQ(child_process_exit_code.get(), 0)
         << "Process " << executable_to_run << " exited with an error: " << err_output.get();
 
     aktualizr_info_output = output.get();
@@ -76,7 +79,8 @@ class AktualizrInfoTest : public ::testing::Test {
   std::shared_ptr<INvStorage> db_storage_;
 
   const std::string executable_to_run = "./aktualizr-info";
-  std::vector<std::string> executable_args = {"-c", test_conf_file_.string()};
+  std::vector<std::string> default_executable_args = {"-c", test_conf_file_.string()};
+  std::vector<std::string> executable_args;
   std::string aktualizr_info_output;
 
   std::string device_id;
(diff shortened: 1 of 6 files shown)


src/aktualizr_info/main.cc
```diff
@@ -24,6 +44,7 @@ int main(int argc, char **argv) {
     ("ecu-keys",  "Outputs UPTANE keys")
     ("images-root",  "Outputs root.json from images repo")
     ("images-target",  "Outputs targets.json from images repo")
+    (delegation_opt_name.c_str(),  "Outputs metadata of image repo targets' delegations")
```
lbonn: I see the point but having this option using a different format here is distracting. And we don't go and change options names everyday, so the benefit of this indirection compared to the readability cost is disputable.

Also, a `std::string` looks overkill here.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -561,6 +561,36 @@ bool SQLStorage::loadDelegation(std::string* data, const Uptane::Role role) {
...
+      LOG_TRACE << "Delegations metadata are not present";
+      return true;
+    } else if (statement_state != SQLITE_ROW) {
+      LOG_ERROR << "Can't get delegations metadata: " << db.errmsg();
+      return false;
+    }
+
+    do {
+      data.emplace_back(
+          DelegationRecord({statement.get_result_col_blob(0).value(), statement.get_result_col_str(1).value()}));
```
lbonn: This is missing the benefits of using `emplace_back` and doesn't compile on Xenial according to CI.

This is enough:

`data.emplace_back(statement.get_result_col_blob(0).value(), statement.get_result_col_str(1).value());`

src/aktualizr_info/main.cc
mike-sul: makes sense. will update it to make it consistent with the other options.

src/libaktualizr/storage/sqlstorage.cc
mike-sul: Yeah, your right, will fix it.

src/libaktualizr/storage/invstorage.h
```diff
@@ -107,6 +107,10 @@ enum class InstalledVersionUpdateMode { kNone, kCurrent, kPending };
 // Functions loading/storing multiple pieces of data are supposed to do so atomically as far as implementation makes it
 // possible
 class INvStorage {
+ public:
+  using DelegationRecord = std::tuple<std::string, std::string>;
```
lbonn: Any reason not to use `std::pair`? And shouldn't `role_name` be the first field?

Probably a matter of test but also not sure if it's worth it adding this type synonym in the header: in client code, we'd have one `std::vector<std::pair<Uptane::Role, std::string>>`, and the rest can be deduced via `auto`.

src/aktualizr_info/aktualizr_info_test.cc
```diff
@@ -484,6 +488,75 @@ TEST_F(AktualizrInfoTest, PrintDeviceNameOnly) {
...
+      std::get<1>(delegation_rec) = delegation_role_val;
+      db_storage->storeDelegation(delegation_val_str, Uptane::Role::Delegation(delegation_role_val));
+
+      ++indx;
+    }
+  };
+
+  auto verify_delegations = [](const std::string& output_str, INvStorage::DelegationArray& delegation_records) {
+    for (INvStorage::DelegationRecord& delegation_rec : delegation_records) {
+      ASSERT_NE(output_str.find(std::get<1>(delegation_rec)), std::string::npos);
```
lbonn: These two lines are identical?

mike-sul: Indeed, they are :)

src/libaktualizr/storage/invstorage.h
mike-sul: >  And shouldn't role_name be the first field?
Why should it be the first ? I think, it makes sense to preserve the same field order as in the DB table.

> Any reason not to use std::pair?
The reason, I decided to go with std::tuple over std::pair is the fact that the later can accommodate just two fields while the former can have arbitrary number of them what can be handy in case if the table schema changes (more than two fields or just one field).


>Probably a matter of test but also not sure if it's worth it adding this type synonym in the header: > in client code, we'd have one std::vector<std::pair<Uptane::Role, std::string>>, and the rest can > be deduced via auto

I think, it makes sense to use Uptane::Role regardless of the type been declared or not, it will make the type more clear and explicit.

Regarding introduction of the new type. IMHO, there are several pros over not doing it:

1. if the type is changed then you don't need to change it in every spot where it's used;
2. if adding some methods is required then the type declaration can be replaced with corresponding struct/class;
3. less typing (at the moment it's used in the storage interface, DB storage interface, DB storage implementation and at least once for each client (although `auto` cannot be used in function declarations/signature)

In general, it improves Extensibility and Maintainability.

Having said that, I don't mind to get rid of the type declaration since this is very unlikely that loadAllDelegations() will have more than two clients, i.e. aktualizr-info and its tests.

mike-sul: Pushed code with the changes implied here.

mike-sul:
> LGTM. @mike-sul do you want to merge now despite the TODO in SpawnProcess? This is higher-priority anyway and if you come back to fix that up later than I don't see a problem.

Yes, anyway the TODO depends on the other PR that hasn't been merged yet.

PR #1139, opened 2019-03-15
Bump googletest version

lbonn:
Dev activity seems to have sped up recently, should help us get rid of some cruft.

I've just checked and it doesn't emit warnings with clang-tidy anymore, so we could check our test files as well (after a cleanup).


patrickvacek:
Oh wow, I literally just did the same thing (bumping gtest) yesterday to get some new functionality so I could improve uptane_test_vectors.

PR #1141, opened 2019-03-15
Silence codecov.io upload

lbonn:


lbonn:
Done simpler in https://github.com/advancedtelematic/aktualizr/pull/1131

PR #1142, opened 2019-03-16
Move target iterator logic into uptane module

doanac:
Iterating targets really isn't dependent on the SotaUptaneClient and
future code in the package manager will need to be able to iterate
through targets to function. Moving it now will help remove a future
circular dependency.

### DIFF
--- a/src/libaktualizr/uptane/iterator.h
+++ b/src/libaktualizr/uptane/iterator.h
@@ -1,0 +1,65 @@
+#ifndef AKTUALIZR_UPTANE_ITERATOR_H_
+#define AKTUALIZR_UPTANE_ITERATOR_H_
+
+#include "fetcher.h"
+#include "imagesrepository.h"
+
+namespace Uptane {
+
+Targets getTrustedDelegation(const Role &delegate_role, const Targets &parent_targets,
+                             const ImagesRepository &images_repo, INvStorage &storage, Fetcher &fetcher);
+
+class LazyTargetsList {
+ public:
+  struct DelegatedTargetTreeNode {
+    Role role{Role::Targets()};
+    DelegatedTargetTreeNode *parent{nullptr};
+    std::vector<std::shared_ptr<DelegatedTargetTreeNode>>::size_type parent_idx{0};
+    std::vector<std::shared_ptr<DelegatedTargetTreeNode>> children;
+  };
+
+  class DelegationIterator {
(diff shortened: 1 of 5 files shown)


doanac:
This is related to issue #1118. The DockerAppManager needs to be able to look up a docker-app target by iterating over the image repository targets. eg you might have something like:
~~~
{
   "signed" : {
      "version" : 12,
      "targets" : {
         "lwm2m2-53.dockerapp" : {
            "hashes" : {
               "sha256" : "07701e25528fc8db2ab8d76df9ddad1bc1c50ede825cc27ec8a624a44144e582"
            },
            "length" : 46,
            "custom" : {
               "version" : "53",
               "targetFormat" : "BINARY",
               "name" : "lwm2m.dockerapp",
             }
         },
         "44-raspberrypi3-64-44" : {
            "length" : 0,
            "custom" : {
               "version" : "44",
               "targetFormat" : "OSTREE",    
               "name" : "44-raspberrypi3-64",
               "hardwareIds" : [
                  "raspberrypi3-64"
               ],
               "dockerApps" : {
                  "lwm2m": {
		    "filename": "lwm2m2-53.dockerapp"  # I need to look this up via the targets iterator
		  }
               }
            },
            "hashes" : {
               "sha256" : "1ebde51292fa3edb7eab4a83759074024f358d378025f1a4b3d7bbdbcea07939"
            }
         }
      },
}
~~~

I tried to avoid having to look through the targets, but without code like this you basically have to embedded the whole "target" into each docker app which gets really repetitive, error-prone, and impossible to manage.

src/libaktualizr/uptane/iterator.h
```diff
@@ -0,0 +1,65 @@
+#ifndef AKTUALIZR_UPTANE_ITERATOR_H_
+#define AKTUALIZR_UPTANE_ITERATOR_H_
+
+#include "fetcher.h"
+#include "imagesrepository.h"
+
+namespace Uptane {
+
+Targets getTrustedDelegation(const Role &delegate_role, const Targets &parent_targets,
+                             const ImagesRepository &images_repo, INvStorage &storage, Fetcher &fetcher);
```
patrickvacek: Can this be static?

patrickvacek: Never mind, I overlooked that this is outside of a class, so the benefit of `static` that I was hoping for is already there.

PR #1143, opened 2019-03-18
Align executables' exit status code across the project

mike-sul:

### DIFF
--- a/src/aktualizr_primary/main.cc
+++ b/src/aktualizr_primary/main.cc
@@ -102,8 +102,7 @@ int main(int argc, char *argv[]) {
 
   LOG_INFO << "Aktualizr version " AKTUALIZR_VERSION " starting";
 
-  int r = -1;
-  boost::signals2::connection conn;
+  int r = EXIT_FAILURE;
 
   try {
     if (geteuid() != 0) {
@@ -118,6 +117,8 @@ int main(int argc, char *argv[]) {
 
     Aktualizr aktualizr(config);
     std::function<void(std::shared_ptr<event::BaseEvent> event)> f_cb = process_event;
+    boost::signals2::scoped_connection conn;
+
     conn = aktualizr.SetSignalHandler(f_cb);
     aktualizr.Initialize();
 
@@ -147,11 +148,10 @@ int main(int argc, char *argv[]) {
     } else {
       aktualizr.RunForever().get();
     }
-    r = 0;
+    r = EXIT_SUCCESS;
   } catch (const std::exception &ex) {
     LOG_ERROR << ex.what();
-    r = -1;
   }
-  conn.disconnect();
+
   return r;
 }
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -189,7 +189,7 @@ int main(int argc, char *argv[]) {
 
   bpo::variables_map commandline_map = parse_options(argc, argv);
 
-  int r = 0;
+  int r = EXIT_FAILURE;
   try {
     if (geteuid() != 0) {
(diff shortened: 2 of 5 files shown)


src/aktualizr_primary/main.cc
```diff
@@ -147,10 +147,10 @@ int main(int argc, char *argv[]) {
     } else {
       aktualizr.RunForever().get();
     }
-    r = 0;
+    r = EXIT_SUCCESS;
   } catch (const std::exception &ex) {
     LOG_ERROR << ex.what();
-    r = -1;
+    r = EXIT_FAILURE;
```
eu-smirnov: It's already `= EXIT_FAILURE`, right?

src/aktualizr_primary/main.cc
```diff
@@ -147,10 +147,10 @@ int main(int argc, char *argv[]) {
...
   }
   conn.disconnect();
```
eu-smirnov: This may throw, I think. Should we move it under the try block?

src/aktualizr_primary/main.cc
mike-sul: Yeah, this line can be removed.

src/aktualizr_primary/main.cc
mike-sul: I guess, you mean line #155 (disconnect) ? If it's moved under the try block then the disconnect() won't be called if one of the previous statements throws exception.

eu-smirnov: Yes, this line. I know, it's not the scope of this PR, but since we already discussing it, maybe change it to `scoped_connection` then?

mike-sul: NP, IMHO, scoped_connection may it little bit better but it does not solve the original problem - i.e. proper tear-down/exit if disconnect() throws.

eu-smirnov: Why? If we change it to scoped_connection and move into the try block, then it will automatically disconnect if an exception is thrown in this block.

mike-sul: 'This(disconnect) may throw' was the original issue statement and it still the issue if it's called in scope of exception handling flow.

patrickvacek:
Why is the coverage getting cut so badly? Is it just because it went up with the recent unit test additions and this needs to be rebased?

mike-sul:
> Why is the coverage getting cut so badly? Is it just because it went up with the recent unit test additions and this needs to be rebased?

I've just updated the PR with the rebase from master, let's wait and see the coverage report on it.

src/aktualizr_primary/main.cc
eu-smirnov: Ah, sorry, I got your previous comment wrong. With `scoped_connection` there is no `disconnect` call, it will be disconnected on the object destruction and destructors don't throw.
Also, I don't think it's a problem that `disconnect` may throw (which I'm not 100% sure, btw), but that it will be an uncaught exception and we don't print anything.

mike-sul: ~scoped_connection()
      {
        disconnect();
      }

If disconnect() throws then scoped_connection's ctor throws too.

src/aktualizr_primary/main.cc
eu-smirnov: Do we need the `int r` at all? There are only two exit points, so maybe just `return EXIT_FAILURE`?

src/aktualizr_primary/main.cc
eu-smirnov: I'm pretty sure that boost follows the guideline that destructor cannot throw, which leads us to the conclusion that disconnect doesn't throw ;) But it would be nicer if they used `noexcept`.

eu-smirnov:
@patrickvacek LGTM

PR #1144, opened 2019-03-18
Fix/misc sql scripts

lbonn:


patrickvacek:
While you're at it... is there a way to silence the "No backward migration from version 0 to -1" (all the way up to "14 to 13")? Last I looked into it, I gave up, since that code is no longer specific to libaktualizr and I didn't want to break other users.

lbonn:
It's already in `LOG_TRACE`, don't really know what to add to that.

patrickvacek:
> It's already in `LOG_TRACE`, don't really know what to add to that.

It just dumps a lot of messages when debugging uptane vector tests. Maybe it's just me. But for a start, we should never try to go from 0 to -1.

lbonn:
Ok I've just removed the log then :)

PR #1145, opened 2019-03-18
Feat/ota 2282/more failure simulation

lbonn:
Based on #1144.

In a nutshell:

- keep a stable ecu order (what's given in StoreEcuSerials)
- device installation result code contains a concatenation of `ecu_type:error` in case of error
- added functionality on top of libfiu, uses temporary files as storage for strings

Will add a test for the changes in `SotaUptaneClient::computeDeviceInstallationResult` when ready

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -40,6 +40,23 @@ void verifyNothingInstalled(const Json::Value& manifest) {
       "noimage");
 }
 
+static Uptane::SecondaryConfig virtual_configuration(const boost::filesystem::path& client_dir) {
+  Uptane::SecondaryConfig ecu_config;
+
+  ecu_config.secondary_type = Uptane::SecondaryType::kVirtual;
+  ecu_config.partial_verifying = false;
+  ecu_config.full_client_dir = client_dir;
+  ecu_config.ecu_serial = "ecuserial3";
+  ecu_config.ecu_hardware_id = "hw_id3";
+  ecu_config.ecu_private_key = "sec.priv";
+  ecu_config.ecu_public_key = "sec.pub";
+  ecu_config.firmware_path = client_dir / "firmware.txt";
+  ecu_config.target_name_path = client_dir / "firmware_name.txt";
+  ecu_config.metadata_path = client_dir / "secondary_metadata";
+
+  return ecu_config;
+}
+
 int num_events_FullNoUpdates = 0;
 std::future<void> future_FullNoUpdates{};
 std::promise<void> promise_FullNoUpdates{};
@@ -114,17 +131,7 @@ TEST(Aktualizr, AddSecondary) {
   auto storage = INvStorage::newStorage(conf.storage);
   Aktualizr aktualizr(conf, storage, http);
 
-  Uptane::SecondaryConfig ecu_config;
-  ecu_config.secondary_type = Uptane::SecondaryType::kVirtual;
-  ecu_config.partial_verifying = false;
-  ecu_config.full_client_dir = temp_dir.Path();
-  ecu_config.ecu_serial = "ecuserial3";
-  ecu_config.ecu_hardware_id = "hw_id3";
-  ecu_config.ecu_private_key = "sec.priv";
-  ecu_config.ecu_public_key = "sec.pub";
-  ecu_config.firmware_path = temp_dir / "firmware.txt";
-  ecu_config.target_name_path = temp_dir / "firmware_name.txt";
-  ecu_config.metadata_path = temp_dir / "secondary_metadata";
+  Uptane::SecondaryConfig ecu_config = virtual_configuration(temp_dir.Path());
 
   aktualizr.AddSecondary(Uptane::SecondaryFactory::makeSecondary(ecu_config));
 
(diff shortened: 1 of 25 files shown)


OYTIS:
This pull request **introduces 2 alerts** when merging ad132cc9f840de9dbf2e03713a54c1e352a71a70 into 935e0ce455539388089f74e5bd60c281df2b7563 - [view on LGTM.com](https://lgtm.com/projects/g/advancedtelematic/aktualizr/rev/pr-882f3ac82a9b865c332c7ca0763693dfd2dcfd87)

**new alerts:**

* 2 for Illegal raise

---

lbonn:
Ok I've cleaned a bit the CMake invocations for running `test_fiuinfo`.

Last things on my list:

- cleaning up the unit test (last commit)
- actions.md
- changelog

But the bulk of the work is ready for review.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -783,10 +783,10 @@ void SQLStorage::storeEcuSerials(const EcuSerials& serials) {
...
     }
 
-    EcuSerials::const_iterator it;
-    for (it = serials.begin() + 1; it != serials.end(); it++) {
-      auto statement = db.prepareStatement<std::string, std::string>("INSERT INTO ecu_serials VALUES (?,?,0);",
-                                                                     it->first.ToString(), it->second.ToString());
+    for (auto it = serials.cbegin() + 1; it != serials.cend(); it++) {
+      auto statement = db.prepareStatement<int64_t, std::string, std::string>(
+          "INSERT INTO ecu_serials(id,serial,hardware_id) VALUES (?,?,?);", it - serials.cbegin(), it->first.ToString(),
+          it->second.ToString());
```
patrickvacek: So the ID is just purely for sequencing, and we just trust that what we receive from the server is the order we should maintain? Not a bad idea to finally make it explicit instead of assuming primary is always first.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -103,6 +103,61 @@ TEST(Aktualizr, FullNoUpdates) {
...
+
+  EcuSerials serials{
+      {Uptane::EcuSerial("primary"), Uptane::HardwareIdentifier("primary_hw")},
+      {Uptane::EcuSerial("ecuserial3"), Uptane::HardwareIdentifier("hw_id3")},
+  };
+  storage->storeEcuSerials(serials);
+
+  Aktualizr aktualizr(conf, storage, http);
+
+  // TODO: was stolen from other test, refactor
```
patrickvacek: uh-oh, a TODO!

tests/fiuinfo_test.cc
```diff
@@ -0,0 +1,32 @@
...
+
+TEST(Fiuinfo, PassInfoWithRun) {
+  EXPECT_NE(fiu_fail("fail"), 0);
+  EXPECT_EQ(fault_injection_last_info(), "test");
+}
+
+TEST(Fiuinfo, PassInfoWithCtrl) {
+  EXPECT_EQ(fiu_fail("failctrl"), 0);
+  Utils::shell(fiu_script + " ctrl -c 'enable name=failctrl,failinfo=test_ctrl' " + std::to_string(getpid()), nullptr);
+  EXPECT_NE(fiu_fail("failctrl"), 0);
```
patrickvacek: As discussed, I think checking that we can unset the failure and recover is worthwhile.

lbonn:
Pushed minor changes in the fault injection part, as it was needlessly depending on having `unsigned long` strictly bigger than 32 bits. Plus, in-line comments to make the inner workings clearer.

tests/fiuinfo_test.cc
lbonn: Should be covered now.

src/libaktualizr/storage/sqlstorage.cc
lbonn: Yes, the id is for sequencing.

I don't believe the order comes from the server, but the local config: https://github.com/advancedtelematic/aktualizr/blob/a51704b6fac25ce10f1fe024a81a01411225226f/src/libaktualizr/primary/initializer.cc#L68.

Hope I haven't overlooked anything but the intent is that any two devices with the same configuration (i.e: with same sets of ecu, deducted from configuration files named the same) would have their ecus in the same order, so that we can safely match them with the same device error code if several ecus fail.

e.g: `ecu1:fail1|ecu2:fail2` and never `ecu2:fail2:ecu1:fail1`

lbonn: > Not a bad idea to finally make it explicit instead of assuming primary is always first.

The api still expects a list of serials that start with the primary.

tests/fiuinfo_test.cc
```diff
@@ -0,0 +1,48 @@
...
+#include <unistd.h>
+
+#include "utilities/fault_injection.h"
+#include "utilities/utils.h"
+
+std::string fiu_script;
+
+TEST(Fiuinfo, PassInfoWithRun) {
+  EXPECT_TRUE(fiu_fail("fail"));
+  EXPECT_EQ(fault_injection_last_info(), "test");
```
patrickvacek: Where does "test" come from here?

lbonn: Here: https://github.com/advancedtelematic/aktualizr/blob/a51704b6fac25ce10f1fe024a81a01411225226f/tests/CMakeLists.txt#L77

patrickvacek: Thanks. It could perhaps be a more greppable string, but not a big deal at all.

src/libaktualizr/primary/aktualizr_test.cc
lbonn: ok, refactored a bit

tests/fiuinfo_test.cc
lbonn: good point, changed it to something more unique

PR #1147, opened 2019-03-20
Support custom URI for downloading targets.

patrickvacek:


doanac:
Thanks @patrickvacek! This could be a handy feature for some stuff I've been looking at.

src/libaktualizr/uptane/fetcher.cc
```diff
@@ -116,6 +116,12 @@ bool Fetcher::fetchVerifyTarget(const Target& target) {
       } else {
         ds.fhandle = storage->allocateTargetFile(false, target);
       }
+
+      std::string target_url = target.uri();
+      if (target_url.empty() || target_url.find("example.com") != std::string::npos) {
```
lbonn: What does "example.com" do here?

patrickvacek: If "example.com" is part of the custom URL, we ignore it and fall back to the default fileserver based on the autoprov URL in the credentials. This is to support backwards compatibility with the server, which often fills that field with "https://example.com/" despite that that is obviously not where we should go looking for files. It's a hack, but I'm working around something out of my control.

jochenschneider: Does the server really do that? Maybe it's just in a tutorial or something?

jochenschneider: Where is this being used? https://github.com/advancedtelematic/aktualizr/blob/master/src/sota_tools/deploy.cc#L91

patrickvacek: > Does the server really do that? Maybe it's just in a tutorial or something?

Yes, last I checked, which was yesterday.

> Where is this being used? https://github.com/advancedtelematic/aktualizr/blob/master/src/sota_tools/deploy.cc#L91

garage-deploy when credentials indicate offline signing is required.

simao: as far as I am aware the server does not put that url, I don't know how/why that is always included.

patrickvacek: IIRC we added the "example.com" bit there because there was a problem with that field being null. That was a long time ago and I'd forgotten about it. Perhaps it can be removed now?

patrickvacek: Here's where it comes from: https://github.com/advancedtelematic/meta-updater/blob/master/classes/sota.bbclass#L39

patrickvacek: Actually, it's been there for as long as meta-updater has supported garage-sign: https://github.com/advancedtelematic/meta-updater/commit/e7d4fbf5cbe8f7b89df1a047ce891ecd4ecef55a

patrickvacek: I'm going to change the URL check here to match *exactly* what we've been using ("https://example.com/") to reduce the likelihood of accidental matches, and I will open a ticket to look into removing that URL as the default.

jochenschneider: 👍

PR #1151, opened 2019-03-22
Example secondary failure simulation

lbonn:


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1391,6 +1392,15 @@ std::vector<result::Install::EcuReport> SotaUptaneClient::sendImagesToEcus(const
   }
 
   for (auto &f : firmwareFutures) {
+    // failure
```
eu-smirnov: I'm not sure I understand why we need this code here. Shouldn't we just take the failure code path below?

lbonn: For now, it's quite ugly to make it fail in two places but that's the quick way I've found. Virtual secondaries have no way to report a reason right now but if we only keep this failure, we have a discrepancy between the version reported in the manifest and the one in EcuInstallationResult (quite worrying in fact...).

That's because the manifest takes the secondary ecus version from `Uptane::SecondaryInterface::getManifest()`.

patrickvacek: It is ugly, but I also think that the problem is larger than this PR.

EpicNuts: > CI is complaining for some reason in `fetcher_test.cc`. Either way, a test for the new functionality would be a good idea.
> 
> By the way, if someone is urgently waiting on this PR, I can pick it up while you (Laurent) are away.

It would be great to have this done, as the User Story ticket about Retry is partly dependent on this. Do you have bandwidth @patrickvacek ?

patrickvacek: I rebased it and will see what CI says. If it's all good, we can merge this, but the task isn't really done, because it still needs tests. But if I can unblock you, I'll try.

PR #1156, opened 2019-03-25
Move updateImagesMeta logic into ImagesRepository

doanac:
This logic has no real dependencies on the SotaUptaneClient and helps
decouples other module's need for this class.


doanac:
I'm not sure if you guys really want a change like this right now. Its something that will be needed by the docker package manager worker I'm doing (the package manager needs to iterate targets, can't depend on sotauptaneclient since sotauptaneclient depends on packagemanager). I'd be happy to bundle this change with that patch-set if that's easier for you.

patrickvacek:
Actually, yes, this change would be very welcome, although I don't follow why the SotaUptaneClient dependency on PackageManager is a problem. However, if you're going to do this, you should also move `checkImagesMetaOffline` and the equivalent functions for the Director metadata.

I have some plans to refactor some parts of that code as well, but it shouldn't conflict with your objectives here as far as I can tell.

Travis is failing for a few reasons that seem bogus, but this one is valid:
```
/aktualizr/src/libaktualizr/uptane/imagesrepository.cc:142:63: error: the parameter 'storage' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
bool ImagesRepository::updateMeta(std::shared_ptr<INvStorage> storage, std::shared_ptr<Fetcher> fetcher) {
                                  ~~~                         ^
                                  const                      &
/aktualizr/src/libaktualizr/uptane/imagesrepository.cc:142:97: error: the parameter 'fetcher' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
bool ImagesRepository::updateMeta(std::shared_ptr<INvStorage> storage, std::shared_ptr<Fetcher> fetcher) {
                                                                       ~~~                      ^
                                                                       const                   &
```

doanac:
> although I don't follow why the SotaUptaneClient dependency on PackageManager is a problem.

Its fine for SotaUptaneClient to depend on PackageManageInterface. The problem I hit was that the DockerPackageManager I'm working on needs to iterate targets and I couldn't make it depend on SotaUptaneClient.

I'll get the PR cleaned up today. Thanks for the input.

doanac:
rebased on master and pushed with comments addressed ... hopefully

eu-smirnov:
LGTM

patrickvacek:
BTW @doanac there's a problem with our CI right now. We'll get it fixed soon and can probably merge this after that.

patrickvacek:
@doanac do you mind rebasing? We've fixed the CI issue now and hopefully there won't be any other issues once that's cleared up.

doanac:
@patrickvacek - no problem. fingers crossed.

PR #1157, opened 2019-03-26
Test/tuf test vector delegations

patrickvacek:
Please review https://github.com/advancedtelematic/tuf-test-vectors/pull/48 as well. That has to get merged first so that the version of the submodule can get bumped here. (Update: it is merged and the submodule version has been bumped accordingly.)


src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -76,11 +79,11 @@ class SotaUptaneClient {
   FRIEND_TEST(Uptane, offlineIteration);
   FRIEND_TEST(Uptane, IgnoreUnknownUpdate);
   FRIEND_TEST(Uptane, kRejectAllTest);
-  FRIEND_TEST(Uptane, Vector);  // Note hacky name (see uptane_vector_tests.cc)
   FRIEND_TEST(UptaneCI, ProvisionAndPutManifest);
   FRIEND_TEST(UptaneCI, CheckKeys);
   FRIEND_TEST(UptaneKey, Check);  // Note hacky name
   FRIEND_TEST(UptaneNetwork, DownloadFailure);
+  FRIEND_TEST(UptaneVector, Test);
```
mike-sul: At first sight, looks like too generic name.

tests/uptane_vector_tests.cc
```diff
@@ -15,178 +15,152 @@
 #include "storage/invstorage.h"
 #include "utilities/utils.h"
 
+std::string address;
+
+bool match_error(const Json::Value& error, const Uptane::Exception& e) {
+  if (error["director"]["update"]["err_msg"].asString() == e.what() ||
+      error["director"]["targets"][e.getName()]["err_msg"].asString() == e.what() ||
+      error["image_repo"]["update"]["err_msg"].asString() == e.what() ||
+      error["image_repo"]["targets"][e.getName()]["err_msg"].asString() == e.what()) {
```
mike-sul: Perhaps, it makes sense to introduce a new class-wrapper that would wrap the vector's json and introduce methods required by its clients. As result of it, business logic of accessing and interpreting the vector's json fields (including comparison with a given exception) will be encapsulated in one place in contrast to spreading it throughout the code base in each vector client.

tests/uptane_vector_tests.cc
```diff
@@ -15,178 +15,152 @@
...
+      }
+      std::vector<Uptane::Target> updates;
+      ASSERT_TRUE(uptane_client->uptaneOfflineIteration(&updates, nullptr))
+          << "uptaneOfflineIteration unexpectedly failed.";
+      if (updates.size()) {
+        /* Download a binary package.
+         * Verify a binary package. */
+        result::Download result = uptane_client->downloadImages(updates);
+        if (result.status != result::DownloadStatus::kSuccess) {
+          ASSERT_TRUE(should_fail) << "downloadImages unexpectedly failed.";
```
mike-sul: What about a case when should_fail is true and result.status is kSuccess ? Should such case ne handled too ?

src/libaktualizr/primary/sotauptaneclient.h
patrickvacek: Yes, I struggled with the naming here. The problem is that the test case is generic by definition, because it is dynamically reused to run each individual Uptane vector test. If you look at `tests/uptane_vector_tests.cc` you'll note that I have to name a class with the first part of that, and I think naming the class `Uptane` (so that the test case name is "VectorTest" or something) is too generic and potentially conflicts with the "Uptane" namespace. If you have a suggestion, I'll take it.

tests/uptane_vector_tests.cc
patrickvacek: Not a bad suggestion. As a first step, I moved all the parts that actually parse the json into a new class, which I think goes most of the way towards cleaning that up. I think getting more fine-grained is probably not necessary at this time.

mike-sul: ok.

src/libaktualizr/primary/sotauptaneclient.h
mike-sul: Yeah, I understand now better why it's called Test and don't have better name though :), perhaps it's enough to have UptaneVector as a namespace. 
Another idea, that I think, I already mentioned is to add python binding to libaktualizr (e.g. https://www.boost.org/doc/libs/1_68_0/libs/python/doc/html/tutorial/index.html) which can be useful for two things:

1) Integration & Functional tests (e.g. the uptane vector tests) written in Python so each test business logic is encapsulated in one place (a single test) instead of being spread across different files written on different languages.

2) For community/customers/app developers who would prefer to develop apps on top of libaktualizr in Python. 

This is just IMHO and I am not sure if it's feasible and there is any business & effort justification to do it all.

patrickvacek: > Another idea, that I think, I already mentioned is to add python binding to libaktualizr

That is interesting, but probably not worth the investment at present. Generally, when we use Python for testing, we are calling an executable directly and actually do not want to reference classes directly. If we were to want to do that, why not just use C++?

PR #1158, opened 2019-03-26
OTA-1684: Test installation and installation finalization failures

mike-sul:

1. Added  tests for two cases:
    a) installation failure;
    b) installation finalization failure.

2. Fixed two bugs
   a) taking into account  the `need completion` case  in the procedure that computes and stores an installation result;
   b) return false if there is no any ECU installation result records in DB in order to make it consistent with other DB methods.

### DIFF
--- a/src/libaktualizr/package_manager/packagemanagerfake.cc
+++ b/src/libaktualizr/package_manager/packagemanagerfake.cc
@@ -64,8 +64,21 @@ data::InstallationResult PackageManagerFake::finalizeInstall(const Uptane::Targe
   }
 
   data::InstallationResult install_res;
+
   if (target == targets[pending_version]) {
-    install_res = data::InstallationResult(data::ResultCode::Numeric::kOk, "Installing fake package was successful");
+    if (fiu_fail("fake_install_finalization_failure") != 0) {
+      std::string failure_cause = fault_injection_last_info();
+      if (failure_cause.empty()) {
+        install_res = data::InstallationResult(data::ResultCode::Numeric::kInstallFailed, "");
+      } else {
+        install_res =
+            data::InstallationResult(data::ResultCode(data::ResultCode::Numeric::kInstallFailed, failure_cause),
+                                     "Failed to finalize the pending update installation");
+      }
+    } else {
+      install_res = data::InstallationResult(data::ResultCode::Numeric::kOk, "Installing fake package was successful");
+    }
+
   } else {
     install_res =
         data::InstallationResult(data::ResultCode::Numeric::kInternalError, "Pending and new target do not match");
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -17,6 +17,14 @@
 #include "uptane_test_common.h"
 #include "utilities/utils.h"
 
+#ifdef FIU_ENABLE
+
+#include <fiu-control.h>
+#include <unordered_map>
+#include "utilities/fault_injection.h"
+
+#endif  // FIU_ENABLE
+
 boost::filesystem::path uptane_repos_dir;
 
 void verifyNothingInstalled(const Json::Value& manifest) {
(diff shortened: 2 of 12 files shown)


src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    size_t current_version{0};
+    size_t pending_version{0};
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+
+    // for some reason there is no any installed version at initial Aktualizr boot/run
+    // IMHO it should return currently installed version
+    ASSERT_TRUE(installed_versions.empty());
+    ASSERT_EQ(pending_version, 0);
+    ASSERT_EQ(current_version, 0);
```
patrickvacek: Historically this has been because this includes Uptane metadata that we don't have available until we've communicated with the server. We could consider storing that metadata the first time that we receive Targets metadata. However, that is out of scope for this. Feel free to make a ticket for that, though.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+
+    auto aktualizr_cycle_thread = aktualizr.RunForever();
+    auto aktualizr_cycle_thread_status = aktualizr_cycle_thread.wait_for(std::chrono::seconds(10));
```
patrickvacek: Is there a way we can do this without requiring a 10-second limit? This is both a long time if we have a lot of system resources, so we will sit idle, and too short if we are resource-starved, so the test will fail unnecessarily. `RunForever` might not be the best choice here. It also disagrees with what was described in the comment preceding this test (it mentions `UptaneCycle`, which is what I would probably prefer).

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+
+    ASSERT_EQ(aktualizr_cycle_thread_status, std::future_status::ready);
+    ASSERT_TRUE(aktualizr.uptane_client_->bootloader->rebootDetected());
+    ASSERT_TRUE(event_hdlr.checkRecievedEvents(expected_event_order));
+    ASSERT_TRUE(aktualizr.uptane_client_->hasPendingUpdates());
+    ASSERT_TRUE(http_server_mock->checkRecievedReports(expected_report_order));
+    // Aktualizr reports to a server that installation was successfull in spite the fact that
+    // the installation requires `finalization`
+    ASSERT_TRUE(http_server_mock->wasInstallSuccessfull());
```
patrickvacek: Even with your fix? That seems like a bug.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+
+    data::InstallationResult dev_installation_res;
+    std::string report;
+    std::string correlation_id;
+    ASSERT_TRUE(storage->loadDeviceInstallationResult(&dev_installation_res, &report, &correlation_id));
+    // Used to return kInstallFailed while it should be kNeedCompletion
+    EXPECT_EQ(dev_installation_res.result_code.num_code, data::ResultCode::Numeric::kNeedCompletion);
```
patrickvacek: Agreed, this looks correct.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    }
+
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+    ASSERT_EQ(installed_versions.size(), 1);
+    // why valid
+    ASSERT_TRUE(installed_versions[0].IsValid());
+    // why 0 if should be 1/true
+    ASSERT_EQ(pending_version, 0);
+    ASSERT_EQ(current_version, 0);
```
patrickvacek: Yes, it looks like there is a bug here. `IsValid` is probably not interesting here, though. It is only used in a specific circumstance to denote an unrecognized target.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(secondary_ecu_id, &installed_versions, &current_version, &pending_version));
+    ASSERT_EQ(installed_versions.size(), 1);
+    ASSERT_TRUE(installed_versions[0].IsValid());
+    ASSERT_EQ(pending_version, 0);
+    // why 0, this is secondary that was succesfully installed, so it should 1/true ???
+    ASSERT_EQ(current_version, 0);
```
patrickvacek: Probably a bug, although perhaps this doesn't get set until the entire install operation (for all ECUs) has succeeded?

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    result::UpdateCheck update_res = aktualizr.CheckUpdates().get();
+    // Is it correct expectations ?????
+    // it fails because it sees that aktualizr got higher version (2) than the one that server returns 1
+    // moreover it takes version out of the stored metadata but not from successfully installed target
+    // see sotauptaneclient.cc::413
+    ASSERT_EQ(update_res.status, result::UpdateStatus::kError);
+    ASSERT_FALSE(aktualizr.uptane_client_->hasPendingUpdates());
+    ASSERT_TRUE(event_hdlr.checkRecievedEvents({"UpdateCheckComplete"}));
+    ASSERT_TRUE(http_server_mock->checkRecievedReports({"EcuInstallationCompleted"}));
+    ASSERT_FALSE(http_server_mock->wasInstallSuccessfull());
```
patrickvacek: To be fair, this is kind of a weird case. So normally, finalization failure would imply that we couldn't boot the newly installed version, so we either rolled back to the previous version or are stuck in an unknown state. The fact that we faked failure here means that we are in a state that probably shouldn't be possible. If you are going to fake failure, then we should probably force libaktualizr to "roll back" to the previous version. Then perhaps this would make more sense.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+    ASSERT_EQ(installed_versions.size(), 1);
+    // why valid , BTW this `valid` value is neither read from DB nor calculated and it's true by default ???
+    ASSERT_TRUE(installed_versions[0].IsValid());
+    ASSERT_EQ(pending_version, 0);
+    ASSERT_EQ(current_version, 0);
+
+    current_version = 0;  // why it's not boolean
+    pending_version = 0;  // why it's not boolean
```
patrickvacek: I thought this was an SQL limitation or something like that.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    // verify currently installed version
+    std::vector<Uptane::Target> installed_versions;
+    size_t current_version{0};  // why it's not boolean
+    size_t pending_version{0};  // why it's not boolean
+
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+    ASSERT_EQ(installed_versions.size(), 1);
+    // why valid , BTW this `valid` value is neither read from DB nor calculated and it's true by default ???
+    ASSERT_TRUE(installed_versions[0].IsValid());
```
patrickvacek: See above; this doesn't really mean what you'd think it would. Perhaps it should be renamed.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+ * Checks actions:
+ *
+ * - [x] Update is not in pending state anymore after failed installation
+ * - [x] Store negative device installation result when an ECU installation failed
+ */
+TEST(Aktualizr, InstallationFailure) {
+  TemporaryDirectory temp_dir;
+  auto http_server_mock = std::make_shared<HttpInstallationFailed>(temp_dir.Path());
+  Config conf = UptaneTestCommon::makeTestConfig(temp_dir, http_server_mock->tls_server);
+  conf.uptane.polling_sec = 0;
```
patrickvacek: Why set `polling_sec`? Shouldn't be necessary if you aren't using `UptaneCycle`.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    EXPECT_EQ(ecu_installation_res.size(), 0);
+
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+    // it says that no any installed version found,
+    // which is, on one hand is correct since installation of the found update failed hence nothing was installed,
+    // on the other hand some version should have been installed prior to the failed update
+    EXPECT_EQ(installed_versions.size(), 0);
+    EXPECT_EQ(current_version, 0);
+    EXPECT_EQ(pending_version, 0);
```
patrickvacek: Yes, this is a bit odd, but since nothing got stored to that table, we should expect it to be empty. So for better or worse, for how it works now, this is probably correct.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+
+    // for some reason there is no any installed version at initial Aktualizr boot/run
+    // IMHO it should return currently installed version
+    ASSERT_TRUE(installed_versions.empty());
+    ASSERT_EQ(pending_version, 0);
+    ASSERT_EQ(current_version, 0);
+
+    aktualizr.UptaneCycle();
+    aktualizr.uptane_client_->completeInstall();
```
patrickvacek: Is it necessary to even try calling `completeInstall` if the installation failed?

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +492,355 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+  HttpInstallationFailed(const boost::filesystem::path& test_dir_in) : HttpFake(test_dir_in, "hasupdates") {}
+
+  HttpResponse handle_event(const std::string& url, const Json::Value& data) override {
+    (void)url;
+
+    for (auto& event_data : data) {
+      auto event_id = event_data["eventType"]["id"].asString();
+
+      if (event_id == "EcuInstallationCompleted") {
+        install_completion_status_ = event_data["event"]["success"].asBool();
```
patrickvacek: I think there should be an `EXPECT_FALSE` here if possible. We need to explicitly check that the reports we send have the values we expect, not just that what we read out of the database is what we expect. The low level stuff is worth checking, but it's the high level that I'm actually most concerned with.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -714,43 +714,74 @@ bool SotaUptaneClient::checkImagesMetaOffline() {
...
-        dev_result.description = "Unable to get installation results from ecus";
+      auto ecu_serial = r.first;
+      auto installation_res = r.second;
+
+      if (hw_ids.find(ecu_serial) == hw_ids.end()) {
+        // couldn't find any ECU with the given serial/ID
+        device_installation_result = data::InstallationResult(data::ResultCode::Numeric::kInternalError,
+                                                              "Unable to get installation results from ecus");
+
+        raw_installation_report = "Couldn't find any ECU with the given serial/ID: " + ecu_serial.ToString();
```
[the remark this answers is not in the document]
mike-sul: This is a serial, will correct it.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: What about storing metadata on the first boot, i.e. store metadata about packages that are pre-installed. In case of OSTree, an image does have 'installed package' which  is ostree-managed rootfs prior to communication with server. I am just thinking loudly here, not sure if there is any practical need in storing "installed versions" an initial boot time though :)

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: It actually doesn't wait under normal circumstances since RunForever() should exit just after a new update installation if force_install_completion/reboot is ON. I just wanted to check edge case when RunForever() hangs for some reason.

src/libaktualizr/primary/aktualizr_test.cc
patrickvacek: The problem is we don't have the Uptane metadata at first boot. We could consider trying to get it via garage-push/garage-sign somehow during bitbake, but that's a lot of work for something that doesn't seem critical. You are still correct that we should do _something_ to prevent installing something already installed, but pre-loading the metadata is probably unrealistic.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: It's actually a bug in my test. Aktualizr reports about both ECUs and installation for the secondary ECU succeeds.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: Actually, it's bug in my test and my incorrect understanding of loadInstalledVersions() API.
I fixed it and now everything looks logical. Just didn't get loadInstalledVersions() for the first time.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: See the above comment.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: If a failure at the installation finalization has happened then a bootloader had switched back to the previous ostree-based revision of rootfs at boot time before running aktualizr, so we don't need to emulate "rolling back" here.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: solved & understood.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: ok

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: yes, you're right.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: ok

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: Yes, if we would like emulate Aktualizr app's behavior we expect on a device. If the installation failed completeInstall will be called from RunForever().

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: Yes, it implicitly tested by changing the flavor value. If the value is not correct/expected then the multiple assertions/expectation will fail.
But, yes I agree on adding explicit check.

src/libaktualizr/primary/aktualizr_test.cc
patrickvacek: But then the version shouldn't be 2, if I understand correctly. We should have rolled back to running version 1.

mike-sul: Yes version should not be 2, but still aktualizr doesn't do any roll back related actions also it actually does know what was the previous version since installation results' life span just within UptaneCycle. 

In this specific case the 'version' is obtained from director's targets metadata what was successfully stored and has never been removed from DB in spite the fact that their installation has failed :).

CheckUpdates()->fetchMeta()->updateMeta()->uptaneIteration()->updateDirectorMeta()->{Update Director Targets Metadata}

mike-sul: Removed CheckUpdates() since it's out of scope of this test.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +500,358 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+
+    std::vector<std::pair<Uptane::EcuSerial, data::InstallationResult>> ecu_installation_res;
+    EXPECT_FALSE(storage->loadEcuInstallationResults(&ecu_installation_res));
+    EXPECT_EQ(ecu_installation_res.size(), 0);
+
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+    // it says that no any installed version found,
+    // which is, on one hand is correct since installation of the found update failed hence nothing was installed,
+    // on the other hand some version should have been installed prior to the failed update
```
patrickvacek: True. Thanks for leaving this comment. If we change this functionality in the future, this should help make it more obvious what to do with this test.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +500,358 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+    }
+
+    pending_version = SIZE_MAX;
+    current_version = SIZE_MAX;
+
+    ASSERT_TRUE(
+        storage->loadInstalledVersions(primary_ecu_id, &installed_versions, &current_version, &pending_version));
+    ASSERT_EQ(installed_versions.size(), 1);
+    ASSERT_TRUE(installed_versions[0].IsValid());
+    // if pending_version equals 0 then it means that this is a pending version
```
patrickvacek: So 0 really means "yes" for these? That is weird, or at least, not what I would've expected. Not that it's your fault or anything.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -492,6 +500,358 @@ TEST(Aktualizr, FullWithUpdatesNeedReboot) {
...
+
+    for (auto& event_data : data) {
+      auto event_id = event_data["eventType"]["id"].asString();
+      auto ecu = event_data["event"]["ecu"].asString();
+      auto correlationID = event_data["event"]["correlationId"].asString();
+
+      if (event_id == "EcuInstallationCompleted") {
+        install_completion_status_[ecu] = event_data["event"]["success"].asBool();
+        if (!install_completion_status_[ecu]) {
+          setFlavor("noupdates");
```
patrickvacek: I still think changing the flavor in the middle of a test is not a good idea. It won't really test what you expect it to, since that will result in unexpected server behavior. Is it still necessary to do something like this here?

mike-sul:
> Did you mean to change the version of tuf-test-vectors used? That looks to me to be an error. Unfortunately, if you rebased after a change to a submodule is commited, git doesn't automatically update it, and `git commit -a` will include a revert of that change, which is not ideal.
> 
> The other thing missing is that I still think we need to explicitly check the API call return values for installation and the event reports to verify that they contain what we expect them to. Unless I'm overlooking something, we are checking the database-level but not the level of what the user will actually see.

1. I didn't mean to change the tuf-test-vectors version, I did accidentally, will fix it.
2. "to explicitly check the API call return values for installation" - I don't call the installation API (neither Aktualizr::Install nor uptane_client_->uptaneInstall) directly, the  installation procedure is called indirectly from UptaneCycle(), so I cannot check the installation API return value explicitly. 
3. I added verification of values of the report events sent to the backend, see HttpInstallationFailed::wasInstallSuccessful() and its usage. Unless, you mean something else.

src/libaktualizr/package_manager/packagemanagerfake.cc
```diff
@@ -65,7 +65,13 @@ data::InstallationResult PackageManagerFake::finalizeInstall(const Uptane::Targe
 
   data::InstallationResult install_res;
   if (target == targets[pending_version]) {
-    install_res = data::InstallationResult(data::ResultCode::Numeric::kOk, "Installing fake package was successful");
+    if (fiu_fail("fake_install_finalization_failure") != 0) {
+      install_res = data::InstallationResult(data::ResultCode::Numeric::kInstallFailed,
+                                             "Failed to finalize the pending update installation");
```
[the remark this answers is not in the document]
mike-sul: What benefit does it bring in terms of installation failure ?

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: Yes, 0 means "yes" in this case since loadInstalledVersions() returns index of 'current_version' and 'pending_version' in the returned array 'installed_versions'. If there is no such version (pending and/or current) then SIZE_MAX is returned as its index value.

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: Yes, I need to prevent a client from trying to install new updates before running aktualizr again (another UptaneCycle) after installation finalization failure. 
It's possible to remove setFlavor("noupdates") and call just aktualizr.Initialize() after the installation finalization failure, but IMHO, this test won't fully corresponds to what would happen in real life scenario, in the end, the server is not supposed to advertise updates installation of which has failed.

src/libaktualizr/package_manager/packagemanagerfake.cc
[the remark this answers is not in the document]
patrickvacek: None, really. I just wanted to be consistent with the other failure points and the docs that imply that that's how it should work, and it may be useful for QA and dev/testing of other parts of the system.

src/libaktualizr/primary/aktualizr_test.cc
patrickvacek: Now I understand. Thanks!

src/libaktualizr/primary/aktualizr_test.cc
patrickvacek: > IMHO, this test won't fully corresponds to what would happen in real life scenario, in the end, the server is not supposed to advertise updates installation of which has failed.

Ironically, that is exactly what the server currently does. It continually tells us to keep retrying to install the same file that failed to install until someone explicitly cancels the update. But in theory, I agree, that really shouldn't be the case.

But the bigger issue here is that I don't think changing the flavor mid-test will do what you want it to do. I don't think the metadata in question was generated such that that will work, although I could be wrong. I think what would happen is that the http server will start serving metadata that is entirely unrelated to what libaktualizr previously received, and libaktualizr will reject all of it because it doesn't make any sense. Since the keys, hashes, lengths, signatures, and versions will all change, libaktualizr should just reject it all and assume the server got hacked. I don't think that's what you are trying to test here.

patrickvacek:
> I didn't mean to change the tuf-test-vectors version, I did accidentally, will fix it.

Looks good, thanks!

>     2. I don't call the installation API (neither Aktualizr::Install nor uptane_client_->uptaneInstall) directly, the  installation procedure is called indirectly from UptaneCycle(), so I cannot check the installation API return value explicitly.

Ideally, that would be nice to test, but it's true, that wasn't a requirement, so maybe we don't need to worry about that. I think I was just being pedantic, sorry.

> I added verification of values of the report events sent to the backend, see HttpInstallationFailed::wasInstallSuccessful() and its usage. Unless, you mean something else.

Yes, you're right, I see that now. That should be good to go. Thanks!

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: You're right saying that aktualizr will reject the metadata and doesn't try to do any updates so I can check installation results of the previous installation, otherwise they will be rewritten by results of the new update installation. Anyway, I can get rid of  setFlavor("noupdates") by not running aktualizr.UptaneCycle() for the second time just after the installation finalization failure, it suffices to run just aktualizr.Initialize() in order to verify what's required by this specific test.

src/libaktualizr/package_manager/packagemanagerfake.cc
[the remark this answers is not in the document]
mike-sul: Yes, agree that it can be useful for QA and dev testing.

mike-sul:
Updated according to the last comments.

PR #1160, opened 2019-03-27
Switch to Stretch Debian release since Jessie is not supported

mike-sul:

### DIFF
--- a/scripts/build_ubuntu.sh
+++ b/scripts/build_ubuntu.sh
@@ -23,5 +23,5 @@ cp -rf "$GITREPO_ROOT/tests/test_data/fake_root" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/create_repo.sh" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/serve_repo.py" "$TEST_INSTALL_DESTDIR"
 
-git -C "$GITREPO_ROOT" fetch --tags --unshallow || true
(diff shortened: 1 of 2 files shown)


scripts/build_ubuntu.sh
```diff
@@ -23,5 +23,5 @@ cp -rf "$GITREPO_ROOT/tests/test_data/fake_root" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/create_repo.sh" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/serve_repo.py" "$TEST_INSTALL_DESTDIR"
 
-git -C "$GITREPO_ROOT" fetch --tags --unshallow || true
+git -C "$GITREPO_ROOT" fetch --tags --depth=10000 || true
```
patrickvacek: This should be fine, but what was the motivation?

mike-sul: for some reason `unshallow`  didn't work on my local docker container.
let's wait and see if it passes through the CI pipelines.

patrickvacek: [This](https://github.com/advancedtelematic/aktualizr/commit/36429d9ddda527b6c1a90bf16b1c53fedffcd145) was the original commit that added the line. So if Travis is happy, then it's probably fine.

mike-sul: Travis is happy so let's merge it so we can pull it to another PRs.
Regarding `--unshallow`, if you run git fetch with this option against already "unshallowed" repo the you get an error, which is case if you run this docker locally.

xaviergully:
Thanks @mike-sul for bringing the ci back to normal

PR #1161, opened 2019-03-29
Fix/more ed25519

patrickvacek:
Some minor fixes found while investigating an ed25519 issue. The real issue has not actually been resolved. We still need to determine if the fix belongs on the server or client side.

### DIFF
--- a/src/libaktualizr/uptane/uptane_delegation_test.cc
+++ b/src/libaktualizr/uptane/uptane_delegation_test.cc
@@ -20,7 +20,8 @@ void delegation_basic(const boost::filesystem::path& delegation_path, bool revok
   if (revoke) {
     cmd += " revoke";
   }
-  Utils::shell(cmd, &output);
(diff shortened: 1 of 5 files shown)


src/libaktualizr/uptane/uptane_delegation_test.cc
```diff
@@ -30,7 +31,8 @@ void delegation_nested(const boost::filesystem::path& delegation_path, bool revo
   if (revoke) {
     cmd += " revoke";
   }
-  Utils::shell(cmd, &output);
+  int retval = Utils::shell(cmd, &output, true);
+  EXPECT_EQ(retval, 0) << output;
```
mike-sul: EXIT_SUCCESS ? (really minor comment)

patrickvacek: Good point, done.

PR #1162, opened 2019-03-29
 Don't require hashes and sizes of Targets objects in Snapshot metadata.

patrickvacek:
Uptane (and TUF) no longer require them. If they are there, we will use
them, and if not, we skip them. This provides no real security benefit,
but may help detect implementation faults.

For more information, see:
https://github.com/uptane/uptane-standard/issues/90
https://github.com/uptane/uptane-standard/pull/92

Please also review https://github.com/advancedtelematic/tuf-test-vectors/pull/49. That needs to get merged first.

### DIFF
--- a/src/libaktualizr/uptane/imagesrepository.cc
+++ b/src/libaktualizr/uptane/imagesrepository.cc
@@ -65,7 +65,8 @@ bool ImagesRepository::verifySnapshot(const std::string& snapshot_raw) {
 
 bool ImagesRepository::verifyRoleHashes(const std::string& role_data, const Uptane::Role& role) const {
   const std::string canonical = Utils::jsonToCanonicalStr(Utils::parseJSON(role_data));
-  bool hash_exists = false;
+  // Hashes are not required. If present, however, we may as well check them.
+  // This provides no security benefit, but may help with fault detection.
   for (const auto& it : snapshot.role_hashes(role)) {
     switch (it.type()) {
       case Hash::Type::kSha256:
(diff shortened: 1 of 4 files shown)


simao:
+1 for the idea, can't comment on the code.

src/libaktualizr/uptane/imagesrepository.cc
```diff
@@ -65,33 +65,28 @@ bool ImagesRepository::verifySnapshot(const std::string& snapshot_raw) {
...
-        hash_exists = true;
         break;
       case Hash::Type::kSha512:
         if (Hash(Hash::Type::kSha512, boost::algorithm::hex(Crypto::sha512digest(canonical))) != it) {
           LOG_ERROR << "Hash verification for " << role.ToString() << " metadata failed";
           return false;
         }
-        hash_exists = true;
         break;
       default:
```
mike-sul: The previous code version returns false in 'default' case while the PR version returns true. Is it desired behavior ?

patrickvacek: Yes, hence the [comment above](https://github.com/advancedtelematic/aktualizr/pull/1162/files#diff-55cff2ab63b7a704e8b46b05a052f4b6R69). The goal of this PR is to make this very hash check non-required, hence it returns true by default. I could've removed it entirely, but for now I decided to leave it.

src/libaktualizr/uptane/tuf.cc
```diff
@@ -361,13 +361,19 @@ void Uptane::Snapshot::init(const Json::Value &json) {
...
     auto role_object = Role(role_name, !Role::IsReserved(role_name));
 
-    role_size_[role_object] = meta_size.asInt64();
     if (meta_version.isIntegral()) {
       role_version_[role_object] = meta_version.asInt();
     } else {
       role_version_[role_object] = -1;
     }
 
+    // Size and hashes are not required, but we may as well record them if
```
mike-sul: Can we assume that hashes_list and meta_size won't return true for isObject() if there is no "hashes" or "length" in the json/meta_list ?

patrickvacek: Yep, `isObject` is false if the field is not present.

lbonn:
Maybe have the commit title and/or message mention that it concerns the Snapshot metadata?

patrickvacek:
> Maybe have the commit title and/or message mention that it concerns the Snapshot metadata?

Sure, I can do that. Can you or someone else review the tuf-test-vectors change, though, so I can merge that first? Then I can rebase this and finally get it ready.

PR #1007, opened 2018-11-14
Implement ostree pause/resume

patriotyk:

PR #1163, opened 2019-03-29
fetcher_test: create temporary repo for OSTree

eu-smirnov:
Currently, we are copying an ostree sysroot directory created by make_ostree_sysroot CMake target to use as a destination local repo for OSTree pull. For some reason, sometimes it was causing a problem on some deployments similar to the problem described and fixed by @patriotyk in https://github.com/advancedtelematic/aktualizr/pull/1007#issuecomment-441233660

This PR replaces the copy, as it's should be even faster to create an empty sysroot in a temp dir, and it also seems to fix the problem.

### DIFF
--- a/src/libaktualizr/package_manager/fetcher_test.cc
+++ b/src/libaktualizr/package_manager/fetcher_test.cc
@@ -231,11 +231,6 @@ int main(int argc, char** argv) {
   logger_init();
   logger_set_threshold(boost::log::trivial::debug);
 
-  if (argc != 2) {
-    std::cerr << "Error: " << argv[0] << " requires the path to an OSTree repository as an input argument.\n";
-    return EXIT_FAILURE;
-  }
-
   std::string port = TestUtils::getFreePort();
   server += port;
   boost::process::child http_server_process("tests/fake_http_server/fake_http_server.py", port);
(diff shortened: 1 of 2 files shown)


src/libaktualizr/package_manager/fetcher_test.cc
```diff
@@ -249,11 +249,17 @@ int main(int argc, char** argv) {
                                               std::string("--create"));
   TemporaryDirectory temp_dir;
   // Utils::copyDir doesn't work here. Complaints about non existent symlink path
```
patrickvacek: This comment about `copyDir` can be removed now.

src/libaktualizr/package_manager/fetcher_test.cc
```diff
@@ -249,11 +249,17 @@ int main(int argc, char** argv) {
...
-  int r = system((std::string("cp -r ") + argv[1] + std::string(" ") + temp_dir.PathString()).c_str());
```
patrickvacek: Does this mean we are no longer using `argv[1]`? If so, we should stop passing it in.

eu-smirnov:
@patrickvacek yes, sorry, I did a quick fix and forgot to clean it up. Thanks for noticing those things!

PR #1165, opened 2019-04-01
OTA-1684: Test events sending if target downloading fails

mike-sul:

1. Test event sending if download fails;
2. Fix the download result if partial download failure.


mike-sul:
> It looks like you are reverting tuf-test-vectors, which is probably not ideal. Still reviewing the rest, but looks good on the whole.

Yeah, did it accidentally, will fix it.

PR #1168, opened 2019-04-03
Fix null pointer bug in aktualizr-lite

doanac:
I was testing aktualizr-lite today and noticed what I think may be a new regression. If the repo server is offline, you'll get a core-dump like:
~~~
   info: Refreshing target metadata
  error: curl error 7 (http code 0): Couldn't connect to server
  error: curl error 7 (http code 0): Couldn't connect to server
  error: curl error 7 (http code 0): Couldn't connect to server
  error: Unable to update latest metadata, using local copy
   info: Updates available to raspberrypi3-64:
Segmentation fault (core dumped)
~~~

This fix does 2 things:
1) Adds exception handling for the situation in iterator.cc to make debugging future issues easier
2) Updates aktualizr-lite to be able to use the offline data if possible

### DIFF
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -31,7 +31,11 @@ static int list_main(Config &config, const bpo::variables_map &unused) {
 
   LOG_INFO << "Refreshing target metadata";
   if (!client->updateImagesMeta()) {
-    LOG_ERROR << "Unable to update latest metadata, using local copy";
+    LOG_WARNING << "Unable to update latest metadata, using local copy";
+    if (!client->checkImagesMetaOffline()) {
+      LOG_ERROR << "Unable to use local copy of TUF data";
+      return 1;
+    }
   }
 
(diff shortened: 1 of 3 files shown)


doanac:
> it's fine with me if you aren't worried about stale metadata. 

Although its potentially stale, it still *valid* correct? - ie - I'm not creating an attack vector here?

I almost wonder if this fix is useful, at least in the case of the "update" path. If we can't reach the repo server it means we don't have network access or the OTA connect server is having troubles. Either way, the odds of an update being able to apply seem low.

I like your idea of cleaning up the API. Making a complex security related API easier to get right seems like a win.

patrickvacek:
> Although its potentially stale, it still _valid_ correct? - ie - I'm not creating an attack vector here?

Yes, it is still valid, and no, it's not an attack vector.

> I almost wonder if this fix is useful

That's what I was wondering, too. :) I don't think it hurts at the moment, though.

PR #1169, opened 2019-04-04
OTA-2418: Remove example.com URL from automated garage-sign usage

mike-sul:

Verification steps:
make check & make qa

Bitbaking and running 'happy path' scenarios on qemu, see this PR https://github.com/advancedtelematic/meta-updater/pull/510 for more details.

### DIFF
--- a/src/sota_tools/deploy.cc
+++ b/src/sota_tools/deploy.cc
@@ -88,7 +88,7 @@ bool OfflineSignRepo(const ServerCredentials &push_credentials, const std::strin
     return false;
   }
 
-  std::string cmd("garage-sign targets add --repo aktualizr --format OSTREE --length 0 --url \"https://example.com/\"");
+  std::string cmd("garage-sign targets add --repo aktualizr --format OSTREE --length 0");
   cmd += " --name " + name + " --version " + hash.string() + " --sha256 " + hash.string();
   cmd += " --hardwareids " + hardwareids;
   if (system(cmd.c_str()) != 0) {


src/libaktualizr/uptane/tuf.cc
```diff
@@ -137,10 +137,7 @@ Target::Target(std::string filename, const Json::Value &content) : filename_(std
 
     if (custom.isMember("uri")) {
       std::string custom_uri = custom["uri"].asString();
-      // Ignore this exact URL for backwards compatibility with old defaults that inserted it.
-      if (custom_uri != "https://example.com/") {
-        uri_ = std::move(custom_uri);
-      }
+      uri_ = std::move(custom_uri);
```
patrickvacek: Unfortunately we probably have to keep this special rule in place for backwards compatibility. I wish we could remove it but it will break things if we do.

src/libaktualizr/package_manager/fetcher_test.cc
```diff
@@ -210,16 +210,6 @@ TEST(Fetcher, DownloadDefaultUri) {
...
-    EXPECT_TRUE(pacman->fetchTarget(target, fetcher, keys, progress_cb, nullptr));
-  }
-  {
-    // example.com (default) custom uri.
-    Json::Value target_json;
-    target_json["hashes"]["sha256"] = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855";
-    target_json["custom"]["uri"] = "https://example.com/";
-    target_json["length"] = 1;
-    Uptane::Target target("fake_file", target_json);
-
```
patrickvacek: We'll have to keep this test case, too.

src/libaktualizr/uptane/tuf.cc
mike-sul: All tests passed, so I am wondering what are the things it will break ?

patrickvacek: There are thousands of objects in production/SIT whose Target metadata lists a URL as "example.com". If someone were to use a version of libaktualizr that has my recent fix to actually use that URL field but does not know to special-case and ignore that specific URL, they won't be able to update to those targets any more.

mike-sul: thanks for explanation.

mike-sul:
Updated according to the comments.

PR #1172, opened 2019-04-09
Fix/no etc dir install

lbonn:
The fix here https://github.com/advancedtelematic/aktualizr/commit/0eccec9f5d8aeba9f409e943a9a7dcc666829b85 was incorrect as it led the `/usr/etc/sota/conf.d` directory to be installed on yocto images.

We actually don't need to create this directory from CMake, so let's remove it altogether.

Also, a typo fix


lbonn:
Note: successfully built an image with these changes

PR #1176, opened 2019-04-10
Add a separate API call to send the device manifest

eu-smirnov:


patrickvacek:
This looks fine, but can you explain the context a bit more? And we should probably add an item to the actions.md list, although it's probably about time to revisit if that list is actually useful anymore considering that the requirements management process seems to be ignoring it.

eu-smirnov:
I've added a sub-item to the `actions.md`, as there was already a section for sending device manifest.
UptaneCycle doesn't call CheckUpdates for sending the manifest but uses uptane_client->sendManifest() instead. Were you mentioning that place, @patrickvacek?

patrickvacek:
> UptaneCycle doesn't call CheckUpdates for sending the manifest but uses uptane_client->sendManifest() instead. Were you mentioning that place, @patrickvacek?

Precisely!

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -122,6 +122,12 @@ std::future<result::Install> Aktualizr::Install(const std::vector<Uptane::Target
   return api_queue_.enqueue(task);
 }
 
+std::future<bool> Aktualizr::SendManifest(const Json::Value &custom) {
+  auto promise = std::make_shared<std::promise<bool>>();
```
eu-smirnov: Funny that no code analyzer is complaining about an unused variable.

lbonn: It's probably more difficult to flag that realiably in C++, because it could just rely on side effects of constructors and destructors. For example, a mutex guard is typically not referenced outside its declaration.

patrickvacek:
Might not be strictly necessary, but can you change the `UptaneCycle` fuction to call the new `SendManifest` function instead of `uptane_client_->putManifest()`?

lbonn:
How would it look like right now, with all the other calls `putManifest` that we have around our code? To use this method, an user would have to make sure that these will never be called even if `SendManifest(xxx)` fails once, for example.

eu-smirnov:
@lbonn Could you explain why `putManifest` shouldn't be called if `SendManifest` fails?
Right now we call putManifest from `Aktualizr::CheckUpdates()`, `Aktualizr::SendDeviceData()` and `SotaUptaneClient::finalizeAfterReboot()`.

lbonn:
The three calls you mentioned won't include the custom data supplied to `SendManifest`. In the end, it could lead to the custom data to never reach the server because it was superseded by another manifest send.

eu-smirnov:
@lbonn Right, but that doesn't prevent a user from sending the manifest with custom data again, does it?

lbonn:
Sure but I'm worried that it would be ignored by backend. Would be good (but maybe impractical?) if the custom data would be sent along with all attempts after the call to `sendManifest`.

eu-smirnov:
I don't think we should try to be extra smart here and send potentially outdated custom data each time to the backend. In my view, it is better to be explicit and don't interact with the custom data in any way.
But I should probably ask Vladimir when he's back if he has the same view on this.
Still, even if we decide to add such logic later, that won't make this PR incorrect, what do you think, @lbonn?

lbonn:
Sure we can merge it now.

Just that it doesn't sound easy to use reliably in this state. Or should we have a way to inhibit other manifest sends, so that the user has really full control on that?

eu-smirnov:
@lbonn That would be a more consistent approach IMO. But, to be honest, I'm still not sure if this is what we want. I've added a description of this (inconsistent?) behavior to the API call description, so we can later return to this problem.
Any other concerns regarding this PR? @patrickvacek?
BTW, `UptaneCycle` now uses API calls instead of sotauptaneclient methods where it is possible.

eu-smirnov:
Cool, I'm merging it then, thank you @lbonn and @patrickvacek!

PR #1177, opened 2019-04-11
OTA-2480: Remove OPC-UA and move IpUptaneSecondary out of libaktualizr

mike-sul:


patrickvacek:
OPC-UA removal looks good. Thanks!

CMakeLists.txt
```diff
@@ -59,7 +58,7 @@ unset(AKTUALIZR_CHECKED_SRCS CACHE)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # find all required libraries
-set(BOOST_COMPONENTS log_setup log filesystem program_options)
+set(BOOST_COMPONENTS log_setup log system filesystem program_options)
```
patrickvacek: Why did we need to re-add the `system` boost library?

mike-sul: Oh, sorry, I forgot to answer to this question in the first place.
I couldn't link aktualizr-check-discovery without it, got linkage error saying that some boost symbol is not found.
I will double check it.

PR #1178, opened 2019-04-11
Unit test for secondary failures simulation

lbonn:


src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -807,26 +799,93 @@ TEST(Aktualizr, InstallationFailure) {
...
+  // primary and secondary failure
+  {
+    TemporaryDirectory temp_dir;
+    auto http_server_mock = std::make_shared<HttpInstallationFailed>(temp_dir.Path());
+    Config conf = UptaneTestCommon::makeTestConfig(temp_dir, http_server_mock->tls_server);
+    conf.bootloader.reboot_sentinel_dir = temp_dir.Path();
+    auto storage = INvStorage::newStorage(conf.storage);
+
+    fiu_init(0);
+    fault_injection_enable("fake_package_install", 1, "PRIMFAIL", 0);
```
mike-sul: Maybe, makes sense to define variable for "PRIMFAIL" and "SECFAIL" as it used at three places.

lbonn: frankly, I think it won't help as building `primary_hw:PRIMFAIL|secondary_hw:SECFAIL` will be pretty verbose. I prefer to have readable than concise testing code (and it would actually increase code size probably).

PR #1179, opened 2019-04-12
Use more recent glib suppression list

lbonn:
From flatpack project.

An example of these was found when running the suite under ptest.

### DIFF
--- a/scripts/build_ubuntu.sh
+++ b/scripts/build_ubuntu.sh
@@ -23,5 +23,5 @@ cp -rf "$GITREPO_ROOT/tests/test_data/fake_root" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/create_repo.sh" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/serve_repo.py" "$TEST_INSTALL_DESTDIR"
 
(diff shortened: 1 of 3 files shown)


scripts/build_ubuntu.sh
```diff
@@ -23,5 +23,5 @@ cp -rf "$GITREPO_ROOT/tests/test_data/fake_root" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/create_repo.sh" "$TEST_INSTALL_DESTDIR"
 cp -rf "$GITREPO_ROOT/src/aktualizr_repo/run/serve_repo.py" "$TEST_INSTALL_DESTDIR"
 
-git -C "$GITREPO_ROOT" fetch --tags --depth=10000 || true
+git -C "$GITREPO_ROOT" fetch --tags --unshallow || true
```
patrickvacek: Too bad, I was worried that wouldn't work.

lbonn: Yes, @mike-sul mentioned that it wouldn't work on his docker environment: https://github.com/advancedtelematic/aktualizr/pull/1160#discussion_r269637305. But shouldn't be a problem, since it's suffixed by `|| true`

lbonn:
In the end, I disabled shallow clones on Travis. A full clone is relatively fast compared to the docker container builds.

PR #1180, opened 2019-04-15
Fix/misc

patrickvacek:
Things found while doing other things.

### DIFF
--- a/src/aktualizr_repo/director_repo.cc
+++ b/src/aktualizr_repo/director_repo.cc
@@ -57,11 +57,21 @@ void DirectorRepo::signTargets() {
   boost::filesystem::remove(path_ / "repo/director/staging/targets.json");
   updateRepo();
 }
+
 void DirectorRepo::emptyTargets() {
+  const boost::filesystem::path current = path_ / "repo/director/targets.json";
   const boost::filesystem::path staging = path_ / "repo/director/staging/targets.json";
+
+  Json::Value targets_current = Utils::parseJSONFile(current);
   Json::Value targets_unsigned;
   targets_unsigned = Utils::parseJSONFile(staging);
-  targets_unsigned["targets"].clear();
+  targets_unsigned["_type"] = "Targets";
+  targets_unsigned["expires"] = expiration_time_;
+  targets_unsigned["version"] = (targets_current["signed"]["version"].asUInt()) + 1;
+  targets_unsigned["targets"] = Json::objectValue;
+  if (repo_type_ == Uptane::RepositoryType::Director() && correlation_id_ != "") {
+    targets_unsigned["custom"]["correlationId"] = correlation_id_;
+  }
   Utils::writeFile(staging, Utils::jsonToCanonicalStr(targets_unsigned));
 }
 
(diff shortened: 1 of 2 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -628,7 +628,7 @@ std::pair<bool, Uptane::Target> SotaUptaneClient::downloadImage(Uptane::Target t
     }
   }
   if (!success) {
-    LOG_ERROR << "Download unsuccessful after " << tries << " attempts.";
+    LOG_ERROR << "Download unsuccessful after " << tries - 1 << " attempts.";
```
eu-smirnov: How about `for (; tries < max_tries; tries++)` ?

src/aktualizr_repo/director_repo.cc
```diff
@@ -57,11 +57,21 @@ void DirectorRepo::signTargets() {
...
+
+  Json::Value targets_current = Utils::parseJSONFile(current);
   Json::Value targets_unsigned;
   targets_unsigned = Utils::parseJSONFile(staging);
-  targets_unsigned["targets"].clear();
+  targets_unsigned["_type"] = "Targets";
+  targets_unsigned["expires"] = expiration_time_;
+  targets_unsigned["version"] = (targets_current["signed"]["version"].asUInt()) + 1;
+  targets_unsigned["targets"] = Json::objectValue;
+  if (repo_type_ == Uptane::RepositoryType::Director() && correlation_id_ != "") {
```
mike-sul: Is this a fix for the super bug ?

src/libaktualizr/primary/sotauptaneclient.cc

src/aktualizr_repo/director_repo.cc
patrickvacek: The super bug? The fix here is just that the generated metadata was incorrect. It had the wrong meaning of "empty".

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -607,22 +607,22 @@ std::pair<bool, Uptane::Target> SotaUptaneClient::downloadImage(Uptane::Target t
...
+  const int max_tries = 3;
+  int tries = 0;
+  std::chrono::milliseconds wait(500);
+
+  for (; tries < max_tries; tries++) {
     success = package_manager_->fetchTarget(target, *uptane_fetcher, keys, prog_cb, token);
     if (success) {
       break;
-    } else if (tries < max_tries) {
+    } else if (tries < max_tries - 1) {
```
eu-smirnov: So we won't sleep in the last iteration (i = 2)? My guess would be that we don't need an `else if` here.

patrickvacek: Right. Why bother sleeping if we've tried three times and failed? No need to sleep before moving on to the next thing.

eu-smirnov: Sure. Somehow I always get confused by this simple loop.

patrickvacek: Me too. This is at least the third time we've tried to get it right despite it being so deceptively simple.

PR #1182, opened 2019-04-17
Use credentials in GitLab CI

lbonn:
OTA-2461

Thanks to @kmanikandan-kmk, who did most of the work, it was just a matter of plugging things together.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -1,5 +1,5 @@
 stages:
-  - Setup Docker
+  - Docker Setup
   - Build and Test
 
 variables:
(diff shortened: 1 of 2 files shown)


lbonn:
OTA Connect credentials are sitting in one S3 bucket and are fetched using the awscli python tool and the aws secrets supplied as environment variables in the gitlab pipelines settings (Settings>CI/CD>Environment Variables).

We can probably decommission the aktualizr side of Jenkins really soon, I'm looking at how gitlab can also display more coverage information.

The master core-image-minimal build test is still useful there IMO but it should be possible to port it as well.

patrickvacek:
> The master core-image-minimal build test is still useful there IMO but it should be possible to port it as well.

Good point. To be honest, I don't use it, but I probably should be. Even just decommissioning the aktualizr part of Jenkins now will be progress, though!

PR #1183, opened 2019-04-17
OTA-2536: Preconfigure IpUptaneSecondary with IP address

mike-sul:
1. Tiny IpUptaneSecondary refactoring 
 - to make it working with a fake package manager;
 - to get serial & hw-id from Secondary during initialization;

2. Pre-configure Secondary with IP addresses by using a json configuration file;
3. Initialize Aktualizr app with Secondaries.

### DIFF
--- a/src/aktualizr_primary/main.cc
+++ b/src/aktualizr_primary/main.cc
@@ -9,6 +9,7 @@
 #include "config/config.h"
 #include "logging/logging.h"
 #include "primary/aktualizr.h"
+#include "secondary.h"
 #include "utilities/utils.h"
 
 namespace bpo = boost::program_options;
@@ -42,6 +43,7 @@ bpo::variables_map parse_options(int argc, char *argv[]) {
       ("primary-ecu-serial", bpo::value<std::string>(), "serial number of primary ecu")
       ("primary-ecu-hardware-id", bpo::value<std::string>(), "hardware ID of primary ecu")
       ("secondary-configs-dir", bpo::value<boost::filesystem::path>(), "directory containing secondary ECU configuration files")
+      ("secondary-config-file", bpo::value<boost::filesystem::path>(), "secondary ECUs configuration file")
       ("campaign-id", bpo::value<std::string>(), "id of the campaign to act on");
   // clang-format on
 
@@ -120,6 +122,11 @@ int main(int argc, char *argv[]) {
     boost::signals2::scoped_connection conn;
 
     conn = aktualizr.SetSignalHandler(f_cb);
+
+    if (!config.uptane.secondary_config_file.empty()) {
+      Primary::initSecondaries(aktualizr, config.uptane.secondary_config_file);
+    }
+
     aktualizr.Initialize();
 
     std::string run_mode;
--- a/src/aktualizr_primary/secondary.cc
+++ b/src/aktualizr_primary/secondary.cc
@@ -1,0 +1,51 @@
+#include <unordered_map>
+
+#include "secondary.h"
+#include "secondary_config.h"
+
+#include "ipuptanesecondary.h"
+
+namespace Primary {
+
+using SecondaryFactoryRegistry =
+    std::unordered_map<std::string, std::function<std::shared_ptr<Uptane::SecondaryInterface>(const SecondaryConfig&)>>;
+
+static SecondaryFactoryRegistry sec_factory_registry = {
+    {IPSecondaryConfig::Type,
+     [](const SecondaryConfig& config) {
+       auto ip_sec_cgf = dynamic_cast<const IPSecondaryConfig&>(config);
+       return Uptane::IpUptaneSecondary::create(ip_sec_cgf.ip, ip_sec_cgf.port);
+     }},
+    //  {
+    //     Add another secondary factory here
+    //  }
+};
+
+static std::shared_ptr<Uptane::SecondaryInterface> createSecondary(const SecondaryConfig& config) {
+  return (sec_factory_registry.at(config.type()))(config);
+}
+
+void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config_file) {
+  if (!boost::filesystem::exists(config_file)) {
+    throw std::invalid_argument("Secondary ECUs config file does not exist: " + config_file.string());
+  }
+
+  auto secondary_configs = SecondaryConfigParser::parse_config_file(config_file);
+
+  for (auto& config : secondary_configs) {
+    try {
+      LOG_INFO << "Creating Secondary of type: " << config->type();
+      std::shared_ptr<Uptane::SecondaryInterface> secondary = createSecondary(*config);
+
+      LOG_INFO << "Adding Secondary to Aktualizr."
+               << "HW_ID: " << secondary->getHwId() << " Serial: " << secondary->getSerial();
+      aktualizr.AddSecondary(secondary);
+
+    } catch (const std::exception& exc) {
+      LOG_ERROR << "Failed to initialize a secondary: " << exc.what();
+      LOG_ERROR << "Continue with initialization of the remaining secondaries, if any left.";
+      // otherwise rethrow the exception
+    }
+  }
+}
+}  // namespace Primary
--- a/src/aktualizr_primary/secondary_config.cc
+++ b/src/aktualizr_primary/secondary_config.cc
@@ -1,0 +1,91 @@
+#include <fstream>
+#include <iostream>
+#include <unordered_map>
+
+#include "logging/logging.h"
+#include "secondary_config.h"
+
+namespace Primary {
+
+const char* const IPSecondaryConfig::Type = "ip";
+
+SecondaryConfigParser::Configs SecondaryConfigParser::parse_config_file(const boost::filesystem::path& config_file) {
+  if (!boost::filesystem::exists(config_file)) {
+    throw std::invalid_argument("Specified config file doesn't exist: " + config_file.string());
+  }
+
+  auto cfg_file_ext = boost::filesystem::extension(config_file);
+  std::shared_ptr<SecondaryConfigParser> cfg_parser;
+
+  if (cfg_file_ext == ".json") {
+    cfg_parser = std::make_shared<JsonConfigParser>(config_file);
+  } else {  // add your format of configuration file + implement SecondaryConfigParser specialization
+    throw std::invalid_argument("Unsupported type of config format: " + cfg_file_ext);
+  }
+
+  return cfg_parser->parse();
+}
+
+/*
+config file example
+
+{
+  "ip": [
+          {"addr": "127.0.0.1:9031"},
+          {"addr": "127.0.0.1:9032"}
+  ],
+  "socketcan": [
+          {"key": "value", "key1": "value1"},
+          {"key": "value", "key1": "value1"}
+  ]
+}
+
+*/
+
+JsonConfigParser::JsonConfigParser(const boost::filesystem::path& config_file) {
+  assert(boost::filesystem::exists(config_file));
+  std::ifstream json_file_stream(config_file.string());
+  Json::Reader json_reader;
+
+  if (!json_reader.parse(json_file_stream, root_, false)) {
+    throw std::invalid_argument("Failed to parse secondary config file: " + config_file.string() + ": " +
+                                json_reader.getFormattedErrorMessages());
+  }
+}
+
+SecondaryConfigParser::Configs JsonConfigParser::parse() {
+  Configs res_sec_cfg;
+
+  for (Json::ValueIterator it = root_.begin(); it != root_.end(); ++it) {
+    std::string secondary_type = it.key().asString();
+
+    if (sec_cfg_factory_registry_.find(secondary_type) == sec_cfg_factory_registry_.end()) {
+      LOG_ERROR << "Unsupported type of sescondary config was found: `" << secondary_type
+                << "`. Ignoring it and continuing with parsing of other secondary configs";
+    } else {
+      (sec_cfg_factory_registry_.at(secondary_type))(res_sec_cfg, *it);
+    }
+  }
+
+  return res_sec_cfg;
+}
+
+static std::pair<std::string, uint16_t> getIPAndPort(const std::string& addr) {
+  auto del_pos = addr.find_first_of(':');
+  if (del_pos == std::string::npos) {
+    throw std::invalid_argument("Incorrect address string, couldn't find port delimeter: " + addr);
+  }
+  std::string ip = addr.substr(0, del_pos);
+  uint16_t port = static_cast<uint16_t>(std::stoul(addr.substr(del_pos + 1)));
+
+  return std::make_pair(ip, port);
+}
+
+void JsonConfigParser::createIPSecondaryConfig(Configs& configs, Json::Value& ip_sec_cfgs) {
+  for (const auto& ip_sec_cfg : ip_sec_cfgs) {
+    auto addr = getIPAndPort(ip_sec_cfg[IPSecondaryConfig::AddrField].asString());
+    configs.emplace_back(std::make_shared<IPSecondaryConfig>(addr.first, addr.second));
+  }
+}
+
+}  // namespace Primary
--- a/src/aktualizr_secondary/socket_server.cc
+++ b/src/aktualizr_secondary/socket_server.cc
@@ -61,6 +61,14 @@ void SocketServer::HandleOneConnection(int socket) {
     // Figure out what to do with the message
     Asn1Message::Ptr resp = Asn1Message::Empty();
     switch (msg->present()) {
+      case AKIpUptaneMes_PR_getInfoReq: {
+        Uptane::EcuSerial serial = impl_->getSerial();
+        Uptane::HardwareIdentifier hw_id = impl_->getHwId();
+        resp->present(AKIpUptaneMes_PR_getInfoResp);
+        auto r = resp->getInfoResp();
+        SetString(&r->ecuSerial, serial.ToString());
+        SetString(&r->hwId, hw_id.ToString());
+      } break;
       case AKIpUptaneMes_PR_publicKeyReq: {
         PublicKey pk = impl_->getPublicKey();
         resp->present(AKIpUptaneMes_PR_publicKeyResp);
--- a/src/libaktualizr-posix/asn1/asn1_message.cc
+++ b/src/libaktualizr-posix/asn1/asn1_message.cc
@@ -4,6 +4,7 @@
 #include "utilities/sockaddr_io.h"
 #include "utilities/utils.h"
 
+#include <arpa/inet.h>
 #include <netinet/tcp.h>
 
 #ifndef MSG_NOSIGNAL
@@ -53,14 +54,16 @@ void SetString(OCTET_STRING_t* dest, const std::string& str) {
   OCTET_STRING_fromBuf(dest, str.c_str(), static_cast<int>(str.size()));
 }
 
-Asn1Message::Ptr Asn1Rpc(const Asn1Message::Ptr& tx, const struct sockaddr_storage& client) {
-  int socket_fd = socket(AF_INET6, SOCK_STREAM, 0);
+Asn1Message::Ptr Asn1Rpc(const Asn1Message::Ptr& tx, const struct sockaddr_in& server_sock_addr) {
+  int socket_fd = socket(AF_INET, SOCK_STREAM, 0);
   if (socket_fd < 0) {
     throw std::system_error(errno, std::system_category(), "socket");
   }
+  // what's the point in usage of SocketHandle
   SocketHandle hdl(new int(socket_fd));
-  if (connect(*hdl, reinterpret_cast<const struct sockaddr*>(&client), sizeof(sockaddr_in6)) < 0) {
-    LOG_ERROR << "connect to " << client << " failed:" << std::strerror(errno);
+
+  if (connect(*hdl, reinterpret_cast<const struct sockaddr*>(&server_sock_addr), sizeof(server_sock_addr)) < 0) {
+    LOG_ERROR << "connect to  client failed:" << std::strerror(errno);
     return Asn1Message::Empty();
   }
   der_encode(&asn_DEF_AKIpUptaneMes, &tx->msg_, Asn1SocketWriteCallback, hdl.get());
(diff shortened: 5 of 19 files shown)


OYTIS:
This pull request **introduces 1 alert** when merging f5b19d3e94c9ebfbf3c995bc26b8b332d4349107 into 99bfcf95409c7e373797097572662ba368bac09e - [view on LGTM.com](https://lgtm.com/projects/g/advancedtelematic/aktualizr/rev/pr-9ed3a9bf560e83fbb216962427b872d72fbb6a38)

**new alerts:**

* 1 for FIXME comment

---

OYTIS:
This pull request **introduces 1 alert** when merging 87325ea4c58ed29bc871be60dbba15ce96190d23 into 99bfcf95409c7e373797097572662ba368bac09e - [view on LGTM.com](https://lgtm.com/projects/g/advancedtelematic/aktualizr/rev/pr-c98e2082e1f1a69936b4606398d2eb81235c6877)

**new alerts:**

* 1 for FIXME comment

---

src/aktualizr_primary/secondary.cc
```diff
@@ -0,0 +1,51 @@
...
+#include "secondary_config.h"
+
+#include "ipuptanesecondary.h"
+
+namespace Primary {
+
+using SecondaryFactoryRegistry =
+    std::unordered_map<std::string, std::function<std::shared_ptr<Uptane::SecondaryInterface>(const SecondaryConfig&)>>;
+
+static SecondaryFactoryRegistry sec_factory_regisrty = {
```
patrickvacek: Typo: `sec_factory_registry.

src/libaktualizr-posix/asn1/messages/ipuptane_message.asn1
```diff
@@ -31,10 +31,20 @@ IpUptane DEFINITIONS ::= BEGIN
...
+  AKGetInfoRespMes ::= SEQUENCE {
+    ecuSerial UTF8String,
+    hwId UTF8String,
+    ...
+  }
+
+  AKDiscoveryReqMes ::= SEQUENCE {
+    port INTEGER(0..65535),
+    ...
+  }
```
patrickvacek: Not critical at all, but the indenting here is inconsistent.

src/libaktualizr-posix/ipuptanesecondary.h
```diff
@@ -7,22 +7,39 @@
...
+
 class IpUptaneSecondary : public SecondaryInterface {
  public:
-  explicit IpUptaneSecondary(const SecondaryConfig& conf) : SecondaryInterface(conf) {}
+  static std::shared_ptr<Uptane::SecondaryInterface> create(const std::string& address, unsigned short port);
+
+  explicit IpUptaneSecondary(const sockaddr_in& sock_addr, EcuSerial serial, HardwareIdentifier hw_id,
+                             PublicKey pub_key);
+
+  // what this method for ? Looks like should be removed out of SecondaryInterface
```
patrickvacek: It's used by `ManagedSecondary`. It's possible that it could be merged into the constructor, but I think there used to be a good reason it was necessary before. It may not be necessary now. Not priority, though. We can look into that further once more of the refactoring is done.

src/libaktualizr-posix/ipuptanesecondary.h
```diff
@@ -7,22 +7,39 @@
...
+  void Initialize() override{};
+
+  // It looks more natural to return const EcuSerial& and const Uptane::HardwareIdentifier&
+  // and they should be 'const' methods
```
patrickvacek: You might be right. In the past we didn't do that in case system calls or other non-const-friendly operations were required to implement these. We can re-evaluate that again soon.

src/libaktualizr/uptane/secondaryconfig.h
```diff
@@ -37,16 +37,18 @@ class SecondaryConfig {
   std::string ecu_public_key;
   KeyType key_type{KeyType::kRSA2048};
 
+  // FIXME: secondary config contains secondary specific params
+  // so introduction of any new type of secondary will require changes here/libaktualizr
+  // what we would like to avoid
```
patrickvacek: Yep, this is not ideal. It might be a necessary evil for now, but we should try to remove them if possible. Not the top priority, though.

src/libaktualizr/uptane/secondaryinterface.h
```diff
@@ -20,21 +20,32 @@ namespace Uptane {
 
 class SecondaryInterface {
  public:
+  // This ctor should be removed as the secondary configuration SecondaryConfig
+  // is the secondaries's specific, see SecondaryConfig declaration
```
patrickvacek: You are probably right. I think a lot of this was done to make `SecondaryFactory` work smoothly, and since we are removing that, we can get rid of some of the associated cruft along with it.

src/libaktualizr/uptane/secondaryinterface.h
```diff
@@ -20,21 +20,32 @@ namespace Uptane {
...
   explicit SecondaryInterface(SecondaryConfig sconfig_in) : sconfig(std::move(sconfig_in)) {}
   virtual ~SecondaryInterface() = default;
+  // not clear what this method for, can be removed
   virtual void Initialize(){};  // optional step, called after device registration
+  // should be pure virtual, since the current implementation reads from the secondaries specific config
```
patrickvacek: I think that makes sense.

src/libaktualizr/uptane/secondaryinterface.h
```diff
@@ -20,21 +20,32 @@ namespace Uptane {
...
   virtual EcuSerial getSerial() { return Uptane::EcuSerial(sconfig.ecu_serial); }
+  // should be pure virtual, since the current implementation reads from the secondaries specific config
   virtual Uptane::HardwareIdentifier getHwId() { return Uptane::HardwareIdentifier(sconfig.ecu_hardware_id); }
   virtual PublicKey getPublicKey() = 0;
 
+  // getSerial(), getHwId() and getPublicKey() can be moved to seperate interface
+  // since their usage pattern differ from the following methods' one
```
patrickvacek: Separate interface? What do you mean?

src/aktualizr_primary/main.cc
```diff
@@ -42,6 +43,7 @@ bpo::variables_map parse_options(int argc, char *argv[]) {
       ("primary-ecu-serial", bpo::value<std::string>(), "serial number of primary ecu")
       ("primary-ecu-hardware-id", bpo::value<std::string>(), "hardware ID of primary ecu")
       ("secondary-configs-dir", bpo::value<boost::filesystem::path>(), "directory containing secondary ECU configuration files")
+      ("secondary-config-file", bpo::value<boost::filesystem::path>(), "secondary ECUs configuration file")
```
patrickvacek: Just noting that we should get rid of `secondary-configs-dir` once we are able to. I think we have to leave both for now, but we should eventually convert virtual secondaries (and anything else still depending on the old factory mechanism) to use the new `secondary-config-file`.

src/aktualizr_primary/secondary_config.cc
```diff
@@ -0,0 +1,91 @@
...
+
+SecondaryConfigParser::Configs JsonConfigParser::parse() {
+  Configs res_sec_cfg;
+
+  for (Json::ValueIterator it = root_.begin(); it != root_.end(); ++it) {
+    std::string secondary_type = it.key().asString();
+
+    if (sec_cfg_factory_registry_.find(secondary_type) == sec_cfg_factory_registry_.end()) {
+      LOG_ERROR << "Unsupported type of sescondary config was found: `" << secondary_type
+                << "`. Ingoring it and continue with parsing of other secondary configs";
```
patrickvacek: Typo: "Ignoring". And actually "continuing" would be better than "continue".

src/libaktualizr-posix/asn1/asn1_message.cc
```diff
@@ -53,14 +54,16 @@ void SetString(OCTET_STRING_t* dest, const std::string& str) {
...
 }
 
-Asn1Message::Ptr Asn1Rpc(const Asn1Message::Ptr& tx, const struct sockaddr_storage& client) {
-  int socket_fd = socket(AF_INET6, SOCK_STREAM, 0);
+Asn1Message::Ptr Asn1Rpc(const Asn1Message::Ptr& tx, const struct sockaddr_in& server_sock_addr) {
+  int socket_fd = socket(AF_INET, SOCK_STREAM, 0);
   if (socket_fd < 0) {
     throw std::system_error(errno, std::system_category(), "socket");
   }
+  // what's the point in usage of SocketHandle
```
patrickvacek: I have no idea.

src/libaktualizr-posix/asn1/asn1_message.cc
```diff
@@ -53,14 +54,16 @@ void SetString(OCTET_STRING_t* dest, const std::string& str) {
   OCTET_STRING_fromBuf(dest, str.c_str(), static_cast<int>(str.size()));
 }
 
-Asn1Message::Ptr Asn1Rpc(const Asn1Message::Ptr& tx, const struct sockaddr_storage& client) {
-  int socket_fd = socket(AF_INET6, SOCK_STREAM, 0);
+Asn1Message::Ptr Asn1Rpc(const Asn1Message::Ptr& tx, const struct sockaddr_in& server_sock_addr) {
+  int socket_fd = socket(AF_INET, SOCK_STREAM, 0);
```
patrickvacek: `AF_INET6` is no good?

src/libaktualizr-posix/ipsecondarydiscovery.cc
```diff
@@ -79,10 +79,10 @@ std::vector<Uptane::SecondaryConfig> IpSecondaryDiscovery::waitDevices() {
     auto resp = msg->discoveryResp();
     conf.ecu_serial = ToString(resp->ecuSerial);
     conf.ecu_hardware_id = ToString(resp->hwId);
-    conf.ip_addr = sec_address;
-    Utils::setSocketPort(&conf.ip_addr, htons(static_cast<uint16_t>(resp->port)));
+    // conf.ip_addr = sec_address;
+    // Utils::setSocketPort(&conf.ip_addr, htons(static_cast<uint16_t>(resp->port)));
 
-    LOG_INFO << "Found secondary:" << conf.ecu_serial << " " << conf.ecu_hardware_id << " " << conf.ip_addr;
+    // LOG_INFO << "Found secondary:" << conf.ecu_serial << " " << conf.ecu_hardware_id << " " << conf.ip_addr;
```
patrickvacek: Is this still a work in progress?

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -1,32 +1,75 @@
...
 
-PublicKey IpUptaneSecondary::getPublicKey() {
-  LOG_INFO << "Getting the public key of a secondary";
-  Asn1Message::Ptr req(Asn1Message::Empty());
+std::shared_ptr<Uptane::SecondaryInterface> IpUptaneSecondary::create(const std::string& address, unsigned short port) {
+  LOG_INFO << "Creating IP Secondary: " << address << ":" << port << "...";
 
-  req->present(AKIpUptaneMes_PR_publicKeyReq);
+  EcuSerial serial{"serial"};
+  HardwareIdentifier hw_id{"hw_id"};
```
patrickvacek: I'd suggest initializing these to empty strings. Otherwise it might be confusing in error conditions. Especially for testing, we often use dummy values that might be easily confused with these defaults.

src/aktualizr_primary/main.cc
mike-sul: Yeah, fully agree.

src/aktualizr_primary/secondary_config.cc
mike-sul: Sure.

src/libaktualizr-posix/asn1/asn1_message.cc
mike-sul: I'll get rid of it in the next PR.

src/libaktualizr-posix/asn1/asn1_message.cc
mike-sul: Benefit of ipv6 usage is questionable in this case (local network within a car).

src/libaktualizr-posix/asn1/messages/ipuptane_message.asn1
mike-sul: I am not sure if I see it.

src/libaktualizr-posix/asn1/asn1_message.cc
lbonn: What are the drawbacks?

src/libaktualizr-posix/ipsecondarydiscovery.cc
mike-sul: That's the discovery stuff we are planning to get rid of in the future, I needed to comment these specifics code lines as they impacted by the new configuration approach implemented in this PR.

src/libaktualizr-posix/asn1/messages/ipuptane_message.asn1
patrickvacek: If you look at the surrounding lines on github, it is more apparent. I think it's a tab vs. spaces thing.

src/libaktualizr-posix/ipsecondarydiscovery.cc
patrickvacek: Sounds good, just checking!

src/libaktualizr-posix/ipuptanesecondary.cc
mike-sul: I tried, but there is an input string length verification in EcuSerial ctor what doesn't allow me to specify an empty string.
Can I remove it as someone removed it for HardwareIdentifier

src/libaktualizr-posix/ipuptanesecondary.h
mike-sul: ok.

src/libaktualizr/uptane/secondaryinterface.h
mike-sul: Well, serial, hw_id and pub_key are constant properties of a secondary and can be determined/obtained before or during a secondary object instantiation, while all other methods represent RPC calls that actually transfer messages between Primary and Secondary. I don't think, that there is  any practical need/benefit in breaking the current interface into two parts because of the aforementioned, but it might appear in the future.

src/libaktualizr-posix/ipuptanesecondary.cc
patrickvacek: I think I remember running into that before, and somehow I was able to work around it. That seems less possible here. We'd need to do a quick check that removing that restriction won't open up any holes, but it's probably fine.

src/libaktualizr-posix/asn1/messages/ipuptane_message.asn1
mike-sul: Yeah, a tab is used for indentation here while my editor inserts spaces on tap pressing.

src/libaktualizr-posix/ipuptanesecondary.cc
mike-sul: Looks like remove it is fine, at least tests passes on my local env.

src/libaktualizr/uptane/secondaryconfig.h
mike-sul: OK

src/libaktualizr-posix/ipuptanesecondary.cc
patrickvacek: I was fairly sure the tests would pass, but I mean we should look at where EcuSerials are instantiated and check if we need to add any new checks for non-emptiness in those places.

src/libaktualizr/uptane/secondaryinterface.h
mike-sul: OK. I can try it in the next PRs since it leads to to many changes across the source code.

src/libaktualizr/uptane/secondaryinterface.h
mike-sul: I'll try in the next PR.

src/aktualizr_primary/secondary.cc
```diff
@@ -0,0 +1,51 @@
...
+void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config_file) {
+  if (!boost::filesystem::exists(config_file)) {
+    throw std::invalid_argument("Secondary ECUs config file does not exist: " + config_file.string());
+  }
+
+  auto secondary_configs = SecondaryConfigParser::parse_config_file(config_file);
+
+  for (auto& config : secondary_configs) {
+    try {
+      LOG_INFO << "Creating Secondary of type: " << config->type();
```
eu-smirnov: "\n"?

src/libaktualizr-posix/asn1/asn1_message.cc
mike-sul: Well, 
1) people get used to IPv4 addresses, so easier to edit a configuration file (::1 instead of  127.0.0.1 and IPv6 slightly longer
2) At the kernel level (IP stack implementation) slightly more work for packets processing 

We consider IPUptaneSecondary as a reference implementation, so I tend to make it simple, unless there is a requirement to support IPv6

src/aktualizr_primary/secondary.cc
mike-sul: Actually LOG_XXX adds "\n" to the end of a message, so I am removing "\n" I added in other places.

src/aktualizr_primary/secondary.cc
mike-sul: ok

src/libaktualizr-posix/asn1/asn1_message.cc
patrickvacek: I actually have no strong feelings here. I'm fine with keeping it simple for now.

eu-smirnov: +1 for ipv4. I can easily imagine ipv6 stack being disabled completely in an embedded system, as it generates additional neighbor discovery multicast traffic. While I'm totally in favor of using ipv6 for large scale networking, it looks like an extra burden for an isolated network in a vehicle.

eu-smirnov: That might be not the case for IoT, though ;)

src/aktualizr_primary/secondary.cc
eu-smirnov: huh, good to know

src/aktualizr_secondary/socket_server.cc
```diff
@@ -61,6 +61,15 @@ void SocketServer::HandleOneConnection(int socket) {
     // Figure out what to do with the message
     Asn1Message::Ptr resp = Asn1Message::Empty();
     switch (msg->present()) {
+      case AKIpUptaneMes_PR_getInfoReq: {
+        std::cout << ">>>>>>>>>>>> Got Info request";
```
eu-smirnov: Debug leftover?

src/libaktualizr-posix/asn1/asn1_message.cc
eu-smirnov: I guess it automatically closes the socket, once it's out of scope.

src/libaktualizr-posix/asn1/asn1_message.cc
lbonn: I don't really buy the performance considerations if not tied to a specific platform/system architecture. Disabling v6 support also seems easier than adding it after the fact to a v4 only code base.

That said, I would have hoped that we wouldn't have to worry about either v4 or v6 by using dual stack sockets. If that's not the case and makes development harder, then ok for dropping it.

src/aktualizr_primary/main.cc
lbonn: Could be helpful to have more different naming and avoid confusion in the meantime (haven't had a great idea though). Would it also make sense to disallow using both options at once?

src/libaktualizr-posix/asn1/asn1_message.cc
patrickvacek: It's worth keeping in mind this whole functionality is just meant to be a reference and demonstration; so it is perfectly acceptable to make a semi-arbitrary choice and just roll with it.

src/aktualizr_primary/main.cc
mike-sul: I'll remove secondary-configs-dir soon (in one of the following PRs)

src/libaktualizr-posix/asn1/asn1_message.cc
eu-smirnov: Fair point, I forgot that it will actually support both AF_INET and AF_INET6. Then I'm for leaving it as it is :)

src/aktualizr_primary/main.cc
lbonn: Ok. I haven't had time to look the details in depth but is the new format really providing the old functionality?
In particular, can virtual secondaries be still used? Some docs refer to them and they are used by qa and more for testing.

src/libaktualizr-posix/asn1/asn1_message.cc
lbonn: @patrickvacek sure, sounds reasonable

src/libaktualizr-posix/asn1/asn1_message.cc
mike-sul: Yeah, your're right. Normally, such RAII objects includes object initialization/creation and deinitialization/deletion at ctor and dtor correspondingly. Also, the dtor's part should include shutdown and preferably verification if a socket was actually opened.

src/aktualizr_secondary/socket_server.cc
mike-sul: Yes.

src/libaktualizr-posix/asn1/asn1_message.cc
mike-sul: The previous version supported only AF_INET6 and it's used the 'discovery' mechanism to get IPv6 addresses. Since, we are going to remove the 'discovery' mechanism and input IP addresses in the config file then there is no need to stay with IPv6. 
I prefer to choose either IPv4 or IPv6, adding support of both will require additional effort, but I am  fine with either choices.

src/aktualizr_primary/main.cc
mike-sul: The plan is to adjust virtual secondaries implementation to the new configuration approach.

mike-sul:
> I think this looks fine unless there's still something you wanted to clean up in this PR.

No, I am fine with it.

PR #1186, opened 2019-04-18
Fix/ota 2451/empty targets

patrickvacek:
Two main fixes:
1. Check for updates even if sending the manifest fails.
2. Persist the last non-empty targets metadata received from the Director to make sure we always know what the last thing we were expected to install was.

There are tests for both in addition to some other much smaller things found while working on this.

### DIFF
--- a/src/libaktualizr/package_manager/packagemanagerfake_test.cc
+++ b/src/libaktualizr/package_manager/packagemanagerfake_test.cc
@@ -29,6 +29,7 @@ TEST(PackageManagerFake, FinalizeAfterReboot) {
   Uptane::Target target("pkg", {Uptane::Hash(Uptane::Hash::Type::kSha256, "hash")}, 1, "");
   auto result = fakepm.install(target);
   EXPECT_EQ(result.result_code, data::ResultCode::Numeric::kNeedCompletion);
+  storage->savePrimaryInstalledVersion(target, InstalledVersionUpdateMode::kPending);
 
   result = fakepm.finalizeInstall(target);
   EXPECT_EQ(result.result_code, data::ResultCode::Numeric::kOk);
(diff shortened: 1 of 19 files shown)


src/libaktualizr/primary/CMakeLists.txt
```diff
@@ -17,5 +17,7 @@ add_aktualizr_test(NAME aktualizr SOURCES aktualizr_test.cc PROJECT_WORKING_DIRE
 add_dependencies(t_aktualizr uptane_repo_full_no_correlation_id)
 
 add_aktualizr_test(NAME reportqueue SOURCES reportqueue_test.cc PROJECT_WORKING_DIRECTORY)
+add_aktualizr_test(NAME emptytargets SOURCES empty_targets_test.cc PROJECT_WORKING_DIRECTORY
```
lbonn: Inconsequential but this source file is introduced in a later commit.

patrickvacek: Oops. If anything else comes up, I can fix that, but I'm not going to rebase just for that.

patrickvacek:
Is something up with gitlab? It still hasn't detected my PR.

src/libaktualizr/package_manager/packagemanagerfake_test.cc
```diff
@@ -29,6 +29,7 @@ TEST(PackageManagerFake, FinalizeAfterReboot) {
   Uptane::Target target("pkg", {Uptane::Hash(Uptane::Hash::Type::kSha256, "hash")}, 1, "");
   auto result = fakepm.install(target);
   EXPECT_EQ(result.result_code, data::ResultCode::Numeric::kNeedCompletion);
+  storage->savePrimaryInstalledVersion(target, InstalledVersionUpdateMode::kPending);
```
mike-sul: I know this is out of scope of this PR, just noticed that savePrimaryInstalledVersion() neither returns error code nor throws exception in case of failure.

eu-smirnov: Yes, that's one of the places that I've mentioned this morning. Feel free to fix it here, of course, or you can wait for me to prepare a PR with few more similar fixes.

PR #1190, opened 2019-04-24
OTA-2536: Get a pub key along with hw_id and serial

mike-sul:
Tiny improvement to avoid two RPC calls during a secondary client initialization, i.e. get serial, hw-id and pub-key within a single RPC call.


lbonn:
Looks good. Can we get rid of `AKIpUptaneMes_PR_publicKeyReq` or can it still be useful somehow?

mike-sul:
> Looks good. Can we get rid of `AKIpUptaneMes_PR_publicKeyReq` or can it still be useful somehow?

Potentially, It can be useful if Secondary updates its keys and Primary needs to get the Secondary's  pub key dynamically without rebooting/restarting. At the moment, it's not needed so I'll remove it.

PR #1192, opened 2019-04-24
Fix issue with never-ending install retries

lbonn:
As backend sends empty director targets in some unexpected situations,
we've implemented a client-side mitigation in
02deeb5cc45e2a1fac4f6e3eb515a396cd14c1c3.

One unwanted side-effect is that after a failed installation, we'll keep
taking into account the targets containing the new version, hence
retrying installations in a loop until the backend asks to install
something new.

Second fix is to drop director targets after installation end, until
we've come to an agreement on a proper solution on both sides.

Note: would probably need an accompanying unit test, on the way :)


patrickvacek:
I think we might need to move `targets` to `latest_targets` instead of just erasing `latest_targets`.

lbonn:
Added a test for the fix. Since it works, I would vote for keeping the simple solution if it doesn't have major drawback, as it's easier to reason about and less likely (IMO) to break in the future.

PR #1195, opened 2019-04-26
Test/ostree full fake install

lbonn:
Finally :).

Well, almost:

- still has funny issues with valgrind
- would like to try to merge `fake_http_server.py` and `fake_test_server.py`, they share most of their code and both have similar non-descriptive names


lbonn:
Oh it actually looks like valgrind doesn't have any problem on CI... I have something on my debian testing, might come back on ptest, as it usually does.

src/libaktualizr/primary/aktualizr_fullostree_test.cc
```diff
@@ -0,0 +1,167 @@
...
+  server += port;
+  boost::process::child http_server_process("tests/fake_http_server/fake_test_server.py", port, "-m", meta_dir.Path());
+  TestUtils::waitForServer(server + "/");
+
+  std::string treehub_port = TestUtils::getFreePort();
+  treehub_server += treehub_port;
+  TemporaryDirectory treehub_dir;
+  boost::process::child ostree_server_process("tests/sota_tools/treehub_server.py", std::string("-p"), treehub_port,
+                                              std::string("-d"), treehub_dir.PathString(), std::string("-s0.5"),
+                                              std::string("--create"));
```
patrickvacek: Do you still need the `treehub_server` if you are using the `fake_test_server`? I'd hoped when I wrote that for it to be able to handle all server tasks in one script.

src/libaktualizr/primary/aktualizr_fullostree_test.cc
```diff
@@ -0,0 +1,167 @@
...
+  boost::filesystem::remove(conf.bootloader.reboot_sentinel_dir / conf.bootloader.reboot_sentinel_name);
+
+  {
+    Aktualizr aktualizr(conf);
+
+    aktualizr.Initialize();
+
+    result::UpdateCheck update_result = aktualizr.CheckUpdates().get();
+    ASSERT_EQ(update_result.status, result::UpdateStatus::kNoUpdatesAvailable);
+  }
```
patrickvacek: I think it would be worth checking somehow that OSTree, the manifest/installation report, or the events sent to the server correctly reflect the expected installed version.

patrickvacek:
Actually, do you think we could convince OSTree to incorporate the ability to fake a deployment somehow? It seems like something not completely unreasonable, although if it is much more complicated than what you've already done, it may or may not be worth it.

src/libaktualizr/primary/aktualizr_fullostree_test.cc
```diff
@@ -0,0 +1,167 @@
...
+  conf.pacman.type = PackageManager::kOstree;
+  conf.pacman.sysroot = sysroot.string();
+  conf.pacman.ostree_server = treehub_server;
+  conf.pacman.os = "dummy-os";
+  conf.provision.device_id = "device_id";
+  conf.provision.ecu_registration_endpoint = server + "/director/ecus";
+  conf.provision.primary_ecu_serial = "CA:FE:A6:D2:84:9D";
+  conf.provision.primary_ecu_hardware_id = "primary_hw";
+  conf.provision.provision_path = "tests/test_data/cred.zip";
+  conf.tls.server = server;
```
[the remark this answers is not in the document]
lbonn: You're right, some of those are populated in makeTestConfig

src/libaktualizr/primary/aktualizr_fullostree_test.cc
lbonn: I haven't tried, to be honest. This test is mostly a mix-and-match of other tests that we have.

However:
- treehub_server also generates a repository at startup which the test uses
- there is the cgi / multipart business that I don't quite understand.

patrickvacek: No problem. It'd be nice to clean up our various test servers, but it doesn't need to be done now. We can come to back to that later if you'd like.

src/libaktualizr/primary/aktualizr_fullostree_test.cc
lbonn: I've added a check using OstreeManager's `getCurrent()`. Would it fit the bill?

lbonn:
> Actually, do you think we could convince OSTree to incorporate the ability to fake a deployment somehow? It seems like something not completely unreasonable, although if it is much more complicated than what you've already done, it may or may not be worth it.

Hm, I don't really see ourselves convincing them to do that work for us, as I don't know if others have similar needs and that's the kind of work that could make the code base more complex without added functionality.

But could be possible if we have a somehow lightweight concrete plan (maybe make some hardcoded paths configurable?) and present a PR in an almost working state. Would require some more investigation on our side, though.

src/libaktualizr/primary/aktualizr_fullostree_test.cc
patrickvacek: Yep, good enough, I think!

lbonn:
Added updates to actions.md.
Will merge if ok, as CI passed on the previous commit.

mike-sul:
Nice work, I think, it's quite useful to fake the ostree work.

PR #1196, opened 2019-04-29
Move Travis build env to Xenial

lbonn:
The new default:

https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

+ remove deprecated sudo flag

https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Would be curious to see if it fixes our problem with deploying the last release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,15 @@ branches:
     - master
     - /^\d\d\d\d\.\d+$/
 env:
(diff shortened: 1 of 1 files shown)


patrickvacek:
Good calls. Assuming this works, when we merge it, we can update the attempted release and see if that does the trick.

lbonn:
The github api key set up in travis belonged to an user who doesn't have write access to the repo anymore. I think this is the most likely explanation.

I've updated that but we'll need an actual release attempt to see if it does work.

PR #1199, opened 2019-05-02
Resurrect provisioning and check for updates load tests

koshelev:


patrickvacek:
Looks like formatting complaints. Code seems fine at first glance.

patrickvacek:
`rebase HEAD~2 --signoff`?

lbonn:
Hm sorry formatting is still off, it needs to be done with clang-format-6.0. When installed, it will be picked up by CMake (will need a reconfigure). It's available on current Ubuntu/Debian versions from apt.

patrickvacek:
@koshelev Would it be helpful if me or someone else did a quick fix of the formatting issues here? I don't want to keep holding this up just because of minor issues.

koshelev:
Seems like I used a wrong version of formatter. Will try again

PR #1200, opened 2019-05-02
Update docs about releases.

patrickvacek:
Mostly just listing 2019.3, but also fixing some small things about the process.


patrickvacek:
Merging despite Travis failure because it's just codecov uploading and this was merely a docs change.

PR #1202, opened 2019-05-06
Doc/debugging

patrickvacek:
I've been meaning to do this for ages. @OYTIS taught me the bitbake debugging trick long ago and it's time to share it more publicly.


docs/debugging-tips.adoc
```diff
@@ -147,11 +57,13 @@ Fetch both of these down. In wireshark preferences set ssl.keylog_file to point
 
 aktualizr can be tested against a dummy repository containing fake images
 
-First, generate a repository using `aktualizr-repo` command:
+First, generate a repository using https://github.com/advancedtelematic/aktualizr/blob/master/src/aktualizr_repo/main.cc[aktualizr-repo] tool:
 
-    `aktualizr-repo generate repo_dir`
+```
+aktualizr-repo generate <repo_dir>
+```
```
mike-sul: maybe makes sense to add steps required to generate a new update(s)/images and trigger their updates:

1. aktualizr-repo image
2. aktualizr-repo addtarget
3. aktualizr-repo signtargets

or reference to this doc https://github.com/advancedtelematic/aktualizr/blob/master/docs/aktualizr-repo.adoc

patrickvacek: Good call. I added a reference to that doc. No need to duplicate the content. I also made all the links relative where possible.

docs/debugging-tips.adoc
```diff
@@ -161,4 +161,21 @@ vgdb --port=2159
...
+INHIBIT_PACKAGE_STRIP = "1"
+```
+
+You also need to make it build with debug symbols, which is recipe-dependent. For akualizr it means specifying
+
+```
+-DCMAKE_BUILD_TYPE=Debug
+```
+
+instead of Release. This may not be the best way, but it's simple and straightforward. However, it doesn't install the sources on the device, so it helps to open the source file on your host machine.
```
lbonn: "This may not be the best way, but it's simple and straightforward."

Subjective but it feels a bit superfluous in a reference doc.

patrickvacek: Agreed. I'll fix it and rebase once your PR is merged.

PR #1203, opened 2019-05-07
Do not generate a ton of secondary keys during tests

lbonn:
About 5x speedup on test_aktualizr!


lbonn:
49 min 37 sec is still a lot...

lbonn:
@patrickvacek I'll merge now, also because of the artful fix

PR #1205, opened 2019-05-08
OTA-2540: IP Secondary integration tests

mike-sul:

### DIFF
--- a/tests/ipsecondary_test.py
+++ b/tests/ipsecondary_test.py
@@ -1,0 +1,523 @@
+#!/usr/bin/env python3
+
+import tempfile
+import subprocess
+import json
+import multiprocessing
+import contextlib
+import logging
+import argparse
+
+from os import path
+from os import urandom
+from os import devnull
+from os import getcwd, chdir
+from uuid import uuid4
+from functools import wraps
+
+from fake_http_server.fake_test_server import ReUseHTTPServer as BaseTestServer
+
+logger = logging.getLogger("IPSecondaryTest")
+
+
+# The following are candidates for fixtures/common/shared components that can be reused by
+# multiple tests/test suits
+class UptaneTestRepo:
+
+    def __init__(self, repo_manager_exe):
+        self.image_rel_dir = 'repo/image'
+        self.target_rel_dir = 'repo/image/targets'
+
+        self._repo_manager_exe = repo_manager_exe
+        self._repo_root_dir = tempfile.TemporaryDirectory()
+
+    @property
+    def root_dir(self):
+        return self._repo_root_dir.name
+
+    @property
+    def image_dir(self):
+        return path.join(self.root_dir, self.image_rel_dir)
+
+    @property
+    def target_dir(self):
+        return path.join(self.root_dir, self.target_rel_dir)
+
+    @property
+    def target_file(self):
+        return path.join(self.image_dir, 'targets.json')
+
+    def add_image(self, id, image_filename, target_name=None, image_size=1024):
+
+        targetname = target_name if target_name else image_filename
+
+        with open(path.join(self.image_dir, image_filename), 'wb') as image_file:
+            image_file.write(urandom(image_size))
+
+        subprocess.run([self._repo_manager_exe, '--path', self.root_dir,
+                        '--command', 'image', '--filename', image_filename, '--targetname', targetname],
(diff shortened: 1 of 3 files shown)


tests/ipsecondary_test.py
```diff
@@ -0,0 +1,537 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._repo_root_dir.__exit__(exc_type, exc_val, exc_tb)
+
+    def _generate_repo(self):
+        subprocess.run([self._repo_manager_exe, '--path', self.root_dir, '--command', 'generate'], check=True)
+
+
+class UptaneTestBackend:
+
+    @property
```
lbonn: style question: what's the benefit of using this `@property` indirection here, compared to just defining `port` and `repo` members? Same question for constants like `image_rel_dir` of `UptaneTestRepo`.

Also, I've seen most __init__ of python classes declared at the top, is it also deliberate to have them in the middle of other methods?

tests/ipsecondary_test.py
```diff
@@ -0,0 +1,537 @@
...
+            json.dump(sec_cfg, config_file)
+
+    def run(self, run_mode):
+        subprocess.run([self._aktualizr_primary_exe, '-c', self._config_file, '--run-mode', run_mode], check=True)
+
+    def get_info(self):
+        return str(subprocess.check_output([self._aktualizr_info_exe, '-c', self._config_file]))
+
+    # ugly stuff that could be removed if Aktualizr had exposed API to check status
+    # or aktializr-info had output status/info in a structured way (e.g. json)
```
patrickvacek: I would not at all be opposed to cleaning up the output of aktualizr-info. I would keep that type of thing out of aktualizr itself, though.

tests/ipsecondary_test.py
```diff
@@ -0,0 +1,537 @@
...
+    # run aktualizr and secondary and wait until the device/aktualizr is registered
+    with aktualizr, secondary:
+        aktualizr.wait_for_completion()
+
+    # the secondary must be registered
+    if not aktualizr.is_ecu_registered(secondary.id):
+        return False
+
+    # run just aktualizr, the previously registered secondary is off
+    # and check if the primary ECU is updatable if the secondary is not connected
```
patrickvacek: How do you control/confirm that the secondary is not running here?

tests/ipsecondary_test.py
mike-sul: > what's the benefit of using this @property indirection here, compared to just defining port and repo members?

make public API explicitly defined, so users/clients don't need to learn implementation in order to figure our a class interface/API

mike-sul: > Also, I've seen most init of python classes declared at the top, is it also deliberate to have them in the middle of other methods?

Yes, your're right, the standard approach/style is to put __init__ at the beginning of a class definition, I'll update my code correspondingly.

tests/ipsecondary_test.py
mike-sul: ok. I'll elaborate my opinion during the stand-up meeting or via email.

tests/ipsecondary_test.py
mike-sul: the context manager of the 'with' statement at line 486 guarantees that IPSecondary::__exit__ is called, but I'll add additional check to make sure that the secondary's process is not running.

tests/ipsecondary_test.py
mike-sul: Just updated the PR, __init__ was moved up and useless properties are  removed

tests/ipsecondary_test.py
mike-sul: Just updated

tests/ipsecondary_test.py
lbonn: I don't want to insist on doing it at all cost but I also fail to see the changes right now. Github only shows a rebase without code changes when inspecting the last force-push.

mike-sul: Your comment is reasonable.
The code should be updated now, somehow I screwed up to commit the changes in the first place.

lbonn: Ok now I see it :+1:

patrickvacek:
Ha! Just barely beat me to it.

PR #1207, opened 2019-05-10
aktualizr-info add parameter '--images-snapshot' and '--images-timest…

xcheng-here:
aktualizr-info add parameter '--images-snapshot' and '--images-timestamp'

### DIFF
--- a/src/aktualizr_info/aktualizr_info_test.cc
+++ b/src/aktualizr_info/aktualizr_info_test.cc
@@ -596,6 +596,59 @@ TEST_F(AktualizrInfoTest, PrintDelegations) {
   }
 }
 
+/**
+ * Verifies aktualizr-info output of snapshot metadata from image repository
+ *
+ * Checks actions:
+ *
+ *  - [x] Print snapshot metadata from image repository
+ */
+TEST_F(AktualizrInfoTest, PrintImageSnapshotMetadata) {
+  Json::Value director_root_json;
+  director_root_json["key-002"] = "value-003";
+  std::string director_root = Utils::jsonToStr(director_root_json);
+  db_storage_->storeRoot(director_root, Uptane::RepositoryType::Director(), Uptane::Version(1));
+
+  Json::Value meta_snapshot;
+  meta_snapshot["signed"]["_type"] = "Snapshot";
+  meta_snapshot["signed"]["expires"] = "2038-01-19T03:14:06Z";
+  meta_snapshot["signed"]["version"] = "2";
+  std::string images_snapshot = Utils::jsonToStr(meta_snapshot);
+  db_storage_->storeNonRoot(images_snapshot, Uptane::RepositoryType::Image(), Uptane::Role::Snapshot());
+
+  aktualizr_info_process_.run({"--images-snapshot"});
+  ASSERT_FALSE(aktualizr_info_output.empty());
+
+  EXPECT_NE(aktualizr_info_output.find("snapshot.json content:"), std::string::npos);
+  EXPECT_NE(aktualizr_info_output.find(images_snapshot), std::string::npos);
+}
+
+/**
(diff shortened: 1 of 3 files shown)


src/aktualizr_info/aktualizr_info_test.cc
```diff
@@ -596,6 +596,59 @@ TEST_F(AktualizrInfoTest, PrintDelegations) {
...
+  db_storage_->storeRoot(director_root, Uptane::RepositoryType::Director(), Uptane::Version(1));
+
+  Json::Value meta_snapshot;
+  meta_snapshot["signed"]["_type"] = "Snapshot";
+  meta_snapshot["signed"]["expires"] = "2038-01-19T03:14:06Z";
+  meta_snapshot["signed"]["version"] = "2";
+  std::string images_snapshot = Utils::jsonToStr(meta_snapshot);
+  db_storage_->storeNonRoot(images_snapshot, Uptane::RepositoryType::Image(), Uptane::Role::Snapshot());
+
+  aktualizr_info_process_.run(std::vector<std::string>{"--snapshot"});
```
mike-sul: Is it really necessary to do an explicit casting here (std::vector<std::string>) ? If not, then it can be removed.

src/aktualizr_info/aktualizr_info_test.cc
```diff
@@ -596,6 +596,59 @@ TEST_F(AktualizrInfoTest, PrintDelegations) {
...
+  std::string director_root = Utils::jsonToStr(director_root_json);
+  db_storage_->storeRoot(director_root, Uptane::RepositoryType::Director(), Uptane::Version(1));
+
+  Json::Value meta_timestamp;
+  meta_timestamp["signed"]["_type"] = "Timestamp";
+  meta_timestamp["signed"]["expires"] = "2038-01-19T03:14:06Z";
+  std::string images_timestamp = Utils::jsonToStr(meta_timestamp);
+  db_storage_->storeNonRoot(images_timestamp, Uptane::RepositoryType::Image(), Uptane::Role::Timestamp());
+
+  aktualizr_info_process_.run(std::vector<std::string>{"--timestamp"});
```
mike-sul: The same as above, explicit casting can be avoided (std::vector<std::string>)

xcheng-here: Thank you Mike, I believe you are right, they should be removed.

patrickvacek:
Two minor issues that I see:
1. The DCO check is failing. You need to run `git rebase HEAD^ --signoff` to sign the DCO.
2. To be consistent with the other metadata output commands, I think we should name these commands "images-timestamp" and "images-snapshot". It's not strictly necessary since we don't support those on the director, but we may change that, so we might as well be specific about it now. (Delegations are different, because they can only exist in the images repo according to the standard.)

Thanks for the work! I've tested it and it works as expected.

src/aktualizr_info/main.cc
```diff
@@ -46,7 +46,9 @@ int main(int argc, char **argv) {
     ("delegation",  "Outputs metadata of image repo targets' delegations")
     ("director-root",  "Outputs root.json from director repo")
     ("director-target",  "Outputs targets.json from director repo")
-    ("allow-migrate", "Opens database in read/write mode to make possible to migrate database if needed");
+    ("allow-migrate", "Opens database in read/write mode to make possible to migrate database if needed")
+    ("images-snapshot", "Outputs snapshot.json from images repo")
+    ("images-timestamp", "Outputs timestamp.json from images repo");
```
patrickvacek: Very small complaint: could you move these two items between "images-root" and "images-target"? The ideal order would be: root, timestamp, snapshot, targets. That more or less matches the order they are downloaded by libaktualizr and the order in which they sign each other.

xcheng-here: Of cause. I moved them.

PR #1209, opened 2019-05-14
make docs automatically

Zee314159:
Build docs in CI to identify missing/incorrect doc issues.


lbonn:
Cool! Though we probably need to build them only on one CI docker environment. We currently have a problem with coverage (on bionic) taking too much time on Travis (despite recent improvements), so I'd suggest only building docs on debian testing where we also run the static tests.

Same idea for Jenkins and gitlab.

scripts/test.sh
```diff
@@ -36,6 +36,8 @@ TEST_PKCS11_MODULE_PATH=${TEST_PKCS11_MODULE_PATH:-/usr/lib/softhsm/libsofthsm2.
 # note: on Ubuntu bionic, use /usr/lib/engines/engine_pkcs11.so on xenial
 TEST_PKCS11_ENGINE_PATH=${TEST_PKCS11_ENGINE_PATH:-/usr/lib/x86_64-linux-gnu/engines-1.1/libpkcs11.so}
 
+BUILD_DOCS=${BUILD_DOCS:-0}
```
lbonn: Would be best if the variable followed the same format as the others (TEST_WITH_DOCS ?) and be grouped with the others , around line 10-14

scripts/test.sh
```diff
@@ -85,6 +87,15 @@ run_make() {
     CTEST_OUTPUT_ON_FAILURE=1 CTEST_PARALLEL_LEVEL="${TEST_PARALLEL_LEVEL}" make "-j${TEST_PARALLEL_LEVEL}" "$@"
 }
 
+#Make doxygen
+if [[$BUILD_DOCS = 1]]; then
```
lbonn: It does not work because there is no space between `[[` and `$`. (don't know why CI doesn't shout louder though...). Also, it's before the CMake step which shouldn't work.

Please try to follow the same pattern as the other stages: `echo ">> xxx"`, the `$TEST_DRYRUN` check, using `run_make`...
I would also put the docs building at the very end  to follow the general practice that tests more likely to fail should precede others. As it is not a test itself and should not fail, we might as well do it at the end.

lbonn:
Another note, in general we advise running [shellcheck](https://www.shellcheck.net/) against bash scripts, it's often useful to find common pitfalls. It's available on current versions of ubuntu `apt install shellcheck`

scripts/test.sh
Zee314159: Thanks. Good advise! I'll try it later.

docker/Dockerfile.debian.testing
```diff
@@ -57,7 +57,10 @@ RUN apt-get update && apt-get -y install \
   softhsm2 \
   sqlite3 \
   valgrind \
-  wget
+  wget \
+  doxygen \
+  graphviz
```
patrickvacek: Would you mind inserting these in the alphabetically-appropriate places? For comparison purposes, it makes it much easier to keep this list sorted!

Zee314159: Sure, thanks! That makes sense.

patrickvacek: I just realized this is still an issue! Do you mind fixing this?

Zee314159: I fixed this issue on the zee_test branch. Maybe I should just simply close this pull request?

patrickvacek: Sure, that's fine! So for future reference, see PR #1223.

PR #1210, opened 2019-05-14
Get rid of check-formatting.sh script

eu-smirnov:
The script location in CMakeLists.txt is relative to the aktualizr project root, which makes the check-format target unusable from upper-level projects.


CMakeLists.txt
```diff
@@ -229,8 +229,8 @@ endif()
 add_custom_target(build_tests)
 
 # clang-check and clang-format
-find_program(CLANG_FORMAT NAMES clang-format-6.0 clang-format)
-find_program(CLANG_TIDY NAMES clang-tidy-6.0 clang-tidy)
+find_program(CLANG_FORMAT NAMES clang-format clang-format-7 clang-format-6.0)
```
lbonn: The first choice should be a specific version. By taking `clang-format` in priority, we'll have eternal problems with people formatting with different versions of clang-format.

And if you'd like to switch to clang-format-7 (to be discussed, it's not available on xenial), it should be reflected on CI and the code base would have to be reformated.

eu-smirnov: I'm fine with 6.0 in CI, but locally I'm using clang-format-7 for quite a while already. Unversioned symlinks break from time to time, so it's annoying that we support only 6.0 or unversioned.
I thought it's the right thing to give the user the freedom to choose which version they want to use, but if you think that it only leads to problems, then I'm for removing it completely.

lbonn: > I'm fine with 6.0 in CI, but locally I'm using clang-format-7 for quite a while already.
The problem is that they often disagree. There shouldn't be "freedom" as we check on CI that the code complies with clang-format-6.0

eu-smirnov: Usually, they do agree, except for a couple of weird formattings from 6.0. I think we should at least give the freedom to fix those rare cases manually in case it breaks in CI and not forcing everybody to install clang-6.0.

lbonn: But it such a loss of time for everybody... Recent example: https://github.com/advancedtelematic/aktualizr/pull/1199, still unmerged.

Why is it such a problem to install clang-format 6.0 in addition to whichever favorite version? They can coexist on a same system. Even then, formatting can be run using the Debian testing docker image.

eu-smirnov: Ok, I think I'm convinced on requiring a single specific version of clang. But why not require clang-8 then? It's not a problem to install it even on Trusty: https://apt.llvm.org/ and it does a much better job then 6.0.

eu-smirnov: For now, I just remove this commit.

lbonn: Yes we can discuss about bumping the version, I agree. Didn't know about this apt.llvm.org, which seems well maintained, that's a good argument :).

PR #1211, opened 2019-05-15
Probabilistic IO failure test

lbonn:
What the test is doing: run a provisioning + update cycle with possible failing syscall (should be ~ 1 io syscall per run), then if an error occured, run it again with the same state but without injected error. The update should go through.

Right now the test uses the fake package manager, it could be more interesting to run it with ostree somehow (though more challenging).

I'll soon launch a long test with an atomicity bug and  see if it's caught as expected!

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -201,11 +201,15 @@ add_test(NAME test_ip_secondary
 add_executable(aktualizr-cycle-simple aktualizr_cycle_simple.cc)
 target_link_libraries(aktualizr-cycle-simple aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
 aktualizr_source_file_checks(aktualizr_cycle_simple.cc)
+add_dependencies(build_tests aktualizr-cycle-simple)
 
-# disabled by default, can be run with `ctest -C disabled ...`
-add_test(NAME test_io_failure CONFIGURATIONS ignored COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_io_failure.py
-    --akt-repo $<TARGET_FILE:aktualizr-repo> --akt-test $<TARGET_FILE:aktualizr-cycle-simple>
-    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+if(FAULT_INJECTION)
+    # run with a very small amount of tests on CI, should be more useful when
+    # run for several hours
+    add_test(NAME test_io_failure COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_io_failure.py
+        --akt-repo $<TARGET_FILE:aktualizr-repo> --akt-test $<TARGET_FILE:aktualizr-cycle-simple> -n 50 -j 10
+        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+endif(FAULT_INJECTION)
 
 add_dependencies(qa check)
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -198,6 +198,15 @@ set_tests_properties(test_log_negative
 add_test(NAME test_ip_secondary
          COMMAND ${PROJECT_SOURCE_DIR}/tests/ipsecondary_test.py --build-dir ${PROJECT_BINARY_DIR})
 
+add_executable(aktualizr-cycle-simple aktualizr_cycle_simple.cc)
+target_link_libraries(aktualizr-cycle-simple aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
+aktualizr_source_file_checks(aktualizr_cycle_simple.cc)
+
+# disabled by default, can be run with `ctest -C disabled ...`
+add_test(NAME test_io_failure CONFIGURATIONS ignored COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_io_failure.py
+    --akt-repo $<TARGET_FILE:aktualizr-repo> --akt-test $<TARGET_FILE:aktualizr-cycle-simple>
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+
 add_dependencies(qa check)
 
 aktualizr_source_file_checks(${TEST_SOURCES})
--- a/tests/aktualizr_cycle_simple.cc
+++ b/tests/aktualizr_cycle_simple.cc
@@ -1,0 +1,90 @@
+#include <gtest/gtest.h>
+
+#include <future>
+#include <iostream>
+#include <string>
+#include <thread>
+
+#include "config/config.h"
+#include "logging/logging.h"
+#include "package_manager/ostreemanager.h"
+#include "primary/aktualizr.h"
+#include "storage/sqlstorage.h"
+
+int updateOneCycle(const boost::filesystem::path &storage_dir, const std::string &server) {
+  Config conf;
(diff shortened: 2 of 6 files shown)


lbonn:
I've let it run about 50000 times at night. Hasn't found any issue with metadata handling. Though, I've only recently worked out an acceptable parallel performance.

With an injected bug (currently living in https://github.com/advancedtelematic/aktualizr/compare/test/OTA-2145/io-failure...wip/io-failure-with-bug), it seems to trigger a failure. I'll make measurement to see if they appear at the expected rate.

lbonn:
So the single system call bug was triggered 3 times in about 5000 runs.That's on the same order of what we expected. Running it around 50000 on a program of this size should give us good confidence.

tests/aktualizr_cycle_simple.cc
```diff
@@ -0,0 +1,90 @@
...
+  if (argc != 3) {
+    std::cerr << "Error: " << argv[0] << " requires the path to the storage directory "
+              << "and url of uptane server\n";
+    return EXIT_FAILURE;
+  }
+  boost::filesystem::path storage_dir = argv[1];
+  std::string server = argv[2];
+
+  return updateOneCycle(storage_dir, server);
+}
```
mike-sul: What's the difference between this aktualizr_cycle_simple.cc and running 'aktualizr --run-mode once' ?

tests/fake_http_server/fake_test_server.py
```diff
@@ -160,28 +169,67 @@ def do_PUT(self):
...
+        self._server_process.start()
+        logger.debug("Uptane server running and listening: {}".format(self.port))
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._server_process.terminate()
+        self._server_process.join(timeout=10)
+        logger.debug("Uptane server has been stopped")
+
+
```
mike-sul: It looks quite similar to UptaneTestBackend, perhaps it makes sense to bring it to common 'denominator' ?

tests/aktualizr_cycle_simple.cc
lbonn: The configuration is self contained, it does the step after (fake) reboot in the same process and it is set to return a success return code in two cases (no update found and installation finished).

Though it's maybe poorly named.

tests/ipsecondary_test.py
```diff
@@ -314,8 +282,10 @@ def decorator(test):
         @wraps(test)
         def wrapper(*args, **kwargs):
             repo_manager_exe_abs_path = path.abspath(repo_manager_exe)
-            with UptaneTestBackend(repo_manager_exe=repo_manager_exe_abs_path) as uptane_server:
-                result = test(*args, **kwargs, uptane_server=uptane_server)
+            with UptaneTestRepo(repo_manager_exe_abs_path) as repo, \
+                    FakeTestServerBackground(meta_path=repo.root_dir,
+                                             target_path=repo.target_dir) as uptane_server:
```
mike-sul: Oh, you actually did it :)

lbonn: Hehe yes, I've changed some details to make it easier to adapt without the UptaneTestRepo class.

tests/test_io_failure.py
```diff
@@ -0,0 +1,121 @@
...
+    parser.add_argument('--akt-test', help='path to aktualizr cycle test')
+    parser.add_argument('--serve-only', action='store_true',
+                        help='only serve metadata, do not run tests')
+    args = parser.parse_args()
+
+    srcdir = path.abspath(args.akt_srcdir) if args.akt_srcdir is not None else os.getcwd()
+
+    print(f'srcdir {srcdir}')
+
+    with tempfile.TemporaryDirectory() as repo_dir:
```
mike-sul: Why the temp dir and repo generation can't be hidden/done within FakeTestServerBackground ?, it's not important for a test, though.

mike-sul:
Great test, is this test going to be executed nightly/periodically or on each PR ?

tests/test_io_failure.py
lbonn: Keeping it loosely coupled seemed more straightforward and easier to refactor/reuse.

lbonn:
Reworked the python part slightly, to be more stylish.

@mike-sul right now, having sufficient confidence that it tried all possible faults takes too long for a CI job. It's only meant to be run manually for now but could be added as nightly job if deemed sufficiently useful. Would perhaps be nice to adapt it to use ostree and/or using less random error injections (e.g. mechanical exhaustive single fault exploration).

patrickvacek:
> It's only meant to be run manually for now but could be added as nightly job if deemed sufficiently useful.

A nightly job is not a bad idea, although it probably only needs to run if something actually changed. Alternatively, maybe even weekly is enough.

Also: we might want to consider running a really basic sample test in CI (much I'd like to with the load tests) just to make sure we don't accidentally break the process and not immediately notice.

tests/CMakeLists.txt
```diff
@@ -198,6 +198,18 @@ set_tests_properties(test_log_negative
...
 
+add_executable(aktualizr-cycle-simple aktualizr_cycle_simple.cc)
+target_link_libraries(aktualizr-cycle-simple aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
+aktualizr_source_file_checks(aktualizr_cycle_simple.cc)
+
+if(FAULT_INJECTION)
+    # run with a very small amount of tests on CI, should be more useful when
+    # run for several hours
+    add_test(NAME test_io_failure COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_io_failure.py
+        --akt-repo $<TARGET_FILE:aktualizr-repo> --akt-test $<TARGET_FILE:aktualizr-cycle-simple> -n 50 -j 10
```
patrickvacek: How long will 50 runs take? I tried to look in CI but I can't find it.

lbonn: CI failed sorry, forgot something. I've just relaunched it

lbonn: Took 10s on gitlab, 15s on jenkins. Let's see on Travis :)

patrickvacek:
Actually, one more request: can we throw a really loud warning if aktualizr is started with the dummy key type? This isn't necessary for the fake package manager, because it should be obvious that installation won't work as intended if you use it, but I really don't want someone to accidentally use the fake key type.

lbonn:
> Actually, one more request: can we throw a really loud warning if aktualizr is started with the dummy key type? This isn't necessary for the fake package manager, because it should be obvious that installation won't work as intended if you use it, but I really don't want someone to accidentally use the fake key type.

Yes you're right, I'll double check everything...

lbonn:
I've removed the dummy key type in the end, tests are not much slower with ed25519. Maybe not worth it to add these suspicious code paths.

mike-sul:
I know it's out of scope of this PR but it kind of relates to such testing as this one.
I think, it makes sense to introduce a test that interrupts an image download at random moment and random cause (timeout, force connection, etc) and verifies whether aktualizr is "updatable" and healthy after such interruption. The test should run many iterations/tries as this one (> 50k-100k). I think, it's quite important taking into account a nature of connectivity (3/4/5G) between connected cars and Cloud which implies unstable connectivity and network throughput jumps.

patrickvacek:
@mike-sul Not a bad suggestion for something to try much like this one. Want to add a ticket to the robustness epic in Jira with your suggestions?

PR #1212, opened 2019-05-15
Increase polling_sec from 10s to 3600s

alexhumphreys:
This is probably woefully inadequate, but I figured I'd get the conversation started in PR form. 

Basically it looks like people are aktualizr into production with the default, so I figure the default should be pretty high and we can set it lower when we need it. Maybe this is too high? Or still too low? Let me know what you think.

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -81,7 +81,7 @@ Options for Uptane.
 [options="header"]
 |==========================================================================================
 | Name                      | Default      | Description
-| `polling_sec`             | `10`         | Interval between polls (in seconds).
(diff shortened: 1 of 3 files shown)


patrickvacek:
After discussions with @smilingwombat, we're debating between your suggestion (once an hour) and once a day. Either way, I think we need to build some more support so that developers/testers and people following the quickstart guide can easily figure out how to change the default to a much lower value. This could be another example config in aktualizr and/or a config fragment in meta-updater, plus some tips in the docs.

We may also need to update meta-updater, e2e, and QA tests to make sure tests pass.

lbonn:
We're looking at smarter options right now.

PR #1215, opened 2019-05-22
Refactor/ota 2633/aktualizr info output

kbushgit:
Refactoring an output of aktualizr-info app.
If the user does not provide any arguments the help will be printed.
To print general information with human readable formatting, use --info.
--ecu-keys and --tls-keys was left, this commands print result more in h2m format, which requires additional parsing.
All other commands print pure information into stdout without garbage, which can be used directly in the pipe.

### DIFF
--- a/src/aktualizr_info/main.cc
+++ b/src/aktualizr_info/main.cc
@@ -38,7 +38,7 @@ int main(int argc, char **argv) {
     ("help,h", "print usage")
     ("config,c", po::value<std::vector<boost::filesystem::path> >()->composing(), "configuration file or directory")
     ("loglevel", po::value<int>(), "set log level 0-5 (trace, debug, info, warning, error, fatal)")
-    ("name-only",  "Only output device name (intended for scripting)")
+    ("name-only",  "Only output device name (intended for scripting). Cannot be used in combination with other arguments.")
     ("tls-creds",  "Outputs TLS credentials")
     ("tls-root-ca", "Outputs TLS Root CA")
     ("tls-cert", "Outputs TLS client certificate")
@@ -78,15 +78,15 @@ int main(int argc, char **argv) {
     }
 
     std::shared_ptr<INvStorage> storage = INvStorage::newStorage(config.storage, readonly);
-
+    bool cmd_trigger = false;
     std::string device_id;
-    if (!storage->loadDeviceId(&device_id)) {
-      std::cout << "Couldn't load device ID" << std::endl;
-      return EXIT_SUCCESS;
-    }
-
+    bool load_id = storage->loadDeviceId(&device_id);
     if (vm.count("name-only") != 0u) {
-      std::cout << device_id << std::endl;
+      if (!load_id) {
+        std::cout << "Couldn't load device ID\n";
+      } else {
+        std::cout << device_id << std::endl;
+      }
       return EXIT_SUCCESS;
     }
 
@@ -100,28 +100,28 @@ int main(int argc, char **argv) {
       std::cout << "Root CA certificate:" << std::endl << ca << std::endl;
       std::cout << "Client certificate:" << std::endl << cert << std::endl;
       std::cout << "Client private key:" << std::endl << pkey << std::endl;
-      return EXIT_SUCCESS;
+      cmd_trigger = true;
     }
 
     if (vm.count("tls-root-ca") != 0u) {
       std::string ca;
       storage->loadTlsCa(&ca);
-      std::cout << ca;
-      return EXIT_SUCCESS;
+      std::cout << ca << std::endl;
+      cmd_trigger = true;
     }
 
     if (vm.count("tls-cert") != 0u) {
       std::string cert;
       storage->loadTlsCert(&cert);
-      std::cout << cert;
-      return EXIT_SUCCESS;
+      std::cout << cert << std::endl;
+      cmd_trigger = true;
     }
 
     if (vm.count("tls-prv-key")) {
       std::string key;
       storage->loadTlsPkey(&key);
-      std::cout << key;
-      return EXIT_SUCCESS;
+      std::cout << key << std::endl;
+      cmd_trigger = true;
     }
 
     // ECU credentials
@@ -132,21 +132,21 @@ int main(int argc, char **argv) {
       storage->loadPrimaryKeys(&pub, &priv);
       std::cout << "Public key:" << std::endl << pub << std::endl;
       std::cout << "Private key:" << std::endl << priv << std::endl;
-      return EXIT_SUCCESS;
+      cmd_trigger = true;
     }
 
     if (vm.count("ecu-pub-key") != 0u) {
       std::string key;
       storage->loadPrimaryPublic(&key);
-      std::cout << key;
+      std::cout << key << std::endl;
       return EXIT_SUCCESS;
     }
 
     if (vm.count("ecu-prv-key") != 0u) {
       std::string key;
       storage->loadPrimaryPrivate(&key);
-      std::cout << key;
-      return EXIT_SUCCESS;
+      std::cout << key << std::endl;
+      cmd_trigger = true;
     }
 
     std::string director_root;
@@ -162,7 +162,7 @@ int main(int argc, char **argv) {
         storage->loadLatestRoot(&images_root, Uptane::RepositoryType::Image());
         std::cout << images_root << std::endl;
       }
-      return EXIT_SUCCESS;
+      cmd_trigger = true;
     }
 
     if (vm.count("images-target") != 0u) {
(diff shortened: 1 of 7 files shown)


src/aktualizr_info/main.cc
```diff
@@ -73,122 +81,66 @@ int main(int argc, char **argv) {
...
+    if (vm.count("name-only") != 0u) {
+      std::cout << device_id << std::endl;
+      return EXIT_SUCCESS;
     }
 
-    auto registered = storage->loadEcuRegistered();
-    std::cout << "Provisioned on server: " << (registered ? "yes" : "no") << std::endl;
-    std::cout << "Fetched metadata: " << (has_metadata ? "yes" : "no") << std::endl;
-
     if (has_metadata) {
```
lbonn: Maybe it was already the case before, but if the user specified one of the following options and `has_metadata` is false, only the generic help would be displayed which won't be too useful for diagnostic. It should probably fail and display an appropriate error message.

lbonn:
There are a couple of clang-tidy warnings.

Also, I'm a bit wary of the change that requires the `--info` flag to display the human readable info summary. At the moment it would break a lot of existing uses + docs. Have you found it was absolutely necessary?

src/aktualizr_info/main.cc
kbushgit: Is it mean that we should consider the case when the user chose an argument which depends on metadata and shows error message in case the has_metadat=false. 
In that case, the flow will be:
If the user chose the option which depends on has_metadata and has_metadata=false we print general info and application terminate with error code. 
In this case, the --info option will not be necessary.

lbonn: > If the user chose the option which depends on has_metadata and has_metadata=false we print general info and application terminate with error code.

Yes that's what I meant. Though, I'm not sure why it would display general info if called with `--image-targets` (for example)? Maybe an error message is enough?

kbushgit: Yes I agree with you, if you look into the code we have "if" statement if(ha_metadata) {...}
I think we can remove that and check the has_metadata variable individually for each argument which is related to it, 
for example:
if (vm.count("images-root") != 0u) {
  if (!has_metadata) {
     // print message
     return EXIT_FAILURE;
  }
  ...
}

The general info will be printed in case the user leaves arguments empty.

lbonn: yes, sounds good!

kbushgit: One question regarding Delegations, currently the aktualizr info test is designed when it expects that process will terminate always with EXIT_SUCCESS status, but actually, it's not always true, since we going to run the app as a child process we can analyze status code EXIT_SUCCESS\EXIT_FAILURE, and based on the exit status decide what we should do with stdout.
For that reason, I guess it makes sense to check exit status for each particular case.
The main idea that we must not analyze the STDOUT but the status code.
In the case of "Delegations", currently, app return EXIT_SUCESS and the message "Delegations are not present".
https://github.com/advancedtelematic/aktualizr/blob/d69f4f60d5ecc26de3ccb30a702bac693db50e54/src/aktualizr_info/aktualizr_info_test.cc#L56

patrickvacek: No delegations is a valid (successful) state, so if we are testing the case where there are delegations, we have to check stdout to verify that. @mike-sul wrote that IIRC and can comment if I'm mistaken.

lbonn:
Are you using another version of clang-tidy than 6.0 (or with other settings)? It looks like you're trying to fix warnings that are not flagged on CI.

src/aktualizr_info/main.cc
kbushgit: Ok, I understand, (no delegation it is a valid state),  
the (EXIT_SUCCESS\EXIT_FAILURE) is actually nothing more than just 0/1, and we can interpret this as (IS_DELEGATION/NO_DELEGATION).

What it gives us!
In the previous version, if there are no delegations, we can know about that only comparing string from stdout: 
if stdout == "Delegations are not present"

With  different exit code we can:
if exit code == 1

And then only if we really want to deal with that and check what happens, we can compare string fropm stdout.

Let me know if it makes sense.

patrickvacek: > we can interpret this as (IS_DELEGATION/NO_DELEGATION).

Actually, I don't think so. `EXIT_FAILURE` should only be used if there was a problem accessing the database or parsing its contents (or something similarly erroneous). `EXIT_SUCCESS` is returned whether or not there are delegations present.

>  if exit code == 1

You are correct that generally `EXIT_SUCCESS` and `EXIT_FAILURE` normally resolve to 0 and 1, but that is not guaranteed, which is why we recommend using the provided variables for comparison.

kbushgit: Ok, it's clear, thanks for the clarification, could you also check other cases with exit FAILURE, in particular, loading root metadata?

patrickvacek: I think you can see that from the source code of aktualizr-info's main.cc. The only time `EXIT_FAILURE` occurs is when there is some significant problem, like inability to read the database.

mike-sul:
Once, this PR is merged corresponding changes must be done in meta-updater as it relies on aktualizr-info in its tests (https://github.com/advancedtelematic/meta-updater/search?q=aktualizr-info&unscoped_q=aktualizr-info)

src/aktualizr_info/main.cc
```diff
@@ -72,32 +79,156 @@ int main(int argc, char **argv) {
...
+      if (!has_metadata) {
+        std::cout << msg_metadata_fail << std::endl;
+      } else {
+        std::string timestamp;
+        storage->loadNonRoot(&timestamp, Uptane::RepositoryType::Image(), Uptane::Role::Timestamp());
+        std::cout << timestamp << std::endl;
+      }
+      return EXIT_SUCCESS;
+    }
+
```
mike-sul: Regarding these multiple "if (<option-foo>) { do_foo(); return EXIT_SUCCESS;}", does it mean that it's not possible to get output of a few types of info/metadata within a single call to aktualizr-info ?

Also, these blocks of "if ... return EXIT_SUCCESS" statements look like ideal candidates for defining a map <std::string, std::function>  and having a single block of code that works for all options, e.g. "for op in options { if is_specified(op) { foo_map[op](); return EXIT_SUCCESS;}. This is rather matter of style and is minor comment that can be ignored.

patrickvacek:
>     1. What should the customer application do if aktualizr-info fails to fetch requested info ? Do polling/launch it multiple times ? Or maybe, aktualizr-info should guarantee somehow that the requested info is fetched ?

aktualizr-info is used primarily internally, so I'm not too worried about this. It would be better to do it right, but no one should be depending on this for production use cases.

>     2\. It's more convenient and reliable to use some structured format for passing info between apps, e.g. json. Perhaps, makes sense to add such functionality into aktualizr-info ?

Yes, we've considered that and I'm open to it. I believe some e2e tests rely on aktualizr-info output, though, so it's worth coordinating with them. (@rdanitz or @stevana can you confirm?)

src/aktualizr_info/main.cc
patrickvacek: The early exit is probably undesirable. Whether the suggested refactoring is worth it is debatable; I have no strong feelings on that.

stevana:
Yes, structured `aktualizr-info` output would be appreciated! (Or a way to get a specific field, e.g. `aktualizr-info --device-name`.)

src/aktualizr_info/main.cc
kbushgit: > Regarding these multiple "if () { do_foo(); return EXIT_SUCCESS;}", does it mean that it's not possible to get output of a few types of info/metadata within a single call to aktualizr-info ?
> 
> Also, these blocks of "if ... return EXIT_SUCCESS" statements look like ideal candidates for defining a map <std::string, std::function> and having a single block of code that works for all options, e.g. "for op in options { if is_specified(op) { foo_mapop; return EXIT_SUCCESS;}. This is rather matter of style and is minor comment that can be ignored.



We do not need multiple outputs since we going to use the actualizr-info output in PIPE.

The block with "if .. return" looks ugly, but it does the same job you propose but generates less of CPU instructions.

Also, can we consider our main function as a single block of code? We can use exit() vs return but this particular case it's the same.

By the way, I using this tool https://godbolt.org/ to check the compiler output, time to time

patrickvacek: > We do not need multiple outputs since we going to use the actualizr-info output in PIPE.

I disagree. There is no reason not to support multiple outputs if the user asks for it. We've supported that until now, and I see no reason to remove it.

kbushgit: Since we enable output for multiple arguments,  should we rename command --name-only?

patrickvacek: No, that is the one specific exception. I had forgotten about that; I made that as a special option for one of the backend devs. That one should indeed only print the name.

patrickvacek:
> Yes, structured `aktualizr-info` output would be appreciated!

@stevana This should be a big step in that direction. Uptane metadata is now printed so that it can be piped directly to `jq`. Keys are now printed without any other output, so it may be possible to pipe them to other tools as well.

> Or a way to get a specific field, e.g. `aktualizr-info --device-name`.

That actually already exists. `--name-only` was added by request from @jochenschneider for scripting purposes. Now it's even better; you previously had to add `--loglevel 4` to prevent any other output from being generated, but now it does that by default. The name of the flag could perhaps be better, but whatever, the functionality is there.

stevana:
@patrickvacek: Nice!

/cc @rdanitz @balajirrao

PR #1216, opened 2019-05-22
remove test_data/config_dirs, create it on the fly

xcheng-here:
those data are used for unit test of config module,
easy to create in unit test, not necessary put these
files in git repository

### DIFF
--- a/src/libaktualizr/config/config_test.cc
+++ b/src/libaktualizr/config/config_test.cc
@@ -169,9 +169,59 @@ TEST(config, TomlConsistentNonempty) {
   EXPECT_EQ(conf_str1, conf_str2);
 }
 
+static std::vector<boost::filesystem::path> generate_multi_config(TemporaryDirectory &temp_dir) {
+  std::string content;
+  {
+    content += "[storage]\n";
+    content += "path = \"path_a\"\n";
+    content += "\n";
+    content += "[pacman]\n";
+    content += "os = \"os_a\"";
+  }
+  Utils::writeFile((temp_dir / "a_dir/a.toml"), content);
+
+  content.clear();
+  {
+    content += "[storage]\n";
+    content += "path = \"path_z\"\n";
+    content += "\n";
+    content += "[pacman]\n";
+    content += "sysroot = \"sysroot_z\"";
+  }
+  Utils::writeFile((temp_dir / "a_dir/z.toml"), content);
+
+  content.clear();
+  {
+    content += "[storage]\n";
+    content += "path = \"cecond_a\"";
+  }
+  Utils::writeFile((temp_dir / "b_dir/a.toml"), content);
+
+  content.clear();
+  {
+    content += "[storage]\n";
+    content += "path = \"latest_path\"\n";
+    content += "\n";
+    content += "[provision]\n";
+    content += "provision_path = \"y_prov_path\"";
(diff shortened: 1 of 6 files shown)


src/libaktualizr/config/config_test.cc
```diff
@@ -169,19 +169,35 @@ TEST(config, TomlConsistentNonempty) {
   EXPECT_EQ(conf_str1, conf_str2);
 }
 
+static std::vector<std::string> generate_multi_config(TemporaryDirectory &temp_dir) {
+  Utils::writeFile((temp_dir / "a_dir/a.toml"), std::string("[storage]\npath = \"path_a\"\n\n[pacman]\nos = \"os_a\""));
```
lbonn: For a bit more readability, it could use the same style as in the `TwoTomlCorrectness` test, where new lines are reflected in the source.

src/libaktualizr/config/config_test.cc
```diff
@@ -169,19 +169,35 @@ TEST(config, TomlConsistentNonempty) {
...
+  Utils::writeFile((temp_dir / "a_dir/z.toml"),
+                   std::string("[storage]\npath = \"path_z\"\n\n[pacman]\nsysroot = \"sysroot_z\""));
+
+  Utils::writeFile((temp_dir / "b_dir/a.toml"), std::string("[storage]\npath = \"cecond_a\""));
+  Utils::writeFile((temp_dir / "b_dir/y.toml"),
+                   std::string("[storage]\npath = \"latest_path\"\n\n[provision]\nprovision_path = \"y_prov_path\""));
+  Utils::writeFile((temp_dir / "b_dir/z.txt"), std::string("[storage]\npath = \"this is path from text file\""));
+
+  return std::vector<std::string>{(temp_dir / "a_dir").c_str(), (temp_dir / "b_dir").c_str()};
```
lbonn: boost::filesystem::path has a `.string()` method which converts into std::string. If you use c_str(), it would probably do two conversions needlessly.

lbonn:
It looks like the reason codecov is showing a big coverage decrease is that your branch is based on fef78ccdd3f63e8b7959cf554d8c65c823a89ffe, which is a few days old and did not include Mike's new test. If you rebase your branch on master, it should show better numbers.

src/libaktualizr/config/config_test.cc
```diff
@@ -169,19 +169,35 @@ TEST(config, TomlConsistentNonempty) {
...
+}
 /* Parse multiple config files in a directory. */
 TEST(config, OneDir) {
-  Config config(std::vector<boost::filesystem::path>{"tests/test_data/config_dirs/one_dir"});
+  TemporaryDirectory temp_dir;
+  std::vector<std::string> &&dirs = generate_multi_config(temp_dir);
```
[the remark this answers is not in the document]
xcheng-here: No special reason. In my understanding, r-value reference could avoid an unnecessary copy and that function could return an unamed variant in stack. I believe it also works if I declare variant out side the generate_multi_config() function and pass it's reference into that function, basically the cost of those two ways are same.

src/libaktualizr/config/config_test.cc
xcheng-here: OK, thanks, I'll modify it.

PR #1219, opened 2019-05-27
OTA-2543: Docs on config and usage of local fake secondary

mike-sul:


docs/configuration.adoc
```diff
@@ -87,6 +87,7 @@ Options for Uptane.
 | `key_source`              | `"file"`     | Where to read the device's private key from. Options: `"file"`, `"pkcs11"`.
 | `key_type`                | `"RSA2048"`  | Type of cryptographic keys to use. Options: `"ED25519"`, `"RSA2048"`, `"RSA3072"` or `"RSA4096"`.
 | `secondary_configs_dir`   | `""`         | Directory containing individual secondary json configuration files. Example here: link:{aktualizr-github-url}/config/secondary/virtualsec.json[]
+| `secondary_config_file`   | `""`         | A path to a json file containing configuration of Secondary ECU(s) for Primary
```
patrickvacek: Perhaps worth pointing to the example and also explaining why we have both this and `secondary_configs_dir`. It's probably worth stating that we want to get rid of `secondary_configs_dir`, too.

config/posix_secondary.toml
```diff
@@ -0,0 +1,18 @@
...
+[network]
+port = 9050
+primary_ip = "127.0.0.1"
+primary_port = 9040
+
+[pacman]
+type = "fake"
+
+[logger]
+loglevel = 0
```
patrickvacek: Will anything be using this config, or is it just an example? Is it possible to have meta-updater use these?

docs/posix-secondaries.adoc
```diff
@@ -0,0 +1,86 @@
+= Posix Secondaries aka IP Secondaries
+
+Secondary is a term introduced by link:https://uptane.github.io/uptane-standard/uptane-standard.html[the uptane spec].
```
patrickvacek: Uptane isn't the origin of the term "secondary", but I think referencing the spec is not a bad idea. I'd suggest moving this first paragraph after the second ("The goal of this doc...") and then saying that the term secondary is being used as it is defined by the Uptane standard.

docs/posix-secondaries.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+The goal of this doc is to guide a reader on configuring and running two executables that can be build out of the given repository source code and act as Primary and Secondary correspondingly.
+
+A key component of this repository is libaktualizr which is a library implementing Uptane protocol and providing customers and community with capability to build application(s) one of the purposes of which is to update software/firmware/packages in compliance with the Uptane standard. In addition to libaktualizr, an user can find in this repo another two components:
+
+. *aktualizr*, aka aktualizr primary - a reference implementation of an application utilizing libaktualizr, its binary/executable is called aktualizr. Aktualizr is used to emulate/"fake" Primary locally on an user's host;
+. *aktualizr-secondary*, aka posix or IP secondary - a reference implementation intended for running on Secondary ECU and capable to communicate with Primary ECU aktualizr via TCP/IP protocol (implemenation relies on Posix API). Aktualizr-secondary is used to "fake" Secondary ECU on an user's host.
```
patrickvacek: I would make it clear that "Aktualizr can be used to emulate a Primary locally...". The *can be* is critical, since it can also be used to update a real primary! Same goes for the secondary.

docs/posix-secondaries.adoc
```diff
@@ -0,0 +1,86 @@
+= Posix Secondaries aka IP Secondaries
+
+Secondary is a term introduced by link:https://uptane.github.io/uptane-standard/uptane-standard.html[the uptane spec].
+It refers to ECU software/firmware/package updates of which are managed by another ECU that hosts and runs Uptane compliant software. The later ECU is called Primary and the sotfware it runs in our case is called aktualizr.
+
+The goal of this doc is to guide a reader on configuring and running two executables that can be build out of the given repository source code and act as Primary and Secondary correspondingly.
```
patrickvacek: "can be build" -> "can be built"

docs/posix-secondaries.adoc
```diff
@@ -0,0 +1,86 @@
...
+* `secondaries_wait_port` - TCP port aktualizr listen on for connections from Secondaries
+* `secondaries_wait_timeout` - timeout (in sec) of waiting for connections from Secondaries. Primary/aktualizr waits for a connection from those secondaries that it failed to connect to at the startup time.
+* `secondaries` -  a list of Secondary TCP/IP addresses
+
+Please, put your credential.zip file into the current working directory or update `[provision] provision_path` in link:../config/sota_local_with_secondaries.toml[the config] so it specifies a full path to your credential file.
+
+*Run*
+
+``<build-dir>/src/aktualizr_primary/aktualizr -c <src-root>/config/sota_local_with_secondaries.toml``
+Once it's running, you should see a lot of messages, *_'Adding Secondary to Aktualizr'_* and *_'Provisioned successfully on Device Gateway'_* are the key one and are signs of successful start. As result, you should see that a new device has been registered at the backend and it(the device) includes two ECUs, you local/"fake" Primary and Secondary, so you can play with it now...
```
patrickvacek: I might rephrase this just a bit:
Once aktualizr is running, check the output for *_'Adding Secondary to Aktualizr'_* and *_'Provisioned successfully on Device Gateway'_*. You should then see that a new device has been registered on the server and that it includes two ECUs: your local Primary and your fake Secondary. You can now send updates to either ECU.

docs/posix-secondaries.adoc
mike-sul: Yeah, the origin of the term is lost in the mists of time... :). 
NP, I'll update the doc as you suggested.

docs/posix-secondaries.adoc
mike-sul: agree

docs/posix-secondaries.adoc
mike-sul: ok

merltron:
Sorry if I'm late to the game here, but does this content overlap with what is already described here? https://docs.ota.here.com/quickstarts/install-a-client-locally-with-fake-secondaries.html

docs/posix-secondaries.adoc
mike-sul: NP, I would add at the end - "and hope it will work..." :)

config/posix_secondary.toml
mike-sul: Nothing uses this conf at the moment, it's rather intended for an user/developer to run a fake secondary, so they can use it as it is or modify it according to their needs. If no such conf provided in the repo then a developer will need to figure out a fake secondary config by himself/herself which is additional burden, IMHO.

docs/configuration.adoc
mike-sul: ok.

mike-sul:
> Sorry if I'm late to the game here, but does this content overlap with what is already described here? https://docs.ota.here.com/quickstarts/install-a-client-locally-with-fake-secondaries.html

The mentioned doc describes usage of so-called "virtual secondaries" by using the old approach to secondaries configuration while the doc being reviewed describes posix/IP secondaries usage locally by using the new approach to secondaries configuration.
I think, https://docs.ota.here.com/quickstarts/install-a-client-locally-with-fake-secondaries.html should be updated once the virtual secondaries implementation is moved to the new configuration approach, at the moment the doc is valid/correct and doesn't overlap with this one.

mike-sul:
Updated according to the Patrick's comments.

patrickvacek:
`docs/linux-secondaries.adoc` is obsolete now, right? Can it be deleted now?

mike-sul:
> `docs/linux-secondaries.adoc` is obsolete now, right? Can it be deleted now?

I was planning to delete it as part of a commit that adds another doc, a doc on bitbaking, configuring and running Primary and Secondary for QEMU, which I am working on now.

docs/configuration.adoc
```diff
@@ -87,6 +87,7 @@ Options for Uptane.
 | `key_source`              | `"file"`     | Where to read the device's private key from. Options: `"file"`, `"pkcs11"`.
 | `key_type`                | `"RSA2048"`  | Type of cryptographic keys to use. Options: `"ED25519"`, `"RSA2048"`, `"RSA3072"` or `"RSA4096"`.
 | `secondary_configs_dir`   | `""`         | Directory containing individual secondary json configuration files. Example here: link:{aktualizr-github-url}/config/secondary/virtualsec.json[]
+| `secondary_config_file`   | `""`         | A path to a json file containing configuration of Secondary ECU(s) for Primary. link:../config/posix_secondary_config.json[Example is here]. It's a new way to configure secondaries and is replacement of `secondary_configs_dir`. The later is going to be removed once the virtual secondary implementation is fully ported to the new approach.
```
lbonn: - "for Primary"  -> "registered on the primary"?
- the link format `link:{aktualizr-github-url}` is also supposed to be more robust and works even if the file is out of the repo.

mike-sul: > * the link format `link:{aktualizr-github-url}` is also supposed to be more robust and works even if the file is out of the repo.
ok.

mike-sul: > * "for Primary"  -> "registered on the primary"?

Not sure, the IP secondaries details of which are specified in the config, are not registered at the moment of initial aktualizr start. By "for Primary" I meant that this conf file is intended for Primary. Maybe we can rephrase it as "A path to a json file containing configuration of Secondary ECU(s) to be registered at/on/in/with Primary." Not sure what is the correct preposition here :).

patrickvacek: Almost all of the those prepositions work. "With" might be best.

docs/posix-secondaries-bitbaking.adoc
```diff
@@ -0,0 +1,114 @@
...
+It is assumed that a reader is familiar with details on running of bitbaked images targeted for QEMU, such information can be found in the following docs:
+
+* link:https://github.com/advancedtelematic/meta-updater/blob/master/README.adoc[meta-updater README].
+* https://docs.ota.here.com/quickstarts/qemuvirtualbox.html[How to build 'core-image-minimal' image for QEMU]
+
+=== Primary
+
+To launch QEMU VM acting as Primary run the following from your build directory
+....
+sudo ../meta-updater/scripts/run-qemu-ota --no-gui --secondary-network primary-image
```
lbonn: Is sudo required because of the network setup?

docs/posix-secondaries.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+*Build*
+
+The Secondary executable is built by default, thus following build instructions outlined in link:../README.adoc[the repo's README] should do the trick. If the build procedure is completed successfully then aktualizr-secondary can be found in ``<build-dir>/src/aktualizr_secondary/aktualizr-secondary``.
+
+*Configure*
+
+A default configuration of aktualizr-secondary can be found in link:../config/posix_secondary.toml[<src-root>/config/posix_secondary.toml].
+
+Configuration params that are worth mentioning are located in [uptane] section:
```
lbonn: "parameters" wouldn't hurt

docs/posix-secondaries-bitbaking.adoc
```diff
@@ -0,0 +1,114 @@
+= Posix Secondaries: bitbaking and usage of Primary and Secondary images
+
+The goal of this doc is to guide a reader on bitbaking of two type of images `primary` and `secondary` that are targeted for QEMU (and RPi TBD)
+and running of which on the target makes it act as `Primary` and `Secondary` ECU of a single device.
+
+Please, refer to link:https://uptane.github.io/uptane-standard/uptane-standard.html[the Uptane standard] in order to grok a meaning of the `Primary` and `Secondary` terms in the given context from a theoretical standpoint.
+Please, check out link:posix-secondaries.doc[this doc] to understand these terms from a practical standpoint and
+to learn how `Primary` and `Secondary` can be emulated locally on an user's host.
+It is highly advisable to follow the doc instructions and play with faked/emulated `Primary` and `Secondary` prior to running steps described further in this doc.
```
lbonn: As I mentioned, I am not sure it's warranted to call this setup fake or emulated as it's basically the same as the yocto setup minus the qemu networking part (which is arguably more fake? :smile:)

Maybe we could talk about primary-secondary communication on localhost / single machine network or similar as an introduction and drop most of the "fake" and "emulated" references later in posix-secondaries.adoc?

docs/posix-secondaries-bitbaking.adoc
mike-sul: No, it's not required, I still use it due to some historical reason. Will remove 'sudo' from the doc.

docs/configuration.adoc
mike-sul: ok. Will update accordingly

docs/posix-secondaries.adoc
mike-sul: yes.

docs/posix-secondaries-bitbaking.adoc
mike-sul: Yeah, I understand your point, not sure if I agree with usage of word 'network' in the term we are looking for.
This doc https://docs.ota.here.com/quickstarts/install-a-client-locally-with-fake-secondaries.html is titled as "Simulate a device on your workstation" so 'simulate device' is used while a link/document is named as 'install-a-client-locally with fake secondaries'.

IMHO, what we do here is simulation of a composite device that includes one primary ECU and one single ECU on a local host (although it can be more than one secondary). Perhaps a term 'multi-ECU device simulation on a local host would do the trick ?

mike-sul:
Updated according to the latest comments.

docs/posix-secondaries-bitbaking.adoc
patrickvacek: >  This doc https://docs.ota.here.com/quickstarts/install-a-client-locally-with-fake-secondaries.html is titled as "Simulate a device on your workstation" so 'simulate device' is used while a link/document is named as 'install-a-client-locally with fake secondaries'.

Yes, but those are virtual secondaries, which are truly and completely fake. These are not that level of fake.

patrickvacek:
Rebased and fixed a few minor things I noticed since Mike is now out of the office for a couple weeks. I'll probably merge once CI passes if no one objects. It's just docs and examples, so if anyone is upset about something, it's easy to fix.

PR #1220, opened 2019-05-27
Refactor/ota 2797/provision rename

patrickvacek:


lbonn:
Seems to be in order. I think a note in the changelog would help though.

patrickvacek:
> I think a note in the changelog would help though.

Thanks for the reminder! Done.

PR #1222, opened 2019-05-28
Fix some easy lintian warnings and errors

lbonn:
Warnings about non-stripped executables, package maintainer name and some file permissions


CMakeLists.txt
```diff
@@ -334,7 +337,7 @@ add_library(jsoncpp OBJECT third_party/jsoncpp/jsoncpp.cpp)
 
 # General packaging configuration
 set(CPACK_GENERATOR "DEB")
-set(CPACK_DEBIAN_PACKAGE_MAINTAINER "anton.gerasimov@here.com")
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Patrick Vacek <patrick.vacek@here.com>")
```
patrickvacek: lol again

config/CMakeLists.txt
```diff
@@ -1,19 +1,19 @@
 if(BUILD_DEB)
     install(FILES systemd/aktualizr-ubuntu.service
-        DESTINATION lib/systemd/system
+        DESTINATION /lib/systemd/system
```
patrickvacek: Why the leading slash? I thought there was a reason we didn't have it, and it isn't present for the other files.

lbonn:
> > set(CMAKE_C_FLAGS_DEBUG "-Og -g")
> 
> Why change from CXX to C?

The `set(CMAKE_CXX_FLAGS_DEBUG "-Og -g")` line is still there. The debug flags were only added to c++ builds, that PR fixes that.

CMakeLists.txt
lbonn: Yes this should be the final version, the linter was complaining about missing name of maintainer...

config/CMakeLists.txt
lbonn: So, with a slash it's an absolute file on the target file system. Lintian complains about installing systemd services in `/usr/lib` instead of `/lib` which is what is documented in `systemd.unit(5)`: https://lintian.debian.org/tags/systemd-service-file-outside-lib.html

More precisely when running `make/ninja install`, `/xx` should install in `$DESTDIR/xx` and `xx` in `$DESTDIR/$PREFIX/xx`. `$DESTDIR` defaults to `/` and `$PREFIX` to `/usr/local`.

I know that we've removed absolute paths in cmake's installs, notably some starting with `/etc`, and this is going in the other direction.

patrickvacek: Fair enough. I'll trust you and the linter!

PR #1223, opened 2019-05-31
Zee test

Zee314159:


scripts/test.sh
```diff
@@ -180,4 +181,10 @@ if [[ $TEST_WITH_TESTSUITE = 1 ]]; then
     fi
 fi
 
+#Make docs
+if [[ $TEST_WITH_DOCS = 1 ]]; then
+   echo ">> Running make docs"
+   make docs
```
lbonn: The `if [[ $TEST_DRYRUN != 1 ]]; then` check is missing

patrickvacek: For consistency with the other cases, it would be good to use `run_make`, `add_failure`, and setting `+x`/`-x`.

Zee314159: Sure.That makes sense.

lbonn:
Cool! Could you rebase everything to one or two commits?

scripts/test.sh
Zee314159: SORRY...so many unintended commits...still on my way to getting along with git. I'll practice somewhere else first next time, and I'll clean this mess up later.

patrickvacek: Sounds fine, let us know if you need help!

Zee314159: ok, I'll try to figure it out by myself first. If it takes too much time unnecessarily, I'll just stop and turn to you for help. Thank you for your patience!

patrickvacek:
@zabbal What is the connection to this PR?

zabbal:
> @zabbal What is the connection to this PR?

Seemed relevant because googletest version is bumped in one of the commits and the ticket is about handling it as a dependency.

patrickvacek:
> Seemed relevant because googletest version is bumped in one of the commits

Strange, we did bump the version relatively recently, but in a different, unrelated PR. I think we can safely move the conversation to the Jira ticket!

PR #1224, opened 2019-06-04
Split pkg tests

lbonn:
Initially made the change because using docker volumes was broken on gitlab CI yesterday (seems to have been fixed now). But I also think it's a little bit cleaner now anyway.

I'm also thinking about removing these install tests from Jenkins and Travis, so that we can continue to clean it up some more and not have to deal with the `/persistent` business.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -1,6 +1,7 @@
 stages:
   - docker
   - test
+  - pkg-test
   - deploy
 
 variables:
@@ -17,7 +18,6 @@ variables:
 
   GIT_SUBMODULE_STRATEGY: recursive
 
-# Build the Docker image that will be used in subsequent CI/CD stages
 Docker Setup:
   image: docker:latest
   stage: docker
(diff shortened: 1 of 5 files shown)


patrickvacek:
> Initially made the change because using docker volumes was broken on gitlab CI yesterday (seems to have been fixed now). But I also think it's a little bit cleaner now anyway.

Looks like it. Functionally, did anything change here?

> I'm also thinking about removing these install tests from Jenkins and Travis, so that we can continue to clean it up some more and not have to deal with the `/persistent` business.

And just doing them on gitlab? Why lose that level of consistency?

And +1 on bumping googletest.

lbonn:
> Looks like it. Functionally, did anything change here?

Shouldn't have. The difference is that the packages are built in one step and packages are passed to the pkg-test step through artifacts. Seee https://docs.gitlab.com/ee/ci/yaml/#dependencies and key quote:

"Note that artifacts from all previous stages are passed by default."

We used to do both in one step, probably relying on docker-in-docker.

> And just doing them on gitlab? Why lose that level of consistency?

It's more about losing unnecessary variability. On other CIs, some scripts needs to be copied in the persistent volume: https://github.com/advancedtelematic/aktualizr/blob/06ca003a5a90f684e14521f3fb668bbaea96b8fd/scripts/build_ubuntu.sh#L19

whereas these are accessible out-of-the-box in gitlab docker containers.

So it could simplify these scripts a bit. Also, in travis and Jenkins, the docker build and run is actually in another bash script: https://github.com/advancedtelematic/aktualizr/blob/master/scripts/test_aktualizr_deb_ubuntu.sh, another distracting layer of indirection.

patrickvacek:
> It's more about losing unnecessary variability.

No objection there. As long as we aren't losing functionality, I trust your refactoring goals. If you do want to cut something out, I'm open to it as long as we discuss the pros and cons.

lbonn:
> If you do want to cut something out, I'm open to it as long as we discuss the pros and cons.

Sure, but don't worry it's not very critical, just an idea for future cleanups.

PR #1225, opened 2019-06-04
campaign consent declined/postponed

lbonn:


lbonn:
Just spotted a typo actually haha

PR #1228, opened 2019-06-07
Fix some test flakiness

lbonn:
Some of the most recent failures: some tight timeouts and scripts racing for tcp ports.

In theory, `TestUtils::getFreePort` have the same problem but I'm not sure if it occurred in practice (also, could be annoying to fix in C++).


src/aktualizr_lite/test_lite.sh
```diff
@@ -49,10 +48,16 @@ add_target foo2
...
-
-# curl http://localhost:$port/repo/image/targets.json | json_pp
+port=
+try=0
+while [[ -z $port  && $try -le 100 ]]; do
+    port=$(netstat -lpan 2> /dev/null | awk -F '[ :]+' "/$pid\\/python3/ {print \$5}")
+    sleep 0.2
+    try=$((try+1))
+done
+echo "http server listening on $port"
```
patrickvacek: Would it be better to move this port-finding logic into a reusable script?

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -635,7 +635,7 @@ TEST(Aktualizr, FinalizationFailure) {
     EXPECT_EQ(current_version, SIZE_MAX);
 
     auto aktualizr_cycle_thread = aktualizr.RunForever();
-    auto aktualizr_cycle_thread_status = aktualizr_cycle_thread.wait_for(std::chrono::seconds(10));
+    auto aktualizr_cycle_thread_status = aktualizr_cycle_thread.wait_for(std::chrono::seconds(20));
```
patrickvacek: :( Is there a way we can get rid of the explicit times? I guess 20 seconds isn't so bad, but still.

lbonn: Hm I don't really know how to do that better. Waiting forever and relying on something else to timeout would probably lead to worse error messages.

Putting less load on CI machines is the best bet I would say and it's ongoing.

src/aktualizr_lite/test_lite.sh
lbonn: yes that's fair, I'll try something

src/libaktualizr/primary/aktualizr_test.cc
patrickvacek: No worries, it's fine. I just got triggered by the hardcoded numbers, you know, but they aren't always evil.

lbonn:
Someone for another review after last changes? :)

PR #1232, opened 2019-06-17
Add 2019.4 doc link and fix some bits of the process.

patrickvacek:
It seems like every time I do this I find something wrong with the steps.

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -43,6 +43,7 @@ The link above is for the doxygen docs on master. Doxygen docs for the following
 * https://advancedtelematic.github.io/aktualizr/2019.1/index.html[2019.1]
 * https://advancedtelematic.github.io/aktualizr/2019.2/index.html[2019.2]
 * https://advancedtelematic.github.io/aktualizr/2019.3/index.html[2019.3]
+* https://advancedtelematic.github.io/aktualizr/2019.4/index.html[2019.4]
 ====
 
 === Updating doxygen on github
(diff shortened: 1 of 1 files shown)


docs/README.adoc
```diff
@@ -43,6 +43,7 @@ The link above is for the doxygen docs on master. Doxygen docs for the following
 * https://advancedtelematic.github.io/aktualizr/2019.1/index.html[2019.1]
 * https://advancedtelematic.github.io/aktualizr/2019.2/index.html[2019.2]
 * https://advancedtelematic.github.io/aktualizr/2019.3/index.html[2019.3]
+* https://advancedtelematic.github.io/aktualizr/2019.3/index.html[2019.4]
```
lbonn: wrong link :)

patrickvacek: Ah! Good catch, fixed. Also rebased.

zabbal:
> It seems like every time I do this I find something wrong with the steps.

Btw, do we have release process/steps documented somewhere?

lbonn:
> > It seems like every time I do this I find something wrong with the steps.
> 
> Btw, do we have release process/steps documented somewhere?

That's exactly what this PR is touching actually. But the steps were slightly incorrect it seems.

patrickvacek:
> > > It seems like every time I do this I find something wrong with the steps.
> > 
> > 
> > Btw, do we have release process/steps documented somewhere?
> 
> That's exactly what this PR is touching actually. But the steps were slightly incorrect it seems.

Exactly. I have a ticket to automate this work, but it's rather low priority because we don't issue new releases too often, so the potential savings aren't great. But still, it would be nice to not have to think about this so much anymore.

PR #1234, opened 2019-06-17
Revert stripping of executable in release mode

lbonn:
Was added in 299475fac6c29ff969ecc06a12b2833d92bf2538 to fix some .deb
lint warning (lintian), but bitbake doesn't like it.


patrickvacek:
I'm in favor of this, but even better would be to do the stripping only for the debian packaging. You think that's possible?

lbonn:
Ok, I've added a slightly ugly way to do it

PR #1237, opened 2019-06-18
aktualizr-lite: Support using TLS keys

doanac:
This is a bit of an edge case, but I've found it useful to run
devices in the lite mode, but to still restrict who can access
the tuf/treehub repos via TLS.

### DIFF
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -11,6 +11,21 @@
 
 namespace bpo = boost::program_options;
 
+static std::shared_ptr<SotaUptaneClient> liteClient(Config &config) {
+  std::string pkey;
+  auto storage = INvStorage::newStorage(config.storage);
+  storage->importData(config.import);
+
+  auto http_client = std::make_shared<HttpClient>();
+  auto bootloader = std::make_shared<Bootloader>(config.bootloader, *storage);
+  auto report_queue = std::make_shared<ReportQueue>(config, http_client);
+
+  KeyManager keys(storage, config.keymanagerConfig());
+  keys.copyCertsToCurl(*http_client);
+
+  return std::make_shared<SotaUptaneClient>(config, storage, http_client, bootloader, report_queue);
+}
+
 static int status_main(Config &config, const bpo::variables_map &unused) {
   (void)unused;
   GObjectUniquePtr<OstreeSysroot> sysroot_smart = OstreeManager::LoadSysroot(config.pacman.sysroot);
@@ -25,8 +40,7 @@ static int status_main(Config &config, const bpo::variables_map &unused) {
 
 static int list_main(Config &config, const bpo::variables_map &unused) {
   (void)unused;
-  auto storage = INvStorage::newStorage(config.storage);
-  auto client = SotaUptaneClient::newDefaultClient(config, storage);
+  auto client = liteClient(config);
   Uptane::HardwareIdentifier hwid(config.provision.primary_ecu_hardware_id);
 
   LOG_INFO << "Refreshing target metadata";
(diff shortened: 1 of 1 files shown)


src/aktualizr_lite/main.cc
```diff
@@ -11,6 +11,34 @@
...
+  EcuSerials ecu_serials;
+  if (!storage->loadEcuSerials(&ecu_serials)) {
+    // Set a "random" serial so we don't get warning messages.
+    std::string serial = config.provision.primary_ecu_serial;
+    std::string hwid = config.provision.primary_ecu_hardware_id;
+    if (hwid.empty()) {
+      hwid = Utils::getHostname();
+    }
+    if (serial.empty()) {
+      serial = Utils::getHostname();
```
patrickvacek: I'm not sure if I understand this logic for generating a serial. Reusing hostname seems like a problem. Serials should be unique, at least within a namespace, and hostnames are rather likely to be non-unique. Our standard technique is to use the key ID of the ECU's Uptane key.

src/libaktualizr/primary/initializer.cc
```diff
@@ -205,7 +205,7 @@ InitRetCode Initializer::initEcuRegister() {
 Initializer::Initializer(
     const ProvisionConfig& config_in, std::shared_ptr<INvStorage> storage_in,
     std::shared_ptr<HttpInterface> http_client_in, KeyManager& keys_in,
-    const std::map<Uptane::EcuSerial, std::shared_ptr<Uptane::SecondaryInterface> >& secondary_info_in)
+    const std::map<Uptane::EcuSerial, std::shared_ptr<Uptane::SecondaryInterface> >& secondary_info_in, bool lite_mode)
```
patrickvacek: I don't particularly like punching holes in our initialization logic, but I haven't yet thought of a better way that isn't just duplicating code. But can we at least change the name? `lite_mode` seems rather hard to understand without already knowing the context. Perhaps `tls_only` would be better? I do like that it defaults to false, though!

src/aktualizr_lite/main.cc
doanac: The problem there is that in "lite" mode there is no uptane key. I guess I could grab the UUID from the client.pem and use that since it would be unique? However, aktualizr-lite doesn't require TLS, so that might not be there either. I guess I could just generate a UUID and use that?

src/libaktualizr/primary/initializer.cc
doanac: I'm a bit uneasy about this as well. Adding conditional logic into a security related flow could cause a subtle bug some day. I think there are two other options here:

1) Keep this patch out-of-tree.
Its kind of specific to something we are doing. I can understand if this isn't interesting to you guys.

2) Have aktualizr-lite use the SotatUptaneClient constructor rather than calling `newDefaultClient`
If we changed things to this, then I could actually just call what I needed directly, ` keys_.copyCertsToCurl(*http_client_);`

I haven't tried out option 2 but I think that might work better here and not leak details of the lite stuff into sotauptaneclient. Sound reasonable?

doanac:
I think this new version is a bit better. I can understand if you guys don't want the 2nd commit "Set a default ECU Serial", but I think the first commit is probably worth merging.

patrickvacek:
Generating a random serial is fine if you have nothing else to go off. But I'd still prefer if we could rename `lite_mode` to something like `tls_only` in order to make it more descriptive.

doanac:
@patrickvacek - sorry - i need to back out the changes to the sotauptaneclient and initializr. The latest revision doens't call those paths.

doanac:
This new diff should be better.

PR #1239, opened 2019-06-20
For OTA-2712, remove test repo meta

xcheng-here:
remove test repo meta from test/test_data/repo,
generate data from test code using aktualizr-repo


tests/metafake.h
```diff
@@ -0,0 +1,149 @@
...
+#include "logging/logging.h"
+#include "utilities/utils.h"
+
+#define SHELL_CMD(cmd) if (system(cmd)) {LOG_ERROR << "system command failed: " << cmd;}
+
+class MetaFake {
+  public:
+  MetaFake(const boost::filesystem::path &meta_dir_in): meta_dir(meta_dir_in) {
+      // generate original repo data
+      cmd_head = "build/src/aktualizr_repo/aktualizr-repo --path " + meta_dir.string() + " --command ";
```
lbonn: this is breaking on ci, because it depends on the cmake build dir being named `build` and sitting in the repo's root.

It would maybe be nicer to use the native c++ api of aktualizr-repo: https://github.com/advancedtelematic/aktualizr/blob/master/src/aktualizr_repo/uptane_repo.h as it wouldn't require shelling out some processes.

tests/httpfake.h
```diff
@@ -26,10 +27,13 @@ class HttpFake : public HttpInterface {
       : test_dir(test_dir_in), flavor_(std::move(flavor)), meta_dir(meta_dir_in) {
     if (meta_dir.empty()) {
       meta_dir = temp_meta_dir.Path();
-      Utils::copyDir("tests/test_data/repo/repo/image", meta_dir / "repo");
-      Utils::copyDir("tests/test_data/repo/repo/director", meta_dir / "director");
-      if (boost::filesystem::is_directory("tests/test_data/repo/campaigner")) {
-        Utils::copyDir("tests/test_data/repo/campaigner", meta_dir / "campaigner");
+
+      boost::filesystem::path gen_dir = meta_dir / "meta_fake";
+      MetaFake meta(gen_dir);
```
lbonn: Unfortunately, this is making the test times to explode and even times out in CI.

My suggestion to use the native lib actually made it worse because it then runs under valgrind in the test suite, which slows it down even more.

But in any case, `MetaFake()` should be moved somewhere outside this constructor so that it can be shared between test cases. The output directory can then just be passed as an argument to `HttpFake::HttpFake()`.

tests/metafake.h
```diff
@@ -0,0 +1,126 @@
...
+
+enum BackupList {
+};
+
+class MetaFake {
+  public:
+  MetaFake(const boost::filesystem::path &meta_dir_in)
+      : meta_dir(meta_dir_in), repo(meta_dir_in, "2021-07-04T16:33:27Z", "id0") {
+      // generate original repo data
+      repo.generateRepo(KeyType::kRSA2048);
```
lbonn: Also for speed: switch to `KeyType::kED25519`

lbonn:
For the failure in `test_uptane_secondary`, it looks like we missed some checks in `Crypto::ED25519Verify`. I would try with that:

```
diff --git a/src/libaktualizr/crypto/crypto.cc b/src/libaktualizr/crypto/crypto.cc
index 8f9f16db..33027b44 100644
--- a/src/libaktualizr/crypto/crypto.cc
+++ b/src/libaktualizr/crypto/crypto.cc
@@ -226,6 +226,9 @@ bool Crypto::RSAPSSVerify(const std::string &public_key, const std::string &sign
   return status == 1;
 }
 bool Crypto::ED25519Verify(const std::string &public_key, const std::string &signature, const std::string &message) {
+  if (public_key.size() < crypto_sign_PUBLICKEYBYTES || signature.size() < crypto_sign_BYTES) {
+    return false;
+  }
   return crypto_sign_verify_detached(reinterpret_cast<const unsigned char *>(signature.c_str()),
                                      reinterpret_cast<const unsigned char *>(message.c_str()), message.size(),
                                      reinterpret_cast<const unsigned char *>(public_key.c_str())) == 0;
```

Would also be good to unit test this case in `libaktualizr/crypto/crypto_test.cc`

tests/httpfake.h
```diff
@@ -14,22 +14,35 @@
...
       : test_dir(test_dir_in), flavor_(std::move(flavor)), meta_dir(meta_dir_in) {
     if (meta_dir.empty()) {
       meta_dir = temp_meta_dir.Path();
-      Utils::copyDir("tests/test_data/repo/repo/image", meta_dir / "repo");
-      Utils::copyDir("tests/test_data/repo/repo/director", meta_dir / "director");
-      if (boost::filesystem::is_directory("tests/test_data/repo/campaigner")) {
-        Utils::copyDir("tests/test_data/repo/campaigner", meta_dir / "campaigner");
+      boost::filesystem::path fake_meta_path;
+
+      if (meta_orig_in.empty()) {
```
lbonn: Hm, unfortunately it seems to not help much for clarity. Having meta_dir_in and meta_orig_in is confusing and it's worsened by the fact that a lot of the `HttpFake` constructions now give an empty `""` in the middle of the arguments.

A big part of this mess predates this PR though, legacy of all the different ways we had to use metadata in tests.

I would expect that the directory generated from `MetaFake` could be directly be passed in `meta_dir_in`.

As it would be great to get rid of these: https://github.com/advancedtelematic/aktualizr/blob/12f8558788dd23b6ce36fedb5c3517822a865a66/src/libaktualizr/primary/aktualizr_fullostree_test.cc#L159 (there are 4 instances in the code base), maybe it's worth changing this line https://github.com/advancedtelematic/aktualizr/blob/12f8558788dd23b6ce36fedb5c3517822a865a66/tests/uptane_test_common.h#L42 to map to what aktualizr-repo spits out.

Then hopefully, we can simplify all of this.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -1908,6 +1926,7 @@ int main(int argc, char** argv) {
   logger_init();
   logger_set_threshold(boost::log::trivial::trace);
 
+  meta_temp = new MetaFakeInTemp;
```
lbonn: We avoid creating objects with `new` in general.

It's not needed in this case, you can just declare `TemporaryDirectory dir;`  and `MetaFake meta_fake;` in main (or grouped in an anonymous struct if you prefer) + copy the path to a global `boost::filesystem`.

Everything will stay in scope while the tests run.

lbonn:
Great, we can merge it after fixing the formatting error.

lbonn:
There are now some merge conflicts, it would need a last rebase before merging.

PR #1241, opened 2019-06-21
OTA-2481/2488: Get rid of SecondaryFactory and VirtualSecondary…

mike-sul:


src/libaktualizr/uptane/secondaryconfig.h
```diff
@@ -16,8 +16,6 @@ enum class SecondaryType {
 
   kVirtual,  // Virtual secondary (in-process fake implementation).
 
-  kLegacy,  // Deprecated. Do not use.
-
```
patrickvacek: Can we get rid of this whole `SecondaryType` concept? The type should be opaque to libaktualizr, right? This probably depends on getting virtual secondaries and the rest out of libaktualizr, though.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -42,8 +42,17 @@ std::shared_ptr<SotaUptaneClient> SotaUptaneClient::newTestClient(Config &config
                                                                   std::shared_ptr<event::Channel> events_channel_in) {
   std::shared_ptr<Bootloader> bootloader_in = std::make_shared<Bootloader>(config_in.bootloader, *storage_in);
   std::shared_ptr<ReportQueue> report_queue_in = std::make_shared<ReportQueue>(config_in, http_client_in);
-  return std::make_shared<SotaUptaneClient>(config_in, storage_in, http_client_in, bootloader_in, report_queue_in,
-                                            events_channel_in);
+  auto sota_uptane_client = std::make_shared<SotaUptaneClient>(config_in, storage_in, http_client_in, bootloader_in,
+                                                               report_queue_in, events_channel_in);
+
+  // This factory method is only used for tests, so calling VirtualSecondary ctor directly here is fine
+  // for now until the virtual secondary implementation is not moved out of libaktualizr
```
patrickvacek: "not moved"? I think the "not" is extraneous. :)

I like this idea, but doesn't that mean that manual usage of aktualizr with virtual secondaries won't work? We still need that for various reasons. It can be temporarily solved (maybe?) by basically copying this into aktualizr-primary's main.cc.

src/libaktualizr/uptane/secondaryconfig.h
mike-sul: Yes, you're assumption is correct, it depends on the virtual secondaries removal out of libaktualizr. I am planning to do it in scope of OTA-2488.

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: Yes, I added 'not' by mistake, will correct it.

>doesn't that mean that manual usage of aktualizr with virtual secondaries won't work? We still need that for various reasons. It can be temporarily solved (maybe?) by basically copying this into aktualizr-primary's main.cc.

Yeas, you're comment is reasonable, I am planning to solve it in scope of OTA-2488. The plan is to get rid of the secondary instantiation functionality out of sotauptaneclient.cc and have it out of libaktualizr so aktualizr binary executed with configuration containing secondaries, including the virtual one, will actually create and add them to the sotaclient.

lbonn: Yes, it's breaking this part of the guide https://docs.ota.here.com/quickstarts/install-a-client-locally-with-fake-secondaries.html. I would be hesitant to merge this. until there is a replacement solution

mike-sul: Adding to the same factory method/iterator to SotaUptaneClient::newDefaultClient() as the one in SotaUptaneClient::newTestClient() can be a temporal solution.

lbonn: Thanks, it's working again when I've tried with a more recent version of your branch.

src/virtual_secondary/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+set(SOURCES managedsecondary.cc virtualsecondary.cc)
+
+set(HEADERS managedsecondary.h virtualsecondary.h)
+
+set(TARGET virtual-secondary)
```
lbonn: Small naming matter: we have the convention of using dashes for executables (kebab-case) and underscores for libraries (snake_case)

src/libaktualizr/primary/CMakeLists.txt
```diff
@@ -12,9 +12,12 @@ set(HEADERS aktualizr.h
 
 
 add_library(primary OBJECT ${SOURCES})
+target_include_directories(primary PUBLIC ${PROJECT_SOURCE_DIR}/src/virtual_secondary)
 
 add_aktualizr_test(NAME aktualizr SOURCES aktualizr_test.cc PROJECT_WORKING_DIRECTORY ARGS ${PROJECT_BINARY_DIR}/uptane_repos)
 add_dependencies(t_aktualizr uptane_repo_full_no_correlation_id)
+target_include_directories(t_aktualizr PUBLIC ${PROJECT_SOURCE_DIR}/src/virtual_secondary)
```
lbonn: cleaner CMake way: don't add the include directories in the library consumers manually but do it for the `virtual-secondary` library and link with them

src/libaktualizr/primary/initializer.cc
```diff
@@ -252,9 +252,9 @@ Initializer::Initializer(
       return;
     }
 
-    for (auto it = secondary_info_.begin(); it != secondary_info_.end(); ++it) {
-      it->second->Initialize();
-    }
+    //    for (auto it = secondary_info_.begin(); it != secondary_info_.end(); ++it) {
```
lbonn: Should these be removed instead of commented-out?

src/libaktualizr/primary/uptane_key_test.cc
```diff
@@ -87,7 +87,7 @@ class UptaneKey_Check_Test {
     // Verify that each secondary has valid keys.
     std::map<Uptane::EcuSerial, std::shared_ptr<Uptane::SecondaryInterface> >::iterator it;
     for (it = sota_client->secondaries.begin(); it != sota_client->secondaries.end(); it++) {
-      EXPECT_TRUE(it->second->sconfig.secondary_type == Uptane::SecondaryType::kVirtual);
+      // EXPECT_TRUE(it->second->sconfig.secondary_type == Uptane::SecondaryType::kVirtual);
```
lbonn: same: just remove it?

src/virtual_secondary/CMakeLists.txt
mike-sul: ok

src/libaktualizr/primary/uptane_key_test.cc
mike-sul: yeah, I'll just remove it since it won't needed once I move the virtual secondaries on the new way of configuration.

src/libaktualizr/primary/initializer.cc
mike-sul: Yeah.

PR #1242, opened 2019-06-24
Some more lintian fixes

lbonn:


tests/CMakeLists.txt
```diff
@@ -59,6 +53,7 @@ set(TEST_SOURCES httpfake.h test_utils.cc test_utils.h uptane_vector_tests.cc)
 include(CMakeParseArguments)
 
 add_library(testutilities test_utils.cc)
+list(APPEND TEST_LIBS testutilities)
```
patrickvacek: Do `gtest` and `gmock` still get linked in somehow?

config/systemd/debian-control/postrm
```diff
@@ -0,0 +1,20 @@
...
+if [ "$1" = "remove" ]; then
+    if [ -x "/usr/bin/deb-systemd-helper" ]; then
+        deb-systemd-helper mask 'aktualizr.service' >/dev/null || true
+    fi
+fi
+
+if [ "$1" = "purge" ]; then
+    if [ -x "/usr/bin/deb-systemd-helper" ]; then
+        deb-systemd-helper purge 'aktualizr.service' >/dev/null || true
+        deb-systemd-helper unmask 'aktualizr.service' >/dev/null || true
```
patrickvacek: Is this all just to remove it if it is still running? We're no longer running it as a daemon anymore, right?

tests/CMakeLists.txt
lbonn: `TEST_LIBS` is defined in the parent scope here: https://github.com/advancedtelematic/aktualizr/blob/8b94dbd8957b99065b473794ca1a86203aa76df8/CMakeLists.txt#L393, that's how.

patrickvacek: Ah, thanks, didn't realize it was redundant.

config/systemd/debian-control/postrm
lbonn: Yes that's right. I've actually been using the network-manager debian package as a model. They use an automated debian tool to generate these steps actually but I think it would be impractical to do that through CMake/CPack which is quite rudimentary.

PR #1243, opened 2019-06-25
Feat/ota 2666/garage tools mutual tls

patrickvacek:
Short version: it works! Well, it works as far as the unit tests are concerned. The certs were generated by the Makefile that @simao provided and @koshelev, @jochenschneider, and @zabbal consulted with.

Long version: This still hasn't been tested with any real server implementation. Also, this PR currently violates my preference to keep binaries out of git. I was able to add some scripts to generate the zip files based on the p12, but that could be taken further to start from just the certs, or we could go even further and basically incorporate Simao's Makefile. That would be cool, actually. Also, it's annoying that in the current branch, I delete the old zips, add new ones, delete them, and add p12s. It should probably be refactored/rebased to just delete the old zips and add the p12s directly. I'm not fixing it right now in case there are other changes based on code review or testing with a real server.

Also: there should be more unit tests for various the combinations of client_auth.p12, api_gateway.url, and treehub.json. We should test that we prefer the p12 and URL if available.


lbonn:
Pushed a trivial fix for the debian testing ci

src/sota_tools/authenticate_test.cc
```diff
@@ -42,12 +44,12 @@ TEST(authenticate, good_cert_zip) {
 
   EXPECT_EQ(CURLE_OK, rc);
 }
-*/
 
 /* Authenticate with nothing (no auth).
  * Parse authentication information from treehub.json.
  * Parse images repository URL from a provided archive. */
 TEST(authenticate, good_cert_noauth_zip) {
+  // Authenticates with ssl_noauth_server on port 2443.
```
zabbal: Just a minor nitpicking: would be better to use "tls" instead of "ssl" everywhere to be consistent with kTls and such.

src/sota_tools/authenticate_test.cc
```diff
@@ -11,9 +11,12 @@
 #include "treehub_server.h"
 #include "utilities/utils.h"
 
+boost::filesystem::path certs_dir;
+
 /* Authenticate with OAuth2.
  * Parse authentication information from treehub.json. */
 TEST(authenticate, good_zip) {
+  // Authenticates with the ATS portal to the SaaS instance.
```
mike-sul: What is "the ATS portal", is it connect.ota.here.com  ?

src/sota_tools/authenticate_test.cc
```diff
@@ -23,36 +26,32 @@ TEST(authenticate, good_zip) {
...
 TEST(authenticate, good_cert_zip) {
-  boost::filesystem::path filepath = "tests/sota_tools/auth_test_cert_good.zip";
+  // Authenticates with ssl_server on port 1443.
+  boost::filesystem::path filepath = certs_dir / "good.zip";
   ServerCredentials creds(filepath);
-  EXPECT_EQ(creds.GetMethod(), AuthMethod::kCert);
+  EXPECT_EQ(creds.GetMethod(), AuthMethod::kTls);
   TreehubServer treehub;
-  int r = authenticate("", creds, treehub);
+  int r = authenticate("tests/fake_http_server/server.crt", creds, treehub);
```
mike-sul: Maybe, makes sense to pass "tests/fake_http_server/server.crt"  as the test parameter ?

src/sota_tools/authenticate_test.cc
```diff
@@ -23,36 +26,32 @@ TEST(authenticate, good_zip) {
...
-
   EXPECT_EQ(CURLE_OK, rc);
 }
-*/
 
 /* Authenticate with nothing (no auth).
  * Parse authentication information from treehub.json.
  * Parse images repository URL from a provided archive. */
 TEST(authenticate, good_cert_noauth_zip) {
+  // Authenticates with ssl_noauth_server on port 2443.
```
mike-sul: Is this port number 2443 actually hard-coded value or it's a parameter or a configuration variable ?

src/sota_tools/authenticate_test.cc
```diff
@@ -23,36 +26,32 @@ TEST(authenticate, good_zip) {
...
   boost::filesystem::path filepath = "tests/sota_tools/auth_test_noauth_good.zip";
   ServerCredentials creds(filepath);
   EXPECT_EQ(creds.GetMethod(), AuthMethod::kNone);
   TreehubServer treehub;
-  int r = authenticate("tests/fake_http_server/client.crt", creds, treehub);
+  int r = authenticate("tests/fake_http_server/server.crt", creds, treehub);
```
mike-sul: Maybe, makes sense to pass "tests/fake_http_server/server.crt" as the test parameter ?

src/sota_tools/authenticate_test.cc
```diff
@@ -63,9 +62,14 @@ TEST(authenticate, good_cert_noauth_zip) {
...
+  // by a different root cert.
+  boost::filesystem::path filepath = certs_dir / "bad.zip";
+  ServerCredentials creds(filepath);
+  EXPECT_EQ(creds.GetMethod(), AuthMethod::kTls);
   TreehubServer treehub;
-  int r = authenticate("", ServerCredentials(filepath), treehub);
+  int r = authenticate("", creds, treehub);
   EXPECT_EQ(0, r);
   CurlEasyWrapper curl_handle;
   curlEasySetoptWrapper(curl_handle.get(), CURLOPT_VERBOSE, 1);
```
mike-sul: Does it make sense to continue the test control flow if the authentication fails, if not then ASSERT_EQ should serve better here.

src/sota_tools/server_credentials.cc
```diff
@@ -65,12 +65,13 @@ ServerCredentials::ServerCredentials(const boost::filesystem::path &credentials_
...
-        client_key_ = readArchiveFile(a)->str();
-      } else if (strcmp(filename, "root.crt") == 0) {
-        root_cert_ = readArchiveFile(a)->str();
+      } else if (strcmp(filename, "client_auth.p12") == 0) {
+        client_p12_ = readArchiveFile(a)->str();
+        method_ = AuthMethod::kTls;
+      } else if (strcmp(filename, "api_gateway.url") == 0) {
+        use_api_gateway = true;
+        ostree_server_ = readArchiveFile(a)->str();
+        boost::trim_if(ostree_server_, boost::is_any_of(" \t\r\n"));
```
mike-sul: Design of ServerCredentials class is really confusing as it encapsulates in a single class context of each type of credentials while effectively just a single type context is used within specific class instance.

src/sota_tools/server_credentials.cc
```diff
@@ -65,12 +65,13 @@ ServerCredentials::ServerCredentials(const boost::filesystem::path &credentials_
...
-      } else if (strcmp(filename, "client.key") == 0) {
-        client_key_ = readArchiveFile(a)->str();
-      } else if (strcmp(filename, "root.crt") == 0) {
-        root_cert_ = readArchiveFile(a)->str();
+      } else if (strcmp(filename, "client_auth.p12") == 0) {
+        client_p12_ = readArchiveFile(a)->str();
+        method_ = AuthMethod::kTls;
+      } else if (strcmp(filename, "api_gateway.url") == 0) {
+        use_api_gateway = true;
+        ostree_server_ = readArchiveFile(a)->str();
```
mike-sul: I read this like `ostree_server` and `api_gateway` are the same entity.

src/sota_tools/server_credentials.cc
```diff
@@ -81,45 +82,45 @@ ServerCredentials::ServerCredentials(const boost::filesystem::path &credentials_
...
                                   credentials_path.string());
     }
   } else {
     archive_read_free(a);
   }
 
-  try {
-    ptree pt;
+  if (use_api_gateway) {
+    repo_url_ = ostree_server_;
```
mike-sul: Now, `ostree_server` becomes repo_url :)

src/sota_tools/authenticate_test.cc
patrickvacek: Nope, that's the normal portal. https://app.atsgarage.com/ is the old portal that surprisingly still works but basically serves the same content.

src/sota_tools/authenticate_test.cc
patrickvacek: Sorry, where should it get passed from?

src/sota_tools/authenticate_test.cc
patrickvacek: It's hardcoded. It probably shouldn't be anymore. I'll look into that.

src/sota_tools/server_credentials.cc
patrickvacek: Unfortunately, for backwards compatibility, we have to support the case where there is no API gateway for some period. Going forward, there should always be such a gateway, but if we don't have the URL in the credentials.zip (or there wasn't such a gateway deployed), we still need to be able to talk to treehub directly.

patrickvacek:
> What would be the most efficient way to figure out design behind this different schema to authentication ?

Some of the answers can be gleaned from https://github.com/advancedtelematic/aktualizr/blob/master/docs/provisioning-methods-and-credentialszip.adoc. What exactly are you looking for, though? Do you want a sequence diagram or a better explanation of the options?

> How can I authenticate at and use the treehub server from a command line by utilizing ostree utility ?

You mean you want to use the ostree commandline tool? If so, I don't think you can. We are operating on OSTree objects directly and parsing them with glib ourselves to walk the object trees. (We do that to avoid a dependency on libostree.) Or is there something else you want to do?

src/sota_tools/authenticate_test.cc
patrickvacek: The `EXPECT_EQ` shouldn't hurt and it's not like we'll save a huge amount of test time if we fail earlier. We normally use `EXPECT` for everything unless there's a good reason not to. If there is consensus around changing that norm, I'm open to revising it, though.

src/sota_tools/authenticate_test.cc
patrickvacek: Fixed.

mike-sul:
> > What would be the most efficient way to figure out design behind this different schema to authentication ?
> 
> Some of the answers can be gleaned from https://github.com/advancedtelematic/aktualizr/blob/master/docs/provisioning-methods-and-credentialszip.adoc. What exactly are you looking for, though? Do you want a sequence diagram or a better explanation of the options?
> 
> > How can I authenticate at and use the treehub server from a command line by utilizing ostree utility ?
> 
> You mean you want to use the ostree commandline tool? If so, I don't think you can. We are operating on OSTree objects directly and parsing them with glib ourselves to walk the object trees. (We do that to avoid a dependency on libostree.) Or is there something else you want to do?

I've already figured out what I wanted by debugging through the code. Thanks.

src/sota_tools/authenticate_test.cc
mike-sul: From corrsponding CMakeLists.txt

add_aktualizr_test(NAME <test-name> SOURCES <> ARGS <test-args>)

mike-sul: but, it's really not important, not sure if it's worth bothering

src/sota_tools/authenticate_test.cc
mike-sul: No worries, I don't think that there is much practical benefits in it.

src/sota_tools/authenticate_test.cc
patrickvacek: Seems like that's just moving the hardcoded value to a more obscure place. I'm not sure I see the benefit right now. That said, I'm interested in redoing this logic so that we can generate all the certs and such dynamically, and once that is done, something like what you are suggesting will probably be necessary.

mike-sul: Yeah, + for the dynamic certs generation, also using TEST_F/test fixtures/base class is an ideal for sharing the same context across multiple tests.

patrickvacek:
Should we merge this or wait until we can test with a "real" server? I had also considered doing some more work to make the test data more dynamically generated, but I didn't want to do it before I knew for sure that it worked with the final server implementation.

lbonn:
If it doesn't break something that's already working (which I think it's the case), we can probably merge it.

patrickvacek:
> If it doesn't break something that's already working (which I think it's the case), we can probably merge it.

Yeah, fair enough. I don't want to have to keep rebasing this, and I'm pretty sure this is on the right track for the eventual integration with the real server implementation. There's still some improvements that could be made here, especially in the automation of the tests, but I think it's a good start.

PR #1244, opened 2019-06-26
 using garage-check automatically at the end of garage-deploy

Zee314159:
Draft code.

### DIFF
--- a/src/sota_tools/check.cc
+++ b/src/sota_tools/check.cc
@@ -1,0 +1,135 @@
+#include <curl/curl.h>
+
+#include "authenticate.h"
+#include "check.h"
+#include "garage_common.h"
+#include "json/json.h"
+#include "logging/logging.h"
+#include "ostree_http_repo.h"
+#include "ostree_object.h"
+#include "rate_controller.h"
+#include "request_pool.h"
+#include "treehub_server.h"
+#include "utilities/types.h"
+#include "utilities/utils.h"
+
+// helper function to download data to a string
+static size_t writeString(void *contents, size_t size, size_t nmemb, void *userp) {
+  assert(userp);
+  // append the writeback data to the provided string
+  (static_cast<std::string *>(userp))->append(static_cast<char *>(contents), size * nmemb);
+
+  // return size of written data
+  return size * nmemb;
+}
+
+int CheckRefValid(const TreehubServer &treehub, const std::string &ref, RunMode mode, int max_curl_requests) {
+  // Check if the ref is present on treehub. The traditional use case is that it
+  // should be a commit object, but we allow walking the tree given any OSTree
+  // ref.
+  CurlEasyWrapper curl;
+  if (curl.get() == nullptr) {
+    LOG_FATAL << "Error initializing curl";
+    return EXIT_FAILURE;
+  }
+  curlEasySetoptWrapper(curl.get(), CURLOPT_VERBOSE, get_curlopt_verbose());
+  curlEasySetoptWrapper(curl.get(), CURLOPT_NOBODY, 1L);  // HEAD
+
+  treehub.InjectIntoCurl("objects/" + ref.substr(0, 2) + "/" + ref.substr(2) + ".commit", curl.get());
+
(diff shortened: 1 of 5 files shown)


src/sota_tools/garage_deploy.cc
```diff
@@ -122,6 +122,14 @@ int main(int argc, char **argv) {
     LOG_FATAL << e.what();
     return EXIT_FAILURE;
   }
+
```
lbonn: Would be nice to have a check for dry run, like in the rest of the code

src/sota_tools/garage_deploy.cc
```diff
@@ -122,6 +122,14 @@ int main(int argc, char **argv) {
...
+  std::string cmd("garage check");
```
lbonn: It seems to have a typo, `garage-check` is the correct tool name.

But also, it would be best to not launch another process and just call some C++ code as garage-check's code lives in this repo. This way, garage-check wouldn't need to be in the `$PATH`.

It would probably require a small refactor to move garage-check's logic in a small library that would be used by `garage-check` and `garage-deploy`

src/sota_tools/garage_deploy.cc
Zee314159: Yeah, that makes sense! Thanks.

src/sota_tools/garage_deploy.cc
Zee314159: Seems a bit complicated than I thought, but this looks like the right way to do it. Thank you!

src/sota_tools/garage_deploy.cc
Zee314159: I haven't tested the code yet, just pushed it here to get your advices. Am I on the right track?

Zee314159:
I tested the garage-deploy tool, looks good. The log showed the process succeed. I checked my other test account, the image I deployed  was present on the server. Is there anything else I should test or fix?

src/sota_tools/deploy.cc
```diff
@@ -77,18 +81,18 @@ bool OfflineSignRepo(const ServerCredentials &push_credentials, const std::strin
     boost::filesystem::remove_all(local_repo);
   }
 
-  std::string init_cmd("garage-sign init --repo aktualizr --credentials ");
+  std::string init_cmd("./garage-sign/bin/garage-sign init --repo aktualizr --credentials ");
```
lbonn: I would switch back to `init_cmd("garage-sign init --repo aktualizr --credentials ");`, this would make problems in the actual released code.

src/sota_tools/deploy.h
```diff
@@ -43,4 +43,10 @@ bool OfflineSignRepo(const ServerCredentials& push_credentials, const std::strin
 bool PushRootRef(const ServerCredentials& push_credentials, const OSTreeRef& ref, const std::string& cacerts,
                  RunMode mode);
 
+/**
```
lbonn: Ok that works but it's maybe a bit confusing to have this function as part of `deploy.cc`. Maybe just add it to a new translation unit `check.cc` and `check.h`.

src/sota_tools/garage_deploy.cc
lbonn: Yes, it's going in the right direction. I would just move `CheckRefValid` in a new source file.

src/sota_tools/check.cc
```diff
@@ -0,0 +1,136 @@
...
+  assert(userp);
+  // append the writeback data to the provided string
+  (static_cast<std::string *>(userp))->append(static_cast<char *>(contents), size * nmemb);
+
+  // return size of written data
+  return size * nmemb;
+}
+
+int CheckRefValid(const TreehubServer &treehub, const std::string &ref, RunMode mode, int max_curl_requests) {
+  // vim: set tabstop=2 shiftwidth=2 expandtab:
```
lbonn: This vim comment shouldn't probably be here.

lbonn:
Only a very small nitpick and it should be good.

Issue #959, opened 2018-10-04
aktualizr release tarballs do not include sub modules


agners:
The sub modules are missing in the release tarball. This does not allow to build from release tarballs...

agners:
Besides the current issue with gitsm, we used to have race conditions before when running bitbake instances in parallel with shared DL_DIR. Maybe they are resolved with the latest changes, we will see.

Anyway, using tarballs would sidestep the whole issue and also lead to smaller download size.

It also just seems a bit pointless providing a release tarball which can't be properly built. I guess the deb files get built on a custom build system anyway? So creating a proper release tarball should be possible there?

patrickvacek:
>  used to have race conditions before when running bitbake instances in parallel with shared DL_DIR.

Interesting. We generally don't do that ourselves, but do you only have problems with meta-updater/aktualizr, or does that also happen to other layers/recipes? It could indeed be related to the submodule issue.

> using tarballs would sidestep the whole issue and also lead to smaller download size.

While that is true, we currently update the aktualizr recipe in meta-updater far more often than we issue releases on github. We are considering changing that process a bit, but especially for the master branch of meta-updater, it is likely to occasionally be ahead of the last github release.

> I guess the deb files get built on a custom build system anyway? So creating a proper release tarball should be possible there?

Yes, that is true, and that is what we've looked into. So far, there hasn't been a need for it, and now we are in the midst of considering a change in CI, so if there isn't an urgent need, we might want to wait on that.

agners:
> > we used to have race conditions before when running bitbake instances in parallel with shared DL_DIR.
> 
> Interesting. We generally don't do that ourselves, but do you only have problems with meta-updater/aktualizr, or does that also happen to other layers/recipes? It could indeed be related to the submodule issue.

In one of our builds it was quite reproducible, seen it in three builds in a row. So far only with aktualizr, but afaict we only use gitsm with ostree and aktualizr.

> 
> > using tarballs would sidestep the whole issue and also lead to smaller download size.
> 
> While that is true, we currently update the aktualizr recipe in meta-updater far more often than we issue releases on github. We are considering changing that process a bit, but especially for the master branch of meta-updater, it is likely to occasionally be ahead of the last github release.
> 
> > I guess the deb files get built on a custom build system anyway? So creating a proper release tarball should be possible there?
> 
[...]

PR #1245, opened 2019-06-28
Add static code scan instrumentation

zabbal:
This ensures that we don't export 3rd party dependencies as part of our own source archive and adds necessary instrumentation for static scan code submissions.

Note: corresponding gitlab pipelines will only be activated via schedules so normal CI shouldn't be affected.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,5 @@
 .gitignore export-ignore 
 .gitmodules export-ignore
 partial/extern/RIOT export-ignore
+third_party export-ignore
+.vscode export-ignore
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -75,6 +75,46 @@ nop11:
   script:
     - ./scripts/test.sh
 
+scan-build:
+  stage: test
+  only:
+    - schedules
+  # prepare build to be submitted for static code analysis
+  variables:
+    TEST_BUILD_DIR: 'scan-build'
+    TEST_CMAKE_BUILD_TYPE: 'Debug'
+    TEST_INSTALL_DESTDIR: "$CI_PROJECT_DIR/scan-build/bin"
+    TEST_WITH_TESTSUITE: '0'
+    TEST_WITH_SOTA_TOOLS: '0'
+    TEST_WITH_OSTREE: '0'
+    TEST_WITH_DEB: '0'
+    TEST_WITH_ISOTP: '0'
+    TEST_WITH_PARTIAL: '0'
+  image: "$UBUNTU_BIONIC_PR_IMAGE"
+  script:
+    - mkdir -p $TEST_INSTALL_DESTDIR
+    - ./scripts/test.sh
+    - git archive -o scan.tar HEAD
+  artifacts:
+    paths:
+      - scan-build/bin
+      - scan.tar
+
+veracode-scan:
+  stage: deploy
+  only:
+    - schedules
+  dependencies:
+    - scan-build
+  before_script:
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -2,3 +2,5 @@
 .gitignore export-ignore 
 .gitmodules export-ignore
 partial/extern/RIOT export-ignore
+third_party export-ignore
```
lbonn: What's the idea behind this? Is it about licensing?
We've added this file specifically to include the third_party code in source archives.

lbonn:
Have you managed to run the pipeline from end to end? If so, do you have a log output available?

zabbal:
The CI fails due to timeout.

zabbal:
> Have you managed to run the pipeline from end to end? If so, do you have a log output available?

Not in aktualizr repo. I've tried similar one with veracode-scan/pipelines in our gitlab.

ci/gitlab/.gitlab-ci.yml
```diff
@@ -75,6 +75,46 @@ nop11:
...
+    - schedules
+  dependencies:
+    - scan-build
+  before_script:
+# The lastest wrapper version can be found in https://repo1.maven.org/maven2/com/veracode/vosp/api/wrappers/vosp-api-wrappers-java/
+    - wget -q -O veracode-wrapper.jar https://repo1.maven.org/maven2/com/veracode/vosp/api/wrappers/vosp-api-wrappers-java/${VERACODE_WRAPPER_VERSION}/vosp-api-wrappers-java-${VERACODE_WRAPPER_VERSION}.jar
+  script:
+    - tar -f scan.tar --append scan-build/bin
+    - java -jar --add-modules java.se.ee veracode-wrapper.jar -vid ${VERACODE_API_ID} -vkey ${VERACODE_API_KEY}
+      -action UploadAndScan -appname "OTA Client" -createprofile true -autoscan true
```
alexhumphreys: extra indent

zabbal: > extra indent

You mean "-action"? That's intentional because it's not a separate command but part of java app parameters.

.gitattributes
zabbal: That's due to SPC policy against scanning 3rd-party components.
How are source archives actually used? Don't want to break smth by accident.

lbonn: It's used when building the source archives for releases: https://github.com/advancedtelematic/aktualizr/releases, https://github.com/advancedtelematic/aktualizr/blob/master/scripts/make_src_archive.sh

zabbal: And how are those used afterwards? I mean we already exclude some of the external dependencies but not others - I'd like to understand where this difference comes from.

lbonn: The first one is automatically generated by GitHub (no opt-out) and contains the complete git repository except submodules. It's not really useful as the code in third_party is necessary to compile aktualizr.

The second comes from the script which is run on CI for releases and fixes #959. The point is that aktualizr can be built using this tarball.

Some third party code (RIOT) is excluded because it's really big and actually not used for the main aktualizr but some PoC code that we'd like to move to another repo someday.

zabbal: I think I've lost the train of thought in here:
* jsoncpp is 3rd party dependency necessary to build aktualizr so it's included into release source code archive
* liboost is 3rd party dependency necessary to build aktualizr so it's NOT included into release source code archive
How do we make this distinction and why?

lbonn: Main reason is that we depend on a specific version of jsoncpp but we support any boost > $SOME_VERSION.

Second distinction is that jsoncpp doesn't compare with boost in term of size and complexity of deployment, having it in the repo doesn't do much harm.

zabbal:
Will be split into separate PRs based on feature branch instead of fork. Thanks for review.

PR #1246, opened 2019-06-28
OTA-2488: Make Virtual Secondary use the new configuration way.

mike-sul:


mike-sul:
1. VirtualSecondary has been moved out of libaktualizr, as well as ManagedSecondary and PartialVerification since the first is used just by the virtual one and the second one is not used by any component or use case;
2. the new approach to secondaries configuration was applied to the virtual secondary;
3. the old approach to secondaries configuration has been removed;
4. SecondaryInterface has been freed from dependency on the old SecondaryConfig containing each secondary type specifics.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -1002,10 +1000,14 @@ TEST(Aktualizr, FullMultipleSecondaries) {
 
   TemporaryDirectory temp_dir2;
   UptaneTestCommon::addDefaultSecondary(conf, temp_dir, "sec_serial1", "sec_hwid1");
-  UptaneTestCommon::addDefaultSecondary(conf, temp_dir2, "sec_serial2", "sec_hwid2");
 
   auto storage = INvStorage::newStorage(conf.storage);
-  Aktualizr aktualizr(conf, storage, http);
+  UptaneTestCommon::TestAktualizr aktualizr(conf, storage, http);
+  UptaneTestCommon::addDefaultSecondary(conf, temp_dir2, "sec_serial2", "sec_hwid2");
+  if (boost::filesystem::exists(conf.uptane.secondary_config_file)) {
```
lbonn: Why is this `if` in an unit test? Shouldn't it fail if the file does not exist?

src/virtual_secondary/managedsecondary.h
```diff
@@ -1,5 +1,5 @@
-#ifndef UPTANE_MANAGEDSECONDARY_H_
-#define UPTANE_MANAGEDSECONDARY_H_
+#ifndef PRIMARY_MANAGEDSECONDARY_H_
```
lbonn: include-guard was not updated. `VIRTUAL_SECONDARY_MANAGEDSECONDARY_H_`?

src/virtual_secondary/virtualsecondary.h
```diff
@@ -1,21 +1,35 @@
-#ifndef UPTANE_VIRTUALSECONDARY_H_
-#define UPTANE_VIRTUALSECONDARY_H_
+#ifndef PRIMARY_VIRTUALSECONDARY_H_
```
lbonn: Same problem with include guard.

src/libaktualizr/primary/aktualizr_test.cc
lbonn: Ah I saw that a similar code lies in `uptane_common_test.h`. If I understand correctly, this is a test to verify that adding a secondary works after `aktualizr` has been instantiated.

If so, a small explanatory comment would be helpful.

config/sota-local.toml
```diff
@@ -13,3 +10,6 @@ path = "storage"
 
 [pacman]
 type = "none"
+
+[uptane]
+secondary_config_file = virtualsec.json
```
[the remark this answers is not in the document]
mike-sul: sure

src/virtual_secondary/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ add_library(${TARGET} STATIC
 )
 
 target_include_directories(${TARGET} PUBLIC ${PROJECT_SOURCE_DIR}/src/virtual_secondary)
```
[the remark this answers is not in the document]
mike-sul: yeah

src/aktualizr_primary/CMakeLists.txt
```diff
@@ -2,9 +2,10 @@ set(TARGET aktualizr)
 set(SOURCES main.cc secondary_config.h secondary_config.cc secondary.h secondary.cc)
 
 add_executable(${TARGET} ${SOURCES})
-target_link_libraries(${TARGET} aktualizr_static_lib aktualizr-posix ${AKTUALIZR_EXTERNAL_LIBS})
+target_link_libraries(${TARGET} aktualizr_static_lib aktualizr-posix virtual_secondary ${AKTUALIZR_EXTERNAL_LIBS})
 target_include_directories(${TARGET} PUBLIC
   ${PROJECT_SOURCE_DIR}/src/libaktualizr-posix
+  ${PROJECT_SOURCE_DIR}/src/virtual_secondary
```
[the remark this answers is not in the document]
mike-sul: yeas, it should be, I'll double check and update it accordingly.

src/aktualizr_primary/secondary_config.h
```diff
@@ -73,13 +66,14 @@ class JsonConfigParser : public SecondaryConfigParser {
 
  private:
   static void createIPSecondariesCfg(Configs& configs, Json::Value& json_ip_sec_cfg);
+  static void createVirtualSecondariesCfg(Configs& configs, Json::Value& json_virtual_sec_cfg);
```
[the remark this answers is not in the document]
mike-sul: Yep

src/libaktualizr/config/config.h
```diff
@@ -102,7 +99,7 @@ class Config : public BaseConfig {
  private:
   void updateFromPropertyTree(const boost::property_tree::ptree& pt) override;
   void updateFromCommandLine(const boost::program_options::variables_map& cmd);
-  void readSecondaryConfigs(const boost::filesystem::path& sconfigs_dir);
+  // void readSecondaryConfigs(const boost::filesystem::path& sconfigs_dir);
```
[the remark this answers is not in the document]
mike-sul: yep

src/libaktualizr/primary/aktualizr_test.cc
mike-sul: It is redundant in the test, we actually should verify if the secondary instantiated successfully.

It's required in uptane_common_test.h as this code is shared among many tests including those that does not use secondaries. I will add explanatory comment.

src/virtual_secondary/managedsecondary.h
mike-sul: Actually, I did, PRIMARY_MANAGEDSECONDARY_H_ is the one I intended to apply, the first word denotes the namespace.

lbonn: Ok, my bad that's from our naming conventions and it ends up being a bit confusing here.
It could be prefixed by the directory name but it would get quite big in this case. Or we could keep that.

mike-sul:
Updated according to the latest comments

PR #1247, opened 2019-06-29
aktualizr-lite: Fix logic for finding latest version

doanac:
You can't assume the targets will be ordered latest->oldest. This
creates a version comparison helper that tries its best to compare
things.

### DIFF
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -1,5 +1,6 @@
 if(BUILD_OSTREE)
-add_executable(aktualizr-lite main.cc)
+set(SOURCES main.cc version.h)
+add_executable(aktualizr-lite ${SOURCES})
 target_link_libraries(aktualizr-lite aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
 
 install(TARGETS aktualizr-lite RUNTIME DESTINATION bin COMPONENT aktualizr-lite)
@@ -18,5 +19,7 @@ add_test(test_aktualizr-lite
 
 endif(BUILD_OSTREE)
 
-aktualizr_source_file_checks(main.cc)
+add_aktualizr_test(NAME lite-version SOURCES version_test.cc)
+
+aktualizr_source_file_checks(${SOURCES} version_test.cc)
 # vim: set tabstop=4 shiftwidth=4 expandtab:
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -1,6 +1,5 @@
 if(BUILD_OSTREE)
-set(SOURCES main.cc version.h)
-add_executable(aktualizr-lite ${SOURCES})
+add_executable(aktualizr-lite main.cc version.h)
 target_link_libraries(aktualizr-lite aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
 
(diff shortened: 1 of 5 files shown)


doanac:
Trying to find a nice way to not use sscanf. Its actually doing what we need and strtol is kind of a pain for this.

lbonn:
If we're confident it's fine, you can still add a lint exception, we already have some.
But in any way, the version class and its comparator are the type of thing that would benefit from unit tests.

doanac:
I think its fine, but I'd like to add a unit test to prove it. That should also help explain what this is capable of comparing.

doanac:
Found an even better more used way to do this. I still think the tests are handy so that we have some idea of the various permutations that can happen and what they product.

src/aktualizr_lite/CMakeLists.txt
```diff
@@ -18,5 +19,7 @@ add_test(test_aktualizr-lite
 
 endif(BUILD_OSTREE)
 
-aktualizr_source_file_checks(main.cc)
+add_aktualizr_test(NAME lite-version SOURCES version_test.cc)
+
+aktualizr_source_file_checks(${SOURCES} version_test.cc)
```
lbonn: `main.cc` is missing here and it causes a configure error when `BUILD_OSTREE` is not defined

src/aktualizr_lite/version.h
```diff
@@ -0,0 +1,10 @@
+#include <string>
+
+#include <string.h>
+
+struct Version {
+  std::string raw_ver;
+  Version(const std::string& version) : raw_ver(std::move(version)) {}
```
lbonn: from clang-tidy:

```
aktualizr/src/aktualizr_lite/version.h:7:49: error: std::move of the const variable 'version' has no effect; remove std::move() or make the variable non-const [performance-move-const-arg,-warnings-as-errors]
  Version(const std::string& version) : raw_ver(std::move(version)) {}
```

doanac: Sorry about always getting clang-tidy stuff wrong. I can't seem to find a way to run clang-tidy that produces the same results as TravisCI. I've tried installing clang-6 into to docker container and I think I've also tried clang-7. From a "clean" code base both versions produces lots of warnings. I need to just sit down and figure out how I can get this to work in the standard docker container.

lbonn: No worries, this version problem is indeed annoying. If you run it from `Dockerfile.debian.testing`, you should be quite close from what's going on on CI.

e.g: `./scripts/run_docker_test.sh docker/Dockerfile.debian.testing`, then `ninja clang-tidy` from a CMake directory or directly `./scripts/test.sh` with the flags used in the CI job.

lbonn:
Sorry it still fails, our clang-tidy setup is sometimes picky.

The fix should be:

```
diff --git a/src/aktualizr_lite/version.h b/src/aktualizr_lite/version.h
index 15e84f8d..6c643e8a 100644
--- a/src/aktualizr_lite/version.h
+++ b/src/aktualizr_lite/version.h
@@ -4,7 +4,7 @@
 
 struct Version {
   std::string raw_ver;
-  Version(const std::string& version) : raw_ver(version) {}
+  Version(std::string version) : raw_ver(std::move(version)) {}
 
   bool operator<(const Version& other) { return strverscmp(raw_ver.c_str(), other.raw_ver.c_str()) < 0; }
 };
```

PR #1249, opened 2019-07-05
Try ccache on gitlab

lbonn:


lbonn:
Good news, it seems to work quite well actually.

PR #1251, opened 2019-07-08
httpclient: Add ability to pass custom headers

doanac:
This allows custom users of libaktualizr to pass custom headers
with HTTP requests sent to the server.

### DIFF
--- a/src/libaktualizr/http/httpclient.cc
+++ b/src/libaktualizr/http/httpclient.cc
@@ -46,7 +46,8 @@ static size_t writeString(void* contents, size_t size, size_t nmemb, void* userp
   return size * nmemb;
 }
 
-HttpClient::HttpClient() : user_agent(std::string("Aktualizr/") + aktualizr_version()) {
+HttpClient::HttpClient(const std::vector<std::string>* extra_headers)
+    : user_agent(std::string("Aktualizr/") + aktualizr_version()) {
   curl = curl_easy_init();
   if (curl == nullptr) {
     throw std::runtime_error("Could not initialize curl");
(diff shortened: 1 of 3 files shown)


doanac:
Feel free to reject this PR. My team has a small set of out-of-tree patches we have that would never be of interest to you guys. I thought this one might be innocent enough to try and get upstream.

patrickvacek:
This seems harmless, and I'm open to it. @lbonn and @eu-smirnov any opinions?

lbonn:
Yes, we can probably merge it

src/libaktualizr/http/httpclient.cc
```diff
@@ -46,7 +46,8 @@ static size_t writeString(void* contents, size_t size, size_t nmemb, void* userp
   return size * nmemb;
 }
 
-HttpClient::HttpClient() : user_agent(std::string("Aktualizr/") + aktualizr_version()) {
+HttpClient::HttpClient(std::vector<std::string>* extra_headers)
```
mike-sul: It could be `**const** std::vector<std::string>*`, just minor comment.

mike-sul:
It harmless as long as an user knows what he/she/it does :), there are  number of headers that can lead to not desired response, e.g. Last-Modified / If-Modified-Since / E-Tag, etc.

doanac:
I made the argument `const` as noted by mike-sul. Again, no big deal, but it helps me manage my out-of-tree patch queue a little easier.

patrickvacek:
Would you mind adding a small unit test to `src/libaktualizr/http/httpclient_test.cc` or somewhere else appropriate? Otherwise I fear that this could easily accidentally break or unintentionally get deleted without anyone but you noticing.

doanac:
Good idea on the test. I've added one that takes advantage of the fact the fake server has a test the checks the Authorization header.

PR #1252, opened 2019-07-08
Ci/apt awscli

lbonn:


lbonn:
By the way, I haven't been able to see the ccache gains on Travis so I might have done something wrong (or something else doesn't work...).

lbonn:
There was a mistake in the CI variables (slightly awkward on our Travis setup). Now fixed and it indeed has a big impact.

lbonn:
Finally a passing build on Travis :). Was pretty tight though.

lbonn:
Once we'll have a ccache on the master branch (next passing build there), everything should be much faster. In fact, compilation time is really the bulk of the job time on Travis...

PR #1253, opened 2019-07-11
Add --wait-until-provisioned option to aktualizr-info

Zee314159:
Draft code.

### DIFF
--- a/src/aktualizr_info/main.cc
+++ b/src/aktualizr_info/main.cc
@@ -53,7 +53,8 @@ int main(int argc, char **argv) {
     ("delegation",  "Outputs metadata of image repo targets' delegations")
     ("director-root",  "Outputs root.json from director repo")
     ("director-target",  "Outputs targets.json from director repo")
-    ("allow-migrate", "Opens database in read/write mode to make possible to migrate database if needed");
+    ("allow-migrate", "Opens database in read/write mode to make possible to migrate database if needed")
+    ("wait-until-provisioned", "Outputs metadata when device already provisioned");
   // clang-format on
 
   try {
@@ -77,6 +78,11 @@ int main(int argc, char **argv) {
       readonly = false;
     }
 
+    bool wait_provisioning = false;
+    if (vm.count("wait-until-provisioned") != 0) {
+      wait_provisioning = true;
+    }
+
     std::shared_ptr<INvStorage> storage = INvStorage::newStorage(config.storage, readonly);
     bool cmd_trigger = false;
     std::string device_id;
@@ -89,6 +95,17 @@ int main(int argc, char **argv) {
         std::cout << device_id << std::endl;
         return EXIT_SUCCESS;
       }
+    }
+
+    std::string director_root;
+    auto registered = storage->loadEcuRegistered();
+    bool has_metadata = storage->loadLatestRoot(&director_root, Uptane::RepositoryType::Director());
+
+    while (wait_provisioning && (!registered || !has_metadata)) {
+      registered = storage->loadEcuRegistered();
+      has_metadata = storage->loadLatestRoot(&director_root, Uptane::RepositoryType::Director());
+
+      sleep(1);
     }
 
     // TLS credentials
@@ -149,9 +166,6 @@ int main(int argc, char **argv) {
       std::cout << key << std::endl;
       cmd_trigger = true;
     }
-
-    std::string director_root;
-    bool has_metadata = storage->loadLatestRoot(&director_root, Uptane::RepositoryType::Director());
 
     // An arguments which depend on metadata.
     std::string msg_metadata_fail = "Metadata is not available";
@@ -286,7 +300,6 @@ int main(int argc, char **argv) {
       }
     }
 
-    auto registered = storage->loadEcuRegistered();
     std::cout << "Provisioned on server: " << (registered ? "yes" : "no") << std::endl;
     std::cout << "Fetched metadata: " << (has_metadata ? "yes" : "no") << std::endl;
 
--- a/tests/prov_test_common.py
+++ b/tests/prov_test_common.py
@@ -6,11 +6,7 @@ from time import sleep
 
 def verify_provisioned(akt_info, conf):
     # Verify that device HAS provisioned.
-    for delay in [5, 5, 5, 5, 10, 10, 10, 10]:
-        sleep(delay)
-        stdout, stderr, retcode = run_subprocess([str(akt_info), '--config', str(conf)])
-        if retcode == 0 and stderr == b'' and 'Fetched metadata: yes' in stdout.decode():
-            break
+    stdout, stderr, retcode = run_subprocess([str(akt_info), '--config', str(conf), '--wait-until-provisioned'])
     machine = platform.node()
     if (b'Device ID: ' not in stdout or
             b'Primary ecu hardware ID: ' + machine.encode() not in stdout or
@@ -28,5 +24,10 @@ def verify_provisioned(akt_info, conf):
 
 def run_subprocess(command, **kwargs):
     print('> Running {}'.format(' '.join(command)))
-    s = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False, **kwargs)
-    return s.stdout, s.stderr, s.returncode
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs)
+    try:
+        stdout, stderr = proc.communicate(timeout=60)
+    except TimeoutExpired:
+        proc.kill()
+        stdout, stderr = proc.communicate()
+    return stdout, stderr, proc.returncode


src/aktualizr_info/main.cc
```diff
@@ -77,219 +78,237 @@ int main(int argc, char **argv) {
...
     }
 
+    bool wait_provision_flag = false;
+    if (vm.count("wait-until-provisioned") != 0) {
+      wait_provision_flag = true;
+    }
+
     std::shared_ptr<INvStorage> storage = INvStorage::newStorage(config.storage, readonly);
     bool cmd_trigger = false;
+    bool provision_trigger = true;
```
patrickvacek: Might be simpler to initialize this to false and just use a `while(!provisioned)` loop.

src/aktualizr_info/main.cc
```diff
@@ -77,219 +78,237 @@ int main(int argc, char **argv) {
...
+      auto registered = storage->loadEcuRegistered();
+      std::cout << "Provisioned on server: " << (registered ? "yes" : "no") << std::endl;
+      std::cout << "Fetched metadata: " << (has_metadata ? "yes" : "no") << std::endl;
+      provision_trigger = (provision_trigger && registered);
+      provision_trigger = (provision_trigger && has_metadata);
 
+      if (provision_trigger) {
+        break;
+      } else {
+        sleep(5);
```
patrickvacek: I think a much smaller sleep time would be better. 1 second is probably plenty. However, that will result in a lot of garbage printed to the screen, so we might want to any output until the loop finishes.

Does anyone else have opinions on what they'd like to see printed while waiting for provisioning to complete? Perhaps most ideal would be to only print the successful parts as they happen.

src/aktualizr_info/main.cc
```diff
@@ -77,219 +78,237 @@ int main(int argc, char **argv) {
...
-    }
 
-    auto registered = storage->loadEcuRegistered();
-    std::cout << "Provisioned on server: " << (registered ? "yes" : "no") << std::endl;
-    std::cout << "Fetched metadata: " << (has_metadata ? "yes" : "no") << std::endl;
+      auto registered = storage->loadEcuRegistered();
+      std::cout << "Provisioned on server: " << (registered ? "yes" : "no") << std::endl;
+      std::cout << "Fetched metadata: " << (has_metadata ? "yes" : "no") << std::endl;
+      provision_trigger = (provision_trigger && registered);
+      provision_trigger = (provision_trigger && has_metadata);
```
patrickvacek: These can be merged into one line: `provisioned = (registered && has_metadata);`

src/aktualizr_info/main.cc
mike-sul: I am just wondering if there is/are condition(s) under which this 'for' loop can last forever, e.g. device hasn't been registered at the server for some reason.

Zee314159: I might have the same question as Mike said, under what condition(s) would the loop last forever? Now I just assume the conditions are :device ID and ECU serials could be loaded, device has been registered, metadata is available. I actually am not sure about it. Is provisioned means those four conditions are meet? @patrickvacek

patrickvacek: You can't register without device ID and ECU serials, and you can't fetch metadata without provisioning, so the only conditions that need to be checked are registration and metadata successfully fetched. It is true that this loop could last forever if something goes wrong. We've discussed a possible timeout feature, but for a first go of this feature, I don't think we need that yet.

src/aktualizr_info/main.cc
```diff
@@ -77,219 +78,236 @@ int main(int argc, char **argv) {
...
     std::string device_id;
 
-    if (!storage->loadDeviceId(&device_id)) {
-      std::cout << "Couldn't load device ID" << std::endl;
-    } else {
-      // Early return if only printing device ID.
-      if (vm.count("name-only") != 0u) {
-        std::cout << device_id << std::endl;
-        return EXIT_SUCCESS;
+    while (!provisioned && wait_provisioning) {
```
patrickvacek: This will break anything that doesn't use the `--wait-until-provisioned` flag since the loop will get skipped.

Zee314159: Yes, I realized that too. I'm fixing it right now.

src/aktualizr_info/main.cc
lbonn: It probably shouldn't output anything until it's provisioned, so that the output can be parsed without errors / duplicates.

The body of the `while` loop looks quite big as well. Is it possible to have something like that?

```
bool provisioned = false;
while (wait_provisioning && !provisioned) {
  // some simple condition, no output
}

// regular aktualizr-info body
```

For the timeout parameter, as it's usually called as a subprocess by other tools (in python for example), we can use a timeout there for the moment.

Zee314159: Thanks. Much clear now！

Zee314159: > For the timeout parameter, as it's usually called as a subprocess by other tools (in python for example), we can use a timeout there for the moment.

So, how does the other tools（in python for example） outside  aktualizr-info main function body break this loop？

lbonn: For example, in Python:

https://docs.python.org/3/library/subprocess.html#subprocess.run

> The timeout argument is passed to Popen.communicate(). If the timeout expires, the child process will be killed and waited for. The TimeoutExpired exception will be re-raised after the child process has terminated.

Zee314159: Ah, thanks!

src/aktualizr_info/main.cc
```diff
@@ -91,6 +99,18 @@ int main(int argc, char **argv) {
...
     }
 
+    std::string director_root;
+    bool has_metadata = storage->loadLatestRoot(&director_root, Uptane::RepositoryType::Director());
+    auto registered = storage->loadEcuRegistered();
+
+    while (wait_provisioning && !provisioned) {
+      registered = storage->loadEcuRegistered();
+      has_metadata = storage->loadLatestRoot(&director_root, Uptane::RepositoryType::Director());
+      provisioned = (registered && has_metadata);
```
patrickvacek: I'm not sure this `provisioned` variable is doing much anymore. Also, currently `storage->loadEcuRegistered()` and `storage->loadLatestRoot()` will always run at least twice. If the while loop's condition directly used `has_metadata` and `registered`, you wouldn't have that problem.

Zee314159: Sure, thanks!

src/aktualizr_info/main.cc
```diff
@@ -91,6 +97,17 @@ int main(int argc, char **argv) {
...
 
+    std::string director_root;
+    auto registered = false;
+    bool has_metadata = false;
+
+    while (wait_provisioning && (!registered || !has_metadata)) {
+      registered = storage->loadEcuRegistered();
+      has_metadata = storage->loadLatestRoot(&director_root, Uptane::RepositoryType::Director());
+
+      sleep(1);
```
patrickvacek: This will result in sleeping even if registration and loading the metadata is successful, which is undesirable. I think the previous way you had it where you called those two `load` calls outside of the loop first was actually fine.

Zee314159: Ok, I've changed it back and pushed it here. By the way, I didn't change anywhere else but sometimes the CI failed sometimes it passed. It's confusing. I'll check the output again after the CI process is done.

lbonn: About the CI, we've had problem with Travis timing out but it should be much better after some recent changes. If you rebase your PR on master, it should go much faster and fail less often.

Zee314159: > About the CI, we've had problem with Travis timing out but it should be much better after some recent changes. If you rebase your PR on master, it should go much faster and fail less often.

It's not the Travis timing out issue. The confusing part is sometimes the prosess will pass.  I'll check the output later.

src/aktualizr_info/main.cc
Raigi: > Seems cool, looks quite close! Thanks.
> 
> @rdanitz @balajirrao @Raigi Is this useful for your needs?

Currently I have a loop that will check aktualizr-info multiple times for provisioned: yes and metadata: yes if it received false. But it might be faster and better with aktualizr-info-wait-until-provisioned when I am creating like 50 devices.

patrickvacek: Cool, that's exactly the type of thing we're hoping this will help optimize.

src/aktualizr_info/main.cc
Zee314159: CI is passing this time.

When it failed the message for Travis might be:
The command "if [[ $RELEASE_NAME = ubuntu_16.04 ]]; then ./scripts/test_aktualizr_deb_ubuntu.sh /persistent "${INSTALL_DOCKERFILE}"; fi" exited with 1.
If GitLab CI failed , it's mostly with 'coverage'. 

I don't know if this infomation is useful. I'm sure I'll run into the same situation again. Then I can dig deeper.

patrickvacek: Sounds like it might be a race condition with the deb generation, probably not related to your PR. If you do see it again, please do let the rest of the team know, though.

PR #1254, opened 2019-07-11
OTA-2851: Support ostree refs posting by the test treehub server

mike-sul:

### DIFF
--- a/src/sota_tools/deploy_test.cc
+++ b/src/sota_tools/deploy_test.cc
@@ -18,15 +18,25 @@ TEST(deploy, UploadToTreehub) {
   OSTreeRepo::ptr src_repo = std::make_shared<OSTreeDirRepo>("tests/sota_tools/repo");
   boost::filesystem::path filepath = (temp_dir.Path() / "auth.json").string();
   boost::filesystem::path cert_path = "tests/fake_http_server/client.crt";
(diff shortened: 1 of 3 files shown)


mike-sul:
> Looks fine, but would you mind writing a test that uses this feature? Or is there more code to come that will make use of this?

Do we have tests that test treehub_server.py explicitly ?

patrickvacek:
> Do we have tests that test treehub_server.py explicitly ?

Not exactly, but several tests use it, so I was just wondering if it made sense to write a test that used this functionality. If it's just exploratory, that's okay too.

mike-sul:
> > Do we have tests that test treehub_server.py explicitly ?
> 
> Not exactly, but several tests use it, so I was just wondering if it made sense to write a test that used this functionality. If it's just exploratory, that's okay too.

OK, I got it, I'll check out if I can extend one of the existing tests or a new one based on the existing one in such way that the functionality I added is used and implicitly tested.

patrickvacek:
Sounds fine. It's not a big deal, I just wanted some proof that the code works as expected. Since this is just test code, a quick example/demo/whatever is also fine.

mike-sul:
Extended `deploy_test` with functionality that pushes a ref from a source repo to a destination one and implicitly tests the functionality I added to the test treehub server.

PR #1255, opened 2019-07-12
Update CHANGELOG for 2019.5 release

mike-sul:
Can it be automated or maybe we have some tools for it already ?
As alternative, we can introduce a rule requiring a PR owner to update CHANGELOG.md in scope of his/her PR so it will be staying up-to-date all the time ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,33 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 
 ## [??? (unreleased)]
 
+## [2019.5] - 2019-07-12
+
+### Added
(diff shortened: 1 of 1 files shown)


lbonn:
> Can it be automated or maybe we have some tools for it already ?
> As alternative, we can introduce a rule requiring a PR owner to update CHANGELOG.md in scope of his/her PR so it will be staying up-to-date all the time ?

Yes we encourage that sometimes but it's not been strictly followed.
However, we don't require every PR to have a changelog entry, only if they have a significant impact on end users (for example, pure tests/ci fixes are usually not featured). That's mostly why it's not automated.

PR #1259, opened 2019-07-18
CI job: Run tests on ubuntu xenial

kbushgit:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,9 @@ cmake_policy(SET CMP0009 NEW)
 
 project(aktualizr)
 
-include_directories(/usr/include/ostree-1)
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    set(CMAKE_CXX_FLAGS -pthread)
+endif()
 
 if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ cmake_policy(SET CMP0009 NEW)
 
 project(aktualizr)
 
+include_directories(/usr/include/ostree-1)
+
 if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release")
     message(STATUS "No CMAKE_BUILD_TYPE specified, defaulting to ${CMAKE_BUILD_TYPE}")
@@ -101,6 +103,11 @@ find_package(SQLite3 REQUIRED)
 find_package(Git)
 find_package(Asn1c REQUIRED)
 find_package(ZLIB)
+
+pkg_check_modules(GLIB glib-2.0 REQUIRED)
+include_directories(${GLIB_INCLUDE_DIRS})
+link_directories(${GLIB_LIBRARIES})
+add_definitions(${GLIB_CFLAGS_OTHER})
 
 if(NOT AKTUALIZR_VERSION)
     if(GIT_EXECUTABLE)
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -76,6 +76,7 @@ coverage:
     TEST_WITH_P11: '1'
     TEST_WITH_DOCKERAPP: '1'
     TEST_WITH_FAULT_INJECTION: '1'
+    TEST_WITH_OSTREE: '1'
     TEST_SOTA_PACKED_CREDENTIALS: "$CI_PROJECT_DIR/credentials.zip"
   image: "$UBUNTU_BIONIC_PR_IMAGE"
   stage: test
@@ -86,6 +87,28 @@ coverage:
   artifacts:
     paths:
       - build-coverage/coverage/
+  script:
+    - aws s3 cp s3://ota-gitlab-ci/hereotaconnect_prod.zip $CI_PROJECT_DIR/credentials.zip
+    - ./scripts/test.sh
+
+coverage-xenial:
+  variables:
+    TEST_BUILD_DIR: 'build-coverage-xenial'
+    TEST_CMAKE_BUILD_TYPE: 'Valgrind'
+    TEST_WITH_COVERAGE: '1'
+    TEST_WITH_P11: '0'
+    TEST_WITH_FAULT_INJECTION: '0'
+    TEST_WITH_OSTREE: '1'
+    TEST_SOTA_PACKED_CREDENTIALS: "$CI_PROJECT_DIR/credentials.zip"
+  image: "$UBUNTU_XENIAL_PR_IMAGE"
+  stage: test
+  cache:
+    key: "$CI_JOB_NAME"
+    paths:
+      - ccache/
+  artifacts:
+    paths:
+      - build-coverage-xenial/coverage/
   script:
     - aws s3 cp s3://ota-gitlab-ci/hereotaconnect_prod.zip $CI_PROJECT_DIR/credentials.zip
     - ./scripts/test.sh
--- a/docker/Dockerfile-test-install.ubuntu.xenial
+++ b/docker/Dockerfile-test-install.ubuntu.xenial
@@ -6,11 +6,6 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get -y install debian-archive-keyring
 RUN apt-get update && apt-get install -y --no-install-suggests --no-install-recommends \
   libarchive13 \
-  libboost-log1.58.0 \
-  libboost-program-options1.58.0 \
-  libboost-system1.58.0 \
-  libboost-test1.58.0 \
-  libboost-thread1.58.0 \
   libc6 \
   libcurl3 \
   libglib2.0-0 \
@@ -22,5 +17,9 @@ RUN apt-get update && apt-get install -y --no-install-suggests --no-install-reco
   openjdk-8-jre \
   python3 \
   zip
+
+RUN add-apt-repository ppa:mhier/libboost-latest
+RUN apt update
+RUN apt install libboost1.67
 
 ADD ./scripts /scripts
--- a/docker/Dockerfile-test-install.ubuntu.xenial
+++ b/docker/Dockerfile-test-install.ubuntu.xenial
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y --no-install-suggests --no-install-reco
   python3 \
   zip
 
+RUN apt install software-properties-common
 RUN add-apt-repository ppa:mhier/libboost-latest
 RUN apt update
 RUN apt install libboost1.67
--- a/docker/Dockerfile-test-install.ubuntu.xenial
+++ b/docker/Dockerfile-test-install.ubuntu.xenial
@@ -18,9 +18,9 @@ RUN apt-get update && apt-get install -y --no-install-suggests --no-install-reco
   python3 \
   zip
 
-RUN apt install software-properties-common
+RUN apt install -y software-properties-common
 RUN add-apt-repository ppa:mhier/libboost-latest
 RUN apt update
-RUN apt install libboost1.67
+RUN apt install -y libboost1.67
 
 ADD ./scripts /scripts
--- a/docker/Dockerfile.ubuntu.xenial
+++ b/docker/Dockerfile.ubuntu.xenial
@@ -82,6 +82,7 @@ RUN make install
 # RUN ./bootstrap.sh --prefix=/usr/
 # RUN ./b2
 # RUN ./b2 install
+RUN apt install software-properties-common
 RUN add-apt-repository ppa:mhier/libboost-latest
 RUN apt update
 RUN apt install libboost1.67-dev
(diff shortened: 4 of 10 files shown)


docker/Dockerfile.ubuntu.xenial
```diff
@@ -50,14 +45,17 @@ RUN apt-get update && apt-get -y install --no-install-suggests --no-install-reco
...
   python3-gi \
   python3-openssl \
   python3-pip \
   python3-venv \
   sqlite3 \
   wget \
-  zip
+  zip \
+  fiu-utils \
+  libfiu-dev
```
patrickvacek: Ideally, these should be sorted alphabetically.

docker/Dockerfile.ubuntu.xenial
```diff
@@ -50,14 +45,17 @@ RUN apt-get update && apt-get -y install --no-install-suggests --no-install-reco
   opensc \
   pkg-config \
   psmisc \
+  python-dev \
```
patrickvacek: It's pretty annoying that we need python2 and python3. What's python2 required for?

docker/Dockerfile.ubuntu.xenial
```diff
@@ -69,6 +67,23 @@ RUN apt-get update && apt-get -y install \
...
+RUN make -j4
+RUN make install
+
+WORKDIR /boost
+RUN wget https://sourceforge.net/projects/boost/files/boost/1.64.0/boost_1_64_0.tar.gz -nv
+RUN tar -xzf boost_1_64_0.tar.gz
+WORKDIR /boost/boost_1_64_0
+RUN ./bootstrap.sh --prefix=/usr/
+RUN ./b2
+RUN ./b2 install
```
patrickvacek: This is pretty unfortunate. What do we require from the newer versions?

src/libaktualizr/primary/CMakeLists.txt
```diff
@@ -44,3 +44,7 @@ target_include_directories(t_uptane_key PUBLIC ${PROJECT_SOURCE_DIR}/src/virtual
 target_link_libraries(t_uptane_key virtual_secondary)
 
 aktualizr_source_file_checks(${SOURCES} ${HEADERS} ${TEST_SOURCES})
+
+set_target_properties (t_aktualizr PROPERTIES COMPILE_FLAGS -pthread LINK_FLAG -pthread)
+set_target_properties (t_aktualizr_fullostree PROPERTIES COMPILE_FLAGS -pthread LINK_FLAGS -pthread)
+set_target_properties (t_uptane_key PROPERTIES COMPILE_FLAGS -pthread LINK_FLAGS -pthread)
```
patrickvacek: I would prefer if we didn't have to have this hardcoded and duplicated for all these targets. Why is this required?

docker/Dockerfile.ubuntu.xenial
kbushgit: Python2 is required to build boost but in the final PR I will change it to be possible to build boost with Python3

docker/Dockerfile.ubuntu.xenial
```diff
@@ -69,6 +67,23 @@ RUN apt-get update && apt-get -y install \
   softhsm2 \
   valgrind
 
+
+WORKDIR /cmake
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.14.4/cmake-3.14.4.tar.gz -nv
```
lbonn: We might be stuck with having to compile boost but it feels like it should be easier to live with an older CMake version. Do you know what's blocking us from being compatible with CMake on Xenial?

src/libaktualizr/primary/CMakeLists.txt
kbushgit: We have a few targets which use libfiu, for these targets, we get linking error " undefined reference to pthread_atfork" I think it related to GCC version there are some differences in -lpthread and -pthread

docker/Dockerfile.ubuntu.xenial
kbushgit: The CMake version lower than 3.8 does not properly support boost 1.64 or higher, it's related to Finding an installed library.

docker/Dockerfile.ubuntu.xenial
kbushgit: As we discussed in chat currently we have a requirement boost 1.64 only for building tests, for building aktualizr Release we can use min boost version 1.58.
The question is, should we up minimum boost version for the project? 

The minimum possible boost version for us is 1.64

docker/Dockerfile.ubuntu.xenial
lbonn: Ah gotcha, I forgot about that, makes sense.

docker/Dockerfile.ubuntu.xenial
lbonn: Another solution could be to use an alternative to `boost::process` for our tests, maybe get something similar from a third party library. That takes a bit more dev effort but would simplify the CI part.

patrickvacek: `boost::process` is pretty nice though, just sayin'.

kbushgit: I believe it is possible to avoid using boost::process and replace it with analog system call for() exec() etc, but I think we should try to be up to date with the third-party libraries we use, if it's possible.
I believe that the boost::process is not a single feature in the boost 1.64 we can use in further development...

src/libaktualizr/primary/CMakeLists.txt
kbushgit: I had installed GCC 7 on ubuntu xenial and the result is the same as for gcc 5, which is the default for the xenial, so the issue is not related to GCC version...

docker/Dockerfile.ubuntu.xenial
patrickvacek: Sorry, I don't think I understand your comment. What are you proposing?

kbushgit: I think we should use boost 1.64 as a min version.

patrickvacek: Yeah, I think I agree. It's a bit of a pain, but undoing Mike's rather nice code using `boost::process` seems worse.

lbonn: Ok then we're settled I guess, sorry for derailing

patrickvacek: No problem, the discussion is still worthwhile, and I was the one who complained originally!

mike-sul: test_utils.h::Process (the one that relies on boost::process) can be potentially replaced by Utils::shell() that relies on stdio::popen() if it's really required.

ci/gitlab/.gitlab-ci.yml
```diff
@@ -89,6 +90,28 @@ coverage:
     - aws s3 cp s3://ota-gitlab-ci/hereotaconnect_prod.zip $CI_PROJECT_DIR/credentials.zip
     - ./scripts/test.sh
 
+coverage-xenial:
+  variables:
+    TEST_BUILD_DIR: 'build-coverage-xenial'
+    TEST_CMAKE_BUILD_TYPE: 'Valgrind'
+    TEST_WITH_COVERAGE: '1'
+    TEST_WITH_P11: '1'
+    TEST_WITH_FAULT_INJECTION: '1'
```
patrickvacek: Let's try disabling the P11 and fault injection variables and then I think there's just a couple small things to finish up.

lbonn:
I think the remaining failure comes from a change in Openssl's output between 1.0 and 1.1. Should be possible to work around.

mike-sul:
> I think the remaining failure comes from a change in Openssl's output between 1.0 and 1.1. Should be possible to work around.

It most likely my fault since I introduced the test(s) that depends on specific openssl version output.
I can fix it.

docker/Dockerfile.ubuntu.xenial
lbonn: Is this still part of the plan?

src/libaktualizr/primary/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    set(CMAKE_CXX_FLAGS -pthread)
```
lbonn: Does it also work if only added in the root CMakeLists.txt, instead of having to duplicate that in several places?

kbushgit: Probably Yes, the global flag should work for all targets, I'll check it...

docker/Dockerfile.ubuntu.xenial
kbushgit: No, this needed only for building boost with enabled boost::python module, we do not use it right now, so it can be removed

CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ cmake_policy(SET CMP0009 NEW)
 
 project(aktualizr)
 
+include_directories(/usr/include/ostree-1)
```
lbonn: Is this needed? We have this line later: `include_directories(${LIBOSTREE_INCLUDE_DIRS})`

docker/Dockerfile-test-install.ubuntu.xenial
```diff
@@ -23,4 +18,9 @@ RUN apt-get update && apt-get install -y --no-install-suggests --no-install-reco
   python3 \
   zip
 
+RUN apt install -y software-properties-common
+RUN add-apt-repository ppa:mhier/libboost-latest
+RUN apt update
+RUN apt install -y libboost1.67
```
lbonn: Why is this needed for test-install? We install the package built against stock libboost in this case.

docker/Dockerfile.ubuntu.xenial
```diff
@@ -13,26 +13,23 @@ RUN apt-get update && apt-get -y install --no-install-suggests --no-install-reco
   autoconf \
   asn1c \
   automake \
+  awscli \
   bison \
   ccache \
-  cmake \
+  clang-format-6.0 \
```
lbonn: Do we need clang-format here?

docker/Dockerfile-test-install.ubuntu.xenial
kbushgit: Boost shared library must be installed on system where we run aktualizr, this is installation from alternative repository which has version of the boost 1.67, the default repo for xenial allows to install only 1.58

lbonn: Yes but that's what gets compiled on the actual debian package and we probably want to keep it like that. It doesn't make sense to provide a xenial deb package and ask people to install a crazy new boost version.

docker/Dockerfile.ubuntu.xenial
kbushgit: maybe this is for the case when we decide to check the formatting, not sure...

docker/Dockerfile-test-install.ubuntu.xenial
kbushgit: Here we back to the question which is the minimum boost library we should use! 

From boost FAQ
How can the Boost libraries be used successfully for important projects? Many of the Boost libraries are actively maintained and improved, so backward compatibility with prior version isn't always possible. Deal with this by freezing the version of the Boost libraries used by your project. Only upgrade at points in your project's life cycle where a bit of change will not cause problems. Individual bug fixes can always be obtained from the boost repository.

https://www.boost.org/users/faq.html

CMakeLists.txt
kbushgit: Please check this cmake module
https://github.com/advancedtelematic/aktualizr/blob/78e19207414a2373d70d63ae5a4810f9b956cbd8/cmake-modules/FindOSTree.cmake#L1

Normally we install ostree in default path "/usr" so the include dir will be located on /usr/include/ostree-1 and lib in /usr/lib!

Not sure this module can properly find the include directories since it nested to ostree-1, for *.so file it works because it located in default path for the linker

kbushgit: That's what I getting!

In file included from /home/kbushko/work/pr/feat_OTA-2753_CI_ubuntu_xenial_build_tests/aktualizr/tests/aktualizr_cycle_simple.cc:10:0:
/home/kbushko/work/pr/feat_OTA-2753_CI_ubuntu_xenial_build_tests/aktualizr/src/libaktualizr/package_manager/ostreemanager.h:8:10: fatal error: ostree.h: No such file or directory
 #include <ostree.h>
          ^~~~~~~~~~

docker/Dockerfile-test-install.ubuntu.xenial
lbonn: Ah yes I just realized that we also build the xenial package against this new boost then... That makes this .deb package quite useless. It doesn't make sense to say we support xenial if we recompile most of our dependencies there.
And we don't even need a new boost for the application, only the tests!

CMakeLists.txt
kbushgit: Ah, this is not an issue with FindOSTree.cmake, the problem is in the aktualizr_cycle_simpl.cc this test include ostreemanager.h which includes ostree.h and if we build project without OSTREE it will fail since find_package for OSTree is not executed!

This actually is a case when TEST which executed by default has dependencies on the package which is optional, in this particular case -DBUILD_OSTREE

lbonn:
As we've agreed this morning, we'll close this PR as we've already started opening others for the bits we want to keep.

PR #1261, opened 2019-07-18
Add static code analysis submission to CI

zabbal:
Test run: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/49815
Note: it's expected to be used via "schedule" on a weekly basis, not a regular pipeline.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -85,6 +85,49 @@ nop11:
   script:
     - ./scripts/test.sh
 
+scan-build:
+  stage: test
+  only:
+    - schedules
+  # prepare build to be submitted for static code analysis
+  variables:
+    TEST_BUILD_DIR: 'scan-build'
+    TEST_CMAKE_BUILD_TYPE: 'Debug'
+    TEST_INSTALL_DESTDIR: "$CI_PROJECT_DIR/scan-build/bin"
+    TEST_WITH_TESTSUITE: '0'
+    TEST_WITH_SOTA_TOOLS: '0'
+    TEST_WITH_OSTREE: '0'
+    TEST_WITH_DEB: '0'
+    TEST_WITH_ISOTP: '0'
+    TEST_WITH_PARTIAL: '0'
+  image: "$UBUNTU_BIONIC_PR_IMAGE"
+  script:
+    - mkdir -p $TEST_INSTALL_DESTDIR
+    - ./scripts/test.sh
+    - git archive -o scan.tar HEAD
+  artifacts:
+    paths:
+      - scan-build/bin
+      - scan.tar
+
+veracode-scan:
+  stage: deploy
+  only:
+    - schedules
+  dependencies:
+    - scan-build
+  image: "$UBUNTU_BIONIC_PR_IMAGE"
+  before_script:
+    - apt update
+    - apt-get -q -y install openjdk-8-jdk
+# The lastest wrapper version can be found in https://repo1.maven.org/maven2/com/veracode/vosp/api/wrappers/vosp-api-wrappers-java/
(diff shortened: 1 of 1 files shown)


lbonn:
The one you linked does not contain the veracode step, I think you meant https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/48773

zabbal:
> The one you linked does not contain the veracode step, I think you meant https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/48773

Yes, definitely. The difference is in how it's triggered - if it's not via schedule than the step is ignored. Thanks, I'll update comment.

zabbal:
Not sure about failed CI - when I follow the links the job is marked as successful. Any ideas how to re-trigger/fix this?

lbonn:
This one doesn't look successful: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/50015, but it's because of an unrelated issue we have with gitlab.

Don't know why it was triggered after this one though: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/49815

ci/gitlab/.gitlab-ci.yml
```diff
@@ -85,6 +85,49 @@ nop11:
...
+
+veracode-scan:
+  stage: deploy
+  only:
+    - schedules
+  dependencies:
+    - scan-build
+  image: "$UBUNTU_BIONIC_PR_IMAGE"
+  before_script:
+    - apt update
```
lbonn: It looks like you actually do not need our Docker image here, you could pull an image with the correct version of java preinstalled and not worry about installing it with apt in the pipeline.

lbonn:
So, as I understand, this step is only triggering a job that's running somewhere else, right?

Do you think it would be possible to output a link to the triggered job somewhere in the logs, so that it can be inspected easily from here?

ci/gitlab/.gitlab-ci.yml
zabbal: Which image is that? Got link/name?

zabbal:
> So, as I understand, this step is only triggering a job that's running somewhere else, right?

That's correct.

> Do you think it would be possible to output a link to the triggered job somewhere in the logs, so that it can be inspected easily from here?

That won't help - you'll have to request account from SPC to view the scan results.

> If that's indeed the case, the job could also be renamed to highlight the fact that it's a trigger.

Any particular suggestions?

ci/gitlab/.gitlab-ci.yml
lbonn: https://hub.docker.com/_/openjdk ?

lbonn:
> > Do you think it would be possible to output a link to the triggered job somewhere in the logs, so that it can be inspected easily from here?
> 
> That won't help - you'll have to request account from SPC to view the scan results.

Well, that could help people with an SPC account, but that's not critical sure. Do they at least have some information about the commit used for the test there?

> 
> > If that's indeed the case, the job could also be renamed to highlight the fact that it's a trigger.
> 
> Any particular suggestions?

veracode-scan-trigger

ci/gitlab/.gitlab-ci.yml
mike-sul: AFAIK, gitlab can serve as a docker image registry, we could prepare our own images, put them into gitlab so CI pipelines/test can utilize them. Effectively, each aktualizr docker file 'bakes' dynamically/on the fly during test running, while having pre-baked docker image could speed these tests little bit.

lbonn:
Make another proposal on top of your changes here: https://github.com/advancedtelematic/aktualizr/pull/1281

lbonn:
Replaced by #1281

PR #1262, opened 2019-07-22
Launch ptest at the end of pipelines

lbonn:
Will probably only enable it for master merge but it's enabled here for the test.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -185,10 +185,20 @@ xenial-pkg-test:
 OE Docker setup:
   extends: .bb_docker_remote
   stage: docker
+  only:
+    variables:
+      - $OE_PTEST
+      - $CI_COMMIT_REF_NAME == "master"
+      - $CI_COMMIT_REF_NAME =~ /^\d\d\d\d\.\d+(-\w+)?$/
 
 OE Checkout:
   extends: .bb_checkout
   stage: oe-checkout
+  only:
+    variables:
+      - $OE_PTEST
+      - $CI_COMMIT_REF_NAME == "master"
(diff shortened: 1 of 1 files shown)


patrickvacek:
How can I see the results of the ptest run?

lbonn:
Before I've limited it to master and tags: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/49676

lbonn:
Hm maybe it would be cool to also be able to force it with a variable, like some of the other pipelines. (master+tags+variable)
I'll see what I can do.

patrickvacek:
> Hm maybe it would be cool to also be able to force it with a variable, like some of the other pipelines. (master+tags+variable)

Could be cool but don't sweat that. This is pretty cool all is.

> Before I've limited it to master and tags: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/49676

Thanks! I see this bit:
```
+ TEST_AKTUALIZR_BRANCH=ci/OTA-2465/aktualizr-oe
+ TEST_AKTUALIZR_REV=12e26566874682ecb677b597e440f5d99f6b69ef
```
But is there a way to be sure that that's really what ptest ran?

lbonn:
Good point. The way I've tested it is to log in the docker image while it's running and cat-ing `site.conf`. Maybe we should just cat it in the configure script in any case.

lbonn:
https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/50185

lbonn:
Needs https://github.com/advancedtelematic/meta-updater/pull/563 for working with the tag.

lbonn:
Ok, finally shown to work with tags and manual triggers!

patrickvacek:
> I'm assuming whyever Travis failed is bogus.

It worked after a restart!

Issue #23, opened 2016-12-08
JSON processing for server communication


embmk:
JSON encoded response data from the server should be encoded with an appropriate JSON parser.  This  simple C approach seems to be promising: https://github.com/udp/json-parser

codders:
Yup. That looks like one of the better implementations. I'm fine with that.

PR #1263, opened 2019-07-23
C API for UptaneCycle and campaigns

eu-smirnov:


eu-smirnov:
I still need to do something with absolute paths in test C app.

src/libaktualizr-c/libaktualizr-c.cc
```diff
@@ -0,0 +1,80 @@
...
+
+void Aktualizr_destroy(Aktualizr *a) {
+  delete a;
+}
+
+Campaign *Aktualizr_campaign_check(Aktualizr *a) {
+  try {
+    auto r = a->CampaignCheck().get();
+    if (!r.campaigns.empty()) {
+      return new Campaign(r.campaigns[0]);
```
patrickvacek: Is it fair to assume there will never be multiple campaigns? It would indeed be confusing if there were, but a comment to state the assumption might be worthwhile.

src/libaktualizr-c/test/Makefile
```diff
@@ -0,0 +1,11 @@
...
+CFLAGS=-I/home/esm/projects/aktualizr/include -g -O0
+# Run application with LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/esm/projects/aktualizr/build/src/libaktualizr-c
+LIBS=-L/home/esm/projects/aktualizr/build/src/libaktualizr-c -laktualizr-c
+OBJ=api-test.o
+
+%.o: %.c
+	$(CC) -c -o $@ $< $(CFLAGS)
+
+api-test: $(OBJ)
+	$(CC) -o $@ $^ $(CFLAGS) $(LIBS)
```
patrickvacek: Why not use cmake? And either way, is there a way to integrate the test into the ctest framework?

src/libaktualizr-c/libaktualizr-c.cc
eu-smirnov: It's not supported at the moment, so I think it's worth keeping it simple here. I'll add a comment.

src/libaktualizr-c/test/Makefile
eu-smirnov: I started with having this test app completely separate, to be sure that linking works without any implicit cmake magic. But I guess I'll use cmake and add it to the ctest now.

src/libaktualizr/package_manager/CMakeLists.txt
```diff
@@ -17,9 +17,6 @@ add_aktualizr_test(NAME packagemanagerfake SOURCES packagemanagerfake_test.cc)
 # Debian backend
 if(BUILD_DEB)
     set_property(SOURCE packagemanagerfactory.cc packagemanagerfactory_test.cc PROPERTY COMPILE_DEFINITIONS BUILD_DEB)
```
patrickvacek: We probably don't need to set this `BUILD_DEB` definition here anymore, right? In fact, is `BUILD_DEB` meaningful at all anymore? In CI it's built by default, and if it brings in no dependencies, there's hardly a reason not to just have it always on.

lbonn: That's a fair point, though we'd ship a code path that calls dpkg on yocto systems where it doesn't make sense. I don't know if the cost of keeping it (low, I think) warrants larger executables and potential bug surface.

patrickvacek: Probably true. Perhaps we should disable by default on CI and only enable it as necessary?

lbonn: Our approach on CI has been to enable all flags, so that everything can be tested at once. The things that are not covered tend to not compile after refactors.

Also, some unrelated useful tests use the Debian mode. For example memory_test. We'll probably migrate them when we can though.

patrickvacek: Fair enough. I have no objection.

src/libaktualizr-c/libaktualizr-c.cc
```diff
@@ -0,0 +1,80 @@
...
+Aktualizr *Aktualizr_create(const char *config_path) {
+  Aktualizr *a;
+  try {
+    Config cfg(config_path);
+    a = new Aktualizr(cfg);
+  } catch (const std::exception &e) {
+    std::cerr << "Creating exception: " << e.what() << std::endl;
+    return NULL;
+  }
+  return a;
```
mike-sul: Maybe, there will be a need/requirement to return some status/error code to am app using the C version of libaktualizr, otherwise the only way to find out what went wrong is by parsing stderr.

eu-smirnov: @mike-sul Absolutely. But first, we'll need to make sure that exceptions thrown by libaktualizr are consistent and meaningful. I plan to work on that next.

src/libaktualizr/package_manager/CMakeLists.txt
eu-smirnov: @patrickvacek, @lbonn Sorry, I'm a bit confused, what's your conclusion on that?

lbonn: I think just keep it as is?

eu-smirnov: Cool, that's what I thought, just wanted to be sure.

patrickvacek: Yes, sorry for the distraction!

include/libaktualizr-c.h
```diff
@@ -0,0 +1,30 @@
...
+typedef struct Campaign Campaign;
+#endif
+
+Aktualizr *Aktualizr_create(const char *config_path);
+int Aktualizr_initialize(Aktualizr *a);
+int Aktualizr_uptane_cycle(Aktualizr *a);
+void Aktualizr_destroy(Aktualizr *a);
+
+Campaign *Aktualizr_campaign_check(Aktualizr *a);
+int Aktualizr_campaign_accept(Aktualizr *a, Campaign *c);
```
lbonn: That's a bit annoying that we have to pass an Aktualizr instance to all of these. Would be handy if Campaign contained a pointer. But that might need changes in the C++ object, don't know if we want that.

src/libaktualizr-c/test/api-test.c
```diff
@@ -0,0 +1,44 @@
...
+    return EXIT_FAILURE;
+  }
+
+  c = Aktualizr_campaign_check(a);
+  if (c) {
+    printf("Accepting running campaign\n");
+    err = Aktualizr_campaign_accept(a, c);
+    Aktualizr_campaign_free(c);
+    if (err) {
+      return EXIT_FAILURE;
```
mike-sul: Will the aktualizr instance be properly tear-downed and freed in this case as well as in case of line #23 ? Or it's not important taking into account that the process is going to die anyway ?

eu-smirnov: It won't, I forgot how to write proper C ;) But I'm rewriting this piece anyway

include/libaktualizr-c.h
eu-smirnov: I see no problem in this, tbh. That's pretty much idiomatic C and is equivalent to `Aktualizr->CampaignAccept(Campaign *c). If we do just a single arg here, I would expect our C++ API to also change to `Campaign->Accept()`. I'm not sure if it makes it nicer, though.

lbonn: Ok I think you're right.

eu-smirnov: On second thought, maybe it's not a bad thing to make those calls members of Campaign object. This way, we would prevent passing a campaign_id that belongs to another Aktualizr instance (a hypothetical situation, but still). WDYT, @lbonn @patrickvacek @mike-sul ?

eu-smirnov:
Would you mind guys merging this in the current state? It looks like doing the proper test for that will take me a bit longer.

include/libaktualizr-c.h
patrickvacek: If it doesn't require changes to the C++ code, then that seems like the way to go. Even if it does, that seems to me to be a slightly better way to go.

eu-smirnov:
@lbonn Cool, I'll merge it then and address the comments about the test app in the next PR.
We also can discuss again if we want to change the Campaigns API, I'm still not sure about that.

PR #1265, opened 2019-07-24
Fix dockerapp package manager after latest Target changes.

patrickvacek:
The test is failing for me locally and not yet in CI so this may or may not be a good idea to merge right now...


lbonn:
I would say it's still good to merge it if we're confident it's the right fix and work on CI later. That would still be better than the code not compiling.

Or we can ask @doanac to double-check if relevant.

patrickvacek:
> I would say it's still good to merge it if we're confident it's the right fix and work on CI later.

Too late, already did it, and the local failure on my machine was just caused by some lingering garbage in my build dir. Looks like this is good to go.

doanac:
thanks guys. looks good. sorry i missed that.

patrickvacek:
> thanks guys. looks good. sorry i missed that.

No problem, it was my recent PR that broke this, and we hadn't put your test in CI until now. I don't blame you for not messing with our CI configuration!

PR #1266, opened 2019-07-24
Don't download targets with a length of 0.

patrickvacek:
Prevents a divide-by-zero error.

I found this while seeing what happens when aktualizr tries to download an OSTree target with the fake package manager. With this fix, it "works" as much as downloading anything else "works" with the fake package manager.


lbonn:
Does it need an unit test? Would be hard to test for the log message but it can at least check that it doesn't crash.

patrickvacek:
> Does it need an unit test? Would be hard to test for the log message but it can at least check that it doesn't crash.

Done and rebased. Basically I just test that the download function doesn't get called.

PR #1267, opened 2019-07-25
Fix/ota 3057/url determination

patrickvacek:
The most interesting part is in sotauptaneclient.cc, where if the custom URL is not set in the Director target metadata, we check if it is set in the Images repo target metadata and use that if it is.

A lot of the changes relate to aktualizr-repo, where I added parameters to set the custom URL.

### DIFF
--- a/src/aktualizr_repo/director_repo.cc
+++ b/src/aktualizr_repo/director_repo.cc
@@ -1,7 +1,7 @@
 #include "director_repo.h"
 
 void DirectorRepo::addTarget(const std::string &target_name, const Json::Value &target, const std::string &hardware_id,
-                             const std::string &ecu_serial) {
+                             const std::string &ecu_serial, const std::string &url) {
   const boost::filesystem::path current = path_ / "repo/director/targets.json";
   const boost::filesystem::path staging = path_ / "repo/director/staging/targets.json";
 
@@ -17,6 +17,11 @@ void DirectorRepo::addTarget(const std::string &target_name, const Json::Value &
   director_targets["targets"][target_name] = target;
   director_targets["targets"][target_name]["custom"].removeMember("hardwareIds");
   director_targets["targets"][target_name]["custom"]["ecuIdentifiers"][ecu_serial]["hardwareId"] = hardware_id;
+  if (!url.empty()) {
+    director_targets["targets"][target_name]["custom"]["uri"] = url;
+  } else {
+    director_targets["targets"][target_name]["custom"].removeMember("uri");
+  }
   director_targets["version"] = (Utils::parseJSONFile(current)["signed"]["version"].asUInt()) + 1;
   Utils::writeFile(staging, Utils::jsonToCanonicalStr(director_targets));
   updateRepo();
--- a/src/aktualizr_repo/image_repo.cc
+++ b/src/aktualizr_repo/image_repo.cc
@@ -19,7 +19,7 @@ void ImageRepo::addImage(const std::string &name, Json::Value &target, const std
 }
 
 void ImageRepo::addBinaryImage(const boost::filesystem::path &image_path, const boost::filesystem::path &targetname,
-                               const std::string &hardware_id, const Delegation &delegation) {
+                               const std::string &hardware_id, const std::string &url, const Delegation &delegation) {
   boost::filesystem::path repo_dir(path_ / "repo/image");
 
   boost::filesystem::path targets_path = repo_dir / "targets";
@@ -37,11 +37,14 @@ void ImageRepo::addBinaryImage(const boost::filesystem::path &image_path, const 
   target["hashes"]["sha256"] = boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest(image)));
   target["hashes"]["sha512"] = boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha512digest(image)));
   target["custom"]["targetFormat"] = "BINARY";
+  if (!url.empty()) {
+    target["custom"]["uri"] = url;
+  }
   addImage(targetname.string(), target, hardware_id, delegation);
 }
 
 void ImageRepo::addCustomImage(const std::string &name, const Uptane::Hash &hash, const uint64_t length,
-                               const std::string &hardware_id, const Delegation &delegation,
+                               const std::string &hardware_id, const std::string &url, const Delegation &delegation,
                                const Json::Value &custom) {
   Json::Value target;
   target["length"] = Json::UInt(length);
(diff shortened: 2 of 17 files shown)


src/aktualizr_repo/director_repo.cc
```diff
@@ -17,6 +17,11 @@ void DirectorRepo::addTarget(const std::string &target_name, const Json::Value &
   director_targets["targets"][target_name] = target;
   director_targets["targets"][target_name]["custom"].removeMember("hardwareIds");
   director_targets["targets"][target_name]["custom"]["ecuIdentifiers"][ecu_serial]["hardwareId"] = hardware_id;
+  if (!url.empty()) {
+    director_targets["targets"][target_name]["custom"]["uri"] = url;
+  } else {
+    director_targets["targets"][target_name]["custom"].removeMember("uri");
```
mike-sul: Why a nested structure is required for the custom URI, not just a single attribute of `target_name`, to paraphrase why ["custom"]["uri"] not just ["custom-uri"] or there is/are other attribute(s) of ["custom"] ?
Perhaps, just the whole "custom" can be removed ?

src/aktualizr_repo/image_repo.cc
```diff
@@ -37,11 +37,14 @@ void ImageRepo::addBinaryImage(const boost::filesystem::path &image_path, const
   target["hashes"]["sha256"] = boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest(image)));
```
mike-sul: I know that it's not related to this specific PR, but loading the whole file/image into memory doesn't look like optimal way to calculate file hash, unless aktualizr-repo is used just for testing purposes.

src/aktualizr_repo/director_repo.cc
patrickvacek: > there is/are other attribute(s) of ["custom"] ?

Precisely. It's following the recommendation of Uptane to keep implementation-specific custom metadata in a `custom` section.

src/aktualizr_repo/image_repo.cc
patrickvacek: Valid point. This is not what we do in libaktualizr itself, but aktualizr-repo is really just for testing, so it's not so critical.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -528,26 +528,31 @@ std::unique_ptr<Uptane::Target> SotaUptaneClient::findTargetInDelegationTree(con
...
+    // If the URL from the Director is unset, but the URL from the Images repo
+    // is set, use that.
+    if (it.uri().empty() && !images_target->uri().empty()) {
+      it.setUri(images_target->uri());
+    }
   }
-  for (auto it = targets.cbegin(); it != targets.cend(); ++it) {
-    auto res = downloadImage(*it, token);
+  for (const auto &it : targets) {
+    auto res = downloadImage(it, token);
```
mike-sul: for some reason downloadImage() passes a target object by value, maybe it makes sense to pass it by reference ?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -528,26 +528,31 @@ std::unique_ptr<Uptane::Target> SotaUptaneClient::findTargetInDelegationTree(con
...
     if (res.first) {
       downloaded_targets.push_back(res.second);
     }
```
mike-sul: Why do we need to continue downloading if a current target image download fails, don't we consider a singe multi-target update as an atomic operation  ?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -582,10 +587,9 @@ void SotaUptaneClient::reportResume() {
   report_queue->enqueue(std_::make_unique<DeviceResumedReport>(correlation_id));
 }
 
-std::pair<bool, Uptane::Target> SotaUptaneClient::downloadImage(Uptane::Target target,
+std::pair<bool, Uptane::Target> SotaUptaneClient::downloadImage(const Uptane::Target &target,
```
mike-sul: Oh, you did it here :)

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -41,8 +41,7 @@ class SotaUptaneClient {
 
   void initialize();
   void addNewSecondary(const std::shared_ptr<Uptane::SecondaryInterface> &sec);
-  result::Download downloadImages(const std::vector<Uptane::Target> &targets,
-                                  const api::FlowControlToken *token = nullptr);
+  result::Download downloadImages(std::vector<Uptane::Target> targets, const api::FlowControlToken *token = nullptr);
```
mike-sul: Why not by reference, is it because of the move semantics ?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -528,26 +528,31 @@ std::unique_ptr<Uptane::Target> SotaUptaneClient::findTargetInDelegationTree(con
...
+result::Download SotaUptaneClient::downloadImages(std::vector<Uptane::Target> targets,
                                                   const api::FlowControlToken *token) {
   // Uptane step 4 - download all the images and verify them against the metadata (for OSTree - pull without
   // deploying)
   std::lock_guard<std::mutex> guard(download_mutex);
   result::Download result;
   std::vector<Uptane::Target> downloaded_targets;
-  for (auto it = targets.cbegin(); it != targets.cend(); ++it) {
-    auto images_target = findTargetInDelegationTree(*it);
+  for (auto &it : targets) {
```
lbonn: When using this style of `for`, it makes it clearer to rename the variable to something other than `it`, as it's not an iterator anymore.

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: This was changed because the targets inside the vectors are modified with `setUri()` and we probably don't want to change it in the vector that lives in the caller.

That's a bit annoying though, I would maybe suggest passing by const reference and making a copy inside the function so that it's more explicit.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Not necessarily. That logic is still being worked out as far as I understand. That should be up to the user to decide what to do. For now, we attempt to download everything regardless of failures.

mike-sul: Maybe, makes sense to make it configurable.

patrickvacek: Agreed. That requires some integration with the backend, though, and we haven't sorted out the details yet.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: >  I would maybe suggest passing by const reference and making a copy inside the function so that it's more explicit. (+ a comment about that)

Agreed, done.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Good call. Done.

PR #1268, opened 2019-07-25
Send device data before UptaneCycle or RunForever.

patrickvacek:
It was only being sent on the check run mode, but it should also be sent on once and full (the default).

### DIFF
--- a/src/aktualizr_primary/main.cc
+++ b/src/aktualizr_primary/main.cc
@@ -158,8 +158,10 @@ int main(int argc, char *argv[]) {
       result::UpdateCheck update_result = aktualizr.CheckUpdates().get();
       aktualizr.Install(update_result.updates).get();
     } else if (run_mode == "once") {
+      aktualizr.SendDeviceData().get();
       aktualizr.UptaneCycle();
     } else {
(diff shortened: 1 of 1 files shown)


lbonn:
Also if it fails once, it won't be sent again until aktualizr restarts?

patrickvacek:
> Also if it fails once, it won't be sent again until aktualizr restarts?

Yes, for better or worse. At least it gets sent at all.

lbonn:
> > Also if it fails once, it won't be sent again until aktualizr restarts?
> 
> Yes, for better or worse. At least it gets sent at all.

Yes that's at least better for now. But should we plan a retry?

patrickvacek:
> But should we plan a retry?

I have no objection. There's probably a way to be more clever about this. The problem is that currently this is all handled via the API so should the retry logic live in aktualizr_primary or inside libaktualizr?

patrickvacek:
@lbonn the tests are failing because of your test tag releases. :( Shall we merge anyway or is there a quick fix?

lbonn:
Ow oops... I've deleted the tag so I hope that a restart should be enough. Annoying that it's doing that though

PR #1269, opened 2019-07-25
Move GitHub release generation to gitlab

lbonn:


OYTIS:
This pull request **introduces 1 alert** when merging c436e41c91b69ce0c7a3d2a6674cfb9106ec112c into 9f0c189d1f9e4d0238d6ef53738a920ef3112f18 - [view on LGTM.com](https://lgtm.com/projects/g/advancedtelematic/aktualizr/rev/pr-58c1b7e52b979166d844bcc8e7975f0e423617b3)

**new alerts:**

* 1 for Unused local variable

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

scripts/make_src_archive.sh
```diff
@@ -5,7 +5,11 @@ set -eu
 REPO=$1
 OUTPUT=$2
 
-pip3 install --user "https://files.pythonhosted.org/packages/60/44/26791beafa5c2c1f8ac0dd38a7761a0162c0f31bdad91a0091a48a923a44/git_archive_all-1.19.4-py2.py3-none-any.whl"
+python3 -m venv venv
+
+. venv/bin/activate
+
+pip3 install "https://files.pythonhosted.org/packages/60/44/26791beafa5c2c1f8ac0dd38a7761a0162c0f31bdad91a0091a48a923a44/git_archive_all-1.19.4-py2.py3-none-any.whl"
```
mike-sul: Referring to a specific package/wheel version might cause an issue once the package is removed from repo.

lbonn: Pypi strongly discourages removing released packages, so it shouldn't really happen. The bigger risk would be that the pypi account of the maintainer is compromised and serve a malicious new version.

I didn't add a hash check though, so I could add one or maybe I might as well do `pip install 'git_archive_all==1.19.4'`.

mike-sul: Yeah, but I experienced it at the previous job :). Will install 'git_archive_all>=1.19.4' work ?

lbonn: That's exactly what's vulnerable against a compromise.

Was it also a package from PyPi that went missing?

mike-sul: Yeah, you're right from hacking/vulnerability standpoint of view.
Yes, but in the case I mentioned we actually had our own repo/PyPi server and we couldn't keep all versions there for forever and were cleaning older packages time to time, I guess official PyPi/pythonhosted won't keep all versions for forever either... 
Anyway, it's not really an issue, the version can be updated once the current doesn't work.

lbonn: Ah I see. Well I think that PyPi intends to keep every version by design, so that nothing breaks in the future. Though I haven't found that spelled out explicitly after a quick search. Some repositories, like cargo are even more strict and don't even allow the publishers to remove an existing version.

I'll maybe change it to the simpler install syntax, after I've verified that this actually works end-to-end (waiting for a bitbake on CI right now...)

lbonn:
Proofs:

https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/52242
https://github.com/advancedtelematic/aktualizr/releases/tag/2019.6-test (will be removed)

lbonn:
Yes that was definitely annoying when testing. And if ptest breaks later because of upstream, it could become annoying. Maybe the ptest step could be marked as non-fatal?

patrickvacek:
> Maybe the ptest step could be marked as non-fatal?

If that's doable, that's not a bad idea. But we'd still want to know what failed (and that it failed) so that we can look into it.

lbonn:
This should do it https://docs.gitlab.com/ee/ci/yaml/README.html#allow_failure

PR #1270, opened 2019-07-29
Note that wifi SSID is case sensitive.

patrickvacek:
Based on a suggestion from @smilingwombat.


docs/posix-secondaries-bitbaking.adoc
```diff
@@ -102,15 +102,15 @@ By default ethernet NIC is used which implies that Raspberry Pi is connected to
 
 |`RPI_WIFI_ENABLE`
 |`"0"`
-|A flag to enable/disable (default) WiFi on RPi board
+|A flag to enable/disable (default) WiFi on RPi board.
 
 |`RPI_WIFI_SSID`
 |`N/A (mandatory if wifi is enabled)`
-|ID of a WiFi network to connect to
+|ID of a WiFi network to connect to. Be aware this is case sensitive!
```
merltron: I dunno if we need to shout at people...

```suggestion
|SSID of a WiFi network to connect to. The SSID is case sensitive.
```

PR #1271, opened 2019-07-29
Fix/ota 3059/target check sooner

patrickvacek:
Note that this will soon be followed up with a PR that will move `uptaneOfflineIteration` to be used before downloading and installing so that metadata will get quickly re-checked before allowing either of those operations to proceed. We can optionally hold this back until that work is ready. That work will also expand the new target_mismatch_test.cc file.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -534,28 +534,9 @@ result::Download SotaUptaneClient::downloadImages(const std::vector<Uptane::Targ
   // deploying)
   std::lock_guard<std::mutex> guard(download_mutex);
   result::Download result;
-  std::vector<Uptane::Target> checked_targets;
   std::vector<Uptane::Target> downloaded_targets;
-  // Copy the targets while iterating over them so that we can change the URL if
-  // necessary.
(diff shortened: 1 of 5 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -730,29 +711,52 @@ result::UpdateCheck SotaUptaneClient::checkUpdates() {
...
+                                 Utils::parseJSON(director_targets), "");
+    if (updates.size() == 1) {
+      LOG_INFO << "1 new update found in Uptane metadata.";
     } else {
-      result = result::UpdateCheck(updates, ecus_count, result::UpdateStatus::kNoUpdatesAvailable,
-                                   Utils::parseJSON(director_targets), "");
-      LOG_DEBUG << "No new updates found in Uptane metadata.";
+      LOG_INFO << updates.size() << " new updates found in Uptane metadata.";
     }
+  } else {
```
lbonn: Minor readability: do an early exit `if (updates.emtpy()) { return ... }` and no else, like it's done higher up in the function. Would make it a bit easier to follow as it's the style of the rest of the logic and would avoid the `if { for { if {}}}` nesting

PR #1272, opened 2019-07-30
build and deploy docker image to docker hub

rdanitz:


ci/gitlab/.gitlab-ci.yml
```diff
@@ -265,3 +265,11 @@ pages:
   # would be nice to have on a per-PR basis: # https://gitlab.com/gitlab-org/gitlab-ce/issues/35141
   only:
     - master
+
+deploy docker image:
+  stage: deploy
+  script:
+    - ./docker/docker-build.sh
+    - docker push advancedtelematic/aktualizr-app:latest
```
lbonn: As we've talked about it, we've yet haven't added any access to dockerhub from gitlab. If you can work with the gitlab's docker registry that would make things slightly easier.

And it would be good to have a way to try the stage. Maybe with a `only: variables: - $xxx` clause, or pushed without `only: - master` in some other branch.

ci/gitlab/.gitlab-ci.yml
```diff
@@ -5,6 +5,7 @@ stages:
   - oe-checkout
   - oe-test
   - deploy
+  - build downstream
```
lbonn: Maybe the stage can be renamed "trigger" and we can add others there when they arise (sounds like qa wants that too)

ci/gitlab/.gitlab-ci.yml
```diff
@@ -265,3 +266,20 @@ pages:
...
+  stage: deploy
+  allow_failure: true
+  before_script:
+    - docker login -u gitlab-ci-token -p "$CI_JOB_TOKEN" "$CI_REGISTRY"
+  script:
+    - sed 's@$BUILDER@'$UBUNTU_BIONIC_MASTER_IMAGE'@' ./ci/gitlab/Dockerfile.aktualizr > ./ci/gitlab/Dockerfile
+    - docker build -t $CI_REGISTRY_IMAGE:ci-$CI_COMMIT_REF_SLUG -f ./ci/gitlab/Dockerfile .
+    - docker push $CI_REGISTRY_IMAGE:ci-$CI_COMMIT_REF_SLUG
+
+build device-farm:
```
lbonn: What about `trigger-e2e-device-farm` as a stage name?

ci/gitlab/.gitlab-ci.yml
```diff
@@ -265,3 +266,20 @@ pages:
   # would be nice to have on a per-PR basis: # https://gitlab.com/gitlab-org/gitlab-ce/issues/35141
   only:
     - master
+
+deploy docker image:
```
lbonn: Rename to `app-docker-image`?

PR #1277, opened 2019-08-02
Added exclusion of test code.

patrickvacek:
The important part is to ignore the third-party libraries pulled in. ORT
mostly complains about Python libraries used by tuf-test-vectors. The
test code in the src directory is probably fine.


patrickvacek:
This is apparently good enough for ORT and can be merged upon successful review.

PR #1276, opened 2019-08-01
Create new "aktualizr-get" command

doanac:
I find myself carefully crafting `wget` commands to debug issues
between my device and device-gateway. This has gotten more complex
now that my team is using HSMs. This adds a simple tool to work
a bit like wget.

Signed-off-by: Andy Doan <andy@foundries.io>

PR #1279, opened 2019-08-02
Rename aktualizr-repo to uptane-generator.

patrickvacek:
Is this too wordy? I'm open to suggestions or abbreviations, but I want to move away from "aktualizr-repo", which is way to ambiguous and confusing.


lbonn:
There are some variables and function names with remnants of the former tool name: `akrepo`, `akt_repo`.

Zee314159:
> Is this too wordy? I'm open to suggestions or abbreviations, but I want to move away from "aktualizr-repo", which is way to ambiguous and confusing.

aktualizr-uptane-gen  or aktualizr-xxx-gen ?

xcheng-here:
I think it's clear but a bit wordy. Good for customer to understand what it is but maybe a little bit long to us to say it in discussing. Maybe it could have an abbreviation at same time ?

patrickvacek:
> aktualizr-uptane-gen or aktualizr-xxx-gen ?

aktualizr-gen-uptane? Or is even just aktualizr-generator good enough? I'm open to discussion and will rebase/redo once we reach a final conclusion.

lbonn:
If we want a shorter name, we can also drop the "aktualizr" part: "uptane-generator", "make-uptane"...

Zee314159:
Aktualizr-meta-gen？

patrickvacek:
> If we want a shorter name, we can also drop the "aktualizr" part

I thought we had decided that all aktualizr tools should be prefixed by `aktualizr` to prevent name collisions and make it clear that these are related, adjacent tools.

lbonn:
> I thought we had decided that all aktualizr tools should be prefixed by `aktualizr` to prevent name collisions and make it clear that these are related, adjacent tools.

Ah I was not aware of this specific discussion. But then the question is how "related" and "adjacent" the tool is. It's clear for aktualizr-info which really complements it but in this case it's mostly not aktualizr-specific (in theory at least), we just happen to make use of it for our aktualizr test suite.

patrickvacek:
> But then the question is how "related" and "adjacent" the tool is. It's clear for aktualizr-info which really complements it but in this case it's mostly not aktualizr-specific (in theory at least), we just happen to make use of it for our aktualizr test suite.

That's a good point. The only things that really need the prefix are probably bundled tools that are useless without aktualizr. `aktualizr-repo` is intended for use with aktualizr, but it's really specific to it at all, nor is it bundled for release in the deb packages, nor is it packaged for distribution via Yocto.

Hence, perhaps "uptane-generator" is good enough as it is. Any objections?

mike-sul:
what about uptane-repo-gen ? :)

patrickvacek:
Updated to "uptane-generator" after today's conversation. Now's your chance to complain.

patrickvacek:
I cannot get that first Travis test to pass for this. It keeps timing out at 50 minutes. Should we merge anyway, or is there something else to try here?

lbonn:
The problem is that I've been cheating with ccache to make it fast enough recently. Since this PR touches a lot of files, it must have invalidated a lot of cache, so we're back to the old running time + overhead of downloading cache.

But whatever, let's not worry about travis for merging.

patrickvacek:
> The problem is that I've been cheating with ccache to make it fast enough recently. Since this PR touches a lot of files, it must have invalidated a lot of cache, so we're back to the old running time + overhead of downloading cache.

Agreed. But I'm confused on one point: does the ccache not get updated if the Travis build doesn't complete? I'm not surprised the first attempt failed, but I was hoping the incomplete build would've made the next one faster, and that doesn't seem to be the case.

Also, https://github.com/advancedtelematic/aktualizr/pull/1276 is failing for the same reason despite not changing much.

> But whatever, let's not worry about travis for merging.

Agreed, at least for this case.

lbonn:
> Agreed. But I'm confused on one point: does the ccache not get updated if the Travis build doesn't complete? I'm not surprised the first attempt failed, but I was hoping the incomplete build would've made the next one faster, and that doesn't seem to be the case.

Unfortunately no, if the build timeouts nothing is uploaded. There is a trick we could use that was mentioned in a blog post: set a timeout in the test itself so that it fails but never timeouts from the point of view of Travis. We might resort to that at some point.

patrickvacek:
> Unfortunately no, if the build timeouts nothing is uploaded.

Ah, I see. So that means if a build is timing out, we'll get stuck in a cycle like the current one where we have to wait until we just get lucky and it succeeds. Could we at least update the ccache after compilation but before running the tests?

PR #1280, opened 2019-08-05
libaktualizr-c: don't set rpath for api-test

eu-smirnov:
By default, cmake adds all paths to the shared libraries during the linking process to rpath and removes it only on install step. This makes bitbake unhappy.

### DIFF
--- a/src/libaktualizr-c/test/CMakeLists.txt
+++ b/src/libaktualizr-c/test/CMakeLists.txt
@@ -1,6 +1,8 @@
 SET(TARGET_NAME api-test)
 SET(SOURCES api-test.c)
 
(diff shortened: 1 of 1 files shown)


lbonn:
To summarize the offline discussion:

- this only affects `api-test` that links with a dynamic library that we build ourselves
- running the test will probably need setting `LD_LIBRARY_PATH`. One question though, how is it run on CI then?
- CMake strips these when installing targets

Does it sound correct?

eu-smirnov:
@lbonn Yes, everything is correct. `api-test` isn't running on CI now, but later I plan to call it from googletest, and we can set the `LD_LIBRARY_PATH` there as well.

PR #1281, opened 2019-08-06
Veracode bis

lbonn:
Small rework of #1261.

One significant change is that the condition is now on a variable being present and not just "schedule". It makes it a bit more flexible: it's possible to launch this stage manually and it leaves the possibility to have schedules that don't include this test.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -104,6 +104,49 @@ nop11:
   script:
     - ./scripts/test.sh
 
+scan-build:
+  stage: test
+  only:
+    - schedules
+  # prepare build to be submitted for static code analysis
+  variables:
+    TEST_BUILD_DIR: 'scan-build'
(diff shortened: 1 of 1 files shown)


lbonn:
Example of a manual trigger https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/59899

zabbal:
Looks good to me overall. Few cosmetic questions:
- why do we need empty "dependencies" entries?
- why the big move of scan/pkg part? It makes it harder to review the diff

lbonn:
For `dependencies: []`: by default all previous artifacts are downloaded and these stages don't require any https://docs.gitlab.com/ee/ci/yaml/#artifactspaths. This is not related to your changes actually.

And I moved it in its own section because it won't run on PRs, so most aktualizr contributors won't even need to look at it. That's the same thing for e2e's pipeline.
Sorry about the diff with your commit, the point was to have a clean diff with master.

PR #1282, opened 2019-08-07
fix/ota-2421/disallow downloading non-OSTree binaries

Zee314159:
Draft code, just some thoughts to address this issue.

### DIFF
--- a/src/libaktualizr/primary/CMakeLists.txt
+++ b/src/libaktualizr/primary/CMakeLists.txt
@@ -26,8 +26,14 @@ if (BUILD_OSTREE)
         ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:ostree_mock>
         LABELS "noptest")
     target_link_libraries(t_aktualizr_fullostree virtual_secondary)
+    add_aktualizr_test(NAME download_nonostree SOURCES  download_nonostree_test.cc PROJECT_WORKING_DIRECTORY ARGS $<TARGET_FILE:uptane-generator> ${PROJECT_BINARY_DIR}/ostree_repo)
+    add_dependencies(t_download_nonostree uptane-generator make_ostree_sysroot)
+    set_tests_properties(test_download_nonostree PROPERTIES
+        LABELS "noptest")
+    target_link_libraries(t_download_nonostree virtual_secondary)
 else (BUILD_OSTREE)
     aktualizr_source_file_checks(aktualizr_fullostree_test.cc)
+    aktualizr_source_file_checks(download_nonostree_test.cc)
 endif (BUILD_OSTREE)
 
 add_aktualizr_test(NAME reportqueue SOURCES reportqueue_test.cc PROJECT_WORKING_DIRECTORY LIBRARIES PUBLIC uptane_generator_lib)
--- a/src/libaktualizr/primary/download_nonostree_test.cc
+++ b/src/libaktualizr/primary/download_nonostree_test.cc
@@ -1,0 +1,95 @@
+#include <gtest/gtest.h>
+
+#include <boost/process.hpp>
+
+#include "config/config.h"
+#include "logging/logging.h"
+#include "package_manager/ostreemanager.h"
+#include "primary/aktualizr.h"
+#include "storage/sqlstorage.h"
+#include "test_utils.h"
+#include "uptane_test_common.h"
+
+static std::string server = "http://127.0.0.1:";
+static std::string treehub_server = "http://127.0.0.1:";
+static boost::filesystem::path sysroot;
+
+/*
+ * Reject non-OSTree binaries.
+ */
+TEST(Aktualizr, DownloadNonOstreeBin) {
+  TemporaryDirectory temp_dir;
+  Config conf = UptaneTestCommon::makeTestConfig(temp_dir, server);
+  conf.pacman.type = PackageManager::kOstree;
+  conf.pacman.sysroot = sysroot.string();
+  conf.pacman.ostree_server = treehub_server;
+  conf.pacman.os = "dummy-os";
+  conf.provision.device_id = "device_id";
+  conf.provision.ecu_registration_endpoint = server + "/director/ecus";
+  conf.tls.server = server;
+
+  {
+    std::shared_ptr<INvStorage> storage = INvStorage::newStorage(conf.storage);
+    auto uptane_client = SotaUptaneClient::newDefaultClient(conf, storage);
+    uptane_client->initialize();
+    EXPECT_FALSE(uptane_client->uptaneIteration());
+    EXPECT_STREQ(uptane_client->getLastException().what(),
+                 "The target had a non-OSTree package that can not be installed on an OSTree system.");
+  }
+}
+
+#ifndef __NO_MAIN__
+int main(int argc, char **argv) {
+  ::testing::InitGoogleTest(&argc, argv);
+
+  logger_init();
+
+  if (argc != 3) {
+    std::cerr << "Error: " << argv[0] << " requires the path to the uptane-generator utility "
+              << "and an OStree sysroot\n";
+    return EXIT_FAILURE;
+  }
+
+  boost::filesystem::path uptane_generator_path;
+  uptane_generator_path = argv[1];
+
+  TemporaryDirectory meta_dir;
+  TemporaryDirectory temp_sysroot;
+  sysroot = temp_sysroot / "sysroot";
+
+  // uses cp, as boost doesn't like to copy bad symlinks
+  int ret = system((std::string("cp -r ") + argv[2] + std::string(" ") + sysroot.string()).c_str());
+  if (ret != 0) {
+    return -1;
+  }
+
+  Process ostree("ostree");
+
+  std::string new_rev = ostree.lastStdOut();
+  boost::trim_if(new_rev, boost::is_any_of(" \t\r\n"));
+
+  std::string port = TestUtils::getFreePort();
+  server += port;
+  boost::process::child http_server_process("tests/fake_http_server/fake_test_server.py", port, "-m", meta_dir.Path());
+  TestUtils::waitForServer(server + "/");
+  std::string treehub_port = TestUtils::getFreePort();
+  treehub_server += treehub_port;
+  TemporaryDirectory treehub_dir;
+  boost::process::child ostree_server_process("tests/sota_tools/treehub_server.py", std::string("-p"), treehub_port,
+                                              std::string("-d"), treehub_dir.PathString(), std::string("-s0.5"),
+                                              std::string("--create"));
+  TestUtils::waitForServer(treehub_server + "/");
+
+  Process uptane_gen(uptane_generator_path.string());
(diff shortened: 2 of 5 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -447,6 +442,15 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets, u
         is_new = true;
       }
 
+      if (primary_ecu_serial == ecu_serial) {
+        if (!target.IsOstree() &&
+            (config.pacman.type == PackageManager::kOstree || config.pacman.type == PackageManager::kOstreeDockerApp)) {
+          LOG_ERROR << "Cannot install a non-OSTree package on an OSTree system";
+          sendEvent<event::InstallTargetComplete>(primary_ecu_serial, false);
```
mike-sul: IMHO, I'd make this code more generic and extendable, i.e. instead of comparing a specific target type with specific package manager types do something like the following:
`if (target.type != config.pacman.type)`

mike-sul: In other cases when `is_new = false` the event `event::InstallTargetComplete` is not sent, I am just wondering if it's correct place to do it here.

patrickvacek: > IMHO, I'd make this code more generic and extendable

I like the idea but currently we don't really have a good way to support that. The only type that we can recognize in the Target metadata is OSTree, so for now, that's the only one worth checking.

> In other cases when `is_new = false` the event `event::InstallTargetComplete` is not sent, I am just wondering if it's correct place to do it here.

I agree, instead of sending that event, we should just set `last_exception` and return false. The log message is fine, though.

patrickvacek:
The logic looks about right, but we need a test for this. It will need to use the OSTree package manager, but it won't need to actually use it since the update will already fail at the check for updates.

patrickvacek:
> > It will need to use the OSTree package manager,
> 
> https://github.com/advancedtelematic/ostree-basic-pkg
> This tool？

I was referring to https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr/package_manager/ostreemanager.h. You may need to use this script to generate an OSTree sysroot for the test: https://github.com/advancedtelematic/aktualizr/blob/master/scripts/make_ostree_sysroot.sh. For an example of how to use that, see https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr/package_manager/CMakeLists.txt#L36 and https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr/package_manager/ostreemanager_test.cc.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -447,6 +442,15 @@ bool SotaUptaneClient::getNewTargets(std::vector<Uptane::Target> *new_targets, u
         is_new = true;
       }
 
+      if (primary_ecu_serial == ecu_serial) {
+        if (!target.IsOstree() &&
+            (config.pacman.type == PackageManager::kOstree || config.pacman.type == PackageManager::kOstreeDockerApp)) {
+          LOG_ERROR << "Cannot install a non-OSTree package on an OSTree system";
+          last_exception = Uptane::InvalidMetadata(target.filename(), "", "Non-OSTree package");
```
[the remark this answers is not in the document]
patrickvacek: That's not a bad choice, but since it's usually used for problems in parsing the metadata, maybe we should find something else. Nothing else looks better, though, so how about making a new exception? `InvalidTarget` perhaps?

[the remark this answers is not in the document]
Zee314159: Sure，I was consider making a new one too. Thanks!

Zee314159:
Totally raw! Maybe I should use json to make a fake target instead of using aktualizr tool to generate one? And the 'generate ostree sysroot' part also looks not good.

patrickvacek:
> Maybe I should use json to make a fake target instead of using aktualizr tool to generate one?

Nah, using aktualizr-repo is perfect. That looks fine to me.

> And the 'generate ostree sysroot' part also looks not good.

I also think that part is a bit overcomplicated. You don't actually need to mock the sysroot and all that, or at least I don't think so. It looks like you modeled that part off of the `test_aktualizr_fullostree`, but what you actually want is just to use the simpler method used by `ostreemanager_test`, which just uses a sysroot generated by the `make_ostree_sysroot.sh` script I mentioned. Since you don't have to actually install anything, you shouldn't need to do the whole complex mocking bit.

src/libaktualizr/primary/CMakeLists.txt
```diff
@@ -26,8 +26,16 @@ if (BUILD_OSTREE)
         ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:ostree_mock>
         LABELS "noptest")
     target_link_libraries(t_aktualizr_fullostree virtual_secondary)
+    add_aktualizr_test(NAME download_nonostree SOURCES  download_nonostree_test.cc PROJECT_WORKING_DIRECTORY ARGS $<TARGET_FILE:aktualizr-repo> ${PROJECT_BINARY_DIR}/ostree_repo)
+    set_target_properties(t_download_nonostree PROPERTIES LINK_FLAGS -Wl,--export-dynamic)
+    add_dependencies(t_download_nonostree ostree_mock aktualizr-repo make_ostree_sysroot)
+    set_tests_properties(test_download_nonostree PROPERTIES
+        ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:ostree_mock>
+        LABELS "noptest")
```
patrickvacek: I think you can do without the `set_target_properties` and `ostree_mock` references here.

src/libaktualizr/primary/download_nonostree_test.cc
```diff
@@ -0,0 +1,89 @@
...
+  aktualizr_repo_path = argv[1];
+  Process akt_repo(aktualizr_repo_path.string());
+  akt_repo.run({"image", "--path", meta_dir.PathString(), "--targetname", "update_1.0", "--targetsha256", new_rev,
+                "--targetlength", "0", "--targetformat", "BINARY", "--hwid", "primary_hw"});
+  akt_repo.run({"addtarget", "--path", meta_dir.PathString(), "--targetname", "update_1.0", "--hwid", "primary_hw",
+                "--serial", "CA:FE:A6:D2:84:9D"});
+  akt_repo.run({"signtargets", "--path", meta_dir.PathString(), "--correlationid", "abc123"});
+  LOG_INFO << akt_repo.lastStdOut();
+  // Work around inconsistent directory naming.
+  Utils::copyDir(meta_dir.Path() / "repo/image", meta_dir.Path() / "repo/repo");
```
patrickvacek: I think the problem here is that this metadata is not available in the test case. I recommend putting the aktualizr-repo bits in the test case itself. The key is making sure the directory the metadata is generated in is accessible to the python server that serves it and that that server is known to the Aktualizr instance.

patrickvacek:
If you rebase on latest master, you should be able to get the Travis timeout resolved.

I ran your branch locally and noticed this in the output: `image command requires --targetsha256 or --targetsha512, and --targetlength when --filename is not supplied.`. That's `uptane-generator` (former `aktualizr-repo`) complaining and might be part of what's going wrong.

Zee314159:
> If you rebase on latest master

Trapped in git chaos...I'll figure the way out tomorrow. 

> image command requires --targetsha256 or --targetsha512, and --targetlength when --filename is not supplied.. That's uptane-generator (former aktualizr-repo) complaining and might be part of what's going wrong.

I might should start with this issue first.

src/libaktualizr/primary/download_nonostree_test.cc
```diff
@@ -0,0 +1,117 @@
...
+  TestUtils::waitForServer(treehub_server + "/");
+  r = ostree.run({"rev-parse", std::string("--repo"), treehub_dir.PathString(), "master"});
+  if (std::get<0>(r) != 0) {
+    return -1;
+  }
+  new_rev = ostree.lastStdOut();
+  boost::trim_if(new_rev, boost::is_any_of(" \t\r\n"));
+  LOG_INFO << "DEST: " << new_rev;
+
+  Process akt_repo(aktualizr_repo_path.string());
```
patrickvacek: Not at all critical, but I recommend renaming the "aktualizr_repo" references here to "uptane_generator". (I also changed every "akt_repo" to just "gen".)

Zee314159:
> A few minor comments. As far as why the test is failing... perhaps my assumptions about how easy it would be to fake an ostree sysroot for this purpose was wrong. I'm not sure i understand why ostree is complaining, though. Have you been able to trace it with gdb?

No, I was trying to use the whole  mock the sysroot and unptane-generate part from aktuarlizr_fullostree_test. But aktuarlizr_fullostree_test will fail locally too.

src/libaktualizr/primary/download_nonostree_test.cc
```diff
@@ -0,0 +1,111 @@
...
+    std::shared_ptr<INvStorage> storage = INvStorage::newStorage(conf.storage);
+    OstreeManager ostree(conf.pacman, storage, nullptr, nullptr);
+
+    Aktualizr aktualizr(conf);
+
+    aktualizr.Initialize();
+
+    result::UpdateCheck update_result = aktualizr.CheckUpdates().get();
+    ASSERT_EQ(update_result.status, result::UpdateStatus::kError);
+  }
```
patrickvacek: Alright, I think I figured it out. The problem is that `CheckUpdates` calls `PutManifest`, which calls `getManifest`, which calls `getCurrent`, which doesn't work with a fake sysroot. Solution: work at a lower level, since we don't need to do any of that. I suggest replacing the section in braces here with this code:

```
    std::shared_ptr<INvStorage> storage = INvStorage::newStorage(conf.storage);
    auto uptane_client = SotaUptaneClient::newDefaultClient(conf, storage);
    uptane_client->initialize();
    EXPECT_FALSE(uptane_client->uptaneIteration());
    EXPECT_STREQ(uptane_client->getLastException().what(),
                 "The target had a non-OSTree package that can not be installed on an OSTree system.");
```

lbonn:
The code looks good but can we change the commit messages before merging?

src/libaktualizr/primary/download_nonostree_test.cc
```diff
@@ -0,0 +1,109 @@
...
+  // uses cp, as boost doesn't like to copy bad symlinks
+  int ret = system((std::string("cp -r ") + argv[2] + std::string(" ") + sysroot.string()).c_str());
+  if (ret != 0) {
+    return -1;
+  }
+  auto r = ostree.run(
+      {"rev-parse", std::string("--repo"), (sysroot / "/ostree/repo").string(), "generate-remote/generated"});
+  if (std::get<0>(r) != 0) {
+    return -1;
+  }
```
patrickvacek: Is this `ostree` process stuff still necessary? Seems like that's a relic of the `aktualizr_fullostree_test.cc` style.

PR #1283, opened 2019-08-07
TEST Target: aktualizr-cycle-simple depends on OSTree lib

kbushgit:

Test target "aktualizr-cycle-simple" indirectly depends on OSTree library since it use OstreeManager class.


patrickvacek:
Sorry, where does it actually use the `OstreeManager` class? I see the `include package_manager/ostreemanager.h`, but I don't see why that include is actually necessary.

kbushgit:
hm yes, you are right, let me check the reason why the aktualizr_cycle_simple.cc include package_manager/ostreemanager.h

lbonn:
Oops, yes it's probably a mistake it compiles without this include.

PR #1284, opened 2019-08-08
Fix issue with find package for the glib-2.0

kbushgit:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,11 @@ find_package(Git)
 find_package(Asn1c REQUIRED)
 find_package(ZLIB)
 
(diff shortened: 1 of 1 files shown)


lbonn:
It would perhaps help if there was a small comment explaining in which conditions this is needed. Otherwise, we'll miss a lot of context in the future and it risks being just removed at the next refactor.

patrickvacek:
@kbushgit This looks fine, but would you mind adding the request comment? Then we can finally merge this.

kbushgit:
The glib-2 package is added only when we build  SOTA_TOOLS, currently the problem with include path is not reproduced, so I think we can close this PR

PR #1285, opened 2019-08-08
Test/ota 2743/remove redundant ostree mock

mike-sul:
- remove redundant ostree-mock.cc
Next step is to come up with the solution that would allow easily run aktualizr with an ostree support locally.


patrickvacek:
> remove redundant ostree-mock.cc

Sorry, how was it redundant? I don't think I follow how the existing test will work without that.

mike-sul:
> > remove redundant ostree-mock.cc
> 
> Sorry, how was it redundant? I don't think I follow how the existing test will work without that.

The test `aktualizr_fullostree_test.cc` used two approaches to overriding/mocking the two desired  ostree-functions, one by means of the preloaded mock library (ostree_mock.cc) and another just by implementing/overriding it in the test. Why not use just one of the approaches ?

lbonn:
I should probably have tried the simple thing before doing the heavy lifting work. Apparently it's enough to just redefine a function that would be present in a dynamic library to just overload it... Would be curious to see a piece of doc that explains that though.

mike-sul:
I added needless commit into this PR by mistake, will redo this PR

mike-sul:
I'll add this change as part of another PR

PR #1286, opened 2019-08-08
Make the cert provider tests immuned to changes in an openssl format

kbushgit:


mike-sul:
@kbushgit Looks like `make format` should be executed.

PR #1288, opened 2019-08-09
Do not use the api_gateway.url.

patrickvacek:
Continue to use the old URLs provided by treehub.url and the ostree.server field of treehub.json.

This is a partial revert of 67949117.


patrickvacek:
Manual testing with https://github.com/advancedtelematic/meta-updater/pull/532 confirms that the bug is fixed. However, we need to update the creds in CI to be able to test this automatically. I think it's probably not worth holding this up for that unless that is easy to do.

Issue #1273, opened 2019-07-30
aktualizr fails to download reposerver files backed by S3


doanac:
I'm going to try and dig into this, but before I get too far I thought I'd describe the issue in case someone happens to already know the root issue. I have a tufreposerver deployed that's backed by an S3 bucked. When I call `PackageManagerInterface::fetchTarget` it gets the 302 redirect to a signed URL, but then the download fails. It looks like it may be something with the HttpClient sending the gateway certs to my S3 bucked (in my case its a google storage acting as s3):

~~~
   info: Looking for DockerApps to fetch
   info: Fetching shellhttpd -> Target(shellhttpd.dockerapp-98 ecu_identifiers: () length:312 hashes: (Hash: 7A26E16871423505EB0B92272680FCE4EFFBA1316259F91312CE7F27D42F8E2E, ))
*   Trying 35.222.49.182...
* TCP_NODELAY set
* Connected to ota-lite.foundries.io (35.222.49.182) port 8443 (#0)
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /tmp/aktualizr-7dce-ebe6-f48a-0673/011d-e559-tls-ca
  CApath: none
* SSL connection using TLSv1.2 / ECDHE-ECDSA-AES256-GCM-SHA384
* ALPN, server accepted to use http/1.1
* Server certificate:
*  subject: CN=ota-gateway
*  start date: Jun 14 03:35:51 2019 GMT
*  expire date: Jun 11 03:35:51 2029 GMT
*  subjectAltName: host "ota-lite.foundries.io" matched cert's "ota-lite.foundries.io"
*  issuer: CN=ota-server-CA
*  SSL certificate verify ok.
> GET /repo/targets/shellhttpd.dockerapp-98 HTTP/1.1
Host: ota-lite.foundries.io:8443
User-Agent: Aktualizr/1.0+gitAUTOINC+4397bdcfed
Content-Type: application/json
Accept: */*
x-ats-ostreehash: 4b16e43adbfa25eab74eaf5822a99209c085e7d01665a98af951a81b2e9be7a7

< HTTP/1.1 302 Found
< Server: openresty/1.15.8.1
< Date: Tue, 30 Jul 2019 04:23:27 GMT
< Content-Length: 0
< Connection: keep-alive
< x-ats-version: reposerver/0.6.0-19-ge074093
[...]

PR #1289, opened 2019-08-09
Send user agents with all our tools

lbonn:


src/libaktualizr/http/httpclient_test.cc
```diff
@@ -80,6 +80,25 @@ TEST(PostTest, put_performed) {
   EXPECT_EQ(json["data"]["key"].asString(), "val");
 }
 
+TEST(HttpClient, user_agent) {
+  {
+    HttpClient http;
+
+    const auto resp = http.get(server + "/user_agent", HttpInterface::kNoLimit);
+    const auto app = resp.body.substr(0, resp.body.find('/'));
+    EXPECT_EQ(app, "Aktualizr");
```
patrickvacek: Could you provide a brief comment to make it clear that you're testing the expected default here?

tests/test_utils.cc
```diff
@@ -55,19 +55,17 @@ void TestUtils::writePathToConfig(const boost::filesystem::path &toml_in, const
...
-  CURL *handle = curl_easy_init();
-  if (handle == nullptr) {
-    throw std::runtime_error("Could not initialize curl handle");
-  }
-  curlEasySetoptWrapper(handle, CURLOPT_URL, address.c_str());
-  curlEasySetoptWrapper(handle, CURLOPT_CONNECTTIMEOUT, 3L);
-  curlEasySetoptWrapper(handle, CURLOPT_NOBODY, 1L);
-  curlEasySetoptWrapper(handle, CURLOPT_VERBOSE, 1L);
-  curlEasySetoptWrapper(handle, CURLOPT_SSL_VERIFYPEER, 0L);
+  CurlEasyWrapper curl;
```
patrickvacek: +1

src/libaktualizr/http/httpclient_test.cc
lbonn: Good point!
Done.

src/libaktualizr/http/httpclient.cc
```diff
@@ -73,7 +61,7 @@ HttpClient::HttpClient(const std::vector<std::string>* extra_headers)
     }
   }
   curlEasySetoptWrapper(curl, CURLOPT_HTTPHEADER, headers);
-  curlEasySetoptWrapper(curl, CURLOPT_USERAGENT, user_agent.c_str());
+  curlEasySetoptWrapper(curl, CURLOPT_USERAGENT, Utils::getUserAgent());
```
mike-sul: It might override the user agent header specified by an HttpClient user/client in `extra_headers`.
IMHO, statically predefined headers should be specified in a HTTP client factory configuration  (which doesn't exist), so libaktualizr clients could specify desired mandatory headers in explicit way. Utils::set/getUserAgent() is an implicit API for libaktualizr clients/apps they need to be aware and set it before Aktualizr instantiation. 
This is just my two cents and can be ignored since it will require more work :)

lbonn: Yes, I agree this is not especially elegant. The main problem I had is that different garage tools are using some common code that instantiates an HttpClient, that's why I opted for the less intrusive global setting.

I don't see much use case for HttpClient users redefining user agents dynamically though, it's maybe the (rare) kind of thing that can be setup globally.

But then I also question my choice of putting the API in utils.h. I'll try moving it to http client itself.

mike-sul: No problem, I understand that we need to be practical and balance between an ideal but time consuming and a just enough but quicker solutions.
BTW, at first glance, this PR, IMHO, is related to this one https://github.com/advancedtelematic/aktualizr/issues/1273#issuecomment-520352928

doanac: lol - i've been watching this PR to see how things went before pushing my change which will at best produce a merge conflict.

lbonn: Haha well... This PR is not really urgent anyway. Would you also prefer that to be done another way?

doanac: no. I was just wanted to see the approach that was chosen here since there's a good chance I could do the same thing in my change.

lbonn:
Should we go forward anyway?

PR #1290, opened 2019-08-09
Clean all targets

lbonn:
Add api calls to list and delete targets in storage.

Tests in test_aktualizr forthcoming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 - Target object equality requires that hardware IDs match: [PR](https://github.com/advancedtelematic/aktualizr/pull/1258)
 - aktualizr-repo image command now requires a hardware ID: [PR](https://github.com/advancedtelematic/aktualizr/pull/1258)
 
+### Changed
+
+- Renamed `GetStoredTarget` to `OpenStoredTarget` in aktualizr API: [PR](https://github.com/advancedtelematic/aktualizr/pull/1290)
 
 ## [2019.5] - 2019-07-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 - aktualizr-info --wait-until-provisioned flag: [PR](https://github.com/advancedtelematic/aktualizr/pull/1253)
 - Target object equality requires that hardware IDs match: [PR](https://github.com/advancedtelematic/aktualizr/pull/1258)
 - aktualizr-repo image command now requires a hardware ID: [PR](https://github.com/advancedtelematic/aktualizr/pull/1258)
+- `GetStoredTargets` and `DeleteStoredTarget` methods: [PR](https://github.com/advancedtelematic/aktualizr/pull/1290)
 
 ### Changed
 
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -170,14 +170,14 @@ boost::signals2::connection Aktualizr::SetSignalHandler(
   return sig_->connect(handler);
 }
 
+std::vector<Uptane::Target> Aktualizr::GetStoredTargets() { return storage_->getTargetFiles(); }
+
+void Aktualizr::DeleteStoredTarget(const Uptane::Target &target) { storage_->removeTargetFile(target.filename()); }
+
 std::unique_ptr<StorageTargetRHandle> Aktualizr::OpenStoredTarget(const Uptane::Target &target) {
-  try {
-    auto handle = storage_->openTargetFile(target);
-    if (handle->isPartial()) {
-      return std::unique_ptr<StorageTargetRHandle>();
-    }
-    return storage_->openTargetFile(target);
-  } catch (...) {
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 - aktualizr-info --wait-until-provisioned flag: [PR](https://github.com/advancedtelematic/aktualizr/pull/1253)
 - Target object equality requires that hardware IDs match: [PR](https://github.com/advancedtelematic/aktualizr/pull/1258)
 - aktualizr-repo image command now requires a hardware ID: [PR](https://github.com/advancedtelematic/aktualizr/pull/1258)
+- `GetStoredTargets` and `DeleteStoredTarget` methods
+
+### Changed
+
+- Renamed `GetStoredTarget` to `OpenStoredTarget` in aktualizr api
```
patrickvacek: Minor, but I'd capitalize "API", and now that you have a PR, you can already fill in the links.

src/libaktualizr/storage/storage_common_test.cc
```diff
@@ -494,6 +494,47 @@ TEST(storage, store_target) {
   }
 }
 
+TEST(storage, list_remove_targets) {
```
patrickvacek: I'd like a brief comment explaining the precise features you're trying to test.

(I've given up on maintaining the actions list these days, since no one seems to care about it, but I still think the comments for each test that clarify the important points of the test are very useful.)

lbonn: Yes that's too bad actually. Part of the problem might be that it's a bit hard to keep track as it needs to be changed in two places.

If we could just define them where they are tested and have an automated tool to collect them, it would be a bit more helpful. But then, it would maybe take some non-trivial effort.

patrickvacek: > If we could just define them where they are tested and have an automated tool to collect them, it would be a bit more helpful. But then, it would maybe take some non-trivial effort.

100% agreed, and it was my hope that we would eventually do that. The one problem I foresaw was how to structure the information, though. That was the nice thing about the list. In any case, it probably isn't hard to read the comments around the tests; we started doing something similar with zephyr integration, but that also got stalled.

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -170,13 +170,30 @@ boost::signals2::connection Aktualizr::SetSignalHandler(
...
+bool Aktualizr::DeleteStoredTarget(const Uptane::Target &target) {
+  try {
+    storage_->removeTargetFile(target.filename());
+  } catch (...) {
+    return false;
+  }
+  return true;
+}
+
+std::unique_ptr<StorageTargetRHandle> Aktualizr::OpenStoredTarget(const Uptane::Target &target) {
```
eu-smirnov: While we're at it, can we change the return type to something like std::ofstream? This is an external API, and IMO we should try to expose less internal/implementation-specific details here.

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -170,13 +170,30 @@ boost::signals2::connection Aktualizr::SetSignalHandler(
...
   try {
     auto handle = storage_->openTargetFile(target);
     if (handle->isPartial()) {
       return std::unique_ptr<StorageTargetRHandle>();
     }
-    return storage_->openTargetFile(target);
+    return handle;
```
eu-smirnov: Maybe it would be better to throw an exception if somebody tries to open a not or partially downloaded target?

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -170,13 +170,30 @@ boost::signals2::connection Aktualizr::SetSignalHandler(
...
+    return storage_->getTargetFiles();
+  } catch (...) {
+    return {};
+  }
+}
+
+bool Aktualizr::DeleteStoredTarget(const Uptane::Target &target) {
+  try {
+    storage_->removeTargetFile(target.filename());
+  } catch (...) {
```
eu-smirnov: Probably we should discuss that again, but haven't we decided to let exceptions propagate through the C++ API level? That's also what other API calls do.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1425,20 +1459,26 @@ void SQLStorage::removeTargetFile(const std::string& filename) {
...
 
   if (statement.step() != SQLITE_DONE) {
     LOG_ERROR << "Statement step failure: " << db.errmsg();
     throw std::runtime_error("Could not remove target file");
   }
-  boost::filesystem::remove(images_path_ / filename);
+  try {
+    boost::filesystem::remove(images_path_ / filename);
+  } catch (std::exception& e) {
+    LOG_ERROR << "Could not remove target file";
```
eu-smirnov: Only printing an error seems insufficient to me. Maybe we should rethrow?

src/libaktualizr/primary/aktualizr.cc
lbonn: Yes, probably a good idea since we don't store the files in database, so the abstraction is not needed now.

I would leave that for another task/PR though.

lbonn:
@eu-smirnov I think I've addressed your comments.

PR #1291, opened 2019-08-10
aktualizr-lite: Improve logic for pacman->getCurrent()

doanac:
This is a 3 part change I've been carrying out-of-tree for a few weeks that I think makes aktualizr-lite act a little more like aktualizr. The first change allows aktualizr-lite to call getCurrent in CI. The second change finalizes installs in a similar way to the sotauptaneclient which then lets us change aktualizr-lite to use getCurrent the in last commit.

### DIFF
--- a/src/libaktualizr/package_manager/ostreemanager.cc
+++ b/src/libaktualizr/package_manager/ostreemanager.cc
@@ -272,8 +272,8 @@ Uptane::Target OstreeManager::getCurrent() const {
   storage_->loadPrimaryInstalledVersions(&installed_versions, nullptr, nullptr);
 
   // Version should be in installed versions
-  std::vector<Uptane::Target>::iterator it;
-  for (it = installed_versions.begin(); it != installed_versions.end(); it++) {
+  std::vector<Uptane::Target>::reverse_iterator it;
+  for (it = installed_versions.rbegin(); it != installed_versions.rend(); it++) {
     if (it->sha256Hash() == current_hash) {
       return *it;
     }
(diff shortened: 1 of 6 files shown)


src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -272,8 +272,8 @@ Uptane::Target OstreeManager::getCurrent() const {
   storage_->loadPrimaryInstalledVersions(&installed_versions, nullptr, nullptr);
 
   // Version should be in installed versions
-  std::vector<Uptane::Target>::iterator it;
-  for (it = installed_versions.begin(); it != installed_versions.end(); it++) {
+  std::vector<Uptane::Target>::reverse_iterator it;
+  for (it = installed_versions.rbegin(); it != installed_versions.rend(); it++) {
```
patrickvacek: Probably worth a comment to explain the reverse_iterator preference.

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -58,6 +58,7 @@ class SotaUptaneClient {
   bool isInstallCompletionRequired();
   void completeInstall();
   Uptane::LazyTargetsList allTargets();
+  Uptane::Target getCurrent() { return package_manager_->getCurrent(); }
```
patrickvacek: This is fine for now, but going forward it is likely we'll want something like this that is ECU-agnostic, i.e. doesn't just assume the primary. I just wanted to give you a heads-up so you're not surprised if/when that happens!

src/libaktualizr/package_manager/ostreemanager.cc
doanac: good point.

PR #1294, opened 2019-08-13
httpclient: Always set path to CA certs

doanac:
Fixes https://github.com/advancedtelematic/aktualizr/issues/1273

Some versions of OE don't explicitly set a default path to the
CA certs directory. This ensures its set and can be overidden if
needed.


doanac:
This isn't exactly how we discussed in the bug, but I decided to build off of https://github.com/advancedtelematic/aktualizr/pull/1289. It seemed you weren't super excited about that approach, so piling onto the idea might not be the best.

patrickvacek:
This seems fine to me but I want to make sure there's consensus around that. How do others feel?

PR #1295, opened 2019-08-14
Don't send device data twice by default.

patrickvacek:
This is a partial revert of 1df4c1652ee50b467bbcb09d519fb7bd9b558e09.

I overzealously added a SendDeviceData before the call to RunForever,
despite that RunForever also calls SendDeviceData. This removes that
redundancy.

This does not change the behavior of "aktualizr once" (which uses
UptaneCycle). It still calls SendDeviceData before UptaneCycle, which
seems to roughly match user expectations.


patrickvacek:
Merging despite Travis timeout, for better or worse.

PR #1297, opened 2019-08-14
docker-app: Fix clang-tidy warnings

doanac:

### DIFF
--- a/src/libaktualizr/package_manager/dockerappmanager.cc
+++ b/src/libaktualizr/package_manager/dockerappmanager.cc
@@ -3,12 +3,12 @@
 #include <sstream>
 
 struct DockerApp {
-  DockerApp(const std::string app_name, const PackageConfig &config)
+  DockerApp(std::string app_name, const PackageConfig &config)
       : name(std::move(app_name)),
-        app_root(std::move(config.docker_apps_root / app_name)),
-        app_params(std::move(config.docker_app_params)),
-        app_bin(std::move(config.docker_app_bin)),
-        compose_bin(std::move(config.docker_compose_bin)) {}
+        app_root(config.docker_apps_root / name),
+        app_params(config.docker_app_params),
+        app_bin(config.docker_app_bin),
+        compose_bin(config.docker_compose_bin) {}
 
   bool render(const std::string &app_content) {
     auto bin = boost::filesystem::canonical(app_bin).string();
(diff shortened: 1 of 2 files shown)


doanac:
I'm surprised CI hasn't caught this. It looks like the .travis.yml has `-eTEST_WITH_DOCKERAPP=1` when it calls test.sh

src/libaktualizr/package_manager/dockerappmanager.cc
```diff
@@ -3,12 +3,12 @@
...
 
 struct DockerApp {
-  DockerApp(const std::string app_name, const PackageConfig &config)
-      : name(std::move(app_name)),
-        app_root(std::move(config.docker_apps_root / app_name)),
-        app_params(std::move(config.docker_app_params)),
-        app_bin(std::move(config.docker_app_bin)),
-        compose_bin(std::move(config.docker_compose_bin)) {}
+  DockerApp(std::string app_name, const PackageConfig &config)
+      : name(app_name),
```
patrickvacek: I'm also a bit surprised. It all looks fine except for this bit: why not move `app_name`? Seems like an unnecessary copy.

doanac: Great question. I lost about 30 minutes of debugging time on that very issue. The problem is this line of code that comes *after* that std::move: `app_root(config.docker_apps_root / app_name),`. The app_name passed to that becomes an empty string and `app_root` winds up not being correct.

patrickvacek:
Also, if you rebase on latest master, I think we can fix the Travis timeout problem.

src/libaktualizr/package_manager/dockerappmanager.cc
patrickvacek: Oh wow, I see. Can you not use `name` instead of `app_name` in the latter line?

doanac: ah - better idea. i've done a force-push with rebase on master + plus this suggestion.

lbonn:
The reason it was not caught by CI is that `TEST_WITH_STATICTESTS` and `TEST_WITH_DOCKERAPP` are not enabled simultaneously on CI right now.
We could add `TEST_WITH_DOCKERAPP` to the debian testing jobs on Travis and gitlab.

PR #1299, opened 2019-08-15
Add a timeout for Travis runs

lbonn:
So that ccache can be built incrementally


lbonn:
Ok seems to work, here is the hack (stolen from https://blog.esciencecenter.nl/travis-caching-and-incremental-builds-6518b89ee889):

When Travis fails, it still uploads ccache but if it times out after 50s, the cache is not uploaded. So if we make our own command timeout after a well chosen delay, the build will be marked as failed but we'll have most of our c++ objects cache for the next try.

So if a build times out the first time, usually a retry will fix it.

PR #1301, opened 2019-08-19
Report failed downloads in an installation report.

patrickvacek:
This should inform the Director that the update failed.

This appears to fix the bug, but it is not perfect:
1) ~~There is no test, because we don't have a way to fake a download failure. It's probably doable with libfiu and the fake package manager, though.~~ This is now covered by tests.
2) There is the same problem with empty targets metadata that occurs with installation failure. It ~~could be~~ has been handled the same way, ~~although~~ despite that I don't like the solution of dropping metadata.
3) The installation report does not contain any information about how, why, or which download failed. That could be improved if anyone wants to see that.

We might want to consider merging testing and merging this as is anyway just to get the fix out the door, but further consideration is necessary to address the above issues.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -142,11 +142,11 @@ void SotaUptaneClient::finalizeAfterReboot() {
         // go forward again
         storage->saveInstalledVersion(ecu_serial.ToString(), target, InstalledVersionUpdateMode::kNone);
         report_queue->enqueue(std_::make_unique<EcuInstallationCompletedReport>(ecu_serial, correlation_id, false));
+        director_repo.dropTargets(*storage);  // fix for OTA-2587, listen to backend again after end of install
       }
 
       computeDeviceInstallationResult(nullptr, correlation_id);
       putManifestSimple();
-      director_repo.dropTargets(*storage);  // fix for OTA-2587, listen to backend again after end of install
     } else {
       // nothing found on primary
       LOG_ERROR << "Expected reboot after update on primary but no update found";
@@ -836,7 +836,7 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
   computeDeviceInstallationResult(&result.dev_report, correlation_id);
   sendEvent<event::AllInstallsComplete>(result);
 
-  if (result.dev_report.result_code != data::ResultCode::Numeric::kNeedCompletion) {
+  if (!(result.dev_report.isSuccess() || result.dev_report.needCompletion())) {
     director_repo.dropTargets(*storage);  // fix for OTA-2587, listen to backend again after end of install
   }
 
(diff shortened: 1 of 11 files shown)


patrickvacek:
I brought over https://github.com/advancedtelematic/aktualizr/pull/1296/commits/f6a5bd529cc290a8a2a9e4cfd17a26c3de320503 from https://github.com/advancedtelematic/aktualizr/pull/1296 and did the same style of Targets metadata dropping to address issue # 2 despite that I really don't like that solution.

lbonn:
>     1. There is no test, because we don't have a way to fake a download failure. It's probably doable with libfiu and the fake package manager, though.

Don't we just need to fake some http error code with HttpFake or some or our python http scripts?

patrickvacek:
> Don't we just need to fake some http error code with HttpFake or some or our python http scripts?

Oooh, nice idea. I was already halfway through setting up libfiu. I'll take that under consideration as well.

patrickvacek:
Fixed with tests using libfiu after all. It wasn't hard and the additional power is nice to have.

simao:
> This should inform the Director that the update failed.
> 
> This appears to fix the bug, but it is not perfect:
> 
>     1. ~There is no test, because we don't have a way to fake a download failure. It's probably doable with libfiu and the fake package manager, though.~ This is now covered by tests.
> 
>     2. There is the same problem with empty targets metadata that occurs with installation failure. It ~could be~ has been handled the same way, ~although~ despite that I don't like the solution of dropping metadata.
> 
>     3. The installation report does not contain any information about how, why, or which download failed. That could be improved if anyone wants to see that.
> 
What do you mean here? director would need at least the correlation id to make this work..
> 
> We might want to consider merging testing and merging this as is anyway just to get the fix out the door, but further consideration is necessary to address the above issues.

patrickvacek:
> What do you mean here? director would need at least the correlation id to make this work..

Correlation ID is included. I just mean the error code/message is fairly generic. I've tested manually and it works, but the UI says:
```
Result code DOWNLOAD_FAILED
Target download failed
```

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -546,12 +546,23 @@ result::Download SotaUptaneClient::downloadImages(const std::vector<Uptane::Targ
...
+            result::Download(downloaded_targets, result::DownloadStatus::kError, "Each target download has failed");
+      } else {
+        LOG_ERROR << "Only " << downloaded_targets.size() << " of " << targets.size()
+                  << " were successfully downloaded.";
+        result = result::Download(downloaded_targets, result::DownloadStatus::kPartialSuccess, "");
+      }
+      // Store installation report to inform Director of the download failure.
+      const std::string &correlation_id = director_repo.getCorrelationId();
+      data::InstallationResult device_installation_result =
+          data::InstallationResult(data::ResultCode::Numeric::kDownloadFailed, "Target download failed");
```
mike-sul: As far as I understand the mechanism of cancelling/removing of the new targets at Director is a sending manifest to it with a negative installation result in case if all targets downloads fails ?

patrickvacek: If I'm understanding you correctly, yes. If we send an installation report with a failure code, the Director will respond by identifying the installation as a failure, and subsequent Targets metadata will be empty (to indicate that the device should take no further action, i.e. not retry until explicitly told to do so). Let me know if that doesn't make sense or you have further questions.

PR #1305, opened 2019-08-20
fix/ota-2183/Enable streaming of garage-push/deploy uploads

Zee314159:
Draft code.

### DIFF
--- a/src/sota_tools/ostree_object.cc
+++ b/src/sota_tools/ostree_object.cc
@@ -2,6 +2,7 @@
 
 #include <assert.h>
 
+#include <sys/stat.h>
 #include <cstring>
 #include <iostream>
 
@@ -21,7 +22,7 @@ OSTreeObject::OSTreeObject(const OSTreeRepo &repo, const std::string &object_nam
       refcount_(0),
       is_on_server_(PresenceOnServer::kObjectStateUnknown),
       curl_handle_(nullptr),
-      form_post_(nullptr) {
+      fd_(nullptr) {
   if (!boost::filesystem::is_regular_file(file_path_)) {
     throw std::runtime_error(file_path_.native() + " is not a valid OSTree repo.");
   }
@@ -206,7 +207,7 @@ void OSTreeObject::MakeTestRequest(const TreehubServer &push_target, CURLM *curl
   request_start_time_ = std::chrono::steady_clock::now();
 }
 
-void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_handle, const RunMode mode) {
+void OSTreeObject::Upload(TreehubServer &push_target, CURLM *curl_multi_handle, const RunMode mode) {
   if (mode == RunMode::kDefault || mode == RunMode::kPushTree) {
     LOG_INFO << "Uploading " << object_name_;
   } else {
@@ -222,25 +223,27 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
   }
   curlEasySetoptWrapper(curl_handle_, CURLOPT_VERBOSE, get_curlopt_verbose());
   current_operation_ = CurrentOp::kOstreeObjectUploading;
+  push_target.SetContentType("Content-Type: application/octet-stream");
   push_target.InjectIntoCurl(Url(), curl_handle_);
   curlEasySetoptWrapper(curl_handle_, CURLOPT_USERAGENT, Utils::getUserAgent());
   curlEasySetoptWrapper(curl_handle_, CURLOPT_WRITEFUNCTION, &OSTreeObject::curl_handle_write);
   curlEasySetoptWrapper(curl_handle_, CURLOPT_WRITEDATA, this);
   http_response_.str("");  // Empty the response buffer
 
-  assert(form_post_ == nullptr);
-  struct curl_httppost *last_item = nullptr;
-  const CURLFORMcode form_err =
-      curl_formadd(&form_post_, &last_item, CURLFORM_COPYNAME, "file", CURLFORM_FILE, file_path_.c_str(), CURLFORM_END);
-  if (form_err != 0u) {
-    // Apparently there is not strerror for formadd.
-    LOG_ERROR << "curl_formadd error: " << form_err;
+  struct stat file_info {};
+  fd_ = fopen(file_path_.c_str(), "rb");
+  if (fd_ == nullptr) {
+    throw std::runtime_error("could not open file to be uploaded");
+  } else {
+    if (stat(file_path_.c_str(), &file_info) < 0) {
+      throw std::runtime_error("Could not get file information");
+    }
   }
+  curlEasySetoptWrapper(curl_handle_, CURLOPT_READDATA, fd_);
+  curlEasySetoptWrapper(curl_handle_, CURLOPT_POSTFIELDSIZE, file_info.st_size);
   curlEasySetoptWrapper(curl_handle_, CURLOPT_POST, 1);
-  curlEasySetoptWrapper(curl_handle_, CURLOPT_HTTPPOST, form_post_);
 
   curlEasySetoptWrapper(curl_handle_, CURLOPT_PRIVATE, this);  // Used by ostree_object_from_curl
-
   const CURLMcode err = curl_multi_add_handle(curl_multi_handle, curl_handle_);
   if (err != 0) {
     LOG_ERROR << "curl_multi_add_handle error:" << curl_multi_strerror(err);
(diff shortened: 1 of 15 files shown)


Zee314159:
I manually tested, using ‘t_deploy’,the output looks it didn’t work. Content_type is multipart and x-www-form-urlencoded.Maybe I was using the wrong test, or the code I added didn’t work somehow.

patrickvacek:
> Content_type is multipart and x-www-form-urlencoded.

It's possible that there is some other way you have to specify it, perhaps forcing the header value doesn't actually stick.

However, I suspect a bigger problem is that you probably don't want to set that from the `RequestPool`. You only want it for uploads, so `OSTreeObject::Upload` in `src/sota_tools/ostree_object.cc` is probably a safer place to do it.

mike-sul:
> You only want it for uploads, so `OSTreeObject::Upload` in `src/sota_tools/ostree_object.cc`

It will require updating treehub_server.py accordingly.
Also, I am wondering what does  `Out of band upload` described in https://github.com/advancedtelematic/treehub/pull/70 exactly mean and whether there is any benefit in that upload mode for us ? My understanding is that it will help to bypass one of the OTA connect internal services (treehub ?) and stream directly to S3.

patrickvacek:
> It will require updating treehub_server.py accordingly.

Ah, I hadn't considered that. You might be right; the tests may need to get updated once we have it working correctly with the real backend.

>  Also, I am wondering what does `Out of band upload` described in [advancedtelematic/treehub#70](https://github.com/advancedtelematic/treehub/pull/70) exactly mean and whether there is any benefit in that upload mode for us ? My understanding is that it will help to bypass one of the OTA connect internal services (treehub ?) and stream directly to S3.

To be clear, the "out of band upload" is out of scope for the current task that Liping is working on. But it's still worth understanding, and we may want to support it eventually, although I think it is somewhat more work. See https://saeljira.it.here.com/browse/OTA-1813 for some additional detail. Basically, the backend could tell us to upload directly to S3 and bypass treehub completely. The advantage is reduced load on our server and fewer points of failure.

src/sota_tools/ostree_object.h
```diff
@@ -111,6 +111,7 @@ class OSTreeObject {
   std::stringstream http_response_;
   CURL* curl_handle_;
   struct curl_httppost* form_post_;
+  curl_slist* headers;
```
eu-smirnov: We should initialize this to nullptr. Otherwise, it could be anything and will segfault in destructor. This should fix some of the failing tests in CI, I think.

src/sota_tools/ostree_object.cc
```diff
@@ -241,6 +245,10 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
 
   curlEasySetoptWrapper(curl_handle_, CURLOPT_PRIVATE, this);  // Used by ostree_object_from_curl
 
+  headers = nullptr;
```
eu-smirnov: If we call Upload() multiple times this will be a leak. Or is it not possible?

tests/sota_tools/treehub_server.py
```diff
@@ -79,6 +79,17 @@ def do_POST(self):
             self.send_response_only(200)
             self.end_headers()
             return
+        elif ctype == 'application/octet-stream':
```
mike-sul: In case of 'application/octet-stream' we can read the posted file/body just from the socket buffer, so something like that should work

       ```
 elif ctype == 'application/octet-stream':
            length = int(self.headers['content-length'])
            body = self.rfile.read(length)
            with open("body.dat", "wb") as f:
                f.write(body)

            self.send_response_only(200)
            self.end_headers()
```

But, if the given implementation works then it's ok - this is just a test tool

Comment (Zee314159) [reply tests/sota_tools/treehub_server.py]: Thanks Mike, I'll try your way. I was also just gonna try this:

import cgitb; cgitb.enable()

form = cgi.FieldStorage()
fileitem = form['filename']
if fileitem.filename:
   with open("body.dat", "wb") as f:
   f.write(fileitem.file.read())

Comment (Zee314159) [reply tests/sota_tools/treehub_server.py]: Tested Mike's way, t_deploy output:

diff: /tmp/aktualizr-951f-8561-2d1e-5c0a/d5ea-2beb-dir/objects/: No such file or directory
/home/liping/aktualizr/src/sota_tools/deploy_test.cc:32: Failure

File: src/sota_tools/ostree_object.cc [hunk 4]
```diff
@@ -241,6 +246,10 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
 
   curlEasySetoptWrapper(curl_handle_, CURLOPT_PRIVATE, this);  // Used by ostree_object_from_curl
 
+  assert(headers_ == nullptr);
```
mike-sul: If `OSTreeObject::Upload` is called more than one for the same class instance then it won't work. I suppose, in our case it's called only once so it shouldn't be a problem, although, perhaps makes sense to set the headers in ctor ?

tests/sota_tools/treehub_server.py
mike-sul: Interesting, I've just checked and it works for me, I missed an ending `return` statement at the end of the code snippet

Zee314159: Yes, I added the return statement. maybe we can see how it work out In CI test?

mike-sul: Oh, I think, I understand where is a problem. It's not enough just set a header that indicates a type of transferring data, actual data should be transferred according to the declared way. In this particular case we set 'application/octet-stream' but actually transfer data as `'multipart/form-data'`  see OSTreeObject::Upload() code below setting the headers.

Zee314159:
All test passed except 'OstreeObject_test.UploadFail". II'll look into it tomorrow.

Zee314159:
> I am not sure if I see where this fd is closed in case if the file is successfully opened.

I closed the file after the reading is done, but somehow it will cause a failing of the test. Any suggestion where to close the file?

Zee314159:
See this PR: https://github.com/advancedtelematic/aktualizr/pull/1244 , looks there left an unsolved issue.

src/sota_tools/ostree_object.cc
```diff
@@ -228,19 +236,22 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
...
-  assert(form_post_ == nullptr);
-  struct curl_httppost *last_item = nullptr;
-  const CURLFORMcode form_err =
-      curl_formadd(&form_post_, &last_item, CURLFORM_COPYNAME, "file", CURLFORM_FILE, file_path_.c_str(), CURLFORM_END);
-  if (form_err != 0u) {
-    // Apparently there is not strerror for formadd.
-    LOG_ERROR << "curl_formadd error: " << form_err;
+  struct stat file_info {};
+  fd_ = fopen(file_path_.c_str(), "rb");
+  if (fd_ == nullptr) {
```
mike-sul: As far as I understand, if a file opening or getting its stat fails the code control flow will end up at line #333 ?
I am just wondering if throwing an exception (like at line #229) will work in this case, at first glance it won't work (just thinking loudly :))

src/sota_tools/ostree_object.cc
```diff
@@ -228,19 +236,22 @@ void OSTreeObject::Upload(const TreehubServer &push_target, CURLM *curl_multi_ha
...
+    if (stat(file_path_.c_str(), &file_info) < 0) {
+      LOG_ERROR << "Could not get file information";
+    }
   }
+  curlEasySetoptWrapper(curl_handle_, CURLOPT_READDATA, fd_);
+  curlEasySetoptWrapper(curl_handle_, CURLOPT_POSTFIELDSIZE, file_info.st_size);
   curlEasySetoptWrapper(curl_handle_, CURLOPT_POST, 1);
-  curlEasySetoptWrapper(curl_handle_, CURLOPT_HTTPPOST, form_post_);
 
-  curlEasySetoptWrapper(curl_handle_, CURLOPT_PRIVATE, this);  // Used by ostree_object_from_curl
```
patrickvacek: Is it correct to remove this line? It may be, but it seems like it is still used.

Zee314159: Didn't remove the line. See line 254.

mike-sul:
@patrickvacek As far as I understand some of the tests triggered by the CI pipelines do use garag-push/deploy against the prod/SIT treehub ?

src/sota_tools/ostree_object.cc
patrickvacek: Thanks, don't know how I missed that.

patrickvacek:
> As far as I understand some of the tests triggered by the CI pipelines do use garag-push/deploy against the prod/SIT treehub ?

Yes, that is correct. I don't think we *actually* push anything, though, I think it's just testing authentication. I might be misremembering.

mike-sul:
> > As far as I understand some of the tests triggered by the CI pipelines do use garag-push/deploy against the prod/SIT treehub ?
> 
> Yes, that is correct. I don't think we _actually_ push anything, though, I think it's just testing authentication. I might be misremembering.

I suppose, it would be great to test this change against the real treehub, at least manually.

patrickvacek:
> I suppose, it would be great to test this change against the real treehub, at least manually.

Absolutely, this is a requirement. Manual confirmation is enough if we have tests that effectively simulate the real backend.

src/sota_tools/ostree_object.cc
Zee314159: Throwing an exception might be better.

src/sota_tools/request_pool.h
```diff
@@ -42,7 +42,7 @@ class RequestPool {
   RateController rate_controller_;
   int running_requests_;
   int total_requests_made_{0};
-  const TreehubServer& server_;
+  TreehubServer& server_;
```
patrickvacek: It's too bad we lose this const-ness. It doesn't seem so critical, though, and I don't have a better solution available yet, though, so maybe it's fine.

Zee314159: Yes, I feel sorry for the loss too. I have tried.

PR #1308, opened 2019-08-21
Restore the doxygen index/front page.

patrickvacek:
Unintentionally (?) removed in a9a1b49f516ba4d2c9f8bd51427ab80889d666fa, but still quite useful.

To see why it is useful, compare these:
https://advancedtelematic.github.io/aktualizr/2019.5/index.html
https://advancedtelematic.github.io/aktualizr/2019.6/index.html


merltron:
> Unintentionally (?) removed in [a9a1b49]

Well, it was kind of intentional.  I was confused as to why we have "README.adoc" and "INDEX.md" both on the top level. But now I understand why was there.

Can we put it in a separate "doxygen" subfolder and not fix up techdoc and reference doc?

patrickvacek:
> Can we use a subfolder or have another way to make it clear that this is used to build doxygen content? (since our source is usually asciidoc)

Sure. So would you like to see all of the doxygen-specific bits moved to, say, `docs/doxygen`?

merltron:
> Sure. So would you like to see all of the doxygen-specific bits moved to, say, `docs/doxygen`?

Yes that would be perfect - just so the separation of responsibility is clearer, and to reduce the risk of files getting unintentionally removed again.

patrickvacek:
> Yes that would be perfect - just so the separation of responsibility is clearer, and to reduce the risk of files getting unintentionally removed again.

Done. Also rebased and added a commit about the release process, since I've been meaning to update that for a while. @tkfu you'll have to let me know if you can expand the section about updating homebrew. I don't know how to do that. Feel free to push a commit, send me the instructions, or leave it as it is (i.e. vague).

tkfu:
> @tkfu you'll have to let me know if you can expand the section about updating homebrew. I don't know how to do that. Feel free to push a commit, send me the instructions, or leave it as it is (i.e. vague).

I just finished a big PR myself, so I think I can take a few minutes and push a commit detailing the process.

patrickvacek:
So have I understood correctly that all crosslinks within the antora docs should use `xref`? I can fix that if so; I didn't do that in my changes here.

patrickvacek:
@merltron Are you willing to remove your change request if I've satisfactorily resolved your desire with the subdirectory?

And let me know if I should changes the links to `xref` or if you and/or @tkfu want to unifty that in a subsequent PR as discussed in https://github.com/advancedtelematic/aktualizr/pull/1313.

merltron:
> @merltron Are you willing to remove your change request if I've satisfactorily resolved your desire with the subdirectory?

@patrickvacek I thought that *was* my change request, but anyway yes. xrefs we can leave till later

tkfu:
It may actually be important to change the link to xref, at least for the stuff inside a module (i.e. not the readme). Antora changes the targets of xrefs to adocs to point to the HTML file when the site generator runs; I don't know if it does the same for links.

PR #1312, opened 2019-08-21
Remove deprecated function call

eu-smirnov:
Deprecated since openssl 1.1.0

See https://www.openssl.org/docs/man1.1.0/man3/SSL_load_error_strings.html

### DIFF
--- a/src/aktualizr_get/main.cc
+++ b/src/aktualizr_get/main.cc
@@ -1,7 +1,6 @@
 #include <unistd.h>
 #include <iostream>
 
-#include <openssl/ssl.h>
 #include <boost/filesystem.hpp>
 #include <boost/program_options.hpp>
 
@@ -59,10 +58,6 @@ int main(int argc, char *argv[]) {
   int r = EXIT_FAILURE;
   try {
     Config config(commandline_map);
-    if (config.logger.loglevel <= boost::log::trivial::debug) {
-      SSL_load_error_strings();
-    }
-
     std::string body = aktualizrGet(config, commandline_map["url"].as<std::string>());
     std::cout << body;
 
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -1,7 +1,6 @@
 #include <unistd.h>
 #include <iostream>
 
-#include <openssl/ssl.h>
 #include <boost/filesystem.hpp>
 #include <boost/program_options.hpp>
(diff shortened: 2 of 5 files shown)


patrickvacek:
Sadly, rocko still uses openssl-1.0.2o. I don't think we can get rid of this.

eu-smirnov:
> Sadly, rocko still uses openssl-1.0.2o. I don't think we can get rid of this.

Hm, are you sure about that? [Here](https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-connectivity/openssl?h=rocko) I can see both 1.0.2 and 1.1.0.

Also, from the description [here](https://man.openbsd.org/ERR_load_crypto_strings.3), I'm not sure if the functions are particularly useful even on 1.0.2:

> Even though the error strings are already compiled into the object code of the library as static strings, these functions store them again using dynamically allocated memory on the heap. That may fail if insufficient memory is available, but these functions do not report such errors. Instead, they fail silently, possibly having registered none or only a part of the strings requested.

patrickvacek:
> Hm, are you sure about that? [Here](https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-connectivity/openssl?h=rocko) I can see both 1.0.2 and 1.1.0.

True. I had thought the default was still 1.0.2 but after looking into it again, I'm less confident in that. Either way, it is still an option in every release branch, but not in master.

> Also, from the description [here](https://man.openbsd.org/ERR_load_crypto_strings.3), I'm not sure if the functions are particularly useful even on 1.0.2

Huh. I believe they *are* deprecated in 1.1.0, and that link is probably referring to that. openssl.org seems to indicate they are still useful for [1.0.2](https://www.openssl.org/docs/man1.0.2/man3/SSL_load_error_strings.html).

I think the question is just how concerned we are with supporting 1.0.2 going forward.

patrickvacek:
[This](https://wiki.openssl.org/index.php/Library_Initialization) is also a depressing read: "Below is a list of some initialization calls you might encounter in code or documentation. Unfortunately, all the initialization function return a useless values (for example, always 1) or are void functions. There is no way to determine if a failure occurred."

eu-smirnov:
Depressing indeed, maybe dropping 1.0.2 support isn't a bad idea. However, this PR will only result in less debug messages when working with 1.0.2.

PR #1314, opened 2019-08-21
Fix veracode packing

zabbal:
Properly add binaries to the veracode submission archieve.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -300,6 +300,7 @@ trigger-veracode-scan:
       - $VERACODE_API_ID
   dependencies:
     - veracode-pkg
+  allow_failure: true
   image: openjdk:8
   before_script:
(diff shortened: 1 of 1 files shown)


lbonn:
Apparently, some other issues need to be fixed. Holding that off for now.

zabbal:
Ok, we can go ahead with merging. Corresponding schedule is temporary disabled - will be re-enabled once SPC clarify issue with compiler support with veracode.

patrickvacek:
> Ok, we can go ahead with merging.

We need to address the CI failure before we can merge. Does it just need to be an "allow failure" type of thing?

zabbal:
The CI failure is the reason I've disabled the schedule: right now the 1st submission succeed, 2nd fail due to problem on veracode side. So yes, I think it should be "allow failure" - at least until veracode is fixed and stable enough. I think I saw patch from @lbonn which allows this to fail.

lbonn:
Ok, here is it with `allow_failure`

zabbal:
Seems like I'm not authorized to merge this. @patrickvacek - feel free to if it looks fine.

PR #1318, opened 2019-08-23
Add automatic removal of old installed targets

lbonn:
This adds a mechanism to remove old downloaded targets.

Here, "old" is what was before the current and the previous installed version and the cleaning is done at the end of the installation.

It's only enabled when using aktualizr's executable but can be added through the API as well.

~~This first version is simple. I was planning to add a more complex logic based on the previous log of installed versions (spinned off here: https://github.com/advancedtelematic/aktualizr/tree/feat/install-log) but now I'm not sure it's worth the trouble.~~

~~Main problem with this version is that it will probably make problems with rollbacks.~~

Also, we'd probably want to have something similar for ostree. (already here?)

### DIFF
--- a/config/sql/migration/migrate.20.sql
+++ b/config/sql/migration/migrate.20.sql
@@ -1,0 +1,13 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+ALTER TABLE installed_versions RENAME TO installed_versions_old;
+CREATE TABLE installed_versions(id INTEGER PRIMARY KEY, ecu_serial TEXT NOT NULL, sha256 TEXT NOT NULL, name TEXT NOT NULL, hashes TEXT NOT NULL, length INTEGER NOT NULL DEFAULT 0, correlation_id TEXT NOT NULL DEFAULT '', is_current INTEGER NOT NULL CHECK (is_current IN (0,1)) DEFAULT 0, is_pending INTEGER NOT NULL CHECK (is_pending IN (0,1)) DEFAULT 0, was_installed INTEGER NOT NULL CHECK (was_installed IN (0,1)) DEFAULT 0);
+INSERT INTO installed_versions(ecu_serial, sha256, name, hashes, length, correlation_id, is_current, is_pending, was_installed) SELECT installed_versions_old.ecu_serial, installed_versions_old.sha256, installed_versions_old.name, installed_versions_old.hashes, installed_versions_old.length, installed_versions_old.correlation_id, installed_versions_old.is_current, installed_versions_old.is_pending, 1 FROM installed_versions_old ORDER BY rowid;
+
+DROP TABLE installed_versions_old;
+
+DELETE FROM version;
+INSERT INTO version VALUES(20);
+
+RELEASE MIGRATION;
--- a/config/sql/rollback/rollback.20.sql
+++ b/config/sql/rollback/rollback.20.sql
@@ -1,0 +1,13 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT ROLLBACK_MIGRATION;
+
+CREATE TABLE installed_versions_migrate(ecu_serial TEXT NOT NULL, sha256 TEXT NOT NULL, name TEXT NOT NULL, hashes TEXT NOT NULL, length INTEGER NOT NULL DEFAULT 0, correlation_id TEXT NOT NULL DEFAULT '', is_current INTEGER NOT NULL CHECK (is_current IN (0,1)) DEFAULT 0, is_pending INTEGER NOT NULL CHECK (is_pending IN (0,1)) DEFAULT 0, UNIQUE(ecu_serial, sha256, name));
+INSERT INTO installed_versions_migrate(ecu_serial, sha256, name, hashes, length, correlation_id, is_current, is_pending) SELECT installed_versions.ecu_serial, installed_versions.sha256, installed_versions.name, installed_versions.hashes, installed_versions.length, installed_versions.correlation_id, installed_versions.is_current, installed_versions.is_pending FROM installed_versions;
+
+DROP TABLE installed_versions;
+ALTER TABLE installed_versions_migrate RENAME TO installed_versions;
+
+DELETE FROM version;
+INSERT INTO version VALUES(19);
+
+RELEASE ROLLBACK_MIGRATION;
--- a/src/libaktualizr/primary/aktualizr.h
+++ b/src/libaktualizr/primary/aktualizr.h
@@ -83,7 +83,9 @@ class Aktualizr {
   std::future<result::Download> Download(const std::vector<Uptane::Target>& updates);
 
   /**
-   * Get log of installations
+   * Get log of installations. The log is indexed for every ECU and contains
+   * every change of versions ordered by time. It may contain duplicates in
+   * case of rollbacks.
    * @return installation log
    */
   struct InstallationLogEntry {
--- a/src/libaktualizr/primary/aktualizr.h
+++ b/src/libaktualizr/primary/aktualizr.h
@@ -1,7 +1,7 @@
 #ifndef AKTUALIZR_H_
 #define AKTUALIZR_H_
 
-#include <atomic>
+#include <future>
 #include <memory>
 
 #include <boost/signals2.hpp>
@@ -81,6 +81,17 @@ class Aktualizr {
    * @return std::future object with information about download results.
    */
   std::future<result::Download> Download(const std::vector<Uptane::Target>& updates);
+
+  /**
+   * Get log of installations
+   * @return installation log
+   */
+  struct InstallationLogEntry {
+    Uptane::EcuSerial ecu;
+    std::vector<Uptane::Target> installs;
+  };
+  using InstallationLog = std::vector<InstallationLogEntry>;
+  InstallationLog GetInstallationLog();
 
   /**
    * Get list of targets currently in storage. This is intended to be used with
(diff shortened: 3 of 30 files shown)


tests/metafake.h
```diff
@@ -10,9 +10,6 @@
 #include "utilities/utils.h"
 #include "uptane_repo.h"
 
-enum BackupList {
-};
-
```
patrickvacek: BTW I also did some cleanup of this file in https://github.com/advancedtelematic/aktualizr/pull/1296/commits/63ce676d075635447ee3bfcde5de76d07a6d17d8.

lbonn: Ah nice! I'll remove that duplicate.

lbonn:
> Agreed. I like the idea, but I think we need to keep the current and previous targets and only get rid of anything older than that.

Ok, I'll push the installation log on top then and will think about a nice api to reuse that easily.

> Agreed, but does OSTree do that automatically? I've never seen `ostree admin status` return more than two commits. Is it hiding older commits or silently cleaning them up?

Hm ok I didn't remember about that. Makes sense

tests/metafake.h
patrickvacek: No problem, you can also steal the commit if you want, or I can pull it out into a separate PR, since it's been hanging out in my branch for two weeks or something at this point.

lbonn: Yes I'd vote for moving it to its own PR.

patrickvacek: Done, see https://github.com/advancedtelematic/aktualizr/pull/1322.

lbonn:
Ok I think it's close to the end. Was more complicated than expected...

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -1362,6 +1332,103 @@ TEST(Aktualizr, DownloadListRemove) {
...
+    aktualizr.Download(update_result.updates).get();
+
+    EXPECT_EQ(aktualizr.GetStoredTargets().size(), 3);
+
+    result::Install install_result = aktualizr.Install(update_result.updates).get();
+    EXPECT_TRUE(install_result.dev_report.success);
+
+    // only two targets are left: dummy_firmware has been cleaned up
+    std::vector<Uptane::Target> targets = aktualizr.GetStoredTargets();
+    ASSERT_EQ(targets.size(), 2);
```
patrickvacek: This is a pretty awesome test, but do you think it's worth checking that the two remaining are the two we expect at the very end?

lbonn: Yes, I'll add that.

config/sql/rollback/rollback.20.sql
```diff
@@ -0,0 +1,15 @@
...
+DROP TABLE installed_versions_log;
+
+CREATE TABLE installed_versions_migrate(ecu_serial TEXT NOT NULL, sha256 TEXT NOT NULL, name TEXT NOT NULL, hashes TEXT NOT NULL, length INTEGER NOT NULL DEFAULT 0, correlation_id TEXT NOT NULL DEFAULT '', is_current INTEGER NOT NULL CHECK (is_current IN (0,1)) DEFAULT 0, is_pending INTEGER NOT NULL CHECK (is_pending IN (0,1)) DEFAULT 0, UNIQUE(ecu_serial, sha256, name));
+INSERT INTO installed_versions_migrate(ecu_serial, sha256, name, hashes, length, correlation_id, is_current, is_pending) SELECT installed_versions.ecu_serial, installed_versions.sha256, installed_versions.name, installed_versions.hashes, installed_versions.length, installed_versions.correlation_id, installed_versions.is_current, installed_versions.is_pending FROM installed_versions;
+
+DROP TABLE installed_versions;
+ALTER TABLE installed_versions_migrate RENAME TO installed_versions;
+
+DELETE FROM version;
+INSERT INTO version VALUES(19);
```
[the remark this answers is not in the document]
lbonn: Looks like not, according to https://sqlite.org/lang_createtrigger.html

> Triggers are automatically dropped when the table that they are associated with (the table-name table) is dropped. However if the trigger actions reference other tables, the trigger is not dropped or modified if those other tables are dropped or modified.

Maybe I'll double check anyway.

doanac:
My only nit is that I had to really read commit https://github.com/advancedtelematic/aktualizr/pull/1318/commits/eb510dfc184b956076650fac85d24ab94203d1ff to get an understanding of what was going on here. "Feat/ota 3073" doesn't give readers much context. I *think* the intent of this is something along the lines of "The installed_versions table grows each time an update is applied. This change prunes out old rows from the table that are no longer needed"

lbonn:
> My only nit is that I had to really read commit [eb510df](https://github.com/advancedtelematic/aktualizr/commit/eb510dfc184b956076650fac85d24ab94203d1ff) to get an understanding of what was going on here. "Feat/ota 3073" doesn't give readers much context. I _think_ the intent of this is something along the lines of "The installed_versions table grows each time an update is applied. This change prunes out old rows from the table that are no longer needed"

Fair points, I opened the PR long before it was finally ready. I've updated the opening description.

There should be some doc bits missing as well, actually.

lbonn:
After further tests, the installation log actually doesn't work properly... Will push some fixes soon.

lbonn:
There was a bug in the previous version. It should now be fixed and also simpler and better tested.

src/libaktualizr/primary/aktualizr.h
```diff
@@ -82,6 +82,19 @@ class Aktualizr {
    */
   std::future<result::Download> Download(const std::vector<Uptane::Target>& updates);
 
+  /**
+   * Get log of installations. The log is indexed for every ECU and contains
+   * every change of versions ordered by time. It may contain duplicates in
+   * case of rollbacks.
```
patrickvacek: Is this still true? Won't anything rolled back erase the prior generation number for that target?

config/sql/migration/migrate.20.sql
```diff
@@ -0,0 +1,14 @@
...
+CREATE TABLE installed_versions(id INTEGER PRIMARY KEY, ecu_serial TEXT NOT NULL, sha256 TEXT NOT NULL, name TEXT NOT NULL, hashes TEXT NOT NULL, length INTEGER NOT NULL DEFAULT 0, correlation_id TEXT NOT NULL DEFAULT '', generation INTEGER, is_current INTEGER NOT NULL CHECK (is_current IN (0,1)) DEFAULT 0, is_pending INTEGER NOT NULL CHECK (is_pending IN (0,1)) DEFAULT 0, was_installed INTEGER NOT NULL CHECK (was_installed IN (0,1)) DEFAULT 0);
+CREATE TRIGGER trig_installed_versions_gen AFTER INSERT ON installed_versions FOR EACH ROW BEGIN UPDATE installed_versions SET generation = (SELECT IFNULL(MAX(generation), 0) FROM installed_versions WHERE ecu_serial = NEW.ecu_serial)+1 WHERE id = NEW.id; END;
+INSERT INTO installed_versions(ecu_serial, sha256, name, hashes, length, correlation_id, is_current, is_pending, was_installed) SELECT installed_versions_old.ecu_serial, installed_versions_old.sha256, installed_versions_old.name, installed_versions_old.hashes, installed_versions_old.length, installed_versions_old.correlation_id, installed_versions_old.is_current, installed_versions_old.is_pending, 1 FROM installed_versions_old ORDER BY rowid;
```
patrickvacek: What is `was_installed` actually used for? I'm not quite following why that is needed.

src/libaktualizr/storage/sqlstorage_test.cc
```diff
@@ -346,6 +363,72 @@ TEST(sqlstorage, DbMigration18to19) {
     FAIL() << "Too many rows";
   }
 }
+
+TEST(sqlstorage, DbMigration19to20) {
```
patrickvacek: What exactly is this test covering? Is it just making sure we don't erase things that were installed with version 19 after we migrate to version 20 since they don't have generation numbers?

src/libaktualizr/primary/aktualizr.h
lbonn: What you describe is close to the old behavior.
It will now create a new generation and have another entry in the table.

config/sql/migration/migrate.20.sql
lbonn: installed_versions has is_current and is_pending columns to distinguish between what's currently running and staged from the rest.

was_installed is used to distinguish historical versions that were never current from the rest. ie: you set something to pending and end up installing some other version. The actual log would only contain versions that have actually been installed.

lbonn: Additional note: we assume that all versions have been installed when migrating (hence, the 1). It seems to be better than reporting none.

src/libaktualizr/storage/sqlstorage_test.cc
lbonn: It's testing the migration with realistic data indeed. It gives a bit more confidence than just checking the table shapes, like schema_migration_test.sh does.

config/sql/migration/migrate.20.sql
patrickvacek: Could we just determine if it was installed if the generation is > 0 (or >= 0)? Or does the generation value get set even if the target is pending, such that if it isn't actually installed, you'd still want to know that it was pending at one point?

lbonn: That's a good idea but in this code, generation was also incremented for pending version.

In the end I realized it was still complex so I removed the generation field entirely and relies on id which should have sufficient guarantees for this purpose.
was_installed is still there in the new version.

patrickvacek: I didn't even realize that `id` would probably work fine. Makes sense if you're copying entries for rollbacks/reinstalls. I think that means this is good to go, then.

PR #1319, opened 2019-08-26
libaktualizr API: capture parameters by-copy

eu-smirnov:
If API is used in asynchronous mode we cannot be sure
that parameters are still valid when we execute the lambda
function in different thread.


eu-smirnov:
@patrickvacek Probably yes, but nothing really comes to my mind atm.

PR #1323, opened 2019-08-28
Minor linter cleanups.

patrickvacek:


src/aktualizr_primary/secondary.cc
```diff
@@ -65,7 +65,7 @@ void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config
 
 class SecondaryWaiter {
  public:
-  SecondaryWaiter(uint16_t wait_port, size_t wait_timeout, Secondaries& secondaries)
+  SecondaryWaiter(uint16_t wait_port, ssize_t wait_timeout, Secondaries& secondaries)
```
kbushgit: The function definition is not clean from the user perspective, "wait_timeout" the value is interpreted as minutes, seconds, milliseconds? 

"wait_timeout_s" looks more clean.

The size_t or ssize_t is not good candidate to define time interval, since it have different purpose, 
I think we can consider "long" as a type for this purpose.

What is the reason to use signed size_t?

patrickvacek: @mike-sul can probably answer most of your concerns here. I was merely fixing a problem that clang-tidy was complaining about on my machine, specifically that we were converting from `size_t` (unsigned) to `boost::posix_time::seconds`, which is signed. Hence, the least intrusive change was to just use `ssize_t`. We generally avoid `long` since its actual size is dependent on the implementation.

eu-smirnov: I'm also a bit surprised to see `(s)size_t` here. Why can't we use int32_t and json.asInt() in the parser?

mike-sul: IMHO, a type for a duration/timeout should be unsigned since negative duration doesn't make sense. But `boost::posix_time::seconds` accepts `long` as an input/ctor params. 
In order, to avoid casting from unsigned type to signed (long), I suggest, to use actually `long` as Kostya suggested and parse it as .asInt64() with following verification whether it's positive.

Yes, I agree, that `wait_timeout_s` instead of `wait_timeout` will improve code readability.

eu-smirnov: I don't have a strong opinion about unsigned types, but tend more to agree with the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Integer_Types):
> You should not use the unsigned integer types such as uint32_t, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, do not use unsigned types to say a number will never be negative. Instead, use assertions for this.

And while we probably don't care about data models other than LP64, parsing `asInt64()` into `long` still doesn't look correct to me, as in [other data models](https://en.cppreference.com/w/cpp/language/types) it's 32-bit long.

mike-sul: so, I suppose, asInt() would be the way to go for us ?
The Google C++ Guide says not to use exceptions which is debatable and makes or doesn't make sense depending on a specific use case, my point is that it shouldn't be Bible for us :)

eu-smirnov: Yes, I'd vote for `asInt()` and `int`, which is guaranteed to fit into `boost::posix_time::seconds`.
I absolutely agree that we shouldn't just blindly follow their guide, but we still can use it as a baseline.
Regarding the exceptions, I don't think they advise against using them, but merely explaining why they've decided to go without. They even say "Things would probably be different if we had to do it all over again from scratch".

xcheng-here: How about name it as "timeout_s" or "timeout_seconds" ? remove "wait_" from it's name to make it short

patrickvacek: Check out latest version with `int`, `asInt()`, and `timeout_s`.

patrickvacek: (Worth noting: I did not change the actual config parameter, just the names of the internal variables.)

PR #1325, opened 2019-08-30
Bump default Yocto branch in docs to thud.

patrickvacek:
Specify it explicitly instead of relying on the default, since going forward we don't want to surprise users by changing their release branch. That should only be done with careful consideration.

@merltron / @tkfu does it make sense to have a note or a page describing the risks of changing branches? (Things like kernel/bootloader compatibility and the size of the update from one branch to another?)


merltron:
> @merltron / @tkfu does it make sense to have a note or a page describing the risks of changing branches? (Things like kernel/bootloader compatibility and the size of the update from one branch to another?)

@patrickvacek  I don't know...what do you think? I don't know what the risks are....if you think they're worth mentioning maybe you could provide more detail or add a draft note to the docs for us to review?

patrickvacek:
@merltron Will do.

PR #180, opened 2017-05-05
Updated to use standard-readme format

codders:
Nicer README.md, and a new badge to boot. We love badges.

PR #1326, opened 2019-09-02
OTA-2743: Initial set of negative tests aiming Director and Repo

mike-sul:
Signed-off-by: Mykhaylo Sul <ext-mykhaylo.sul@here.com>

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -188,6 +188,10 @@ add_test(NAME test_ip_secondary
          COMMAND ${PROJECT_SOURCE_DIR}/tests/ipsecondary_test.py
          --build-dir ${PROJECT_BINARY_DIR} --src-dir ${PROJECT_SOURCE_DIR})
 
+add_test(NAME test_backend_failure
+         COMMAND ${PROJECT_SOURCE_DIR}/tests/test_backend_failure.py
+         --build-dir ${PROJECT_BINARY_DIR} --src-dir ${PROJECT_SOURCE_DIR})
+
 add_executable(aktualizr-cycle-simple aktualizr_cycle_simple.cc)
 target_link_libraries(aktualizr-cycle-simple aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
 aktualizr_source_file_checks(aktualizr_cycle_simple.cc)
--- a/tests/test_backend_failure.py
+++ b/tests/test_backend_failure.py
@@ -1,0 +1,235 @@
+#!/usr/bin/env python3
+
+import logging
+import argparse
+
+from os import getcwd, chdir
+
+from test_fixtures import with_aktualizr, with_uptane_backend, KeyStore, with_secondary, with_path,\
+    DownloadInterruptionHandler, MalformedJsonHandler, with_director, with_imagerepo, InstallManager,\
+    with_install_manager, with_images, MalformedImageHandler, with_customrepo, SlowRetrievalHandler, \
+    RedirectHandler
+
+
+logger = logging.getLogger(__file__)
+
+"""
+Verifies whether aktualizr is updatable after director metadata download failure
+with follow-up successful metadata download.
+
+Currently, it's tested against two types of metadata download/parsing failure:
+    - download interruption - metadata file download is interrupted once|three times, after that it's successful
+    - malformed json - aktualizr receives malformed json/metadata as a response to the first request for metadata,
+      a response to subsequent request is successful
+
+Note: Aktualizr doesn't send any installation report in manifest in case of metadata download failure
+https://saeljira.it.here.com/browse/OTA-3730
+"""
+@with_uptane_backend(start_generic_server=True, port=8888)
+@with_path(paths=['/1.root.json', '/root.json', '/targets.json'])
+@with_director(handlers=[
+                            DownloadInterruptionHandler(number_of_failures=1),
+                            MalformedJsonHandler(number_of_failures=1),
+                            DownloadInterruptionHandler(number_of_failures=3),
+                        ])
+@with_aktualizr(start=False, run_mode='full')
+@with_install_manager()
+def test_backend_failure_sanity_director_update_after_metadata_download_failure(install_mngr, director,
+                                                                                aktualizr, **kwargs):
+    with aktualizr:
+        director.wait_for_install()
+        # we have to stop director before terminating aktualizr since the later doesn't support graceful shutdown
+        # (doesn't handle any signal (SIGTERM, SIGKILL, etc) what leads to receiving broken requests at director
+        # https://saeljira.it.here.com/browse/OTA-3744
+        director.stop()
+    return director.is_install_successful() and install_mngr.are_images_installed()
+
+
+"""
+Verifies whether aktualizr is updatable after image metadata download failure
+with follow-up successful metadata download.
+
+Currently, it's tested against two types of metadata download/parsing failure:
+    - download interruption - metadata file download is interrupted once|three times, after that it's successful
+    - malformed json - aktualizr receives malformed json/metadata as a response to the first request for metadata,
+      a response to subsequent request is successful
+
+Note: Aktualizr doesn't send any installation report in manifest in case of metadata download failure
+"""
+@with_uptane_backend(start_generic_server=True, port=8888)
+@with_path(paths=['/1.root.json', '/root.json', '/timestamp.json', '/snapshot.json', '/targets.json'])
+@with_imagerepo(handlers=[
+                            DownloadInterruptionHandler(number_of_failures=1),
+                            MalformedJsonHandler(number_of_failures=1),
+                            DownloadInterruptionHandler(number_of_failures=3),
+                        ])
+@with_director()
+@with_aktualizr(run_mode='full')
+@with_install_manager()
+def test_backend_failure_sanity_imagerepo_update_after_metadata_download_failure(install_mngr, director,
+                                                                                 aktualizr, **kwargs):
+    with aktualizr:
+        director.wait_for_install()
+        director.stop()
+    return director.is_install_successful() and install_mngr.are_images_installed()
+
+
+"""
+Verifies whether aktualizr is updatable after image download failure
+with follow-up successful download.
+
+Currently, it's tested against two types of image download failure:
+    - download interruption - file download is interrupted once, after that it's successful
+    - malformed image - image download is successful but it's malformed. It happens once after that it's successful
+"""
+@with_uptane_backend(start_generic_server=True, port=8888)
+@with_images(images_to_install=[(('primary-hw-ID-001', 'primary-ecu-id'), 'primary-image.img')])
+@with_imagerepo(handlers=[
+                            # TODO: test fails because aktualizr issues byte range request
+                            # that are not supported by server
+                            # https://saeljira.it.here.com/browse/OTA-3716
+                            #DownloadInterruptionHandler(number_of_failures=1, url='/targets/primary-image.img'),
+                            MalformedImageHandler(number_of_failures=1, url='/targets/primary-image.img'),
+                        ])
+@with_director()
+@with_aktualizr(run_mode='full', id=('primary-hw-ID-001', 'primary-ecu-id'))
+@with_install_manager()
+def test_backend_failure_sanity_imagerepo_update_after_image_download_failure(install_mngr, director,
+                                                                              aktualizr, **kwargs):
+    with aktualizr:
+        director.wait_for_install()
+        director.stop()
+    return director.is_install_successful() and install_mngr.are_images_installed()
+
+
+"""
+    Verifies whether aktualizr is updatable after malformed image is downloaded
+    from a custom image server with follow-up successful download.
+"""
+@with_uptane_backend(start_generic_server=True, port=8888)
+@with_images(images_to_install=[(('primary-hw-ID-001', 'primary-ecu-id'), 'primary-image.img',
+                                 'http://localhost:8891/primary-image.img')])
+@with_customrepo(handlers=[
+                            # TODO: This test fails because the issue with image download
+                            #  from a server that doesn't support byte range requests
+                            # DownloadInterruptionHandler(number_of_failures=1, url='/primary-image.img'),
+                            # https://saeljira.it.here.com/browse/OTA-3716
+                            MalformedImageHandler(number_of_failures=1, url='/primary-image.img')
+                            # TODO: this test fails too, although httpclient.cc sets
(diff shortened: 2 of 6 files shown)


tests/sota_tools/treehub_server.py
```diff
@@ -80,6 +80,16 @@ def do_POST(self):
             self.end_headers()
             return
 
+        elif ctype == 'application/octet-stream':
+            length = int(self.headers['content-length'])
+            body = self.rfile.read(length)
+            with open("body.dat", "wb") as f:
```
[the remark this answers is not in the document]
mike-sul: Yeah, your're right, this code should be removed at all as it's left over after some experiments I did.
This PR is not ready for review, please, wait for a while until I polish it up

tests/ostree_packman_wrapper.cc
```diff
@@ -0,0 +1,92 @@
+#include "ostree_packman_wrapper.h"
```
patrickvacek: Not a big deal, but it seems like unnecessary deviation to use "packman" here and "pacman" everywhere else.

CMakeLists.txt
```diff
@@ -158,7 +158,7 @@ endif(FAULT_INJECTION)
 
 # flags for different build types
 set(CMAKE_CXX_FLAGS_DEBUG "-Og -g")
-set(CMAKE_C_FLAGS_DEBUG "-Og -g")
+set(CMAKE_C_FLAGS_DEBUG "-Og -g ")
```
patrickvacek: Unnecessary whitespace change.

mike-sul:
> Can you summarize what exactly is being covered here? I'm not following what scenarios are actually being tested. Comments preceding the tests in question would be helpful. We should probably get back to keeping actions.md up to date, and although you don't have to worry about that for now, it's worth putting in a couple brief descriptions that you could imagine being put in a document like that.
> 
> Also, the change to tuf-test-vectors seems like it's rolling back the latest upgrade there.

this PR is not ready for review yet.

mike-sul:
@patrickvacek this PR is ready for review.
The following are tickets derived from this test effort:
https://saeljira.it.here.com/browse/OTA-3716
https://saeljira.it.here.com/browse/OTA-3717
https://saeljira.it.here.com/browse/OTA-3728
https://saeljira.it.here.com/browse/OTA-3729
https://saeljira.it.here.com/browse/OTA-3730

tests/CMakeLists.txt
```diff
@@ -188,6 +188,12 @@ add_test(NAME test_ip_secondary
          COMMAND ${PROJECT_SOURCE_DIR}/tests/ipsecondary_test.py
          --build-dir ${PROJECT_BINARY_DIR} --src-dir ${PROJECT_SOURCE_DIR})
 
+if(BACKEND_FAILURE_TEST)
```
lbonn: Why the new switch? Does the test require new exotic dependencies (that's the usual reason we have these)? If not it can be enabled by default or with another combination of switches.

mike-sul: The goal here is to avoid running this test by default since it takes over 700s to run it and enable this test only on CI or if it's instructed by an user. I am open for any alternatives to what is here.

tests/test_backend_failure.py
```diff
@@ -0,0 +1,219 @@
...
+Currently, it's tested against two types of image download failure:
+    - download interruption - file download is interrupted once, after that it's successful
+    - malformed image - image download is successful but it's malformed. It happens at once after that it's successful
+"""
+@with_uptane_backend(start_generic_server=True, port=8888)
+@with_images(images_to_install=[(('primary-hw-ID-001', 'primary-ecu-id'), 'primary-image.img')])
+@with_director()
+@with_imagerepo(handlers=[
+                            # TODO: test fails because aktualizr issues byte range request
+                            # that are not supported by server
```
patrickvacek: By "not supported by server", you mean the test server being used her, right? I was under the impression the real server handled this, and this is already covered by the `fetcher_test` and `fetcher_death_test` as well, unless those tests don't do what I think they do.

tests/test_backend_failure.py
```diff
@@ -0,0 +1,219 @@
...
+                                 'http://localhost:9001/primary-image.img')])
+@with_director()
+@with_customrepo(handlers=[
+                            # TODO: This test fails because the issue with image download
+                            #  from a server that doesn't support byte range requests
+                            #DownloadInterruptionHandler(number_of_failures=1, url='/primary-image.img'),
+                            MalformedImageHandler(number_of_failures=1, url='/primary-image.img')
+                            # TODO: this test fails too, although httpclient.cc sets
+                            # CURLOPT_LOW_SPEED_TIME and CURLOPT_LOW_SPEED_TIME
+                            #SlowRetrievalHandler(url='/primary-image.img')
```
patrickvacek: Do you know why this test fails? I didn't see a ticket related to this, but it's worth investigating and understanding.

tests/test_fixtures.py
```diff
@@ -0,0 +1,769 @@
...
+            secondary_cfg = json.loads(Aktualizr.SECONDARY_CONFIG_TEMPLATE.format(port=wait_port, timeout=wait_timeout))
+            json.dump(secondary_cfg, secondary_config_file)
+            self._secondary_config_file = secondary_config_file.name
+
+        with open(path.join(self._storage_dir.name, 'config.toml'), 'w+') as config_file:
+            config_file.write(Aktualizr.CONFIG_TEMPLATE.format(server_url=uptane_server.base_url,
+                                                               ca_path=ca, pkey_path=pkey, cert_path=cert,
+                                                               serial=id[1], hw_ID=id[0],
+                                                               storage_dir=self._storage_dir,
+                                                               db_path=path.join(self._storage_dir.name, 'db.sql'),
```
patrickvacek: Normally our database is named "sql.db", but that can be overwritten with `storage.sqldb_path`. Not a big deal, but the deviation seems like an opportunity for confusion.

tests/test_fixtures.py
```diff
@@ -0,0 +1,769 @@
...
+            self._failure_counter += 1
+        else:
+            request_handler.default_get()
+
+    def map(self, url):
+        return {'GET': {url: MalformedJsonHandler(self._number_of_failures)}}
+
+
+# The following are candidates for fixtures/common/shared components that can be reused by
+# multiple tests/test suits
```
patrickvacek: Aren't they already used by multiple tests now? :)

src/aktualizr_primary/main.cc
```diff
@@ -162,6 +162,7 @@ int main(int argc, char *argv[]) {
     r = EXIT_SUCCESS;
   } catch (const std::exception &ex) {
     LOG_ERROR << ex.what();
+    r = EXIT_FAILURE;
```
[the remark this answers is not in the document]
mike-sul: No, I put it there when I was playing with different approaches to ostree update mocking. Agree, that it can be removed.

tests/test_backend_failure.py
mike-sul: > By "not supported by server", you mean the test server being used her, right? I was under the impression the real server handled this, and this is already covered by the `fetcher_test` and `fetcher_death_test` as well, unless those tests don't do what I think they do.

Yes, I mean the test server being used  in the given test suite. That's correct, the current real/OTA Connect image repo does support a byte range request, in case of the `custom URL` usage we don't know in advance if the `custom image server` supports it, so IMHO, aktualizr needs to be able to fallback to regular HTTP GET requests during download retries as long as we need to support `custom URL`.

mike-sul: > this is already covered by the fetcher_test and fetcher_death_test as well, unless those tests don't do what I think they do.
`fetcher_test` doesn't test/simulate the given case, it uses the mocked implementation of http client which ignores a range/offset value.

`fetcher_death_test` tests similar use case (resume after pause/die) but it tests it against a server that does support byte range requests (fake_test_server.py::elif self.path.endswith('/large_file'): )

tests/test_backend_failure.py
mike-sul: No, I don't. I didn't create ticket for this issue because I am not sure if this is an issue or not. I remember testing this case manually and IIRC Aktualizr did detect `slow retrieval attack`. I think, it might to do with specifics of using sockets on `localhost`. If it's so then it's not possible to test the given case in scope of the given test suite at all. Perhaps, it makes sense to create a ticket in order to double check it manually and make recommendations to the QA and/or E2E QA team to come with a test verifying Aktualizr against `slow retrieval attack`.

tests/test_fixtures.py
mike-sul: yeah, I guess, this is just mistyping.

tests/test_fixtures.py
mike-sul: Yes, this comment was introduced back then when I produced initial tests for IP secondary, I'll remove it.

mike-sul:
> Probably worth putting references to those ticket numbers next to the TODOs in the test code. I think once upon a time that was a requirement, but I don't see that specified in the contrib doc anymore. Still a good idea.

OK

tests/test_backend_failure.py
patrickvacek: > aktualizr needs to be able to fallback to regular HTTP GET requests during download retries

I understand and agree. Actually, it sounds like we need to be able to fall back to regular GET requests regardless of the custom URL usage. We should also try to see if we can get byte range requests working in treehub so that we can take advantage of the functionality we already have, too.

tests/test_backend_failure.py
patrickvacek: > Perhaps, it makes sense to create a ticket in order to double check it manually and make recommendations to the QA and/or E2E QA team to come with a test verifying Aktualizr against `slow retrieval attack`.

Yeah, it's worth a ticket to follow up on this. It would be great if we could unit test it ourselves, but if it isn't possible, then a higher level is an option as well. There's a tech debt task for writing a test to catch oversized metadata, and this seems somewhat similar, albeit more challenging.

mike-sul: https://saeljira.it.here.com/browse/OTA-3737

tests/CMakeLists.txt
lbonn: I don't think it's really necessary to disable it by default if it's just 700s and it's run by default on CI.

A softer method would just be to add a "long" label to the test so that it can be skipped when running `ctest -LE long`:

    set_tests_properties(test_xxx PROPERTIES LABELS "long")

mike-sul:
@patrickvacek @lbonn I managed to make the tests to pass as well as decreased their running time from ~700 to less than 200s so I removed that cmake config variable.

patrickvacek:
>  I managed to make the tests to pass as well as decreased their running time from ~700 to less than 200s so I removed that cmake config variable.

Nice! How did you pull that off? Sounds like a good lesson to remember, whatever it was!

mike-sul:
> > I managed to make the tests to pass as well as decreased their running time from ~700 to less than 200s so I removed that cmake config variable.
> 
> Nice! How did you pull that off? Sounds like a good lesson to remember, whatever it was!

It turned out that `polling_sec = "0"` is considered by aktualizr as an empty value so it applies the default 20 seconds in such case, replacing it with `polling_sec = 0` helped to set the inter-polling wait/delay to zero.

patrickvacek:
> It turned out that `polling_sec = "0"` is considered by aktualizr as an empty value so it applies the default 20 seconds in such case, replacing it with `polling_sec = 0` helped to set the inter-polling wait/delay to zero.

Ah yes, the joys of yaml and our custom parser in particular...

tests/test_backend_failure.py
```diff
@@ -0,0 +1,243 @@
...
+                            #MalformedJsonHandler(number_of_failures=3)
+                        ])
+@with_aktualizr(start=False, run_mode='full')
+@with_install_manager()
+def test_backend_failure_sanity_director_update_after_metadata_download_failure(install_mngr, director,
+                                                                                aktualizr, **kwargs):
+    with aktualizr:
+        director.wait_for_install()
+        # we have to stop director before terminating aktualizr since the later doesn't support graceful shutdown
+        # (doesn't handle any signal (SIGTERM, SIGKILL, etc) what leads to receiving broken requests at director
```
patrickvacek: Who/what exactly is trying to send SIGTERM or SIGKILL to aktualizr?

tests/test_backend_failure.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+"""
+Verifies whether aktualizr is updatable after image metadata download failure
+with follow-up successful metadata download.
+
+Currently, it's tested against two types of metadata download/parsing failure:
+    - download interruption - metadata file download is interrupted once|four times, after that it's successful
+    - malformed json - metadata file download is successful but the received metadata are malformed.
+      It happens at once| four time, after that it's successful
```
patrickvacek: Looks like three times now; any significance to 3 vs 4?

tests/test_backend_failure.py
mike-sul: test_fixtures.py::#180 `self._process.terminate()`, it sends `SIGTERM` to the aktualizr process.

tests/test_backend_failure.py
mike-sul: yeah, just removed.

tests/test_backend_failure.py
patrickvacek: Ah yes. It's true, we don't handle it gracefully. I think this came up once before for some reason, but that might've been a couple years ago. It might be worth trying to do better here, maybe we need another ticket in the robustness epic?

tests/test_backend_failure.py
patrickvacek: I'm confused, I still see the discrepancy here and elsewhere.

mike-sul: It should be fine now, the comment was outdated.
No, there is no significance if it's 3,4 or even 5 :). There two use cases here:
1) download succeeds withing a single uptane cycle thanks to download retries, a number of retires is set to 3 at the moment (hardcoded in aktualizr)

2) download fails withing a single uptane cycle even after 3 attempts, aktualizr exits current uptane cycle, starts a new one and try to download the given metadata file again which succeeds.

So, those two DownloadInterruptionHandler(number_of_failures=1), DownloadInterruptionHandler(number_of_failures=3) are aimed to test the aforementioned two scenarios.

In case if malformed json/metadata are received aktualizr doesn't do any retry it just simply exits a current uptane cycle and starts a new one, so just one scenario in this case.

tests/test_backend_failure.py
mike-sul: https://saeljira.it.here.com/browse/OTA-3744

PR #1327, opened 2019-09-02
Remove some NO_VALGRIND test properties

lbonn:
Unclear why they should be needed in most cases


lbonn:
Was just rebsased, merging.

PR #1329, opened 2019-09-03
OTA-3682: using tabbed approach for config variants (.conf vs .toml)

merltron:
- Putting different ways of configuring into
tabs ie("local.conf" vs "sota_local.toml") and adding more content reuse.
- Minor fix to deploy checklist (added xref)
- Minor fix to env setup command
- restored missing ECU events from nav

Example of tabbed approach:
image

Signed-off-by: Merlin Carter <merlin.carter@here.com>

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/_partials/config-descriptions.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/_partials/config-descriptions.adoc
@@ -1,0 +1,62 @@
+//  tag::buildconfig-hint[]
+If you are building a disk image that contains aktualizr, your configuration will be the **local.conf** file.
+
+* You'll find this file in the `conf` subdirectory of the of your build project.
+
+For more information, see the xref:build-configuration.adoc[build configuration reference] and the xref:build-images.adoc[sample build procedures]
+
+//  end::buildconfig-hint[]
+
+//  tag::clientconfig-hint[]
+If you are running standalone aktualizr, your configuration file will be a *.toml* file.
+
+You configure aktualizr by creating and updating a `*.toml` file in one of the following directories:
+
+* `/usr/lib/sota/conf.d`
+* `/etc/sota/conf.d/`
+
+For more information and links to sample configuration files, see the xref:aktualizr-config-options.adoc[client configuration reference].
+
+//  end::clientconfig-hint[]
+
+//  tag::pollconfig-dev[]
+Interval between polls (in seconds).
+
+The default polling internal designed to make it convenient for you test and develop OTA update functions.
+
+//  end::pollconfig-dev[]
+
+//  tag::autorebootconfig-dev[]
+Forces installation completion. Causes a system reboot in case of an ostree package manager. Emulates a reboot in case of a fake package manager.
+
+You'll want to enable this option when developing because it's more convenient.
+
+//  end::autorebootconfig-dev[]
+
+//  tag::pollconfig-prod[]
+When moving to production you'll want to have a much longer interval. 
+In fact, for production, we don't support intervals less the 1 hour (3,600 seconds). Longer internals help you to reduce the internet bandwidth and power consumption for your devices.
+
+We recommend an internal between 1 and 7 days (86,400 to 604,800 seconds)
+
+//  end::pollconfig-prod[]
+
+//  tag::autorebootconfig-prod[]
+If you followed our recommendation to enable automatic rebooting for development, you should disable again for production.
+
+//  end::autorebootconfig-prod[]
+
+
+//  tag::metadata-expires[]
+Use this option to have the metadata expire after a fixed date and time.
+
+Specify the time as a UTC instant like in the following example:
+
+//  end::metadata-expires[]
+
+//  tag::metadata-expireafter[]
+Use this option to have the metadata expire after an elapsed period of time.
+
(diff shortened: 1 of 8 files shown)


docs/ota-client-guide/modules/ROOT/pages/recommended-clientconfig.adoc
```diff
@@ -1,79 +1,89 @@
...
-* If you are running standalone aktualizr, your configuration file will be a **.toml** file.
-+
-You configure aktualizr by creating and updating a `*.toml` file in one of the following directories:
+The configuration file and exact parameters depend on how you use OTA Connect:
 
-** `/usr/lib/sota/conf.d`
-** `/etc/sota/conf.d/`
+* If you're building disk images that contain aktualizr, you need to update your *build configuration*.
+* If you're using a standalone version of aktualizr that you've installed locally, you need to update the *client configuration*.
+
```
xaviergully: what if I am building disk images that contain libaktualizr ?

merltron: I don't know. This PR is more about changing the layout of the existing content. @patrickvacek do the build config parameters work the same way if you add a libaktualizr integration to your build (like this `IMAGE_INSTALL_append = " libaktualizr-demo"`).
For example would the polling interval config work in the same way as building with normal aktualizr?

docs/ota-client-guide/modules/ROOT/pages/_partials/config-descriptions.adoc
```diff
@@ -0,0 +1,62 @@
...
+Specify the time as a UTC instant like in the following example:
+
+//  end::metadata-expires[]
+
+//  tag::metadata-expireafter[]
+Use this option to have the metadata expire after an elapsed period of time.
+
+Specify the number of years, months and days like in the following example:
+
+//  end::metadata-expireafter[]
```
patrickvacek: This page looks pretty weird in github. Not a big deal; maybe just worth putting in the backlog of things to figure out later.

docs/ota-client-guide/modules/ROOT/pages/build-raspberry.adoc
```diff
@@ -70,7 +71,7 @@ Now you can run the following script to get the environment set up:
 
 [subs=+attributes]
 ----
-source meta-updater/scripts/envsetup.sh {machine}
+source meta-updater/scripts/envsetup.sh {meta-env}
```
patrickvacek: This one should still use `machine`. The other change above (referring to `meta-updater-raspberrypi`) is correct, though.

docs/ota-client-guide/modules/ROOT/pages/metadata-expiry.adoc
```diff
@@ -7,57 +7,73 @@ The default expiry dates are as follows:
...
+| 
+include::partial$config-descriptions.adoc[tags=metadata-expireafter]
 ----
-GARAGE_TARGET_EXPIRE_AFTER = "1Y3M5D"
+garage-sign targets sign ----expire-after 1Y3M5D  --repo myimagerepo --key-name mytargets
 ----
 |====================
+
+--
+====
```
patrickvacek: This page also looks pretty weird in github FWIW.

docs/ota-client-guide/modules/ROOT/pages/recommended-clientconfig.adoc
patrickvacek: > do the build config parameters work the same way if you add a libaktualizr integration to your build

That's a good question, I haven't tried it myself. AFAIK, yes, it should. The config fragments in meta-updater do not depend on the aktualizr recipe. However, you can use libaktualizr-demo-app without using meta-updater, so that's up to the user to decide. @eu-smirnov can you confirm my understanding?

docs/ota-client-guide/modules/ROOT/pages/_partials/config-descriptions.adoc
merltron: Anything in the "partials" dir is a partial  - you're not supposed to view it in Github.

docs/ota-client-guide/modules/ROOT/pages/metadata-expiry.adoc
merltron: Yea, that's cos it uses include directives for content reuse - as discussed previously, if we want to have nice interactivity and content reuse in the portal, we're going to have to make sacrifices for GitHub readability. 
We can of course add a "if envgithub" conditional statements but this is going to make the doc extremely complex to maintain and I'd rather ditch the requirement for "if/then" statements.

patrickvacek: No, now that I get what's going on, it's fine. But I'm definitely in favor of doing the "hey maybe you should view this file on the docs portal" message for the github env if that's not a total pain to do.

docs/ota-client-guide/modules/ROOT/pages/build-raspberry.adoc
merltron: fixed

merltron:
Fixed outstanding issues. More complex changes relating to configuring libaktualizr should be in another PR.

PR #1330, opened 2019-09-03
Add doc about the considerations of changing release branches.

patrickvacek:
Please review critically. I'm not sure I've done a great job explaining this.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/yocto-release-branches.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/yocto-release-branches.adoc
@@ -1,0 +1,34 @@
+= Yocto release branches
+
+== Supported branches
+
+Yocto has a number of release branches. Their details are documented in the https://wiki.yoctoproject.org/wiki/Releases[Yocto wiki]. HERE OTA Connect currently actively supports the following branches:
+
+* warrior
+* thud
+
+Several older branches were actively supported in the past. These branches should still be stable, but they have not been updated for some time and are no longer regularly maintained:
+
+* sumo
+* rocko
+* pyro
+* morty
+
+The master branch is the primary branch for development, and if you wish to add new functionality, it probably belongs there. However, we do not recommend using master for anything except development, as most layers are not guaranteed to be stable on master.
+
+== Changing branches
+
+Changing the release branch of an existing Yocto build environment or the software running on a device in production is something that should only be done with careful consideration. If you change a release branch in your build environment, you will mostly like need to re-bitbake every package for your image, which can take several hours. If you then try to update a device from a version built with an older branch to a version built with a newer branch, the size of the update will mostly like be quite large, which runs the risk of the device running out of disk space to download the update. There may also be unanticipated changes to other packages or the bootloader that could cause unexpected problems.
+
+That said, in general such changes are possible. However, due to a change in the bootloader configuration (including initramfs in FIT images), upgrading from sumo (or earlier branches) to thud (or later branches) is not supported.
+
+== Using the default manifest
+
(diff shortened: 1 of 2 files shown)


docs/ota-client-guide/modules/ROOT/pages/yocto-release-branches.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+== Changing branches
+
+Changing the release branch of an existing Yocto build environment or the software running on a device in production is something that should only be done with careful consideration. If you change a release branch in your build environment, you will mostly like need to re-bitbake every package for your image, which can take several hours. If you then try to update a device from a version built with an older branch to a version built with a newer branch, the size of the update will mostly like be quite large, which runs the risk of the device running out of disk space to download the update. There may also be unanticipated changes to other packages or the bootloader that could cause unexpected problems.
+
+That said, in general such changes are possible. However, due to a change in the bootloader configuration, upgrading from rocko (or earlier branches) to sumo (or later branches) is not supported.
```
lbonn: I think the breaking point is between sumo / thud when we started including initramfs in fit images.

patrickvacek: Ah, sorry, I misremembered and I seem to keep forgetting about how this came about. I didn't think there was much in sumo that wasn't in thud, though, so it is a bit surprising. Or did we intentionally keep that out of rocko because we didn't want to break things?

patrickvacek: @lbonn You're talking about https://github.com/advancedtelematic/meta-updater/pull/368, right? You are right, then: it was never backported, so it isn't in rocko or sumo. It is in thud and warrior since they were released after that PR was merged.

patrickvacek:
Also note that nothing in the nav points to this article. @merltron any suggestions where something like this belongs?

merltron:
> Also note that nothing in the nav points to this article. @merltron any suggestions where something like this belongs?

I've not sure if I would even give this it's own page - what procedure would I be following when I would think "oh I want to use a different branch"? - the normal "demo" build procedure (raspberry, qemu). Can you give me a bit of context about when I would most need this info?

patrickvacek:
> I've not sure if I would even give this it's own page - what procedure would I be following when I would think "oh I want to use a different branch"? - the normal "demo" build procedure (raspberry, qemu). Can you give me a bit of context about when I would most need this info?

Yes, that was my concern as well. The supported branches bit is useful for developers but should probably just get merged with https://github.com/advancedtelematic/meta-updater/blob/master/CONTRIBUTING.adoc at some point. The rest is mostly troubleshooting and/or advanced usage for developers.

It indeed might not need to have it's own page. Or maybe this fits in with some of the other topics that have come up with specific boards or board versions working only with certain release branches. (RPi 3 B+ requires thud and newer; RPi 4 will probably require warrior once it's ready.)

merltron:
OK, yes then I would recommend putting as a note in either "supported-boards.adoc" or "add-ota-functonality-existing-yocto-project.adoc"

docs/ota-client-guide/modules/ROOT/pages/yocto-release-branches.adoc
lbonn: yes indeed. Thanks for digging it out!

patrickvacek:
Actually, now I see that nowhere in the docs portal do we talk about supported branches. That is only covered by https://github.com/advancedtelematic/meta-updater/blob/master/CONTRIBUTING.adoc. I suggest we take that section, remove it from the contrib doc and combine it with this doc. This doc should then live alongside the supported boards doc (currently in meta-updater), probably just as a standalone reference doc for whoever cares about such details.

merltron:
Alright, I'll help to add that tomorrow, about to knock off soon.

patrickvacek:
Closing; replaced by https://github.com/advancedtelematic/meta-updater/pull/595.

patrickvacek:
Reopened, restored changes from the meta-updater PR, and added the doc to the nav. What do y'all think?

PR #1331, opened 2019-09-04
Fix bulleted list formatting.

patrickvacek:
The extra newline apparently matters.

See https://docs.ota.here.com/ota-client/dev/add-ota-functonality-existing-yocto-project.html for how it looks now.


merltron:
Thx - I thought I had fixed that. Guess not.

PR #1335, opened 2019-09-06
Only trigger github release on tags

lbonn:
I forgot why I opted for `CI_COMMIT_REF_NAME` but I don't see why it would be useful outside of tags.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -250,7 +250,7 @@ github-release:
       aktualizr_src-$CI_COMMIT_TAG.tar.gz
   only:
     variables:
-      - $CI_COMMIT_REF_NAME =~ /^\d\d\d\d\.\d+(-\w+)?$/
+      - $CI_COMMIT_TAG =~ /^\d\d\d\d\.\d+(-\w+)?$/
 
 # -- publish coverage results on gitlab pages
 


ci/gitlab/.gitlab-ci.yml
```diff
@@ -250,7 +250,7 @@ github-release:
       aktualizr_src-$CI_COMMIT_TAG.tar.gz
   only:
     variables:
-      - $CI_COMMIT_REF_NAME =~ /^\d\d\d\d\.\d+(-\w+)?$/
+      - $CI_COMMIT_TAG =~ /^\d\d\d\d\.\d+(-\w+)?$/
```
tkfu: How about `/^\d\d\d\d\.\d\d?(\.\d\d?)?$/` instead? We shouldn't be doing releases with dashes and arbitrary words anyway, right?

patrickvacek: We sometimes like to test things, so being able to do `2019.7-test` is nice.

lbonn: Also release candidates that we don't really have bothered with in the past. But maybe we should.

patrickvacek: I'm in favor of doing this smaller change for now, and if we need to be even more restrictive later, it's easy enough to change.

PR #1338, opened 2019-09-09
OTA-3746: Adding github-only note

merltron:
Adding note to all topics that directs people to the rendered version of the article in the docs portal.

We discussed this during our versioning meeting. For more details, see the meeting notes: https://confluence.in.here.com/display/OTA/MEETING+MINUTES%3A+Docs+versioning

Signed-off-by: Merlin Carter <merlin.carter@here.com>

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/account-setup.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/account-setup.adoc
@@ -1,4 +1,12 @@
 = Set up multiple accounts
+ifdef::env-github[]
+
+[NOTE]
+====
+We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}.html[view this article in our documentation portal]. Not all of our articles render correctly in GitHub.
+====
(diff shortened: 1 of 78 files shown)


merltron:
> I'm not sure I like the approach of defining a branch with an attribute. That would imply that we would change it for each `version-docs` branch, which would be one extra make-work task for the docs release process. I don't think we lose much by always linking to `latest`.

OK, noporobs - will make it always point to latest

merltron:
@tkfu OK removed branch attr as requested - have another look.

tkfu:
I'd suggest using a squash merge, since the 2nd commit just undoes all the same stuff.

merltron:
@tkfu OK, I squashed the other commit - I can't merge until I get an approve.

tkfu:
Oops, sorry, thought I had approved with that comment. Squash merge is one of the options in the dropdown menu you get from the arrow next to the merge button on github, but the way you did it is also just fine.

PR #1341, opened 2019-09-10
Fix/update secondary config example (2019.05-docs)

patrickvacek:
Backport of https://github.com/advancedtelematic/aktualizr/pull/1340 to the 2019.5 docs.

Does it matter that the release tag is 2019.5 but the docs branch is 2019.05-docs?


merltron:
> Backport of #1340 to the 2019.5 docs.
> 
> Does it matter that the release tag is 2019.5 but the docs branch is 2019.05-docs?

No...I don't think so, at least not for the docs build - Antora only looks at the branch

patrickvacek:
> No...I don't think so, at least not for the docs build - Antora only looks at the branch

Fair enough, but I was more concerned with consistency and keeping track of things.

PR #1342, opened 2019-09-10
Update changelog and docs for 2019.7 release.

patrickvacek:
Removed garage-deploy-version from cross-deploy-images.adoc, since it wasn't actually being used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,30 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 
 ## [??? (unreleased)]
 
+
+## [2019.7] - 2019-09-10
+
 ### Added
 
 - `GetInstallationLog` API method: [PR](https://github.com/advancedtelematic/aktualizr/pull/1318)
 - The aktualizr daemon will now automatically remove old downloaded targets to free up disk space: [PR](https://github.com/advancedtelematic/aktualizr/pull/1318)
+- CA path is now always supplied to curl and can be overwritten: [PR](https://github.com/advancedtelematic/aktualizr/pull/1294)
+
+### Changed
+
+- garage-push and garage-deploy can now stream OSTree objects to [S3 via Treehub](https://github.com/advancedtelematic/treehub/pull/70) (instead of getting copied): [PR](https://github.com/advancedtelematic/aktualizr/pull/1305)
+
+### Removed
+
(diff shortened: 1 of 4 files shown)


merltron:
We also need to create a 2019.6 docs branch since 2019.7 now becomes "latest".

patrickvacek:
> We also need to create a 2019.6 docs branch since 2019.7 now becomes "latest".

Sure. Although actually shouldn't the 2019.6 docs branch already exist, and we should be making the 2019.7 branch now to match the release I'm about to make?

merltron:
> > We also need to create a 2019.6 docs branch since 2019.7 now becomes "latest".
> 
> Sure. Although actually shouldn't the 2019.6 docs branch already exist, and we should be making the 2019.7 branch now to match the release I'm about to make?

I'm sorry, afraid I couldn't parse that sentence. I have no idea why a 2019.6-docs branch "should already exist"....I didn't already exist.....but I bought it into existence just now. Here it is: https://github.com/advancedtelematic/aktualizr/tree/2019.6-docs/docs/ota-client-guide/modules/ROOT/pages

merltron:
> we should be making the 2019.7 branch now to match the release I'm about to make?

No, master is always latest from a docs perspective. Unless I'm completely misunderstanding you?

patrickvacek:
> I'm sorry, afraid I couldn't parse that sentence. I have no idea why a 2019.6-docs branch "should already exist"....It didn't already exist.....but I bought it into existence just now.

Lol. Yes, conceptually, nothing "should" exist per se. But what I meant was that we should be making the docs branches at the same time as the releases. 2019.6 should already have it's own static docs branch that generally shouldn't change, just like the 2019.6 won't change. Same will go for 2019.7 once it's there. "Latest" should only apply if you're really using the latest head of master.

merltron:
> ...what I meant was that we should be making the docs branches at the same time as the releases. 2019.6 should already have it's own static docs branch that generally shouldn't change, just like the 2019.6 won't change. Same will go for 2019.7 once it's there. "Latest" should only apply if you're really using the latest head of master.

Ah...now I got it  (yes, I went into a temporary philosophical black hole, contemplating things that should exist but don't...such as "flying octopus that secretes vanilla ice cream from its tentacles" ) 

Anyway....We continuously, update the docs even though the  branch is released (typos, content improvements, ect...) So when creating docs PRs, we would need to request approval to merge into "master" AND "2019.7-docs" - wouldn't that be making things more complicated?

merltron:
> As `-docs` branches will only have docs changes in them, they will be pretty trivial to merge into master as needed, in advance of the next release.
@tkfu @patrickvacek  Ok fine with me, will create 2019.7-docs as well

PR #1343, opened 2019-09-10
Fix inaccurate homebrew release instructions

tkfu:
Signed-off-by: Jon Oster <jon.oster@here.com>


tkfu:
I have brought shame upon my family by writing inaccurate docs based only on my inaccurate memory of how a thing works. This fixes that.

patrickvacek:
Merging despite Travis, which seems to be having trouble again.

PR #1347, opened 2019-09-11
Fix/garage deploy check push server

patrickvacek:
This was a minor logic problem (with detecting when we can't sign the targets) caught by OTF, but while looking closer, I realized there was a bigger problem: we were running the garage-check logic on the *fetch* repo instead of the *push* repo. I fixed that and addressed the online signing logic flow as well.


src/sota_tools/garage_deploy.cc
```diff
@@ -112,20 +112,24 @@ int main(int argc, char **argv) {
...
+      if (!push_credentials.CanSignOffline()) {
+        LOG_FATAL << "Online signing with garage-deploy is not supported.";
+        return EXIT_FAILURE;
+      }
+      if (!OfflineSignRepo(ServerCredentials(push_credentials.GetPathOnDisk()), name, commit, hardwareids)) {
+        return EXIT_FAILURE;
+      }
+
+      TreehubServer push_server;
+      if (authenticate(cacerts, push_credentials, push_server) != EXIT_SUCCESS) {
```
Zee314159: Shouldn't  push_server authenticate be done before 'UploadToTreehub'?

patrickvacek: Yeah, I've accidentally made it authenticate twice now. I'll try to refactor this.

src/sota_tools/CMakeLists.txt
```diff
@@ -317,7 +317,7 @@ if (BUILD_SOTA_TOOLS)
     add_test(NAME garage-deploy-online-signing
     COMMAND ${PROJECT_SOURCE_DIR}/tests/sota_tools/test-garage-deploy-online-signing $<TARGET_FILE:garage-deploy>
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
-    set_tests_properties(garage-deploy-online-signing PROPERTIES PASS_REGULAR_EXPRESSION "Online signing with garage-deploy is currently unsupported")
+    set_tests_properties(garage-deploy-online-signing PROPERTIES PASS_REGULAR_EXPRESSION "Online signing with garage-deploy is not supported")
```
Zee314159: So this means we are not planning to support online signing anytime soon or at all?

patrickvacek: "Online signing" here has kind of a funny meaning. We just shouldn't use that phrase here, because it's confusing with how we use the term elsewhere. I'll change it.

To explain: normally, your credentials file has a key for the Targets role of the Image repository. This is true even if you haven't rotated that key (e.g. what we normally call "offline signing"). If that key isn't there, it's a problem.

src/sota_tools/garage_deploy.cc
patrickvacek: Fixed.

PR #1357, opened 2019-09-13
Update release process doc.

patrickvacek:
1. Explain how to fix a broken release if absolutely necessary.
2. Improve/simplify the doxygen part.
3. OTF is now automatically triggered, so manual testing is no longer necessary.
4. Add section about updating meta-updater.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/release-process.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/release-process.adoc
@@ -42,7 +42,11 @@ git push github <tag>
 
 Gitlab will build this tag and automatically create a release for it on github.
 
-=== Create a new docs branch
+=== Updating an already-released tag
+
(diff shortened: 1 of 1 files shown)


merltron:
I know this is not strictly part of the PR, but I would recommend putting the toc in conditional tags (`ifdef::env-github[]`).

On the portal, a TOC is already automatically generated, so we don't need two TOCs.

patrickvacek:
> I know this is not strictly part of the PR, but I would recommend putting the toc in conditional tags

I can do that, but it seems that Antora automatically ignores it. What looks wrong to you on https://docs.ota.here.com/ota-client/latest/release-process.html?

merltron:
> > I know this is not strictly part of the PR, but I would recommend putting the toc in conditional tags
> 
> I can do that, but it seems that Antora automatically ignores it. What looks wrong to you on https://docs.ota.here.com/ota-client/latest/release-process.html?

well, this looks wrong to me:

image

patrickvacek:
> well, this looks wrong to me:

Fixed, hopefully.

PR #1358, opened 2019-09-13
Refactor/constructors

eu-smirnov:
While looking into the exception handling in ctors and dtors, I did some refactorings in hope that they will simplify the task.
Main ideas are to bring object initialization and ownership closer to the place they are being used, specifically `Bootloader` to `PackageManager` and `ReportQueue` to `SotaUptaneClient`, as well as to make object creation more transparent and to reduce code duplication.

### DIFF
--- a/src/libaktualizr/bootloader/bootloader.cc
+++ b/src/libaktualizr/bootloader/bootloader.cc
@@ -10,7 +10,7 @@
 #include "utilities/exceptions.h"
 #include "utilities/utils.h"
 
-Bootloader::Bootloader(const BootloaderConfig& config, INvStorage& storage) : config_(config), storage_(storage) {
+Bootloader::Bootloader(BootloaderConfig config, INvStorage& storage) : config_(std::move(config)), storage_(storage) {
   reboot_sentinel_ = config_.reboot_sentinel_dir / config_.reboot_sentinel_name;
   reboot_command_ = config_.reboot_command;
 
--- a/src/libaktualizr/package_manager/ostreemanager.cc
+++ b/src/libaktualizr/package_manager/ostreemanager.cc
@@ -222,12 +222,18 @@ data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &ta
   return data::InstallationResult(data::ResultCode::Numeric::kOk, "Successfully booted on new version");
 }
 
-OstreeManager::OstreeManager(PackageConfig pconfig, std::shared_ptr<INvStorage> storage,
-                             std::shared_ptr<Bootloader> bootloader, std::shared_ptr<HttpInterface> http)
-    : PackageManagerInterface(std::move(pconfig), std::move(storage), std::move(bootloader), std::move(http)) {
+OstreeManager::OstreeManager(PackageConfig pconfig, BootloaderConfig bconfig, std::shared_ptr<INvStorage> storage,
+                             std::shared_ptr<HttpInterface> http)
+    : PackageManagerInterface(std::move(pconfig), std::move(bconfig), std::move(storage), std::move(http)) {
   GObjectUniquePtr<OstreeSysroot> sysroot_smart = OstreeManager::LoadSysroot(config.sysroot);
   if (sysroot_smart == nullptr) {
     throw std::runtime_error("Could not find OSTree sysroot at: " + config.sysroot.string());
+  }
+
+  // consider boot successful as soon as we started, missing internet connection or connection to secondaries are not
+  // proper reasons to roll back
+  if (imageUpdated()) {
+    bootloader_->setBootOK();
   }
 }
 
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -20,7 +20,7 @@ Aktualizr::Aktualizr(Config &config) : config_(config) {
   sig_ = make_shared<boost::signals2::signal<void(shared_ptr<event::BaseEvent>)>>();
   storage_ = INvStorage::newStorage(config_.storage);
   storage_->importData(config_.import);
-  uptane_client_ = std::make_shared<SotaUptaneClient>(config_, storage_, nullptr, nullptr, nullptr, sig_);
+  uptane_client_ = std::make_shared<SotaUptaneClient>(config_, storage_, nullptr, nullptr, sig_);
 }
 
 Aktualizr::Aktualizr(Config &config, std::shared_ptr<INvStorage> storage_in, std::shared_ptr<HttpInterface> http_in)
@@ -31,10 +31,9 @@ Aktualizr::Aktualizr(Config &config, std::shared_ptr<INvStorage> storage_in, std
 
   sig_ = make_shared<event::Channel>();
   storage_ = std::move(storage_in);
-  std::shared_ptr<Bootloader> bootloader_in = std::make_shared<Bootloader>(config_.bootloader, *storage_);
   std::shared_ptr<ReportQueue> report_queue_in = std::make_shared<ReportQueue>(config_, http_in);
 
-  uptane_client_ = std::make_shared<SotaUptaneClient>(config_, storage_, http_in, bootloader_in, report_queue_in, sig_);
+  uptane_client_ = std::make_shared<SotaUptaneClient>(config_, storage_, http_in, report_queue_in, sig_);
 }
 
 void Aktualizr::Initialize() {
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -13,7 +13,10 @@ using std::make_shared;
 using std::shared_ptr;
 
 Aktualizr::Aktualizr(Config &config) : config_(config) {
-  systemSetup();
+  if (sodium_init() == -1) {  // Note that sodium_init doesn't require a matching 'sodium_deinit'
+    throw std::runtime_error("Unable to initialize libsodium");
+  }
+
   sig_ = make_shared<boost::signals2::signal<void(shared_ptr<event::BaseEvent>)>>();
   storage_ = INvStorage::newStorage(config_.storage);
   storage_->importData(config_.import);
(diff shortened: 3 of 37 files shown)


eu-smirnov:
Hi, @doanac! I cannot add you to reviewers, but it would be great if you could also have a look since it touches parts of the code you mostly worked with.

src/libaktualizr/bootloader/bootloader.cc
```diff
@@ -10,7 +10,7 @@
 #include "utilities/exceptions.h"
 #include "utilities/utils.h"
 
-Bootloader::Bootloader(const BootloaderConfig& config, INvStorage& storage) : config_(config), storage_(storage) {
+Bootloader::Bootloader(BootloaderConfig config, INvStorage& storage) : config_(std::move(config)), storage_(storage) {
```
doanac: kind of a c++ learning question for my benefit: Seems like passing as a const by reference would be more ideal. What benefit does this gain?

kbushgit: I guess it comes from the clang-tidy best practice, the reason is when you pass value through const references the move semantic won't have an effect and copy constructor will be called anyway.
This can have an effect when you pass rvalue into constructor, in this case this approach avoids calling the copy constructor.

eu-smirnov: Yes, exactly, it comes from [this clang-tidy warning](https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html).
Which it turn was caused by the change in src/libaktualizr/bootloader/bootloader.h:28
Previously, we stored a reference to the config object, which isn't the right thing to do, IMO.  It makes the behavior of Bootloader class dependent on the state of an external object and that is usually not what we want.

doanac: thanks that was exactly what i was hoping to learn about.

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -12,38 +12,19 @@
...
     throw std::runtime_error("Unable to initialize libsodium");
   }
 
-  LOG_TRACE << "Seeding random number generator from /dev/urandom...";
-  Timer timer;
-  unsigned int seed;
-  std::ifstream urandom("/dev/urandom", std::ios::in | std::ios::binary);
-  urandom.read(reinterpret_cast<char *>(&seed), sizeof(seed));
-  urandom.close();
-  std::srand(seed);  // seeds pseudo random generator with random number
```
lbonn: Are you sure it's not used by one of our dependencies? Could be worth checking with some strace.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -26,21 +26,18 @@ static void report_progress_cb(event::Channel *channel, const Uptane::Target &ta
...
-                                   std::shared_ptr<HttpInterface> http_in, std::shared_ptr<ReportQueue> report_queue_in,
+                                   std::shared_ptr<HttpInterface> http_in,
                                    std::shared_ptr<event::Channel> events_channel_in)
     : config(config_in),
       uptane_manifest(config, storage_in),
       storage(storage_in),
       http(std::move(http_in)),
-      report_queue(std::move(report_queue_in)),
+      uptane_fetcher(new Uptane::Fetcher(config, http)),
+      report_queue(new ReportQueue(config, http)),
```
patrickvacek: This looks promising. However, this means it is no longer possible to *not* have a report queue. That's probably fine, but in the report queue itself, we have this sort of annoying thing where we only know not to use it if we get 404s when sending reports, and we don't store that information. We dump a trace-level log message every time we get a 404. Maybe there's a better way to do this. At any rate, the default case is clearer here, and should work just fine, so that is good.

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -139,14 +157,14 @@ class SotaUptaneClient {
   Uptane::ImagesRepository images_repo;
   Uptane::Manifest uptane_manifest;
   std::shared_ptr<INvStorage> storage;
-  std::shared_ptr<PackageManagerInterface> package_manager_;
   std::shared_ptr<HttpInterface> http;
+  std::shared_ptr<PackageManagerInterface> package_manager;
```
patrickvacek: Why remove the final underscore? It's a small point, but I thought we wanted to actually move to all member variables having the final underscore, which IIRC is recommended by the Google style guide.

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -139,14 +157,14 @@ class SotaUptaneClient {
...
   std::shared_ptr<Uptane::Fetcher> uptane_fetcher;
   std::unique_ptr<ReportQueue> report_queue;
   Json::Value last_network_info_reported;
   Uptane::EcuMap hw_ids;
   std::shared_ptr<event::Channel> events_channel;
-  boost::signals2::connection conn;
+  boost::signals2::scoped_connection conn;
```
patrickvacek: I think there was a reason I didn't originally use a `scoped_connection`... but I can't figure out why not. Seems like a good choice.

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -30,9 +31,26 @@
...
+  SotaUptaneClient(Config &config_in, const std::shared_ptr<INvStorage> &storage_in,
+                   std::shared_ptr<event::Channel> events_channel_in)
+      : SotaUptaneClient(config_in, storage_in, std::make_shared<HttpClient>(), std::move(events_channel_in)) {}
+
+  SotaUptaneClient(Config &config_in, const std::shared_ptr<INvStorage> &storage_in,
+                   std::shared_ptr<HttpInterface> http_in)
+      : SotaUptaneClient(config_in, storage_in, std::move(http_in), nullptr) {}
+
+  SotaUptaneClient(Config &config_in, const std::shared_ptr<INvStorage> &storage_in)
+      : SotaUptaneClient(config_in, storage_in, std::make_shared<HttpClient>(), nullptr) {}
```
patrickvacek: What makes having the multiple constructors better than using default arguments?

patrickvacek:
Oops, because I reviewed the commits one by one, several of the comments are already marked as "Outdated". One was, so I removed it. The others are still valid. Let me know if you can't see them and I can repost them in some way.

lbonn:
Should we refresh this PR?

eu-smirnov:
@lbonn yes, I'll do that tomorrow

src/libaktualizr/primary/aktualizr.cc
eu-smirnov: I think we care only about libsodium and libopenssl, and they are not using that.

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -222,13 +222,19 @@ data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &ta
...
+OstreeManager::OstreeManager(PackageConfig pconfig, BootloaderConfig bconfig, std::shared_ptr<INvStorage> storage,
+                             std::shared_ptr<HttpInterface> http)
+    : PackageManagerInterface(std::move(pconfig), std::move(bconfig), std::move(storage), std::move(http)) {
   GObjectUniquePtr<OstreeSysroot> sysroot_smart = OstreeManager::LoadSysroot(config.sysroot);
   if (sysroot_smart == nullptr) {
     throw std::runtime_error("Could not find OSTree sysroot at: " + config.sysroot.string());
   }
+
+  // consider boot successful as soon as we started, missing internet connection or connection to secondaries are not
+  // proper reasons to roll back
```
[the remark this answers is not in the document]
eu-smirnov: Yeah, we probably should revisit this part.

src/libaktualizr/primary/sotauptaneclient.h
eu-smirnov: Default arguments should be checked at runtime vs compile time for constructors, that makes default constructors a safer option IMO.

src/libaktualizr/primary/sotauptaneclient.h
eu-smirnov: Wanted to make it more consistent with other members, but you are right, we probably should move to the opposite direction. Reverted.

src/libaktualizr/primary/sotauptaneclient.cc
eu-smirnov: We were always creating one, so I see no reason to keep it optional. For the issue you describe, we probably can have a connectivity flag somewhere in sotauptaneclient.

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -12,25 +12,17 @@
...
-
-  sig_ = make_shared<boost::signals2::signal<void(shared_ptr<event::BaseEvent>)>>();
-  storage_ = INvStorage::newStorage(config_.storage);
-  storage_->importData(config_.import);
-  uptane_client_ = std::make_shared<SotaUptaneClient>(config_, storage_, sig_);
-}
+Aktualizr::Aktualizr(Config &config)
+    : Aktualizr(config, INvStorage::newStorage(config.storage), std::make_shared<HttpClient>()) {}
 
 Aktualizr::Aktualizr(Config &config, std::shared_ptr<INvStorage> storage_in, std::shared_ptr<HttpInterface> http_in)
```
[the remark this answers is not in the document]
eu-smirnov: Makes sense. Added to the last commit.

src/libaktualizr/primary/aktualizr.cc
patrickvacek: It looks like we originally added randomness initialization for the selection of pretty names in https://github.com/advancedtelematic/aktualizr/commit/6e32f371fa34c254e159c3a01c125b847a9cad30. Not exactly critical, but we do still use `std::random_device urandom;` in `src/libaktualizr/utilities/utils.cc` for that and for `randomUuid()`.

eu-smirnov: Ah, I see. Initially, it was added together with `std::rand()` for device names. After the change to `std:random_device`, `srand()` is not needed anymore.

eu-smirnov:
> One part I'm still not completely sold on yet: why move the `Bootloader` class to `PackageManagerInterface`? Nothing overloads any of the relevant functions, and the only gain seems to be the bit in `ostreemanager.cc` where you can call `bootloader_->setBootOK();` directly. Seems fine, but I want to make sure this change conceptually makes sense.

Since the bootloader implements installation-related logic, I think it conceptually belongs to the package manager(s) and certainly not to the UptaneClient.
The only place we do something meaningful with the bootloader is the ostree package manager class.  Maybe we should go even further and make it a part of ostree package manager, but I'm not sure that we'll never want to use this functionality in other managers.

patrickvacek:
> Since the bootloader implements installation-related logic, I think it conceptually belongs to the package manager(s) and certainly not to the UptaneClient.
> The only place we do something meaningful with the bootloader is the ostree package manager class. Maybe we should go even further and make it a part of ostree package manager, but I'm not sure that we'll never want to use this functionality in other managers.

I think I'm following you. I also think that for now, it's feel to leave it generic and not just ostree-specific.

src/libaktualizr/primary/aktualizr.cc
patrickvacek: True, no reason to keep around the C functions when we have nice C++ ones, and it looks like you're right that nothin was actually using this anymore!

PR #1361, opened 2019-09-16
OTA-2743: Aktualizr test in case of ostree update with failures

mike-sul:
- free aktualizr_fullostree_test from the ostree_mock library and its preload
- make the ostree_mock universal so it can be used by any test, not just C/C++ written, e.g. by python-based integration tests like test_backend_failure.
- test aktualizr if failures happens during ostree objects download from Treehub in case of ostree updates.

Signed-off-by: Mykhaylo Sul <ext-mykhaylo.sul@here.com>

### DIFF
--- a/src/libaktualizr/primary/CMakeLists.txt
+++ b/src/libaktualizr/primary/CMakeLists.txt
@@ -23,10 +23,7 @@ target_link_libraries(t_aktualizr virtual_secondary)
 if (BUILD_OSTREE)
     add_aktualizr_test(NAME aktualizr_fullostree SOURCES aktualizr_fullostree_test.cc PROJECT_WORKING_DIRECTORY ARGS $<TARGET_FILE:uptane-generator> ${PROJECT_BINARY_DIR}/ostree_repo)
     set_target_properties(t_aktualizr_fullostree PROPERTIES LINK_FLAGS -Wl,--export-dynamic)
-    add_dependencies(t_aktualizr_fullostree ostree_mock uptane-generator make_ostree_sysroot)
-    set_tests_properties(test_aktualizr_fullostree PROPERTIES
-        ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:ostree_mock>
-        LABELS "noptest")
+    add_dependencies(t_aktualizr_fullostree uptane-generator make_ostree_sysroot)
     target_link_libraries(t_aktualizr_fullostree virtual_secondary)
     add_aktualizr_test(NAME download_nonostree SOURCES  download_nonostree_test.cc PROJECT_WORKING_DIRECTORY ARGS $<TARGET_FILE:uptane-generator> ${PROJECT_BINARY_DIR}/ostree_repo)
     add_dependencies(t_download_nonostree uptane-generator make_ostree_sysroot)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -56,10 +56,10 @@ add_library(testutilities test_utils.cc)
 list(APPEND TEST_LIBS testutilities)
 
 if(BUILD_OSTREE)
-add_library(ostree_mock SHARED ostree_mock.c)
-aktualizr_source_file_checks(ostree_mock.c)
-target_include_directories(ostree_mock PUBLIC ${LIBOSTREE_INCLUDE_DIRS})
-add_dependencies(build_tests ostree_mock)
+    add_library(ostree_mock SHARED ostree_mock.c)
+    aktualizr_source_file_checks(ostree_mock.c)
+    target_include_directories(ostree_mock PUBLIC ${LIBOSTREE_INCLUDE_DIRS})
+    add_dependencies(build_tests ostree_mock)
 endif(BUILD_OSTREE)
 
 # Setup coverage
@@ -194,7 +194,7 @@ add_test(NAME test_ip_secondary
 
 add_test(NAME test_backend_failure
          COMMAND ${PROJECT_SOURCE_DIR}/tests/test_backend_failure.py
-         --build-dir ${PROJECT_BINARY_DIR} --src-dir ${PROJECT_SOURCE_DIR})
+         --build-dir ${PROJECT_BINARY_DIR} --src-dir ${PROJECT_SOURCE_DIR} --ostree ${BUILD_OSTREE})
 
 add_executable(aktualizr-cycle-simple aktualizr_cycle_simple.cc)
 target_link_libraries(aktualizr-cycle-simple aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -55,8 +55,12 @@ include(CMakeParseArguments)
 add_library(testutilities test_utils.cc)
 list(APPEND TEST_LIBS testutilities)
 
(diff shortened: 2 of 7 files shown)


src/libaktualizr/primary/CMakeLists.txt
```diff
@@ -23,10 +23,7 @@ target_link_libraries(t_aktualizr virtual_secondary)
 if (BUILD_OSTREE)
     add_aktualizr_test(NAME aktualizr_fullostree SOURCES aktualizr_fullostree_test.cc PROJECT_WORKING_DIRECTORY ARGS $<TARGET_FILE:uptane-generator> ${PROJECT_BINARY_DIR}/ostree_repo)
     set_target_properties(t_aktualizr_fullostree PROPERTIES LINK_FLAGS -Wl,--export-dynamic)
-    add_dependencies(t_aktualizr_fullostree ostree_mock uptane-generator make_ostree_sysroot)
-    set_tests_properties(test_aktualizr_fullostree PROPERTIES
-        ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:ostree_mock>
-        LABELS "noptest")
+    add_dependencies(t_aktualizr_fullostree uptane-generator make_ostree_sysroot)
```
patrickvacek: Why was the `noptest` label removed? And am I understanding correctly that are you able to remove the `ostree_mock` dependency because this test has it's own internal mock?

tests/CMakeLists.txt
```diff
@@ -55,8 +55,12 @@ include(CMakeParseArguments)
 add_library(testutilities test_utils.cc)
 list(APPEND TEST_LIBS testutilities)
 
+if(BUILD_OSTREE)
 add_library(ostree_mock SHARED ostree_mock.c)
 aktualizr_source_file_checks(ostree_mock.c)
+target_include_directories(ostree_mock PUBLIC ${LIBOSTREE_INCLUDE_DIRS})
+add_dependencies(build_tests ostree_mock)
+endif(BUILD_OSTREE)
```
patrickvacek: Should be indented correctly.

tests/ostree_mock.c
```diff
@@ -3,26 +3,137 @@
...
+  struct OstreeDeployment booted;
+
+  if (get_ostree_deployment_info(filename, &deployed, &booted) != 0) {
+    return NULL;
+  }
+
+  OstreeDeployment* new_depl =
+      ostree_deployment_new(0, OSTREE_DEPLOYMENT_OS_NAME, deployed.rev, deployed.serial, booted.rev, booted.serial);
+  return new_depl;
+}
```
patrickvacek: So who is using this version? And what exactly is it doing that the old version couldn't do? I'm having trouble following it.

lbonn: From the PR description, it could be useful for plugging the mocks in other tests apparently coming in other PRs. The downside is that there is nothing guaranteeing that it works in the current state.

doanac:
Not sure its worth trying to consolidate. But I have a simple use for it here: https://github.com/advancedtelematic/aktualizr/blob/master/src/aktualizr_lite/ostree_mock.cc

src/libaktualizr/primary/CMakeLists.txt
mike-sul: The label has been removed by mistake, the goal was to remove dependency on ostree_mock.
BTW, what does this label mean ?

Yes, the `aktualizr_fullostree_test.cc` test can be self-sufficient in this regard and doesn't require preloading of any library.

tests/CMakeLists.txt
mike-sul: Yes, will correct it.

tests/ostree_mock.c
mike-sul: > So who is using this version?
This version of the ostree mock is being used by one of the `test_backend_failure` test suite, specifically the test that test an ostree update in case of download failures. The test is currently under development, it will go as a second commit into this PR.

> And what exactly is it doing that the old version couldn't do?
The old version implies that the `mock` function is actually implemented and exposed/exported which is require from an user to implement the `mock` in some C/C++ file. This fact does not allow (or make it not feasible) to reuse it from more complex integration tests written in other languages, e.g test_backend_failure, or in case of standalone running of aktualizr on a local host configured with the ostree package manager.

The given version does not require writing any mock function from an user, just set LD_PRELOAD environment variable for aktualizr process and control/manage the ostree deployed and booted revisions via a file from other processes/scripts which makes it universally applicable in cases when aktualizr is executed in a separate process.

mike-sul:
> Not sure its worth trying to consolidate. But I have a simple use for it here: https://github.com/advancedtelematic/aktualizr/blob/master/src/aktualizr_lite/ostree_mock.cc

Your solution is simpler than the one in this PR, but I am not sure if it will allow to change a revision dynamically from another process, i.e. change a value of the environment variable of the aktualizr process from another process. Also, there is a need in two managing two revisions, deployed and booted, just after an update these versions differs until aktualizr is rebooted/restarted.

src/libaktualizr/primary/CMakeLists.txt
patrickvacek: AFAIK the labels are only used for grouping tests. We use this in CI to specify tests to run or to exclude from each run.

lbonn: ptest are running on bitbaked images and indeed the preloading was too brittle to work there. So maybe the label can indeed be removed in this case.

tests/ostree_mock.c
```diff
@@ -3,26 +3,135 @@
...
+struct OstreeDeployment {
+  int serial;
+  char rev[255];
+};
+
+static int read_deployment_info(FILE* file, struct OstreeDeployment* deployment) {
+  struct OstreeDeployment deployment_res;
+
+  // TODO: make it more robust and reliable as a file might contain a value exceeding 255
+  if (fscanf(file, "%s\n", deployment_res.rev) != 1) {
```
lbonn: `"%254s\n`?

tests/ostree_mock.c
```diff
@@ -3,26 +3,135 @@
...
+                                      struct OstreeDeployment* booted) {
+  FILE* file = fopen(filename, "r");
+  if (file == NULL) {
+    printf("\n>>>>>>>>>> Failed to open the file: %s\n\n", filename);
+    return -1;
+  }
+
+  int return_code = -1;
+
+  while (1) {
```
lbonn: Why is this a loop? It looks like it cannot continue past the first iteration.

mike-sul: To avoid `return` for each negative case.

tests/ostree_mock.c
mike-sul: Yeah, it should be better, although I don't like fscanf :), I think, it's good enough for the test component.

tests/ostree_mock.c
lbonn: I see, then some `goto fail;` would be even more readable I think. This style calls for an explanatory comment, as it is quite uncommon. If you're really dogmatic against goto, it could be changed to a `do { } while (0);` which at least self-documents that this is not a loop. But it would probably still require a bit more time to grasp.

mike-sul: I am fine with `do { } while (0);`. I've seen this approach on multiple projects (at least two) back in 20XX, perhaps I am getting too old  :)

PR #1367, opened 2019-09-16
Fix asciidoc syntax issues to clean up warnings (2019.5)

tkfu:
This PR is to clean up some asciidoc syntax issues that were making antora barf up a bunch of warnings (and in some cases errors). That looks ugly on CI, and it also means that it's easy to miss a warning or error that we actually care about.

Mainly there were 3 issues:

### 1. Wrong numbers in ordered lists.
This comes from something that's a good practice in Markdown, but a bad habit in asciidoc. In asciidoc, you should either make an ordered list like this:
```
. some item
. something else
. item 3
```
or this:
```
1. some item
2. something else
3. item 3
```
The first option will automatically number the list for you, and you don't have to worry about it. The second option will also auto-number, but if the number you put is different that what got assigned with auto-numbering, it will throw a warning. This is actually good, because it catches errors like this, which would lead to both items being numbered 1, and asciidoctor throwing up a warning that you numbered item two wrong.
```
1. some item

There's an interesting detail about this item

2. something else
```
### 2. Unescaped attribute substitution

Asciidoc has a syntax for defining variables and then using them in the document, except that they're called attributes. You use an attribute by putting the attribute name in curly braces: `{some-attribute}`. If you put something in curly braces that isn't an attribute, you get a warning. You can fix this by escaping the first brace: `\{some-attribute}`

### 3. Heading levels in the wrong sequence

Markdown is lax about heading levels: you can start the page with a `###` section if you want, or put a `####` right after a `##` with no `###` in between. Asciidoc is pickier; your heading levels have to be strictly in order.


patrickvacek:
I merged the branch manually, so this PR is no longer necessary. I've now force-pushed 2019.5-docs and hopefully you agree that it looks correct.

Issue #1292, opened 2019-08-11
How to store custom data for targets?


doanac:
I'm currently carrying a patch out-of-tree I use to keep track of a Target's custom data so that I know what docker-apps to deploy:

 https://github.com/doanac/aktualizr/commit/40c51db83b8060a54a34e8e43ff771a7e61e4201

In its current form there are some things wrong with the patch:
1) I can/should use the ostree-mock approach to avoid my `fakeGetCurrent` hack
2) I'm writing a file directly to disk and this project has put in a decent amount of effort avoiding that by saving things in sqlite.

I'm curious is there's a way I could re-work this patch that might be acceptable to you guys? One idea I had was to add a column named something like `custom-data` to the `installed_versions` table. That would actually solve this problem universally and not just for docker-apps. However, the custom data would increase the size of each entry and I'm not sure any other code in the near/planned future would need this? I *could* ifdef something like that in and just activate for docker-apps, but that seems a little sloppy as well.

I'm open to anything here as it would be nice to get this code maintained/tested upstream.

patrickvacek:
We're still thinking about how best to address this, but one solution might be to make a new table specifically for holding that custom metadata, possibly per ECU.

One point is still confusing me a bit, though: why is re-reading the Target metadata from the json that we store in the `meta` table not an option?

doanac:
> One point is still confusing me a bit, though: why is re-reading the Target metadata from the json that we store in the meta table not an option?

Director sends the target w/o custom metadata. At least this was the situation when I was originally building this. I think it would be a nice improvement if Director could send that.

patrickvacek:
> Director sends the target w/o custom metadata

Was the Director provided with the custom metadata? Or do you mean that the Director isn't copying the custom metadata provided to the Image repository? If the latter, that is by design, although it is presumably possible to change. (@simao may have more insight.) If the former, that is a bug. However, the only way I know to provide custom metadata to the Director is with [ota-cli](https://github.com/advancedtelematic/ota-cli), which probably isn't an option if you're using our SaaS instance, but might be for you're running your own.

On a related note, we just recently implemented [some logic to be smarter with URLs in the custom metadata](https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr/primary/sotauptaneclient.cc#L750) from the Director and Image repos. I don't really see a good way to generalize that at present, but it at least starts the framework of comparing metadata received from the Director and Image repos and synchronizing them as necessary.

[...]

PR #1370, opened 2019-09-18
Store target's custom meta data when installing

doanac:
This is my stab at fixing #1292.

### DIFF
--- a/config/sql/migration/migrate.21.sql
+++ b/config/sql/migration/migrate.21.sql
@@ -1,0 +1,9 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+ALTER TABLE installed_versions ADD COLUMN custom_meta TEXT NOT NULL DEFAULT "";
+
+DELETE FROM version;
+INSERT INTO version VALUES(21);
+
+RELEASE MIGRATION;
--- a/config/sql/rollback/rollback.21.sql
+++ b/config/sql/rollback/rollback.21.sql
@@ -1,0 +1,13 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT ROLLBACK_MIGRATION;
+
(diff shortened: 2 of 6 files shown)


config/sql/migration/migrate.21.sql
```diff
@@ -0,0 +1,13 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+ALTER TABLE installed_versions RENAME TO installed_versions_old;
+CREATE TABLE installed_versions(id INTEGER PRIMARY KEY, ecu_serial TEXT NOT NULL, sha256 TEXT NOT NULL, name TEXT NOT NULL, hashes TEXT NOT NULL, length INTEGER NOT NULL DEFAULT 0, correlation_id TEXT NOT NULL DEFAULT '', is_current INTEGER NOT NULL CHECK (is_current IN (0,1)) DEFAULT 0, is_pending INTEGER NOT NULL CHECK (is_pending IN (0,1)) DEFAULT 0, was_installed INTEGER NOT NULL CHECK (was_installed IN (0,1)) DEFAULT 0, custom_meta TEXT NOT NULL DEFAULT "");
+INSERT INTO installed_versions(ecu_serial, sha256, name, hashes, length, correlation_id, is_current, is_pending, was_installed) SELECT installed_versions_old.ecu_serial, installed_versions_old.sha256, installed_versions_old.name, installed_versions_old.hashes, installed_versions_old.length, installed_versions_old.correlation_id, installed_versions_old.is_current, installed_versions_old.is_pending, installed_versions_old.was_installed FROM installed_versions_old ORDER BY rowid;
```
lbonn: You can directly order by `id` instead of `rowid`.

Also, it looks like a good use case for `ALTER TABLE ADD COLUMN`: https://sqlite.org/lang_altertable.html.

doanac: I could have sworn this was failing a migration test when I tried it yesterday. But I just changed to `alter column` did a *clean* build and it works. I did a force-push update and hopefully CI will be happy.

lbonn:
Sure, I've actually already taken a look and it seems fine on a correctness perspective.

PR #1373, opened 2019-09-19
Adding "last updated" info, plus misc improvements

merltron:
This PR is mainly to add the "last update" attribute to each asciidoc file. For now I'll only add this to the developer guide. There will be a related PR for the UI bundle.

Each doc file should now have the attribute `:page-lastupdated: {docdate}`

I tried to set this as a global in the Antora playbook, but it choked (value of "page-lastupdated" undefined)
I tried to set `page.docdate` and `page.attributes.docdate` in the handlebars template, but according to [this list](https://docs.antora.org/antora-ui-default/templates/#template-variables-table) the docdate is not one of the environment variables that Antora knows about (yet). So this seems to be the only viable solution for now..


merltron:
Noprobs, yes will split the other stuff out in a bit - and hide the "lastupdated" bit from the UI template.

PR #1375, opened 2019-09-19
Docs/backport ci build triggers 2019.6

tkfu:


lbonn:
This is maybe the kind of changes that could be pushed to old branches without a PR by the people responsible for the docs. But hopefully, we won't need that often once all of our setup settles.

PR #1380, opened 2019-09-19
DOCS: Also exclude docs/ branches from CI pipelines

tkfu:
@lbonn: I promise this is the last one :-P. (I will likely backport it, but I'll just push directly rather than spamming with PRs.)

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -43,6 +43,7 @@ Docker Setup:
   stage: docker
   except:
     - /^20\d\d\.\d\d?-docs$/
+    - /^docs\//
   services:
     - docker:18-dind
   before_script:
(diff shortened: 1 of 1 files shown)


ci/gitlab/.gitlab-ci.yml
```diff
@@ -43,6 +43,7 @@ Docker Setup:
   stage: docker
   except:
     - /^20\d\d\.\d\d?-docs$/
+    - /^docs/
```
lbonn: Should we use `/^docs\//`?

tkfu: Good call, fixed.

patrickvacek:
I'm confused, is this preventing the docs getting added from any of the docker images? We just recently explicitly added building the Doxygen to one of our pipelines to make sure that doesn't get accidentally broken. It looks like this broke that check if I'm not mistaken.

patrickvacek:
> is this preventing the docs getting added from any of the docker images?

Ah, nm, sorry, this is filtering based on branch name. My bad.

PR #1381, opened 2019-09-19
Adding CTA for next steps

merltron:
Adding CTA and link to next step which guides the reader to try and upload some updates to sample software.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/_partials/recommended-steps.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/_partials/recommended-steps.adoc
@@ -115,3 +115,12 @@ Once you understand how the libaktualizr demo application works, you'll want to 
 After you've finished integration and development, you'll need to move disk images from one account to another. For example, you might want to send a development build that you’re happy with to the QA team, or send that build to the deployment team once it’s passed QA.
 
 //  end::deploy-steps[]
+
+//  tag::firstbuild-nextstep[]
+[TIP]
+====
+Once you've build your first image, try adding some new software and xref:ota-client::pushing-updates.adoc[push your update to the OTA Connect server].
+====
+//MC: full versioned xref because snippet is displayed in two components.
+
+//  end::firstbuild-nextstep[]
--- a/docs/ota-client-guide/modules/ROOT/pages/pushing-updates.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/pushing-updates.adoc
@@ -1,4 +1,4 @@
-= Upload a sample software version
+= Upload an updated software version
 :page-partial:
 ifdef::env-github[]
 
--- a/docs/ota-client-guide/modules/ROOT/pages/pushing-updates.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/pushing-updates.adoc
@@ -1,4 +1,4 @@
-= Upload an updated software version
+= Add software to your Yocto image
 :page-partial:
 ifdef::env-github[]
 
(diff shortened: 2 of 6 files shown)


docs/ota-client-guide/modules/ROOT/pages/pushing-updates.adoc
```diff
@@ -1,4 +1,4 @@
-= Upload a sample software version
+= Upload an updated software version
```
tkfu: All of the stuff on this page is about modifying a Yocto image for updating; maybe it would be better to have a title that was something like "Create and upload an updated Yocto image"?

docs/ota-client-guide/modules/ROOT/pages/_partials/recommended-steps.adoc
```diff
@@ -115,3 +115,12 @@ Once you understand how the libaktualizr demo application works, you'll want to
 After you've finished integration and development, you'll need to move disk images from one account to another. For example, you might want to send a development build that you’re happy with to the QA team, or send that build to the deployment team once it’s passed QA.
 
 //  end::deploy-steps[]
+
+//  tag::firstbuild-nextstep[]
+[TIP]
+====
+Once you've build your first image, try adding some new software and xref:latest@ota-client::pushing-updates.adoc[push an update to the OTA Connect server].
```
tkfu: Is there some gotcha about having to use the extended xref syntax in partials? Shouldn't this work with just xref:pushingupdates.adoc[]?

docs/ota-client-guide/modules/ROOT/pages/pushing-updates.adoc
merltron: Well, I was trying to highlight the high-level goals in the titles. I'm only modifying the Yocto image so that I can get some new software on the device, and this happens to be the way that I have to do it (as opposed to directly installing a package). 

But we have the "Software management" topic now, so  I don't mind either way. The only other problem is that we now have many modifiers for "image"
https://www.google.com/search?q=site:docs.ota.here.com+*+image

Yocto image, disk image, software image, filesystem image, SD Card image, ect..

Perhaps "Use Yocto to update your image"?

Would be good be consistent...I know we talked about "software image" but I noticed other products talk more often about "disk images". Also, isn't Yocto a tool to BUILD images? 
In the terminology, do we meed to distinguish images built with Yocto from other types of images? Maybe the answer is "yes" do if Yocto is the tool people are using, not really sure myself.

docs/ota-client-guide/modules/ROOT/pages/_partials/recommended-steps.adoc
merltron: This content is included in getstarted page so xref:pushingupdates.adoc[]? shows up as broken when I view the page there.
i.e: https://docs.ota.here.com/getstarted/dev/raspberry-pi.html
Includes content from
https://docs.ota.here.com/ota-client/latest/build-raspberry.html
the the xref has to work in both places

tkfu: Interesting, I would've hoped Antora was smart enough to resolve it correctly, even in includes, since it has all the information it needs at include time.

Does it also fail without the `latest@`?

merltron: > Does it also fail without the `latest@`?

Nope, it works. Looks like just the component is needed,

docs/ota-client-guide/modules/ROOT/pages/pushing-updates.adoc
tkfu: Looking at the actual content of the page, something like "Add software to your Yocto image" seems most appropriate. There's not a single word of instruction about uploading it; the three topics are adding a package to a yocto build, adding a layer to a yocto build, and writing your own package or layer.

tkfu:
Great! If you can just squash out that merge commit in the middle, this is ready to go.

merltron:
> Great! If you can just squash out that merge commit in the middle, this is ready to go.

OK should be good to go now.

PR #1382, opened 2019-09-19
Docs/fix more broken xrefs

tkfu:
Cleaning up the rest of the "dev@" xrefs, plus a stray footnote that was broken because of an unescaped square brace.

@merltron There should never have been a need to include "dev@" in any xref; you only need to specify a version if you want a specific one that is not the default. There are quite a few more links that currently use dev@getstarted or dev@ota-web; I didn't change those yet in case there's some reason for it I don't know about. Is there?


docs/ota-client-guide/modules/ROOT/pages/simulate-device-basic.adoc
```diff
@@ -123,7 +123,7 @@ Each directory should contain the following:
 
 * credentials.zip
 * sota_local.toml -- a config file for aktualizr
-* (Optional) A `virtualsec.json` file containing the configuration for one or more secondary ECUs.footnote:[The terms "primary" and "secondary" ECU are used in the Uptane specification. For more information about the difference between primary and secondary ECUs, see our overview of the dev@ota-client::uptane.adoc#_primary_and_secondary_ecus[Uptane framework].]
+* (Optional) A `virtualsec.json` file containing the configuration for one or more secondary ECUs.footnote:[The terms "primary" and "secondary" ECU are used in the Uptane specification. For more information about the difference between primary and secondary ECUs, see our overview of the xref:uptane.adoc#_primary_and_secondary_ecus[Uptane framework\].]
```
merltron: @tkfu This content is included here https://docs.ota.here.com/getstarted/dev/simulate-device-workstation.html
So a local xref wont work when viewing it from "Get Started"

merltron:
> Cleaning up the rest of the "dev@" xrefs, plus a stray footnote that was broken because of an unescaped square brace.
> 
> @merltron There should never have been a need to include "dev@" in any xref; you only need to specify a version if you want a specific one that is not the default. There are quite a few more links that currently use dev@getstarted or dev@ota-web; I didn't change those yet in case there's some reason for it I don't know about. Is there?

Some of the topics were originally in other components (such as metaupdater aka "ota-build") plus I have a value memory of struggling with "cross-component" xrefs that were missing a version. Anyway, I assume that if we want to always want to point to latest we would need to include "latest" as a version?

merltron:
> Anyway, I assume that if we want to always want to point to latest we would need to include "latest" as a version?

OK, I confirm that You're correct - we don't need to include the version, it DOES default to latest.

tkfu:
I believe the way it works is that it'll default to whatever is at the top of the list according to [antora's ordering rules](https://docs.antora.org/antora/2.1/component-versions/#how-antora-sorts-versions); since those rules always place latest at the top because it's alphabetical, it should always go to latest. Maybe we should do some more research to confirm this, though.

docs/ota-client-guide/modules/ROOT/pages/simulate-device-basic.adoc
tkfu: Okay, I've updated. Am I correct that all pages that are included elsewhere have the `:page-partial:` attribute? (Or are in the _partials directory?)

merltron: Yes, whole pages have the `:page-partial:` attribute and can be anywhere you want. Smaller, content chunks have to live in the partials dir.

PR #1383, opened 2019-09-20
Add github CI action to build and upload macOS package

tkfu:
Github has CI now! And they have agents running macOS!

So, this PR automates almost all of [this process](https://docs.ota.here.com/ota-client/latest/release-process.html#_update_the_homebrew_recipe_for_aktualizr). Normally the only thing you need to change in the build part of the recipe is the version; we can do that with sed. Homebrew provides tooling for generating the bottle, and for modifying the recipe to include the bottle, so we use that to generate a complete recipe that's ready to submit as a PR against https://github.com/advancedtelematic/homebrew-otaconnect. (The recipe shows up in the build log, and is attached to the CI run as an artifact.) The last step automates adding the bottle to the newly-generated release, which makes the generated recipe with the `bottle do` block work.

I used a personal repo to test this out; you can see a successful run over [here](https://github.com/tkfu/homebrew-action-testing/runs/229686121), with the release with the bottle attached [here](https://github.com/tkfu/homebrew-action-testing/releases/tag/2019.8). (We can't do a trial run in the real aktualizr repo, because it only triggers when a new release is created.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,21 +7,10 @@ jobs:
     name: Build macOS homebrew bottle for aktualizr release page
     runs-on: macOS-10.14
     steps:
-    # This first step is here to work around the fact that aktualizr doesn't build on macOS
-    # with boost 1.71. It checks out a known good revision of the boost recipe and
-    # reinstalls it. However, right now the macOS base image GitHub's agents are using has
-    # an older version of boost that builds without modification, so the commit we're
-    # trying to check out doesn't exist (hence the `|| true`). So, this is more-or-less for
-    # future-proofing; we don't know when GitHub will update their agents to the latest
-    # versions.
-    - name: Check out known-working version of boost
-      run: |
-        BREW_GIT_DIR=$(brew --prefix)/Homebrew/Library/Taps/homebrew/homebrew-core
-        git -C "$BREW_GIT_DIR" checkout 5da5895add Formula/boost.rb && brew reinstall boost || true
     - name: Attempt to auto-update recipe and build bottle
       run: |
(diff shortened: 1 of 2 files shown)


tkfu:
I discovered that there's something wrong with the curl command--the file that gets added to the release has some text, including the HTTP headers, prepended, for some reason. Just gotta figure out the right curl incantation and this'll be ready to go.

lbonn:
Just a quick note: we usually avoid having scripts with tons of steps directly into CI definitions and offload most of the complexity in external scripts. One of the advantage is that it makes it easier to run some CI steps manually, for example during debugging.

I don't know if everything can be merged into a single script here and we can always revisit that later. But if you have ideas about how to simplify the steps in the yaml file, it would be appreciated.

tkfu:
Well, just in time to miss the latest release, I have this working now. Two main changes here in 83c76818c2a4781f6d0ba101d802def69ff8da93:

* Removed boost workaround. So, either homebrew's boost packaging changed or something in between 2019.8 and 2019.9 fixed whatever the problem was, because aktualizr now builds just fine on macos with the latest boost. That gets rid of one of the complex bits; if it breaks again later we can worry about it then.
* Used a designed-for-the-purpose tool called [ghr](https://github.com/tcnksm/ghr) to upload the artifact to the release page, instead of manually calling the APIs with curl. This gets rid of some complexity, and handles all of the annoying logic for looking up the release ID &c.

@lbonn, is that enough simplifying to suit you? I could still refactor this out into a script, but it needs some the environment variables github sets to make it work anyway, and it really just amounts to 5 commands; it doesn't quite seem worth it as a separate script.

This should be ready to merge, with any approval.

PR #1384, opened 2019-09-23
Signal handling

lbonn:
@mike-sul Here is a try for handling signals more cleanly. I resurrected a very old branch of mine. Is it close to what you had in mind?

Right now, it's only enabled in aktualizr's `RunForever` loop. Adding it to other parts of the code might be possible but would need more thinking.

### DIFF
--- a/src/aktualizr_primary/main.cc
+++ b/src/aktualizr_primary/main.cc
@@ -11,6 +11,7 @@
 #include "primary/aktualizr_helpers.h"
 #include "secondary.h"
 #include "utilities/aktualizr_version.h"
+#include "utilities/sig_handler.h"
 #include "utilities/utils.h"
 
 namespace bpo = boost::program_options;
@@ -131,6 +132,8 @@ int main(int argc, char *argv[]) {
     }
 
     aktualizr.Initialize();
+
+    SigHandler::get().start([&aktualizr]() { aktualizr.Shutdown(); });
 
     std::string run_mode;
     if (commandline_map.count("run-mode") != 0) {
--- a/src/aktualizr_primary/main.cc
+++ b/src/aktualizr_primary/main.cc
@@ -133,7 +133,14 @@ int main(int argc, char *argv[]) {
 
     aktualizr.Initialize();
 
-    SigHandler::get().start([&aktualizr]() { aktualizr.Shutdown(); });
+    // handle unix signals
+    SigHandler::get().start([&aktualizr]() {
+      aktualizr.Abort();
+      aktualizr.Shutdown();
+    });
+    SigHandler::signal(SIGHUP);
+    SigHandler::signal(SIGINT);
+    SigHandler::signal(SIGTERM);
 
     std::string run_mode;
     if (commandline_map.count("run-mode") != 0) {
@@ -164,7 +171,13 @@ int main(int argc, char *argv[]) {
       boost::signals2::connection ac_conn =
           aktualizr.SetSignalHandler(std::bind(targets_autoclean_cb, std::ref(aktualizr), std::placeholders::_1));
 
-      aktualizr.RunForever().get();
+      try {
+        aktualizr.RunForever().get();
+      } catch (const std::exception &ex) {
+        LOG_ERROR << ex.what();
+      }
+
+      LOG_DEBUG << "Aktualizr daemon exiting...";
     }
     r = EXIT_SUCCESS;
   } catch (const std::exception &ex) {
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -86,7 +86,10 @@ std::future<void> Aktualizr::RunForever() {
 
     std::unique_lock<std::mutex> l(exit_cond_.m);
     while (true) {
-      UptaneCycle();
+      if (!UptaneCycle()) {
+        break;
+      }
+
       if (exit_cond_.cv.wait_for(l, std::chrono::seconds(config_.uptane.polling_sec),
                                  [this] { return exit_cond_.flag; })) {
         break;
@@ -98,7 +101,6 @@ std::future<void> Aktualizr::RunForever() {
 }
 
 void Aktualizr::Shutdown() {
-  Abort();
   {
     std::lock_guard<std::mutex> g(exit_cond_.m);
     exit_cond_.flag = true;
--- a/src/libaktualizr/primary/aktualizr.cc
+++ b/src/libaktualizr/primary/aktualizr.cc
@@ -83,12 +83,27 @@ bool Aktualizr::UptaneCycle() {
 std::future<void> Aktualizr::RunForever() {
   std::future<void> future = std::async(std::launch::async, [&]() {
     SendDeviceData().get();
-    while (UptaneCycle()) {
-      std::this_thread::sleep_for(std::chrono::seconds(config_.uptane.polling_sec));
+
+    std::unique_lock<std::mutex> l(exit_cond_.m);
+    while (true) {
+      UptaneCycle();
+      if (exit_cond_.cv.wait_for(l, std::chrono::seconds(config_.uptane.polling_sec),
+                                 [this] { return exit_cond_.flag; })) {
+        break;
+      }
     }
     uptane_client_->completeInstall();
   });
   return future;
+}
+
+void Aktualizr::Shutdown() {
+  Abort();
+  {
+    std::lock_guard<std::mutex> g(exit_cond_.m);
+    exit_cond_.flag = true;
+  }
+  exit_cond_.cv.notify_all();
 }
 
 void Aktualizr::AddSecondary(const std::shared_ptr<Uptane::SecondaryInterface> &secondary) {
--- a/src/libaktualizr/utilities/sig_handler.cc
+++ b/src/libaktualizr/utilities/sig_handler.cc
@@ -1,14 +1,10 @@
 #include "sig_handler.h"
 #include "logging/logging.h"
 
-void signal_handler(int sig) {
-  (void)sig;
-  bool v = false;
-  // put true if currently set to false
-  SigHandler::signal_marker_.compare_exchange_strong(v, true);
-}
-
 std::atomic<bool> SigHandler::signal_marker_;
+std::mutex SigHandler::exit_m_;
+std::condition_variable SigHandler::exit_cv_;
+bool SigHandler::exit_flag_;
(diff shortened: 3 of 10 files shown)


src/libaktualizr/primary/aktualizr.cc
```diff
@@ -83,8 +84,24 @@ bool Aktualizr::UptaneCycle() {
 std::future<void> Aktualizr::RunForever() {
   std::future<void> future = std::async(std::launch::async, [&]() {
     SendDeviceData().get();
-    while (UptaneCycle()) {
-      std::this_thread::sleep_for(std::chrono::seconds(config_.uptane.polling_sec));
+
+    bool exiting = false;
+    std::mutex exit_m;
+    std::condition_variable exit_cv;
+    SigHandler::get().start([&exit_m, &exit_cv, &exiting]() {
```
mike-sul: IMHO, a single handler should 'live' at an application level not at a library one since an application might incorporate not just libaktualizr but also another libraries and functionality that require triggering specific teardown actions on a system signal. Specifically, I would put a signal handler registration&instantion at aktualizr app (aktualizr-primary) level and triggers teardown functionality via Aktualizr API, it will allow an application developer to extend or override the signal handling functionality.

src/libaktualizr/utilities/sig_handler.cc
```diff
@@ -0,0 +1,77 @@
...
+        signal_pending = true;
+      }
+
+      if (signal_pending && masked_secs_ == 0) {
+        LOG_INFO << "calling signal handler";
+        on_signal();
+        return;
+      }
+
+      boost::this_thread::sleep_for(boost::chrono::seconds(1));
```
mike-sul: Why not just sleep for `masked_secs_` and then run `on_signal()` ? Or this is for case when another signal arrives during masking. Actually the best practice says to return from a single handler as soon as possible, not sure if I fully understand a value in the delayed signal handling

src/libaktualizr/utilities/sig_handler.cc
```diff
@@ -0,0 +1,77 @@
...
+  if (polling_thread_.get_id() != boost::thread::id()) {
+    throw std::runtime_error("SigHandler can only be started once");
+  }
+
+  polling_thread_ = boost::thread([this, on_signal]() {
+    while (true) {
+      bool signal = signal_marker_.exchange(false);
+
+      if (signal) {
+        LOG_INFO << "received KILL request";
```
mike-sul: It's not just `KILL` signal handler

src/libaktualizr/utilities/sig_handler.cc
```diff
@@ -0,0 +1,77 @@
...
+      {
+        boost::unique_lock<boost::mutex> lock(m);
+        if (masked_secs_ > 0) {
+          masked_secs_ -= 1;
+        }
+      }
+    }
+  });
+
+  signal(SIGHUP, signal_handler);
```
mike-sul: I would make it configurable for an application level, i.e. the one who defines a signal handler is supposed to specify what is/are signals this handler is intended for. Ideally, I as an application/executable developer should have possibility to define different signal handlers for different signals.

KostiantynBushko:
I agree with @mike-sul  the signal handler should not be a part of the libaktualizr...
If, for example, we set handler for SIGINT signal and do not perform exit it means we will not be able kill app ever!
I tested it with my client, using RunForever()  feature  I can't stop app by pressing Ctrl-C or using kill <pid> with is obviosly since we handle SIGINT and do not call exit...

lbonn:
Yes, you're both right, this PR is not ready yet.

lbonn:
It should be simpler and more inter-operable now.

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -83,14 +83,32 @@ bool Aktualizr::UptaneCycle() {
...
+        break;
+      }
     }
     uptane_client_->completeInstall();
   });
   return future;
 }
 
+void Aktualizr::Shutdown() {
+  Abort();
```
mike-sul: Abort() stops the command queue threads which is stopped in CommandQueue() dtor which itself is triggered during Aktualizr instance destruction, so calling Abort() in case of signal handling from the app/exe that destroys Aktualizr instance anyway looks like redundant. Having said that, I think it, might be useful if there is a need in calling RunForever() and Shutdown() multiple times within a single Aktualizr instance lifetime. Perhaps, break it into two methods ?

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -83,14 +83,32 @@ bool Aktualizr::UptaneCycle() {
...
+      if (!UptaneCycle()) {
+        break;
+      }
+
+      if (exit_cond_.cv.wait_for(l, std::chrono::seconds(config_.uptane.polling_sec),
+                                 [this] { return exit_cond_.flag; })) {
+        break;
+      }
     }
     uptane_client_->completeInstall();
```
mike-sul: If completeInstall() is not called will it be properly handled during the subsequent Aktualizr/UptaneCycle() ?

src/aktualizr_primary/main.cc
```diff
@@ -132,6 +133,12 @@ int main(int argc, char *argv[]) {
 
     aktualizr.Initialize();
 
+    // handle unix signals
+    SigHandler::get().start([&aktualizr]() { aktualizr.Shutdown(); });
+    SigHandler::signal(SIGHUP);
+    SigHandler::signal(SIGINT);
+    SigHandler::signal(SIGTERM);
```
mike-sul: I hope UptaneCycle() can end its work within systemd's `DefaultTimeoutStopSec= default to 90s` otherwise it will send SIGKILL and causes non-graceful shutdown of Aktualizr. Perhaps, we might consider setting higher value of `TimeoutStopSec=` in https://github.com/advancedtelematic/meta-updater/blob/master/recipes-sota/aktualizr/files/aktualizr-secondary.service

lbonn: I figured that 90s is actually reasonable. How long would you suggest it should wait?

In any way, we cannot avoid all the cases of abrupt shutdown and if aktualizr doesn't recover after one, we consider that a bug.

src/libaktualizr/primary/aktualizr.cc
lbonn: `Abort()` docstring says that it aborts the currently running command if it can be aborted, that's why it sounded appropriate to call it from the signal handling.

So you're suggesting moving away `Abort()` from this method to the signal handler which will call `Abort();Shutdown();`?

lbonn: Or given the docstring I should maybe merge Shutdown() into Abort(): Abort() would also apply to RunForever()?

src/aktualizr_primary/main.cc
mike-sul: > How long would you suggest it should wait?
I suggest to consider changing the default value (90s) if we face any issue during CI/test phase.

src/libaktualizr/primary/aktualizr.cc
lbonn: Right now this method is used for rebooting. So:

- if the system reboots and aktualizr is killed before this method, nothing is lost
- if aktualizr receives SIGKILL or crashes before this method, `isInstallCompletionRequired()` should still return true when systemd would restart aktualizr and it should reboot then. But I don't think we cover that with any unit test.

src/libaktualizr/primary/aktualizr.cc
mike-sul: I think, it depends on what Shutdown() is intended for ? If we would like to stop Aktualizr during an app exit then this method is useless at all because teardown/shutdown happens at Aktualizr dtor.

If Shutdown() is intended for the case when an user/app would like to stop aktualizr's UptaneCycle and then run it again within a single Aktualizr instance lifetime then the user will need to Initialize() if abort is part of Shutdown. BTW, abort doesn't stop the report queue thread.

Whether abort() should be part of the signal handler routine depends on a need in aborting ongoing operation during restart. I don't have strong opinion here, taking into account that the most timing operation is download and we support download resuming I think, I am fine with abort, but it might have negative impact when some write to DB or communication with secondaries is interrupted. I think, I would go with abort() at the moment.

lbonn: I'm not sure what you exactly mean in your first point. You mean that Shutdown() wouldn't be useful as an handler to a gui event for example? And you're saying that calling Abort() there is not useful because it's also called later? But the point is that calling Abort() early can speed up the tear down and trigger the `~Aktualizr()` call in a shorter delay. Or am I missing something

But your second point sounds right, calling `Initialize()` again after `Shutdown()` sounds quite wrong. I'll move `Abort()` to the signal handler.

mike-sul:
Perhaps, it makes sense to add a test that test it from aktualizr client standpoint of view, e.g spawn aktualizr process and send SIGTERM signal to it and wait/join the aktualizr process ?

lbonn:
@mike-sul ok, I've added a test using the python fixtures

tests/test_aktualizr_kill.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
+
+import logging
+import argparse
+
+from os import getcwd, chdir
+
+from test_fixtures import with_aktualizr, with_uptane_backend, KeyStore, with_secondary, with_path,\
+    DownloadInterruptionHandler, MalformedJsonHandler, with_director, with_imagerepo, InstallManager,\
```
mike-sul: *Handler and with_treehub and InstallManager and with_imagerepo can be removed

tests/test_aktualizr_kill.py
```diff
@@ -0,0 +1,54 @@
...
+    with_install_manager, with_images, MalformedImageHandler, with_customrepo, SlowRetrievalHandler, \
+    RedirectHandler, with_sysroot, with_treehub
+
+
+logger = logging.getLogger(__file__)
+
+
+@with_uptane_backend(start_generic_server=True, port=8891)
```
mike-sul: This commit should be synced with this one https://github.com/advancedtelematic/aktualizr/pull/1395 as I changed in it an approach to choosing a port for the mocked servers.

tests/test_aktualizr_kill.py
lbonn: thanks!

lbonn:
Rebased after the merge of #1395

PR #1386, opened 2019-09-23
Docs/refactor resource ids

tkfu:
This is a general cleanup of the xref links; in many places we were unnecessarily using version identifiers or component IDs. This makes the docs more brittle.

However, we **MUST** use component IDs in includes or links that are inside of partials or page-partials, even when linking to a page in the same component; this is because they might be included from another component.

I've also removed some images that are no longer needed. However, this change depends on https://github.com/advancedtelematic/ota-plus-server/pull/2092 being merged first, so please don't merge this until that one's in.


patrickvacek:
Can we write a doc for how to edit these docs? It is actually a bit difficult to keep track of all the details.

patrickvacek:
Also: it would be great if we could find a way to avoid putting even the smallish binary files in git. I don't have a good idea how to avoid that for now, though.

tkfu:
@patrickvacek Yes to the how-to doc. There's a preliminary version of that in the readme of the main docs repo, and I plan to expand/improve it, and put a copy of it here, as soon as I finally finish my review of all the new site content.

tkfu:
As for the binaries, we actually could do a refactor that puts them in their own repo, so as to avoid having it pollute this one.

PR #1388, opened 2019-09-24
DOCS: use a single file for configuring the aktualizr version

tkfu:
Reasoning for this change: easier inter-component behaviour. I was reviewing some things, and discovered that one of our get started pages was still referencing 2019.5 as the most current version of aktualizr. We don't want to have to make changes in any non-aktualizr repo when we bump the version, so we create a file that can be used as an include in other components to make sure they are pointing to the latest version.


patrickvacek:
Worth updating https://docs.ota.here.com/ota-client/latest/release-process.html#_update_the_changelog_and_other_docs to reflect this change as well. Should make that step slightly easier and clearer.

PR #1389, opened 2019-09-25
Disable test_backend_failure on ptest

lbonn:
Currently failing and probably not worth fixing


lbonn:
Sorry, I was way too fast and the CMake was bogus.

PR #1392, opened 2019-09-26
Remove the old .deb files we were carrying around

lbonn:


tests/test_install_aktualizr_and_update.sh
```diff
@@ -33,4 +36,4 @@ if [ $? -ne 0 ];then
     echo "ERROR: $filename does not exist or sha256sum does not match."
```
Zee314159: I also have a bit of concern at line33-line35. what if somehow `aktualizr_info` run into some error and return a blank filename, and `ls $TEMP_DIR/images/$filename` still return to 0, but actually it shouldn't?

lbonn: yes actually this test is weird, it should just be `if [ ! -f ... ]; then`

tests/test_install_aktualizr_and_update.sh
```diff
@@ -33,4 +36,4 @@ if [ $? -ne 0 ];then
...
     exit 1
 fi
-
+rm -rf $TEMP_DIR
```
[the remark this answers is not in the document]
lbonn: yes

lbonn:
Hm I've refactored some bash line into two steps and now it fails because jq is missing on our docker envs.

Looks like the tests passed by mistake before that.

PR #1393, opened 2019-09-26
Move out uptiny from aktualizr's main repository

lbonn:
See https://github.com/advancedtelematic/uptiny


lbonn:
> Does that mean that ATM https://github.com/advancedtelematic/uptiny with have an infinitely-recursing git submodule chain?

Hm maybe, I'll change that as soon as we merge this

lbonn:
Ah actually no, because there is no `uptiny` submodule in aktualizr, it's a simple directory.

patrickvacek:
> Ah actually no, because there is no `uptiny` submodule in aktualizr, it's a simple directory.

Yes, of course. I wonder what would happen if you tried to do that, though.

PR #1394, opened 2019-09-26
aktualizr-lite: Move more code to "helpers" so it can be tested

doanac:
This change nets us a little more test coverage and sets things up so that we can add more in the future. The first change is just renaming a file. The second change adds in a little test coverage to the "liteClient" and the third change does a small refactor that's now a lot easier to test thanks to the first 2 changes.

### DIFF
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -1,5 +1,8 @@
+set(AKTUALIZR_LITE_SRC main.cc helpers.cc)
+set(AKTUALIZR_LITE_HEADERS helpers.h)
+
 if(BUILD_OSTREE)
-add_executable(aktualizr-lite main.cc version.h)
+add_executable(aktualizr-lite ${AKTUALIZR_LITE_SRC})
 target_link_libraries(aktualizr-lite aktualizr_static_lib ${AKTUALIZR_EXTERNAL_LIBS})
 
 install(TARGETS aktualizr-lite RUNTIME DESTINATION bin COMPONENT aktualizr-lite)
@@ -19,7 +22,7 @@ add_library(t_lite-mock SHARED ostree_mock.cc)
 
 endif(BUILD_OSTREE)
 
-add_aktualizr_test(NAME lite-version SOURCES version_test.cc)
+add_aktualizr_test(NAME lite-helpers SOURCES helpers.cc helpers_test.cc)
 
-aktualizr_source_file_checks(main.cc version.h version_test.cc ostree_mock.cc)
+aktualizr_source_file_checks(main.cc ${AKTUALIZR_LITE_SRC} ${AKTUALIZR_LITE_HEADERS} helpers_test.cc ostree_mock.cc)
 # vim: set tabstop=4 shiftwidth=4 expandtab:
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -19,10 +19,12 @@ add_test(test_aktualizr-lite
         ${RUN_VALGRIND}
 )
 add_library(t_lite-mock SHARED ostree_mock.cc)
+add_aktualizr_test(NAME lite-helpers SOURCES helpers.cc helpers_test.cc
+                   ARGS ${PROJECT_BINARY_DIR}/ostree_repo)
+set_tests_properties(test_lite-helpers PROPERTIES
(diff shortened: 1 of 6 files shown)


lbonn:
The changes seem fine. Valgrind reported some memory leaks on CI, in test_lite-helpers though.

doanac:
@lbonn - CI works! I just pushed a small change to fix the memory leak in OstreeManager as well as my bad copy/paste that was based on it.

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -292,7 +292,7 @@ Json::Value OstreeManager::getInstalledPackages() const {
 
 Uptane::Target OstreeManager::getCurrent() const {
   GObjectUniquePtr<OstreeSysroot> sysroot_smart = OstreeManager::LoadSysroot(config.sysroot);
-  OstreeDeployment *booted_deployment = ostree_sysroot_get_booted_deployment(sysroot_smart.get());
+  g_autoptr(OstreeDeployment) booted_deployment = ostree_sysroot_get_booted_deployment(sysroot_smart.get());
```
mike-sul: I think it will cause an issue in case of a real ostree, not mocked one (`ostree-sysroot.c`). I am gonna fix the memory leak in `aktualizr/tests/ostree_mock.c` anyway

mike-sul: I studied the ostree source code with more details and can confirm that ostree expects that OstreeDeployment instance is owned by OstreeSysroot instance. Any client calling ostree_sysroot_get_booted_deployment() should get a pointer to the OstreeDeployment without any ownership transfer and should not do any reference decrement and/or the OstreeDeployment instance deletion, this is a task of the entity that created this instance. So, I am gonna fix the issue here https://github.com/advancedtelematic/aktualizr/blob/ccf304a93df743beb801699401aafdc1122d4cf7/tests/ostree_mock.c#L133

mike-sul:
@doanac We've been considering refactoring around the 'package manager' implementation in order to make it more flexible and support more use cases, e.g. multi-ECU/secondaries, multiple update types on a single ECU, adding new installation/update type without making changes in libaktaulizr, ect.

The use case of "ostree+docker-app" looks like one of the use cases we would like to support smoothly and I would like to learn it better. I am trying to run aktualizr-lite on a local host and trying to figure out what would be the right configuration to make it working, I would really appreciate it if you provide me with an example of an aktualizr-lite configuration to run it on a local/linux host and with "ostree+docker-app" PM.

src/libaktualizr/package_manager/ostreemanager.cc
doanac: ah - i need to do a similar fix here also: https://github.com/advancedtelematic/aktualizr/blob/master/src/aktualizr_lite/ostree_mock.cc

I'll fix that so that this PR passes valgrind

doanac:
I just pushed a small docker app publicly to help make this easier. Here's the sota.toml I use for development:
~~~
[uptane]
repo_server = "https://api.foundries.io/ota/repo/lmp/api/v1/user_repo/"

[pacman]
type = "ostree+docker-app"
ostree_server = "https://api.foundries.io/ota/treehub/lmp/api/v2/"
docker_apps = "shellhttpd"
docker_apps_root = "/tmp/sota/docker-apps"
sysroot = "/tmp/ostree/sysroot"

[storage]
type = "sqlite"
path = "/tmp/sota"

[provision]
primary_ecu_hardware_id = raspberrypi3-64
~~~

src/libaktualizr/package_manager/ostreemanager.cc
doanac: I think i've finally got this working cleanly: https://github.com/advancedtelematic/aktualizr/commit/3a01b3a95aaf7b1548aa30afcd52f195ea25a171

mike-sul:
> I just pushed a small docker app publicly to help make this easier. Here's the sota.toml I use for development:
> 
> ```
> [uptane]
> repo_server = "https://api.foundries.io/ota/repo/lmp/api/v1/user_repo/"
> 
> [pacman]
> type = "ostree+docker-app"
> ostree_server = "https://api.foundries.io/ota/treehub/lmp/api/v2/"
> docker_apps = "shellhttpd"
> docker_apps_root = "/tmp/sota/docker-apps"
> sysroot = "/tmp/ostree/sysroot"
> 
> [storage]
> type = "sqlite"
> path = "/tmp/sota"
> 
> [provision]
> primary_ecu_hardware_id = raspberrypi3-64
> ```

Thanks, I'll try it.

PR #1395, opened 2019-09-27
OTA-3795: Do default verification if a secondary update and an…

mike-sul:

### DIFF
--- a/src/libaktualizr/package_manager/ostreemanager.cc
+++ b/src/libaktualizr/package_manager/ostreemanager.cc
@@ -234,12 +234,23 @@ OstreeManager::OstreeManager(PackageConfig pconfig, std::shared_ptr<INvStorage> 
 bool OstreeManager::fetchTarget(const Uptane::Target &target, Uptane::Fetcher &fetcher, const KeyManager &keys,
                                 FetcherProgressCb progress_cb, const api::FlowControlToken *token) {
   if (!target.IsOstree()) {
+    // The case when the ostree package manager is set as a package manager for aktualizr
+    // while the target is aimed for a secondary ECU that is configured with another/non-ostree package manager
     return PackageManagerInterface::fetchTarget(target, fetcher, keys, progress_cb, token);
   }
   return OstreeManager::pull(config.sysroot, config.ostree_server, keys, target, token, progress_cb).success;
 }
 
 TargetStatus OstreeManager::verifyTarget(const Uptane::Target &target) const {
+  if (!target.IsOstree()) {
+    // The case when the ostree package manager is set as a package manager for aktualizr
+    // while the target is aimed for a secondary ECU that is configured with another/non-ostree package manager
+    return PackageManagerInterface::verifyTarget(target);
+  }
+  return verifyTargetInternal(target);
+}
+
+TargetStatus OstreeManager::verifyTargetInternal(const Uptane::Target &target) const {
   const std::string refhash = target.sha256Hash();
   GError *error = nullptr;
 
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+import logging
+import argparse
+
+from os import getcwd, chdir
+
+from test_fixtures import KeyStore, with_aktualizr, with_uptane_backend, with_secondary, with_director, with_imagerepo,\
+    with_sysroot, with_treehub
+
+
+logger = logging.getLogger(__file__)
+
+
+"""
+ Test update of Primary and Secondary if their package manager differs, `ostree` and `fake` respectively
+ 
+ Aktualizr/Primary's package manager is set to `ostree`
+ Secondary's package manager is set to `fake`
+ Primary goal is to verify whether aktualizr succeeds with a binary/fake update of secondary
+ while aktualizr/primary is configured with ostree package manager   
+"""
+@with_uptane_backend(start_generic_server=True, port=8788)
+@with_director()
+@with_treehub()
+@with_sysroot()
+@with_secondary(start=True)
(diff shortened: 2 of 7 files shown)


src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -234,12 +234,23 @@ OstreeManager::OstreeManager(PackageConfig pconfig, std::shared_ptr<INvStorage>
...
   if (!target.IsOstree()) {
+    // The case when the ostree package manager is set as a package manager for aktualizr
+    // while the target is aimed for a secondary ECU that is configured with another/non-ostree package manager
     return PackageManagerInterface::fetchTarget(target, fetcher, keys, progress_cb, token);
   }
   return OstreeManager::pull(config.sysroot, config.ostree_server, keys, target, token, progress_cb).success;
 }
 
 TargetStatus OstreeManager::verifyTarget(const Uptane::Target &target) const {
+  if (!target.IsOstree()) {
```
doanac: I think we may need a similar check in dockerappmanager.cc? In fact I think the current logic isn't quite right there. Its doing verification for all target types and really should just run when the type is OSTree. I think something like this is probably correct:
~~~
TargetStatus DockerAppManager::verifyTarget(const Uptane::Target &target) const {
  TargetStatus status = OstreeManager::verifyTarget(target);
  if (status != TargetStatus::kGood || !target.IsOstree()) {
      return status;
  }
  auto cb = [this](const std::string &app, const Uptane::Target &app_target) {
    LOG_INFO << "Verifying " << app << " -> " << app_target;
    std::stringstream ss;
    ss << *storage_->openTargetFile(app_target);
    DockerApp dapp(app, config);
    return dapp.render(ss.str(), false);
  };
  if (!iterate_apps(target, cb)) {
    return TargetStatus::kInvalid;
  }
  return TargetStatus::kGood;
}
~~~

mike-sul: If secondaries/multi-ECU update is not required in case of "ostree+docker-app" package manager usage on Primary then the current logic is fine provided that it works now for updating both ostree and docker app on Primary.

My understanding is that the "ostree+docker-app" package manager is intended for supporting of an ostree-based rootfs update and docker apps configuration update on Primary ? If it's so I am not sure if I understand why `auto res = OstreeManager::install(target);` is called in `DockerAppManager::install()`, I suppose, the logic should call the ostree install routine only if the target type is ostree?

mike-sul: What value of `target.type_` will be in case of the docker-app update ?

mike-sul: We are planning to redesign approach to package management as new requirements/use cases appeared:

1) secondaries support
2) multiple type of update on a single ECU/Primary
3) providing an integrator with  possibility  to add a new type of 'package manager' without libaktualizr modification.

doanac: @mike-sul - The docker-app manager is confusing right now because I haven't documented it well at all. I think what I should do is put a PR that includes some code comments about the package manager so people can reason about it. Then I can fix it and add some tests to make sure its following the logic you've done here properly.

doanac: FYI - here's the blog https://foundries.io/insights/2019/09/23/docker-app/

mike-sul: > FYI - here's the blog https://foundries.io/insights/2019/09/23/docker-app/

Thanks for pointing out to it, gonna read it through.

tests/test_update.py
```diff
@@ -0,0 +1,91 @@
...
+    test_suite_run_result = True
+    for test in test_suite:
+        logger.info('>>> Running {}...'.format(test.__name__))
+        test_run_result = test()
+        logger.info('>>> {}: {}'.format('OK' if test_run_result else 'Failed', test.__name__))
+        test_suite_run_result = test_suite_run_result and test_run_result
+
+    chdir(initial_cwd)
+    exit(0 if test_suite_run_result else 1)
+
```
patrickvacek: This is powerful and cool, but I wonder if it would've been possible to test this at a lower level. This is also becoming quickly redundant with `shared_cred_prov_test.py` and we should consider merging them at some point.

Also, minor nitpick: the name of this test/file is quite general. It's worth getting a bit more specific next time we mess with this.

mike-sul: Yes, I agree, the name is too generic.

In terms of 'This is also becoming quickly redundant with shared_cred_prov_test.py',  I think they are very different, ' shared_cred_prov_test.py' tests just provisioning of non-ostree against the real backend, while 'test_update.py' tests a secondary update against the mocked backend if primary is based on ostree. This is the only test at the moment that can catch the issue described here https://saeljira.it.here.com/browse/OTA-3795, none of our tests could catch the issue before.

In retrospect, I think it makes sense to move this test to the IP secondary test suite (ipsecondary_test.py) as it actually tests the secondary update or replace it with lower level tests (C/C++ based) utilizing a virtual secondary. The later, will require duplication of the boilerplate code around ostree mocking located in `aktualizr_fullostree_test.cc` or the test logic itself can be added to aktualizr_fullostree_test.cc, what I tried but failed.

patrickvacek: > In terms of 'This is also becoming quickly redundant with shared_cred_prov_test.py', I think they are very different

Fair enough. Thanks for the explanation!

> I think it makes sense to move this test to the IP secondary test suite (ipsecondary_test.py) as it actually tests the secondary update

Not a bad idea.

> or replace it with lower level tests (C/C++ based) utilizing a virtual secondary

Yeah, we could do that, but I'd say that isn't a priority, since you already have a nice working version in Python!

PR #1397, opened 2019-09-29
Misc Language fixes

merltron:
- Added adjective "conserve": Technically bandwidth isn't something that you want to minimize - your want to have as much of it as possible.
- "Footprint" needs an article, you can't say "it minimizes ... footprint" - added "your".
- Broke up very long sentence (still not happy with the final sentence "...file system deployments" sounds a bit vague but can't think of anything better right now...)

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/developer-tools.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/developer-tools.adoc
@@ -45,7 +45,7 @@ The meta-updater Yocto layer contains "recipes" related to software updating. Yo
 
 Later in this guide, we go into xref:yocto.adoc[more detail about how Yocto works].
 
-| *OSTree* |  OSTree is a tool for managing Linux filesystems, and delivers atomic updates with rollback capability and built-in delta support. OSTree has several advantages over traditional dual-bank systems, but the most important one is that it minimizes network bandwidth and data storage footprint by sharing files with the same contents across file system deployments.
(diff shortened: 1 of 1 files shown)


tkfu:
>     * Added adjective "conserve": Technically bandwidth isn't something that you want to minimize - your want to have as much of it as possible.

I can see where this might be misparsed, but the intent was clear to me: it minimizes the footprint on (both) your network bandwidth and data storage. They are grouped together because they are minimized through the same mechanism. (Hence "the most important one" and not "the most important ones" at the beginning of the sentence.)

>     * "Footprint" needs an article, you can't say "it minimizes ... footprint" - added "your".

I'd be inclined to disagree; it's perfectly acceptable to leave the article implicit with words like usage, efficiency, throughput, or footprint. Compare "The Farnsworth Valve maximizes the throughput of water through the system while open."/"The Farnsworth Valve maximizes the throughput while open."/"The Farnsworth Valve maximizes throughput while open." I find the third option to be clearly best.

merltron:
OK, well generally I was trying to make it easier to read - but since this is my last day maybe you could get a second opinion from another native speaker?

One thing though, since I love a good bout of grammar banter  -  unlike "throughtput" and "efficiency"  - "footprint" is a countable noun (1 footprint, 2 footprints) so it doesn't really that fit in that list. Gimme a good example of where "footprint" is used without an article an I'll be sold.

But I'll confess to you here, that my primary motivation for making this change was to get screenshots for my internal "updating docs for dummies" guide: https://confluence.in.here.com/display/OTA/Updating+OTA+Connect+Docs

I was quickly trying to find something that I could genuinely change as part of my guide, and that sentence just caught my eye.

tkfu:
I got rid of "footprint" altogether and rephrased to make the grouping clearer.

PR #1399, opened 2019-10-01
Aktualizr-lite: return only not-nulled target to avoid crash

mike-sul:


doanac:
good catch

PR #1402, opened 2019-10-08
Store installed version as "non current" before installing

lbonn:
So that the metadata can be re-accessed even if the process is interrupted before the end of installation.

In the infamous "error 20" scenario, the update now shows as successful with error code "0". The reason is that aktualizr doesn't report any installation report but backend accepts it because we now report the correct new version.

It's also maybe not ideal as we would maybe want to report some kind of error (should director error on empty installation reports?), but I think it's a step in the good direction.


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -160,6 +160,8 @@ void SotaUptaneClient::finalizeAfterReboot() {
 data::InstallationResult SotaUptaneClient::PackageInstallSetResult(const Uptane::Target &target) {
   data::InstallationResult result;
   Uptane::EcuSerial ecu_serial = uptane_manifest.getPrimaryEcuSerial();
+  // load the target in installed versions so that it can be re-accessed later
+  storage->saveInstalledVersion(ecu_serial.ToString(), target, InstalledVersionUpdateMode::kNone);
```
patrickvacek: Is it safe to put that in the `installedVersions` table if it hasn't been installed yet?

lbonn:
Yes the signal tests sometimes hangs like that... The issue is hard to reproduce though.
Where do you think we should add a comment?

patrickvacek:
> Where do you think we should add a comment?

Probably in `SotaUptaneClient::PackageInstallSetResult()`, just to clarify how we expect to "recover" if the installation is interrupted and the potential risk if more logic is added after the OSTree deploy that might get skipped.

Is it reasonable to add a test for this scenario? How have you been simulating it, anyway?

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Cool, just checking. Seems fine, then.

lbonn:
> Is it reasonable to add a test for this scenario? How have you been simulating it, anyway?

Yes, testing is a bit tricky though. For verification, I've built a qemu image with an aktualizr that just throws after the ostree step.

patrickvacek:
> For verification, I've built a qemu image with an aktualizr that just throws after the ostree step.

That's probably fine for now. Maybe for now we should just make a tech debt task to figure out how to test it, and then if you add a small comment about the current risks of the installation interruption, I'll approve.

PR #1404, opened 2019-10-10
Fix test_aktualizr_kill

lbonn:
Try to resolve the failures we get on CI.

There might be something else, stay tuned.

### DIFF
--- a/tests/test_fixtures.py
+++ b/tests/test_fixtures.py
@@ -6,6 +6,7 @@ import threading
 import time
 import os
 import shutil
+import signal
 import socket
 
 from os import devnull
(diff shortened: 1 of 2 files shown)


tests/test_fixtures.py
```diff
@@ -233,12 +232,14 @@ def output(self):
     def wait_for_completion(self, timeout=120):
         self._process.wait(timeout)
 
-    def wait_for_provision(self, timeout=60):
-        while True:
-            info = self.get_info()
-            if 'Provisioned on server: yes' in info:
-                break
+    def wait_for_provision(self, timeout=20):
```
mike-sul: What this 'timeout' param is for ?

Frankly speaking, I am getting annoyed by nondeterministic behavior of aktualizr-info, it doesn't yield the desired result even after a device is provisioned, looks like it happens when it tries to access the DB when aktualizr writes some data to it, just guessing here...

patrickvacek: > I am getting annoyed by nondeterministic behavior of aktualizr-info, it doesn't yield the desired result even after a device is provisioned, looks like it happens when it tries to access the DB when aktualizr writes some data to it, just guessing here...

Yeah, that has sporadically cropped up from time to time, and the general suspicion is the database access issue you mention. Perhaps aktualizr-info needs a built-in retry?

lbonn: > What this 'timeout' param is for ?

The method is a `wait_for_xxx`, so it's here to prevent an infinite block if it never provisions.

lbonn: Yes, you're right. Just saw the typo...

PR #1406, opened 2019-10-11
Fix the CMake SQL schema generation rule

lbonn:
It should always run, as the script itself avoids useless rewrites.

Was broken since ca3c1ed7a30fda5ffe5b3dff716d5cb8b9f66806


src/libaktualizr/storage/embed_schemas.py
```diff
@@ -37,43 +38,57 @@ def apend_migration(migration_path, header_file):
...
 
-    pattern = re.compile("//Latest schema timestamp: (\d+)")
+    pattern = re.compile(r"// created on: (\d+)")
     if os.path.exists(schemas_header):
         for i, line in enumerate(open(schemas_header)):
             for match in re.finditer(pattern, line):
                 if int(match.groups()[0]) == int(max_file_stamp):
                     # header up to date, exiting
                     sys.exit(0)
+    """
```
patrickvacek: Why leave this commented out?

patrickvacek: Lol, beat me to it.

lbonn: :)

lbonn:
> It's too bad we have to always re-run the script, but this seems like a decent compromise in preventing the problems we've seen.

The problem is that the dependency list is not fixed. The alternative would be to list all the source files we use in `config/sql` but it's error-prone and tedious.

And it takes 50ms on my machine when there is nothing to change, so the only (minor) issue is noise on stdout.

PR #1408, opened 2019-10-11
Create System Requirements

MatthiasBerchner:
Compare 
* https://confluence.in.here.com/display/OTA/_Customer+Technical+Questionaire+for+Pre-Sales
* https://confluence.in.here.com/pages/viewpage.action?pageId=821264693
Would put it under "Reference"

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/System Requirements
+++ b/docs/ota-client-guide/modules/ROOT/pages/System Requirements
@@ -1,0 +1,138 @@
+= System Requirements 
+ifdef::env-github[]
+
+[NOTE]
+====
+We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}.html[view this article in our documentation portal]. Not all of our articles render correctly in GitHub.
+====
+endif::[]
+
+:page-partial:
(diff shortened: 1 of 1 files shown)


tkfu:
Several bookkeeping changes needed:

1. All docs-only PRs should be from a branch beginning with `docs/`. This is for CI.
2. All docs-only commit messages should start with `DOCS:`, for ease of reading in the commit history.
3. Docs changes should be merged into the currently active docs branch, not master, unless they apply specifically to changes in the aktualizr codebase that are not yet present in the latest release.
4. New pages shouldn't be added to the docs if they aren't in the navigation; adding undisoverable pages is a bad practice. If there's a really strong reason for it we can discuss offline, but in general if you want to add content you need to consider where it fits in the information architecture of the site, and integrate it with existing content. We don't just throw information up there willy-nilly.

I haven't started looking at the content, but I'm going to close this and ask you to open a new one with those corrections so that we can start to review it.

patrickvacek:
FYI see also https://docs.ota.here.com/ota-client/latest/supported-boards.html, which already has a few of the points you are interested in.

PR #1414, opened 2019-10-15
Use `git describe --long` for versions

lbonn:
So we'll have 2019.8-0-g5443f810 instead of just 2019.8


lbonn:
> Seems legit, but do we also need to change meta-updater to work around `-DAKTUALIZR_VERSION=${PV}`?

Yes this PR is a step in this direction as I'm starting to think that the easiest way would just be to remove this bit in meta-updater. It was added before aktualizr had the git version detection procedure, it's probably obsolete now.

PR #1416, opened 2019-10-16
OTA-3716: Full file download If a byte range request fails

mike-sul:
Fallback to a full file download if a byte range request fails (a server does not support it).

Ideally, this logic should be "behind" HttpInterface interface and encapsulated in HttpClient implementation of this interface. Unfortunately, there is an additional user context for the download process that is owned by a client (DownloadMetaStruct in packagemanagerinterface.cc)  which has to be managed/reset in case of download retry and which is opaque within HttpClient.

### DIFF
--- a/src/libaktualizr/package_manager/packagemanagerinterface.cc
+++ b/src/libaktualizr/package_manager/packagemanagerinterface.cc
@@ -96,18 +96,18 @@ bool PackageManagerInterface::fetchTarget(const Uptane::Target& target, Uptane::
       LOG_WARNING << "Skipping download of target with length 0";
       return true;
     }
-    DownloadMetaStruct ds(target, std::move(progress_cb), token);
+    std::unique_ptr<DownloadMetaStruct> ds = std_::make_unique<DownloadMetaStruct>(target, progress_cb, token);
     if (exists == TargetStatus::kIncomplete) {
       auto target_check = storage_->checkTargetFile(target);
-      ds.downloaded_length = target_check->first;
+      ds->downloaded_length = target_check->first;
       auto target_handle = storage_->openTargetFile(target);
-      ::restoreHasherState(ds.hasher(), target_handle.get());
+      ::restoreHasherState(ds->hasher(), target_handle.get());
       target_handle->rclose();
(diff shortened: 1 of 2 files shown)


src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -96,18 +96,19 @@ bool PackageManagerInterface::fetchTarget(const Uptane::Target& target, Uptane::
       LOG_WARNING << "Skipping download of target with length 0";
       return true;
     }
-    DownloadMetaStruct ds(target, std::move(progress_cb), token);
+    std::unique_ptr<DownloadMetaStruct> ds{new DownloadMetaStruct(target, std::move(progress_cb), token)};
```
patrickvacek: `make_unique`?

mike-sul: Yeah, it makes sense to make it unique, however, I have to use the custom make_unique (std_::make_unique) since we don't support c++14 by default.

src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -96,18 +96,19 @@ bool PackageManagerInterface::fetchTarget(const Uptane::Target& target, Uptane::
       LOG_WARNING << "Skipping download of target with length 0";
       return true;
     }
-    DownloadMetaStruct ds(target, std::move(progress_cb), token);
+    std::unique_ptr<DownloadMetaStruct> ds = std_::make_unique<DownloadMetaStruct>(target, progress_cb, token);
+    assert(ds);
```
lbonn: This assert doesn't seem useful here. make_unique should throw an exception if it fails.

Even then, assert would no be compiled in release builds and that's exactly where we'd want to catch allocation failures.

mike-sul: Absolutely agree, just another habit.

mike-sul:
can it be merged ?

lbonn:
Sure :)

PR #1419, opened 2019-10-16
Factor out the get_version logic

lbonn:
DRY


scripts/build_ubuntu.sh
```diff
@@ -17,4 +17,4 @@ mkdir -p "$TEST_INSTALL_DESTDIR"
 LDFLAGS="-s" "$GITREPO_ROOT/scripts/test.sh"
 
 git -C "$GITREPO_ROOT" fetch --tags --unshallow || true
-git -C "$GITREPO_ROOT" describe --long > "$TEST_INSTALL_DESTDIR/aktualizr-version"
+"$GITREPO_ROOT/scripts/get_version.sh" > "$TEST_INSTALL_DESTDIR/aktualizr-version"
```
patrickvacek: Do you still need to provide the git repo path here for this to work?

lbonn: It's computed from the scripts' directory on line 6 if left undefined.

lbonn: It's computed from the script's current directory if left undefined, on line 6.

lbonn: Ok I see. I've added a fix for that.

Also made `make_src_archive.sh` smarter.

scripts/make_src_archive.sh
```diff
@@ -1,24 +1,25 @@
...
 python3 -m venv venv
 
 . venv/bin/activate
 
-pip install 'git_archive_all==1.19.4'
-
-ABS_OUT=$(realpath "$OUTPUT")
+# Use our fork which makes it easier to add extra files
+# pip install 'git_archive_all==1.19.4'
+pip install git+https://github.com/lbonn/git-archive-all.git
```
patrickvacek: Is this something that you can upstream? Alternatively would be better to have in the same namespace (advancedtelematic).

lbonn: - Yes https://github.com/Kentzo/git-archive-all/pull/73
- ok I can do that. Initially thought it was not worth the noise

patrickvacek: > ok I can do that. Initially thought it was not worth the noise

Nah, don't worry about it for now if you've already got the PR to merge upstream. It's fine for now.

PR #1420, opened 2019-10-17
Limit number of HTTP redirects to 10

Zee314159:
Draft code

### DIFF
--- a/src/libaktualizr/http/httpclient.cc
+++ b/src/libaktualizr/http/httpclient.cc
@@ -230,6 +230,7 @@ std::future<HttpResponse> HttpClient::downloadAsync(const std::string& url, curl
   curlEasySetoptWrapper(curl_download, CURLOPT_URL, url.c_str());
   curlEasySetoptWrapper(curl_download, CURLOPT_HTTPGET, 1L);
   curlEasySetoptWrapper(curl_download, CURLOPT_FOLLOWLOCATION, 1L);
+  curlEasySetoptWrapper(curl_download, CURLOPT_MAXREDIRS, 10L);
   curlEasySetoptWrapper(curl_download, CURLOPT_WRITEFUNCTION, write_cb);
   curlEasySetoptWrapper(curl_download, CURLOPT_WRITEDATA, userp);
   if (progress_cb != nullptr) {
--- a/tests/test_backend_failure.py
+++ b/tests/test_backend_failure.py
@@ -202,6 +202,25 @@ def test_backend_failure_sanity_customrepo_update_redirect(aktualizr, uptane_rep
     return install_result and update_hash == aktualizr.get_current_image_info(aktualizr.id)
 
 """
+    Verifies if aktualizr rejects redirects over 10 times - update fails after redirect
+"""
+@with_uptane_backend(start_generic_server=True)
+@with_customrepo(handlers=[
+                            RedirectHandler(number_of_redirects=(11 * 3 + 1), url='/primary-image.img')
+                        ])
+@with_imagerepo()
+@with_director()
+@with_aktualizr(start=False, run_mode='once', output_logs=True)
+def test_backend_failure_sanity_customrepo_unsuccessful_update_redirect(aktualizr, uptane_repo,
+                                                           custom_repo, director, **kwargs):
+    update_hash = uptane_repo.add_image(aktualizr.id, 'primary-image.img',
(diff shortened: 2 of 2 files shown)


src/libaktualizr/http/httpclient.cc
```diff
@@ -53,6 +53,9 @@ HttpClient::HttpClient(const std::vector<std::string>* extra_headers) {
 
   curlEasySetoptWrapper(curl, CURLOPT_VERBOSE, get_curlopt_verbose());
 
+  curlEasySetoptWrapper(curl, CURLOPT_FOLLOWLOCATION, 1L);
+  curlEasySetoptWrapper(curl, CURLOPT_MAXREDIRS, 10L);
```
mike-sul: This curl param should be set in HttpClient::downloadAsync(). But won't work anyway because aktualizr/package_manager tries to download it three times.

tests/test_backend_failure.py
```diff
@@ -201,6 +201,24 @@ def test_backend_failure_sanity_customrepo_update_redirect(aktualizr, uptane_rep
...
+@with_director()
+@with_aktualizr(run_mode='once', output_logs=True)
+def test_backend_failure_sanity_customrepo_unsuccessful_update_redirect(aktualizr, uptane_repo,
+                                                           custom_repo, director, **kwargs):
+    update_hash = uptane_repo.add_image(aktualizr.id, 'primary-image.img',
+                                        custom_url=custom_repo.base_url + '/' + 'primary-image.img')
+    install_result = director.wait_for_install()
+    return not install_result
+
+
```
mike-sul: The test should look like this

`"""
    Verifies if aktualizr rejects redirects over 10 times - update fails after redirect
"""
@with_uptane_backend(start_generic_server=True)
@with_customrepo(handlers=[
                            RedirectHandler(number_of_redirects=(11 * 3 + 1), url='/primary-image.img')
                        ])
@with_imagerepo()
@with_director()
@with_aktualizr(start=False, run_mode='once', output_logs=True)
def test_backend_failure_sanity_customrepo_unsuccessful_update_redirect(aktualizr, uptane_repo,
                                                           custom_repo, director, **kwargs):
    update_hash = uptane_repo.add_image(aktualizr.id, 'primary-image.img',
                                        custom_url=custom_repo.base_url + '/' + 'primary-image.img')
    with aktualizr:
        aktualizr.wait_for_completion()

    return not director.get_install_result()`

A number of requests should be equal to (CURL_MAX_REDIRECTS + 1) * (RETRY_NUMB) + 1, which is in this specific case is (10+1)*3 + 1 (An initial request to the mock repo server is not considered as a redirect request by curl so it starts counting only from the second request, hence +1).

Zee314159: Thanks, @mike-sul . I'll have a try.

src/libaktualizr/http/httpclient.cc
```diff
@@ -241,6 +241,8 @@ std::future<HttpResponse> HttpClient::downloadAsync(const std::string& url, curl
   curlEasySetoptWrapper(curl_download, CURLOPT_LOW_SPEED_TIME, speed_limit_time_interval_);
   curlEasySetoptWrapper(curl_download, CURLOPT_LOW_SPEED_LIMIT, speed_limit_bytes_per_sec_);
   curlEasySetoptWrapper(curl_download, CURLOPT_RESUME_FROM_LARGE, from);
+  curlEasySetoptWrapper(curl_download, CURLOPT_FOLLOWLOCATION, 1L);
```
mike-sul: It's redundant, see line # 232.

doanac:
Just a nitpick, but could you change the commit message to "Limit number of HTTP redirects to 10". Just saves someone from having to read the PR so closely.

tests/test_backend_failure.py
Zee314159: @mike-sul , the reason of using `aktualizr.wait_for_completion()` instand of `director.wait_for_install()` is "limitation of a redirect number should be aktualizr-wide and applied to each outgoing request.", right?

mike-sul: This is a negative test, an installation of a new update is not supposed to happen since a number of redirects exceeds the maxim allowed. Therefore, director.wait_for_install() will never succeed (installation should not happen) and exit with a timeout error. So, the test just simply wait until Aktualizr completes an uptane cycle.

PR #1423, opened 2019-10-21
aktualizr-info: add --version parameter.

patrickvacek:


src/aktualizr_info/main.cc
```diff
@@ -31,13 +33,25 @@ static int loadAndPrintDelegations(const std::shared_ptr<INvStorage> &storage) {
   return EXIT_SUCCESS;
 }
 
+void check_info_options(const bpo::options_description &description, const bpo::variables_map &vm) {
```
Zee314159: Looks good to me. I have a question though with the function name, do we have any rules to comply with to name the functions or variables? Like where or when do we use the type `example_test_func` or `exampleTestFunc` ?

patrickvacek: Thanks for noticing that; I should change it. The guidelines are in https://github.com/advancedtelematic/aktualizr/blob/master/CONTRIBUTING.md#code-quality-and-style, which references the [Google style guide for C++](https://google.github.io/styleguide/cppguide.html). According to the section on [function names](https://google.github.io/styleguide/cppguide.html#Function_Names), we should be doing `ExampleTestFunc`, but in practice we do `exampleTestFunc`, which is probably close enough. In any case, this function I've copied here is misnamed, and I will fix it.

patrickvacek: Fixed. BTW @xcheng-here your review was automatically dismissed when I pushed new commits.

PR #1387, opened 2019-09-24
Feat/c api updates

mchekhovoi:
OTA-3212 SendDeviceData()/SendManifest()/OpenStoredTarget() C APIs

PR #1425, opened 2019-10-22
OTA-3764 Extend uptane-generator tool and fake_test_server.py to supp…

mchekhovoi:
…ort campaigns

### DIFF
--- a/src/libaktualizr/campaign/campaign.h
+++ b/src/libaktualizr/campaign/campaign.h
@@ -31,16 +31,7 @@ static inline Cmd cmdFromName(const std::string &name) {
 class Campaign {
  public:
   Campaign() = default;
-  Campaign(std::string idIn, std::string nameIn, int64_t sizeIn, bool autoAcceptIn, std::string descriptionIn,
-           int estInstallationDurationIn, int estPreparationDurationIn)
-      : id(std::move(idIn)),
-        name(std::move(nameIn)),
-        size(sizeIn),
-        autoAccept(autoAcceptIn),
-        description(std::move(descriptionIn)),
-        estInstallationDuration(estInstallationDurationIn),
-        estPreparationDuration(estPreparationDurationIn) {}
-  static Campaign fromJson(const Json::Value &json);
+  Campaign(const Json::Value &json);
   void getJson(Json::Value &out) const;
 
   std::string id;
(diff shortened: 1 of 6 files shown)


src/uptane_generator/repo.cc
```diff
@@ -221,6 +222,27 @@ void Repo::generateRepo(KeyType key_type) {
   if (repo_type_ == Uptane::RepositoryType::Director()) {
     Utils::writeFile(path_ / DirectorRepo::dir / "manifest", std::string());  // just empty file to work with put method
   }
+
+  generateCampaigns();
```
patrickvacek: It looks like this will cause a campaign to get generated for every user of uptane-generator, regardless of whether they want it. I don't think that's ideal. If we want to create campaign metadata, we need a separate command for it.

src/libaktualizr/campaign/campaign.h
```diff
@@ -30,18 +30,30 @@ static inline Cmd cmdFromName(const std::string &name) {
 // Out of uptane concept: update campaign for a device
 class Campaign {
  public:
+  Campaign() = default;
+  Campaign(std::string idIn, std::string nameIn, int64_t sizeIn, bool autoAcceptIn, std::string descriptionIn,
```
patrickvacek: What is this constructor being used for? If it's really needed, it seems like it might be easier to provide a constructor with json input instead.

src/uptane_generator/repo.cc
mchekhovoi: Agree. I will fix it.

src/libaktualizr/campaign/campaign.h
eu-smirnov: Yep, also not clear for me. And it looks like the static fromJson method can rather be a ctor.

eu-smirnov:
merged into #1387

src/libaktualizr/campaign/campaign.h
lbonn: I usually don't like to put deserialization methods in pure constructors as `fromJson` seems clearer to me. Once another mean of creating an object is needed, there would be no risk of confusion with a simple constructor.

But that's a minor style preference, the constructor solution seems fine here.

PR #1428, opened 2019-10-30
Update mininum RAM requirements.

patrickvacek:
Also fix a syntax error and specify rough estimate of general RAM usage.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/supported-boards.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/supported-boards.adoc
@@ -39,12 +39,12 @@ Your images will also need network connectivity to be able to reach an actual OT
 
 == Minimum hardware requirements for controllers (ECUs)
 
-The aktualizr binary is a lightweight {cpp} application between 2-5 MB in size. It uses a minimum amount of resources when idle.
+The aktualizr binary is a lightweight {cpp} application between 2-5 MB in size. It uses around 20 MB RAM, although some of that comes from shared libraries. It uses a minimum amount of other resources when idle.
 
 The following hardware is required for your primary ECU:
 
-* At a minimum, aktualizr needs **16 MB of RAM** and **128 MB of storage** to run.
-* We recommend that you use a controller with **128 MB of RAM** and **512 MB storage** -- especially if you if your plan to process large, complex sofware updates.
+* At a minimum, the board or device aktualizr is running on will need **64 MB of RAM** and **128 MB of storage**.
+* We recommend using a board with slightly higher specs, however. **256 MB of RAM** and **512 MB storage** are reasonable targets, especially if you plan to process large, complex software updates. The aktualizr process itself won't consume significantly more RAM with larger updates, but the filesystem cache memory usage might increase.
 
 [NOTE]
 ====


docs/ota-client-guide/modules/ROOT/pages/supported-boards.adoc
```diff
@@ -39,12 +39,12 @@ Your images will also need network connectivity to be able to reach an actual OT
...
 == Minimum hardware requirements for controllers (ECUs)
 
-The aktualizr binary is a lightweight {cpp} application between 2-5 MB in size. It uses a minimum amount of resources when idle.
+The aktualizr binary is a lightweight {cpp} application between 2-5 MB in size. It uses around 20 MB RAM, although some of that comes from shared libraries. It uses a minimum amount of other resources when idle.
 
 The following hardware is required for your primary ECU:
 
-* At a minimum, aktualizr needs **16 MB of RAM** and **128 MB of storage** to run.
-* We recommend that you use a controller with **128 MB of RAM** and **512 MB storage** -- especially if you if your plan to process large, complex sofware updates.
+* At a minimum, aktualizr needs **64 MB of RAM** and **128 MB of storage** to run.
```
eu-smirnov: Sorry that didn't notice earlier, but I think the phrasing here is a bit confusing. Should we better say "At a minimum, system running aktualizr..."?

tkfu: I agree.
```suggestion
* At a minimum, the board or device aktualizr is running on will need **64 MB of RAM** and **128 MB of storage**.
```

docs/ota-client-guide/modules/ROOT/pages/supported-boards.adoc
```diff
@@ -39,12 +39,12 @@ Your images will also need network connectivity to be able to reach an actual OT
...
+* We recommend that you use a controller with **128 MB of RAM** and **512 MB storage** -- especially if you plan to process large, complex sofware updates.
```
tkfu:
```suggestion
* We recommend using a board with slightly higher specs, however. **128 MB of RAM** and **512 MB storage** are reasonable targets, especially if you plan to process large, complex software updates.
```

eu-smirnov:
As Patrick is on a vacation for the next few days, I'll takeover this PR. I hope he doesn't mind it.

eu-smirnov:
I'll also raise our recommended RAM requirements to 256 MB, as it seems more reasonable if we are talking about large and complex updates.

tkfu:
> I'll also raise our recommended RAM requirements to 256 MB, as it seems more reasonable if we are talking about large and complex updates.

Is it necessarily the case? What kinds of thing would cause the ram usage to be so dramatically higher?

eu-smirnov:
> Is it necessarily the case? What kinds of thing would cause the ram usage to be so dramatically higher?

It's not that the RAM usage of aktualizr itself will be higher, but working with lager files will certainly eat up more RAM for the caches. Depending of what is done next with the downloaded updates, more space might be needed for additional libraries, etc. 
I think 128 MB is a pretty low number by modern standards, we usually test/demo on systems that have 1-2 GB of RAM.

eu-smirnov:
@tkfu I'm also a bit confused about the target branch. Do we first merge this PR to docs branch and docs branch to master after?

tkfu:
> It's not that the RAM usage of aktualizr itself will be higher, but working with lager files will certainly eat up more RAM for the caches. 

Why? Isn't the size of the download buffer independent of the size of the file? Are we loading the entire file into ram when we hash it or something?
 
> Depending of what is done next with the downloaded updates, more space might be needed for additional libraries, etc.

Sure, but that's a bit outside the scope of aktualizr's system requirements. If we want to make further recommendations based on speculation about what the needs of other libraries or systems might integrate with libaktualizr, we should make it more clear that's what we're doing. I don't think we need to go down that road, though.

> I think 128 MB is a pretty low number by modern standards, we usually test/demo on systems that have 1-2 GB of RAM.

Some companies are extremely sensitive about BOM costs, and potential target ECUs might include TCUs that are basically glorified modems.

tkfu:
> @tkfu I'm also a bit confused about the target branch. Do we first merge this PR to docs branch and docs branch to master after?

See https://github.com/advancedtelematic/adr/blob/master/0007-aktualizr-docs-release-process.md. This branch should be rebased on 2019.9-docs before merging; the `{version}-docs` branches shouldn't have any commits except for docs commits. At some point before the 2019.10 release, 2019.9-docs will be merged into master, but there's no particular need to do it right away.

eu-smirnov:
> Why? Isn't the size of the download buffer independent of the size of the file? Are we loading the entire file into ram when we hash it or something?
We don't. I'm talking here about the filesystem cache, we have no control over it.
> Some companies are extremely sensitive about BOM costs, and potential target ECUs might include TCUs that are basically glorified modems.
That's why we have the minimum requirements, IMO. 
> Sure, but that's a bit outside the scope of aktualizr's system requirements. If we want to make further recommendations based on speculation about what the needs of other libraries or systems might integrate with libaktualizr, we should make it more clear that's what we're doing. I don't think we need to go down that road, though.
That's a fair point. I just don't want us to send a false message, leading the user to a situation when running aktualizr is pretty much all the system can do. But if you think that this number can scare some potential users out, I'll change it back to 128.

We will also need to revisit the storage numbers at some point. In the current state they don't make much sense to me. I think we should specify the space required by aktualizr + libraries and maximum stored metadata size. The rest will be defined by the size of updates, which user should know better.

docs/ota-client-guide/modules/ROOT/pages/supported-boards.adoc
```diff
@@ -39,12 +39,12 @@ Your images will also need network connectivity to be able to reach an actual OT
...
 
-The aktualizr binary is a lightweight {cpp} application between 2-5 MB in size. It uses a minimum amount of resources when idle.
+The aktualizr binary is a lightweight {cpp} application between 2-5 MB in size. It uses around 20 MB RAM, although some of that comes from shared libraries. It uses a minimum amount of other resources when idle.
 
 The following hardware is required for your primary ECU:
 
-* At a minimum, aktualizr needs **16 MB of RAM** and **128 MB of storage** to run.
-* We recommend that you use a controller with **128 MB of RAM** and **512 MB storage** -- especially if you if your plan to process large, complex sofware updates.
+* At a minimum, the board or device aktualizr is running on will need **64 MB of RAM** and **128 MB of storage**.
+* We recommend using a board with slightly higher specs, however. **256 MB of RAM** and **512 MB storage** are reasonable targets, especially if you plan to process large, complex software updates.
```
tkfu:
```suggestion
* We recommend using a board with slightly higher specs, however. **256 MB of RAM** and **512 MB storage** are reasonable targets, especially if you plan to process large, complex software updates. The aktualizr process itself won't consume significantly more RAM with larger updates, but the filesystem cache memory usage might increase.
```

tkfu:
> > Why? Isn't the size of the download buffer independent of the size of the file? Are we loading the entire file into ram when we hash it or something?
> 
> We don't. I'm talking here about the filesystem cache, we have no control over it.

Ah, got it. Thanks.

> > Some companies are extremely sensitive about BOM costs, and potential target ECUs might include TCUs that are basically glorified modems.
> 
> That's why we have the minimum requirements, IMO.

Point taken.

> > Sure, but that's a bit outside the scope of aktualizr's system requirements. If we want to make further recommendations based on speculation about what the needs of other libraries or systems might integrate with libaktualizr, we should make it more clear that's what we're doing. I don't think we need to go down that road, though.
> 
> That's a fair point. I just don't want us to send a false message, leading the user to a situation when running aktualizr is pretty much all the system can do. But if you think that this number can scare some potential users out, I'll change it back to 128.

I think, if you're ok with the suggestion I just made, that leaving at 256 is reasonable and won't scare anyone off.

> We will also need to revisit the storage numbers at some point. In the current state they don't make much sense to me. I think we should specify the space required by aktualizr + libraries and maximum stored metadata size. The rest will be defined by the size of updates, which user should know better.

Agreed. We can do that in a future PR, though.

PR #1429, opened 2019-10-31
OTA-3213 Pause()/Resume()/Abort() C APIs

mchekhovoi:
Signed-off-by: Mykola Chekhovoi <ext-mykola.chekhovoi@here.com>

### DIFF
--- a/src/libaktualizr-c/libaktualizr-c.cc
+++ b/src/libaktualizr-c/libaktualizr-c.cc
@@ -183,7 +183,7 @@ int Aktualizr_send_manifest(Aktualizr *a, const char *manifest) {
 
 int Aktualizr_send_device_data(Aktualizr *a) {
   try {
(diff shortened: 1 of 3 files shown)


src/libaktualizr-c/libaktualizr-c.cc
```diff
@@ -226,3 +227,43 @@ int Aktualizr_close_stored_target(StorageTargetHandle *handle) {
...
+      return Pause_Status_C::kError;
+    }
+    default: {
+      assert(false);
+      return Pause_Status_C::kError;
+    }
+  }
+}
+
+static std::mutex queue_wrapper_mutex;
```
eu-smirnov: Not sure, why do we need a mutex here?

mchekhovoi: Fixed.

PR #1430, opened 2019-10-31
Test failure of corrupted ostree repo pull

lbonn:


lbonn:
It's not testing the right thing right now as it returns malformed `.filez` files. It should fail at the checksum step.

mike-sul:
This test is quite similar to this one https://github.com/advancedtelematic/aktualizr/blob/07e38359c12ada3e3973cd97d184c7e034e13daa/tests/test_backend_failure.py#L144

lbonn:
Hm good point. Then I would maybe modify this one to return a valid `filez` file with a wrong checksum.

PR #1432, opened 2019-11-04
PackageConfig: using fake package manager for non-ostree build

kbushgit:
By default, aktualizr will be built with the PackageConfig::kNone, the PackageConfig::kOstree is used only when the -DBUILD_OSTREE is defined.

This fix has no effect on meta-updater.


lbonn:
I'm normally not a huge fan of these ifdefs but maybe it's quite benign in this case. Is that what you had in mind @patrickvacek @eu-smirnov?

patrickvacek:
I think this is fine, but I'm wondering if there's a good way to test this at a low level. We generally avoid trying to put too many `#ifdef`s in the code, so I wonder if it's worth abstracting this somehow. But this is small and simple, so maybe it's fine as is. I encourage input from others, but in any case it's probably fine to merge as is and re-evaluate later if no good ideas emerge in the meantime.

PR #1434, opened 2019-11-04
sotauptaneclient: Make reportHwInfo smarter

doanac:
Make the reportHwInfo call be like reportNetworkInfo and only send
things when they've changed.


patrickvacek:
Nice, thanks! I was just recently talking about doing this; didn't realize it'd be so easy.

mike-sul:
The report will be resend after aktualizr or a device restart .

patrickvacek:
> The report will be resend after aktualizr or a device restart .

Yep, I think that's just fine for now.

PR #1435, opened 2019-11-04
dockerappmanager: Handle removed apps

doanac:
There's a bug where:

 1) sota.toml is configured with 2 docker apps: "app1, app2"
 2) update is applied, so we are now running both app1 and app2
 3) sota.toml is updated with 1 docker app: "app1"

At this point the current logic will leave that app running forever.
This change finds apps that have been removed, stops them, and removes
them from the filesystem.


doanac:
Sorry guys, bad cherry-pick by me. force-push'd with a fix.

doanac:
I think this CI failure is unrelated to my change

PR #1436, opened 2019-11-04
httpclient: Expose method to update headers

doanac:
We have a custom client that needs to update HTTP headers after the
object has been instantiated. This shouldn't touch any current code
paths in aktualizr itself.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>


mike-sul:
Perhaps adding/removing headers after the httpclient instantiation will be required at some point too.

PR #1437, opened 2019-11-08
Remove the systemd socket activation functionality

mike-sul:


lbonn:
You should be able to remove `BUILD_SYSTEMD` and everything related.

mike-sul:
> You should be able to remove `BUILD_SYSTEMD` and everything related.

Yeah, I hope so :)

thirdparty.spdx
```diff
@@ -208,19 +208,6 @@ FilesAnalyzed: false
 PackageComment: <text>Dynamically linked.</text>
 
 
-PackageName: systemd
```
lbonn: There is related line, close to the top of this file:

```Relationship: SPDXRef-DOCUMENT DESCRIBES SPDXRef-systemd```

mike-sul:
> Looks good, but I still see a reference to `socket_activation` in `docs/doxygen/Doxyfile.in`. And do we still need to keep around the `SocketServer` class in `src/aktualizr_secondary/socket_server.h`?

- removed the `socket_activation` reference from the doxygen file.

- Yes, that's actually a TCP server implementation that the IP secondary relies on.

PR #1438, opened 2019-11-08
repo manifest upload

lbonn:
I don't really like the new third party libraries, including another json support...

Will ideally be rewritten by using a well supported xml library and our jsoncpp.


doanac:
This looks really interesting. Can you elaborate on the "repo manifest"? ie - are you talking about the OE android-repo default.xml type manifest file, something entirely different, or is this sort of a free-form thing to (ab)use?

src/sota_tools/garage_push.cc
```diff
@@ -147,6 +151,27 @@ int main(int argc, char **argv) {
         return EXIT_FAILURE;
       }
     }
+
+    if (manifest_path != "") {
+      std::string manifest_json;
+      manifest_json = xml2json(Utils::readFile(manifest_path).c_str());
```
patrickvacek: I'd originally assumed we could just make a call to `repo manifest -r` from the shell, but did you have a particular motivation to read it from file? Presumably in meta-updater you'd just reuse the file created by the manifest class?

patrickvacek:
> Can you elaborate on the "repo manifest"? ie - are you talking about the OE android-repo default.xml type manifest file, something entirely different, or is this sort of a free-form thing to (ab)use?

@doanac Yes, this is for the Android repo tool, but in theory, I think you could (ab)use it for anything you want to upload like that that might be useful. The idea here is that we want to have a better idea of what was used to build the images we store, which makes tracking bugs in the field easier.

src/sota_tools/garage_push.cc
mike-sul: My understanding is that `xml2json` can throw an exception, if it is correct then such exception won't be caught.

mike-sul:
I  suppose, we could convert the xml to json by means of boost that aktualizr already depends on.

src/sota_tools/garage_push.cc
lbonn: > Presumably in meta-updater you'd just reuse the file created by the manifest class?

Yes, that's what I opted to do. In general it's more flexible to pass a complete path, instead of depending on being able to execute `repo` from where garage-push is launched.

lbonn:
> Can we use a git submodule or just add a new dependency instead of copying the files into our repo? I want to avoid that as much as we can going forward.

Yes I can try to find a better solution. I'll maybe just reimplement the thing with boost as @mike-sul suggested, if possible. Or lxml something, if not.

src/sota_tools/garage_push.cc
patrickvacek: Makes sense the more I think about it. Thanks!

lbonn:
Replaced by #1440

PR #1440, opened 2019-11-11
repo manifest upload (reworked)

lbonn:
New version of #1438, with a custom written conversion routine.

### DIFF
--- a/src/libaktualizr/utilities/CMakeLists.txt
+++ b/src/libaktualizr/utilities/CMakeLists.txt
@@ -17,7 +17,8 @@ set(HEADERS apiqueue.h
             sockaddr_io.h
             timer.h
             types.h
-            utils.h)
+            utils.h
+            xml2json.h)
 
 set_property(SOURCE aktualizr_version.cc PROPERTY COMPILE_DEFINITIONS AKTUALIZR_VERSION="${AKTUALIZR_VERSION}")
 
@@ -29,5 +30,6 @@ add_aktualizr_test(NAME timer SOURCES timer_test.cc)
 add_aktualizr_test(NAME types SOURCES types_test.cc)
 add_aktualizr_test(NAME utils SOURCES utils_test.cc PROJECT_WORKING_DIRECTORY)
 add_aktualizr_test(NAME sighandler SOURCES sighandler_test.cc)
+add_aktualizr_test(NAME xml2json SOURCES xml2json_test.cc)
 
 aktualizr_source_file_checks(${SOURCES} ${HEADERS} ${TEST_SOURCES})
--- a/src/libaktualizr/utilities/xml2json.h
+++ b/src/libaktualizr/utilities/xml2json.h
@@ -1,0 +1,99 @@
+#include <boost/optional.hpp>
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/xml_parser.hpp>
+#include <istream>
+
+#include "json/json.h"
+
+namespace xml2json {
+
+static inline void addSubArray(Json::Value &d, const std::string &key, const Json::Value &arr) {
+  if (arr.size() == 0) {
+    return;
+  } else if (arr.size() == 1) {
+    d[key] = arr[0];
+  } else {
+    d[key] = arr;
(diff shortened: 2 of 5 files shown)


doanac:
Two small things you might want to consider. We've encountered both at Foundries and have resorted to always using the repo tool to tell us the content of the manifest:

1) You can have more than one .xml file. In fact, that's now the way we tell people to build against our LMP. Take our manifest, then create your own "overrides" file which can customize your own layers.
2) Not everyone uses a "pinned manifest", so the contents of the file aren't always precise. You can have repo generate a pinned manifest with something like `repo manifest -r -o $archive/manifest.pinned.xml`

I suppose the solution to the issue I'm raising is to simply generate this file myself and then pass that to this command. However, it might make things easier for users if this command could do the dirty work for them.

lbonn:
> Two small things you might want to consider. We've encountered both at Foundries and have resorted to always using the repo tool to tell us the content of the manifest:
> 
>     1. You can have more than one .xml file. In fact, that's now the way we tell people to build against our LMP. Take our manifest, then create your own "overrides" file which can customize your own layers.
> 
>     2. Not everyone uses a "pinned manifest", so the contents of the file aren't always precise. You can have repo generate a pinned manifest with something like `repo manifest -r -o $archive/manifest.pinned.xml`
> 
> 
> I suppose the solution to the issue I'm raising is to simply generate this file myself and then pass that to this command. However, it might make things easier for users if this command could do the dirty work for them.

If I understand your point correctly, this is already what we are planning to do. The input manifest is generated in this bbclass before being passed to garage-push: https://github.com/advancedtelematic/meta-updater/blob/21a463dc264434dc132ae79526a824e7955e0032/classes/image_repo_manifest.bbclass#L17.

Nevertheless, I hadn't thought about a multi-manifest case. That sounds interesting and we might recommend that as well if it works as expected.

doanac:
>  https://github.com/advancedtelematic/meta-updater/blob/21a463dc264434dc132ae79526a824e7955e0032/classes/image_repo_manifest.bbclass#L17.

Thanks for the pointer!

patrickvacek:
> You can have more than one .xml file. In fact, that's now the way we tell people to build against our LMP. Take our manifest, then create your own "overrides" file which can customize your own layers.

Interesting, can you share a link or explain how that works? And is it possible to merge the manifests in some way for export, or how would you suggest keeping track of that?

doanac:
> Interesting, can you share a link or explain how that works? 

The way we manage it is like this: Each custom forks our [lmp-manifes](https://github.com/foundriesio/lmp-manifest/). They then create a their own "overrides" file. For example I have something like this:
~~~
# andy-corp.xml
<manifest>
  <include name="default.xml" />

  <remote fetch="https://source.foundries.io/partners/andy-corp"
          name="subscriber-overrides" />
  <remote fetch="https://github.com/doanac"
          name="gh" />

  <remove-project name="meta-lmp" />
  <project name="meta-lmp" path="layers/meta-lmp" remote="gh" revision="8608f34dc2c9255839466e73970445e03f220efe" />

  <project name="meta-subscriber-overrides"
           path="layers/meta-subscriber-overrides"
           remote="subscriber-overrides"/>
</manifest>
~~~

To initialize my repository I then run something like: `repo init -u https://source.foundries.io/factories/andy-corp/lmp-manifest.git -m andy-corp.xml`

This is super handy because you can see that I can override and add things to the common platform to make it "my own". And the `repo manifest -r -o $archive/manifest.pinned.xml` understands that and produces a single combined file reduced to the proper content.

lbonn:
Fixed formatting + small lint and added some basic exception handling.

patrickvacek:
> This is super handy because you can see that I can override and add things to the common platform to make it "my own". And the `repo manifest -r -o $archive/manifest.pinned.xml` understands that and produces a single combined file reduced to the proper content.
> 
> The huge benefit is that its pretty easy to merge our changes back into a forked repo without merge conflicts since they are doing thing in their own file.

Cool, thanks! And as I was hoping, if `repo manifest -o` understands it, then it will be compatible with our manifest upload strategy.

PR #1444, opened 2019-11-13
Fix a use-after-move in Aktualizr()

lbonn:
This worked fine on x86-64 but causes fatal issues on ARM/raspberrypi


ricardosalveti:
Interesting, but wonder what would make it behave differently on ARM, there is nothing special here.

lbonn:
> Interesting, but wonder what would make it behave differently on ARM, there is nothing special here.

I haven't looked at the assembly but since it's some kind of undefined behavior, it may manifest differently depending on the compiler, the target architecture, compilation options...

PR #1445, opened 2019-11-13
Remove/correct some bogus openssl includes.

patrickvacek:


patrickvacek:
Travis is being painfully slow and obtuse. Tests are passing in both CIs, though, so I'm merging this.

PR #1449, opened 2019-11-15
OTA-3969: Do the full verification of metadata on the IP Secondary

mike-sul:

### DIFF
--- a/src/aktualizr_secondary/CMakeLists.txt
+++ b/src/aktualizr_secondary/CMakeLists.txt
@@ -53,6 +53,11 @@ list(INSERT TEST_LIBS 0 aktualizr_secondary_static_lib)
 add_aktualizr_test(NAME aktualizr_secondary_config
                    SOURCES aktualizr_secondary_config_test.cc PROJECT_WORKING_DIRECTORY)
 
+add_aktualizr_test(NAME aktualizr_secondary_uptane_verification
+                   SOURCES uptane_verification_test.cc
+                   ARGS "$<TARGET_FILE:uptane-generator>"
+                   PROJECT_WORKING_DIRECTORY)
+
 add_aktualizr_test(NAME aktualizr_secondary_update
                    SOURCES update_test.cc
                    ARGS ${PROJECT_BINARY_DIR}/ostree_repo PROJECT_WORKING_DIRECTORY)
--- a/src/aktualizr_secondary/CMakeLists.txt
+++ b/src/aktualizr_secondary/CMakeLists.txt
@@ -4,6 +4,7 @@ set(AKTUALIZR_SECONDARY_LIB_SRC
     aktualizr_secondary.cc
     aktualizr_secondary_config.cc
     aktualizr_secondary_common.cc
+    aktualizr_secondary_metadata.cc
     socket_server.cc
     )
 
@@ -40,6 +41,7 @@ set(ALL_AKTUALIZR_SECONDARY_HEADERS
     aktualizr_secondary_interface.h
     aktualizr_secondary_config.h
     aktualizr_secondary_common.h
+    aktualizr_secondary_metadata.h
     socket_server.h
     )
 
--- a/src/aktualizr_secondary/aktualizr_secondary.cc
+++ b/src/aktualizr_secondary/aktualizr_secondary.cc
@@ -132,6 +132,13 @@ bool AktualizrSecondary::sendFirmwareResp(const std::shared_ptr<std::string>& fi
     return false;
   }
 
+  // This check is not applicable for the ostree case, just for the binary file case(s)
+  if (targetToApply.length() != firmware->length()) {
+    LOG_ERROR << "The target image size specified in metadata " << targetToApply.length()
+              << " does not match actual size " << firmware->length();
+    return false;
+  }
+
   install_res = pacman->install(targetToApply);
   if (install_res.result_code.num_code != data::ResultCode::Numeric::kOk) {
     LOG_ERROR << "Could not install target (" << install_res.result_code.toString() << "): " << install_res.description;
@@ -182,7 +189,7 @@ bool AktualizrSecondary::doFullVerification(const Metadata& metadata) {
 
   // 1. Load and verify the current time or the most recent securely attested time.
   //
-  //    We trust the time that the given system/OS/ECU provides, In this ECU we trust :)
+  //    We trust the time that the given system/OS/ECU provides, In ECU We Trust :)
   TimeStamp now(TimeStamp::Now());
 
   // 2. Download and check the Root metadata file from the Director repository, following the procedure in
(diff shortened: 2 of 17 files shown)


patrickvacek:
Nice, maybe this is easier than I thought! Let me know when you're ready for a deeper review. But at first glance it's looking good.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -113,14 +86,17 @@ bool AktualizrSecondary::putRootResp(const std::string& root, bool director) {
 }
 
 bool AktualizrSecondary::sendFirmwareResp(const std::shared_ptr<std::string>& firmware) {
-  if (target_ == nullptr) {
-    LOG_ERROR << "No valid installation target found";
+  auto targetsForThisEcu = director_repo_.getTargets(getSerial());
```
patrickvacek: Should we be rechecking the metadata (without fetching) like we do on the Primary before installation? It seems redundant, but it's a protection against unexpected changes in the time since the metadata was received.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -200,3 +176,60 @@ void AktualizrSecondary::connectToPrimary() {
...
+  // procedure in Section 5.4.4.4.
+  // 4. NOT SUPPORTED: Download and check the Snapshot metadata file from the Director repository, following the
+  // procedure in Section 5.4.4.5.
+  //
+  // 5. Download and check the Targets metadata file from the Director repository, following the procedure in
+  // Section 5.4.4.6. DirectorRepository::updateMeta() method implements this verification step
+  //
+  // The followin steps of the Director's target metadata verification are missing in DirectorRepository::updateMeta()
+  //  6. If checking Targets metadata from the Director repository, verify that there are no delegations.
+  //  7. If checking Targets metadata from the Director repository, check that no ECU identifier is represented more
```
patrickvacek: Can you make a tech debt ticket for this and put it in the robustness epic?

It looks like you've already implemented the check for the secondary; is your proposal to move that check into the Uptane code directly so it is applied universally as early as is reasonable?

src/libaktualizr/uptane/directorrepository.cc
```diff
@@ -96,6 +96,10 @@ bool DirectorRepository::updateMeta(INvStorage& storage, Fetcher& fetcher) {
 
   // Update Director Root Metadata
   {
+    // This procedure doesn't exactly follow the spec:
+    // https://uptane.github.io/uptane-standard/uptane-standard.html#check_root At first, it should try to download N+1
+    // version of the Root metadata file instead of downloading the "latest" one how do we know which version is the
+    // latest and why just root.json is downloaded ?
```
patrickvacek: `root.json` without a number is guaranteed to be the latest in our scheme. The design is to fetch the latest, and if it matches what we already have stored, we're good. If not, then we have to go fetch the missing ones by name/number until we catch up. Might be worth a renewed discussion on what the best method is; I don't see the harm in our method.

src/libaktualizr/uptane/directorrepository.cc
```diff
@@ -108,6 +112,12 @@ bool DirectorRepository::updateMeta(INvStorage& storage, Fetcher& fetcher) {
 
     int local_version = rootVersion();
 
+    // if remote_version <= local_version then the director's root metadata are never verified
+    // which leads to two issues
+    // 1. At initial stage (just after provisioning) the root metadata from 1.root.json are not verified
+    // 2. If local_version becomes higher than 1, e.g. 2 than a rollback attack is possible since the business logic
+    // here won't return any error as suggested in #4 of
+    // https://uptane.github.io/uptane-standard/uptane-standard.html#check_root
```
patrickvacek: Good call. Can you make another tech debt for this? (It's the same logic with the Image repo, so it can all be in one ticket.)

src/libaktualizr/uptane/imagesrepository.cc
```diff
@@ -235,6 +241,11 @@ bool ImagesRepository::updateMeta(INvStorage& storage, Fetcher& fetcher) {
       local_version = -1;
     }
 
+    // I am not sure that #6 of https://uptane.github.io/uptane-standard/uptane-standard.html#check_snapshot is
+    // performed
```
patrickvacek: What's your concern? That we don't match the Targets metadata filenames?

src/aktualizr_secondary/CMakeLists.txt
```diff
@@ -61,7 +68,7 @@ if(BUILD_OSTREE)
                        LIBRARIES uptane_generator_lib
                        LIBRARIES aktualizr-posix
                        ARGS ${PROJECT_BINARY_DIR}/ostree_repo PROJECT_WORKING_DIRECTORY)
-    target_link_libraries(t_aktualizr_secondary_uptane virtual_secondary)
+                       target_link_libraries(t_aktualizr_secondary_uptane virtual_secondary)
```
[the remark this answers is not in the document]
mike-sul: Some kind of typo, will move it back.

src/aktualizr_secondary/aktualizr_secondary.cc
mike-sul: Yes, I agree, we should check at least an image file size (binary case) and its hash (for both ostree and binary).

src/aktualizr_secondary/aktualizr_secondary.cc
mike-sul: Sure, I'll do it.

I did #7, not #6, and yes my proposal to do apply it universally.

src/libaktualizr/uptane/directorrepository.cc
mike-sul: Thanks for explanation.

src/libaktualizr/uptane/directorrepository.cc
mike-sul: Sure.

src/libaktualizr/uptane/imagesrepository.cc
mike-sul: Yes, I mean exactly what #6 says

"Check that each Targets metadata filename listed in the previous Snapshot metadata file is also listed in this"

src/aktualizr_secondary/aktualizr_secondary.cc
patrickvacek: I think it's also worth rechecking the whole metadata chain from the metadata, but yeah, checking the object's hash/size is even more important.

src/aktualizr_secondary/aktualizr_secondary.cc
patrickvacek: Understood. We implicitly do step 6 in that we ignore the concept of delegations from the Director, but it's not a bad idea to check for them and throw a warning if found. Since we don't use the Snapshot role on the Director, though, all we can do is just check for them in the Targets metadata.

src/libaktualizr/uptane/imagesrepository.cc
patrickvacek: This might be another thing worth making a ticket for and looking into more. I'm not sure how critical this really is, though, and I'm also not sure that the backend doesn't make some optimizations that would prevent that from working. In fact, I'm not sure this is actually sensible; what happens if a delegation gets (legitimately) removed?

src/aktualizr_secondary/aktualizr_secondary.cc
mike-sul: https://saeljira.it.here.com/browse/OTA-4118

src/libaktualizr/uptane/imagesrepository.cc
patrickvacek: FYI: https://github.com/uptane/uptane-standard/issues/149. Let's wait for resolution on that before making changes on our side.

src/libaktualizr/uptane/directorrepository.cc
mike-sul: https://saeljira.it.here.com/browse/OTA-4119

src/libaktualizr/uptane/imagesrepository.cc
mike-sul: OK

mike-sul:
@patrickvacek Updated the PR according to the comments as well as created the suggested "robustness" tickets.
Also, created a follow-up ticket for more detailed testing as we discussed https://saeljira.it.here.com/browse/OTA-4120.

PR #1451, opened 2019-11-19
Add ecu_version_report counter for primary

xcheng-here:
According to the standard of Uptane, add counter for
ECU version report. Currently, only added for primary.
It's name in the manifest is "report_counter", in the
payload of primary ecu version report.

### DIFF
--- a/config/sql/schema.sql
+++ b/config/sql/schema.sql
@@ -1,5 +1,5 @@
 CREATE TABLE version(version INTEGER);
-INSERT INTO version(rowid,version) VALUES(1,21);
+INSERT INTO version(rowid,version) VALUES(1,22);
 CREATE TABLE device_info(unique_mark INTEGER PRIMARY KEY CHECK (unique_mark = 0), device_id TEXT, is_registered INTEGER NOT NULL DEFAULT 0 CHECK (is_registered IN (0,1)));
 CREATE TABLE ecu_serials(id INTEGER PRIMARY KEY, serial TEXT UNIQUE, hardware_id TEXT NOT NULL, is_primary INTEGER NOT NULL DEFAULT 0 CHECK (is_primary IN (0,1)));
 CREATE TABLE misconfigured_ecus(serial TEXT UNIQUE, hardware_id TEXT NOT NULL, state INTEGER NOT NULL CHECK (state IN (0,1)));
@@ -23,3 +23,4 @@ CREATE TABLE ecu_installation_results(ecu_serial TEXT NOT NULL PRIMARY KEY, succ
 CREATE TABLE need_reboot(unique_mark INTEGER PRIMARY KEY CHECK (unique_mark = 0), flag INTEGER NOT NULL DEFAULT 0);
 CREATE TABLE rollback_migrations(version_from INT PRIMARY KEY, migration TEXT NOT NULL);
 CREATE TABLE delegations(meta BLOB NOT NULL, role_name TEXT NOT NULL, UNIQUE(role_name));
+CREATE TABLE ecu_report_counter(ecu_serial TEXT NOT NULL PRIMARY KEY, counter INTEGER NOT NULL DEFAULT 0);
--- a/src/libaktualizr/primary/initializer.cc
+++ b/src/libaktualizr/primary/initializer.cc
@@ -201,11 +201,28 @@ InitRetCode Initializer::initEcuRegister() {
   return InitRetCode::kOk;
 }
 
+bool Initializer::initEcuReportCounter() {
+  std::vector<std::pair<Uptane::EcuSerial, int64_t>> ecu_cnt;
+
+  if (storage_->loadEcuReportCounter(&ecu_cnt)) {
+    return true;
+  }
+
+  EcuSerials ecu_serials;
+
+  if (!storage_->loadEcuSerials(&ecu_serials) || (ecu_serials.size() == 0)) {
+    return false;
+  }
+
+  storage_->saveEcuReportCounter(Uptane::EcuSerial(ecu_serials[0].first.ToString()), 0);
+
+  return true;
+}
 // Postcondition: "ECUs registered" flag set in the storage
 Initializer::Initializer(
     const ProvisionConfig& config_in, std::shared_ptr<INvStorage> storage_in,
     std::shared_ptr<HttpInterface> http_client_in, KeyManager& keys_in,
-    const std::map<Uptane::EcuSerial, std::shared_ptr<Uptane::SecondaryInterface> >& secondary_info_in)
+    const std::map<Uptane::EcuSerial, std::shared_ptr<Uptane::SecondaryInterface>>& secondary_info_in)
     : config_(config_in),
       storage_(std::move(storage_in)),
       http_client_(std::move(http_client_in)),
@@ -239,6 +256,10 @@ Initializer::Initializer(
     }
     if (!initEcuSerials()) {
       LOG_ERROR << "ECU serial generation failed. Aborting initialization.";
+      return;
+    }
+    if (!initEcuReportCounter()) {
+      LOG_ERROR << "ECU report counter init failed. Aborting initialization.";
       return;
     }
 
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -190,6 +190,13 @@ Json::Value SotaUptaneClient::AssembleManifest() {
   Json::Value version_manifest;
 
   Json::Value primary_ecu_version = package_manager_->getManifest(primary_ecu_serial);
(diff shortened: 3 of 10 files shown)


src/libaktualizr/primary/initializer.cc
```diff
@@ -201,11 +201,27 @@ InitRetCode Initializer::initEcuRegister() {
...
+  EcuSerials ecu_serials;
+
+  if (!storage_->loadEcuSerials(&ecu_serials) || (ecu_serials.size() == 0)) {
+    return false;
+  }
+
+  storage_->saveEcuReportCounter(Uptane::EcuSerial(ecu_serials[0].first.ToString()), 0);
+  return true;
+}
 // Postcondition: "ECUs registered" flag set in the storage
```
patrickvacek: Tiny quibble, but an empty line here between the functions would be good.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1323,6 +1323,57 @@ bool SQLStorage::loadDeviceInstallationResult(data::InstallationResult* result,
...
+
+bool SQLStorage::loadEcuReportCounter(std::vector<std::pair<Uptane::EcuSerial, int64_t>>* results) {
+  SQLite3Guard db = dbConnection();
+
+  std::vector<std::pair<Uptane::EcuSerial, int64_t>> ecu_cnt;
+
+  // keep the same order as in ecu_serials (start with primary)
+  auto statement = db.prepareStatement(
+      "SELECT ecu_serial, counter FROM ecu_report_counter INNER JOIN ecu_serials ON "
+      "ecu_serials.serial = ecu_serial AND ecu_serials.is_primary = 1 ORDER BY ecu_serials.id;");
```
patrickvacek: Why the inner join? Is that just to get the Primary first?

xcheng-here:
Ah, I missed scripts for database migration and rollback. And I didn't test all of the test cases before push it, only tested the case I modified. So stupid.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1323,6 +1323,57 @@ bool SQLStorage::loadDeviceInstallationResult(data::InstallationResult* result,
   return true;
 }
 
+void SQLStorage::saveEcuReportCounter(const Uptane::EcuSerial& ecu_serial, const int64_t counter) {
+  SQLite3Guard db = dbConnection();
+
+  auto statement = db.prepareStatement<std::string, int64_t>(
+      "INSERT OR REPLACE INTO ecu_report_counter (ecu_serial, counter) VALUES "
+      "(?,?);",
+      ecu_serial.ToString(), counter);
```
patrickvacek: I'm not sure `INSERT OR REPLACE` is right here. We should enforce one counter per ECU, so won't we need a `WHERE` clause? The counter itself doesn't need to be unique.

config/sql/schema.sql
```diff
@@ -23,3 +23,4 @@ CREATE TABLE ecu_installation_results(ecu_serial TEXT NOT NULL PRIMARY KEY, succ
 CREATE TABLE need_reboot(unique_mark INTEGER PRIMARY KEY CHECK (unique_mark = 0), flag INTEGER NOT NULL DEFAULT 0);
 CREATE TABLE rollback_migrations(version_from INT PRIMARY KEY, migration TEXT NOT NULL);
 CREATE TABLE delegations(meta BLOB NOT NULL, role_name TEXT NOT NULL, UNIQUE(role_name));
+CREATE TABLE ecu_report_counter(ecu_serial TEXT NOT NULL PRIMARY KEY, counter INTEGER NOT NULL DEFAULT 0);
```
patrickvacek: Good call to include the `ecu_serial` already. Once this ticket is good and merged, I think part two will be adding secondary-specific counters, so it'll be nice not to have to do another migration.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -190,6 +190,13 @@ Json::Value SotaUptaneClient::AssembleManifest() {
   Json::Value version_manifest;
 
   Json::Value primary_ecu_version = package_manager_->getManifest(primary_ecu_serial);
+  std::vector<std::pair<Uptane::EcuSerial, int64_t>> ecu_cnt;
+  if (!storage->loadEcuReportCounter(&ecu_cnt) || (ecu_cnt.size() == 0)) {
+    LOG_ERROR << "No ECU version report counter, please check the database!";
+  } else {
+    primary_ecu_version["report_counter"] = std::to_string(ecu_cnt[0].second + 1);
```
patrickvacek: @simao @jochenschneider If we add the ECU version report counter here, does that look good to you? You can maybe see it better in the test below (https://github.com/advancedtelematic/aktualizr/pull/1451/files#diff-81426acc9fdba515977c8106aaa0eec4R159).

src/libaktualizr/storage/sqlstorage.cc
xcheng-here: Yes, I want this sql help me filter the Primary for now. There is already a loop several lines below, it will put all of the ecu_serial in the vector. So, when I need to implement this feature for all of the ECUs, in this file I only need to modify this line of sql. I'm not very familiar with sql, just learned a few in 15 minutes. Do you think it's not good to use it in this way ?

src/libaktualizr/storage/sqlstorage.cc
xcheng-here: I didn't use `WHERE` because I also use this function in initializer, when it's an empty table. After initialization, every call of this function should use an existing ecu_serial from database, so there will be only one line for each ecu in this ecu_report_counter table.

src/libaktualizr/primary/sotauptaneclient.cc
jochenschneider: Is that field documented somewhere? It is hard to see the context here.

src/libaktualizr/storage/sqlstorage.cc
patrickvacek: Okay, I think this was my mistake. I'm out of practice with SQL! Looking at other examples, this looks fine.

src/libaktualizr/storage/sqlstorage.cc
patrickvacek: Now I get it. I still think the `"ecu_serials.serial = ecu_serial AND ecu_serials.is_primary = 1` is unnecessary, though. You might as well get everything, and ordering by `ecu_serials.id` should mean the Primary is always first. You can see a similar example in `SQLStorage::loadEcuInstallationResults()` that does about the same thing.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Is https://saeljira.it.here.com/browse/OTA-3864 enough context?

But actually, either way it makes sense to make this more obvious. @xcheng-here can you post an example of what a manifest with the counter would look like? You can get it from the logs while running manually or by debugging the test case that uses it.

jochenschneider: Yes, I'd love to have an example.

xcheng-here: Sure, I uploaded a manifest.txt to the jira case, it's at line 94.

src/libaktualizr/storage/sqlstorage.cc
xcheng-here: It seems the SQLStorage::loadEcuInstallationResults() is also using "INNER JOIN". I think the reason might be that we only have "id" column in ecu_serial table, so if we want to order the result by "id", we still have to inner join with ecu_serial table. I think "ORDER BY" might be better than default order of "select".

patrickvacek: Sorry, I misquoted what I thought was unnecessary. I think the `INNER JOIN` and `ORDER BY` are fine, it's just the `AND ecu_serials.is_primary = 1` that I think is unnecessary.

jochenschneider:
You mean checking that it increases? That wouldn't be rocket since. It's just that we are still working on the new version of the director and would preferably implement it there.

patrickvacek:
> You mean checking that it increases? That wouldn't be rocket science. It's just that we are still working on the new version of the director and would preferably implement it there.

Sounds fine. I'd like to also implement this counter for secondaries as well, but I'd like to make sure we have this figured out with the backend side of things before we get too far ahead of ourselves on the embedded side. We can be patient for now, though. :)

patrickvacek:
> It's just that we are still working on the new version of the director and would preferably implement it there.

Should I make a ticket for the BE to capture that work?

jochenschneider:
That sounds like a good idea. Thanks!

PR #1452, opened 2019-11-20
Add python3 to the list of minimum dependencies

lbonn:
It's used at build time via embed_schemas.py

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -41,7 +41,7 @@ This client is aligned with the https://uptane.github.io[Uptane] security framew
 To install the minimal requirements on Debian/Ubuntu, run this:
 
 ----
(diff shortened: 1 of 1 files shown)


patrickvacek:
Travis is failing on the same thing in `test_backed_failure` that fails on my laptop. @mike-sul can you tell what's going wrong?

mike-sul:
> Travis is failing on the same thing in `test_backed_failure` that fails on my laptop. @mike-sul can you tell what's going wrong?

@patrickvacek Please, provide a link to the failing build/run.

patrickvacek:
It's hidden up in the checks for this PR, but here's the link: https://travis-ci.org/advancedtelematic/aktualizr/builds/614627653

mike-sul:
aktualizr-info fails and returns an empty string in stdout. I'll add more retries by default as well as printing stderr in case of failure, perhaps, it will help to figure out a root cause of the issue.

PR #1453, opened 2019-11-21
Update SOTA_CLIENT_FEATURES list.

patrickvacek:


patrickvacek:
> A question though, the combination of the 3 options would be valid too？

Yes. I think that's fairly clear, but is there a way to phrase it to make that more obvious?

Zee314159:
> A question though, the combination of the 3 options would be valid too？
> 
> Yes. I think that's fairly clear, but is there a way to phrase it to make that more obvious?

No need to, fairly clear. This variable is about `FEATURES`, should be a list of options.

PR #1455, opened 2019-11-21
Remove manual doxygen steps from the release docs

lbonn:
This is now automated


patrickvacek:
Probably should be based against 2019.10-docs, right?

patrickvacek:
But looks good at any rate!

PR #1458, opened 2019-11-22
Feat/c api updates

mchekhovoi:


lbonn:
Can be rebased on master to fix the failure in `check_not_expired_targets`

src/libaktualizr-c/test/api-test-utils/api-test-utils.cc
```diff
@@ -32,11 +33,18 @@ Config *Get_test_config(const char *storagePath) {
   config->provision.server = serverAddress;
   config->provision.provision_path = "tests/test_data/cred.zip";
 
-  config->storage.path = storagePath;
+  temp_dir = new TemporaryDirectory;
```
patrickvacek: Can we use a `unique_ptr` here instead of a raw `new` and `delete`?

tests/uptane_repo_generation/generate_repo.sh
```diff
@@ -0,0 +1,64 @@
...
+trap 'rm -rf "$IMAGES"' exit
+PRIMARY_FIRMWARE="$IMAGES/primary.txt"
+echo "primary" > "$PRIMARY_FIRMWARE"
+
+uptane_gen --command generate --expires 2021-07-04T16:33:27Z
+uptane_gen --command image --filename "$PRIMARY_FIRMWARE" --targetname primary.txt --hwid primary_hw
+uptane_gen --command addtarget --hwid primary_hw --serial CA:FE:A6:D2:84:9D --targetname primary.txt
+
+if [ "$ADD_DEFAULT_SECONDARY" = true ]
+then
```
patrickvacek: Not a big deal, but we usually prefer the style `if [ ]; then` as you did above.

src/libaktualizr-c/test/api-test-utils/api-test-utils.cc
mchekhovoi: There is no std::make_unique in C++11.

patrickvacek: True, but we've filled in the gap: https://github.com/advancedtelematic/aktualizr/blob/d5d0c65d7bc24c87c1c40ca9d14747159bf88e17/src/libaktualizr/utilities/utils.h#L169

patrickvacek:
I thought it looked fine but something went wrong when I tried to run this through ptest in CI: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/-/jobs/863441

src/libaktualizr-c/test/CMakeLists.txt
```diff
@@ -3,17 +3,32 @@ SET(SOURCES api-test.c)
 
 SET(CMAKE_SKIP_RPATH TRUE)
 
+SET(REPO_PATH ${PROJECT_BINARY_DIR}/uptane_repos/c-api-test-repo)
+
+if(CMAKE_CROSSCOMPILING)
+    find_program(UPTANE_GENERATOR NAMES uptane-generator)
```
patrickvacek: I don't think we need this `CMAKE_CROSSCOMPILING` clause. We ought to be able to do the same things we do in `src/libaktualizr/primary/CMakeLists.txt` and `tests/uptane_repo_generation/CMakeLists.txt` to create the repos and then make sure the test depends on that task and can find the paths.

patrickvacek:
It probably makes sense for you to run the ptests in CI directly. Basically you need to create a branch in meta-updater that bumps the aktualizr version (you can reuse mine; https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/tree/fix/ptest-c-api) and then run a pipeline on that branch with OE_PTEST set to 1. (You can also set OE_RPI to 0 to save some time.)

mchekhovoi:
https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/-/jobs/866481  - I suppose t_c_api_test is passed now (using "if(CMAKE_CROSSCOMPILING)").

PR #1460, opened 2019-11-25
Test installing updates in separate calls.

patrickvacek:
Note, however, that if you were to check for updates between the two calls, that would most likely confuse the backend.

Note that this is meant to cover the case that was broken when https://github.com/advancedtelematic/aktualizr/blob/test/split-update/src/libaktualizr/primary/sotauptaneclient.cc#L896 was not guarded by the `if` clause. I still don't like dropping Targets metadata, but this proves that we can at least do it fairly  safely only if there are problems.

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -359,6 +359,211 @@ TEST(Aktualizr, FullWithUpdates) {
   EXPECT_EQ(http->events_seen, 8);
 }
 
+class HttpFakeSplit : public HttpFake {
+ public:
+  HttpFakeSplit(const boost::filesystem::path& test_dir_in, const boost::filesystem::path& meta_dir_in)
+      : HttpFake(test_dir_in, "hasupdates", meta_dir_in) {}
+
+  HttpResponse handle_event(const std::string& url, const Json::Value& data) override {
+    (void)url;
+    for (const Json::Value& event : data) {
+      ++events_seen;
+      std::string event_type = event["eventType"]["id"].asString();
+      if (event_type.find("Ecu") == 0) {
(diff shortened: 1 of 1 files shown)


patrickvacek:
Travis failure seems irrelevant:
```
Error: SQLite database migration failed
ERROR:test_fixtures:None
ERROR:test_fixtures:Failed to get aktualizr info/status
INFO:/aktualizr/tests/test_backend_failure.py:>>> Failed: test_backend_failure_sanity_director_update_after_metadata_download_failure
```
@mike-sul any hints on that one?

mike-sul:
> Travis failure seems irrelevant:
> 
> ```
> Error: SQLite database migration failed
> ERROR:test_fixtures:None
> ERROR:test_fixtures:Failed to get aktualizr info/status
> INFO:/aktualizr/tests/test_backend_failure.py:>>> Failed: test_backend_failure_sanity_director_update_after_metadata_download_failure
> ```
> 
> @mike-sul any hints on that one?

According to the logs aktualizr-info didn't output the desired/expected message.

lbonn:
> Note that this is meant to cover the case that was broken when [/split-update/src/libaktualizr/primary/sotauptaneclient.cc@`test`#L896](https://github.com/advancedtelematic/aktualizr/blob/test/split-update/src/libaktualizr/primary/sotauptaneclient.cc#L896) [](/advancedtelematic/aktualizr/blob/HEAD@{2019-11-25T16:03:35Z}/split-update/src/libaktualizr/primary/sotauptaneclient.cc#L896) was not guarded by the `if` clause. I still don't like dropping Targets metadata, but this proves that we can at least do it fairly safely only if there are problems.

Should we include some of this info in a comment in the test, for easy reference on why it exists?

patrickvacek:
> Should we include some of this info in a comment in the test, for easy reference on why it exists?

Done.

PR #1462, opened 2019-11-26
Use jsoncpp as a submodule

lbonn:


CMakeLists.txt
```diff
@@ -386,6 +383,14 @@ ENABLE_TESTING()
 # It would be great to use GTEST_OUTPUT directly, but I couldn't get it to work.
 set(GOOGLE_TEST_OUTPUT --gtest_output=xml:${CMAKE_BINARY_DIR}/results/)
 
+# amalgamate jsoncpp source at compile time
+# note: jsoncpp has a CMake support that they intent to deprecate and is hard to
+# integrate with.
+execute_process(COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/amalgamate-jsoncpp.sh ${CMAKE_CURRENT_SOURCE_DIR}/third_party/jsoncpp ${CMAKE_CURRENT_BINARY_DIR}/jsoncpp)
```
patrickvacek: Why do we have to do this amalgamation?

How else do they build if they are deprecating their CMake support?

src/libaktualizr/CMakeLists.txt
```diff
@@ -24,8 +24,8 @@ add_library(aktualizr_static_lib STATIC
     $<TARGET_OBJECTS:campaign>
     $<TARGET_OBJECTS:config>
     $<TARGET_OBJECTS:crypto>
-    $<TARGET_OBJECTS:http>
     $<TARGET_OBJECTS:jsoncpp>
+    $<TARGET_OBJECTS:http>
```
patrickvacek: Why break the alphabetical ordering?

src/libaktualizr/utilities/utils_test.cc
```diff
@@ -46,6 +46,23 @@ TEST(Utils, PrettyNameOk) {
   EXPECT_FALSE(PrettyNameOk("foo-bar-123&"));
 }
 
+TEST(Utils, parseJSON) {
+  // this should not cause valgrind warnings
```
patrickvacek: Agreed... what warnings might it throw?

src/virtual_secondary/virtualsecondary.cc
```diff
@@ -57,10 +55,13 @@ void VirtualSecondaryConfig::dump(const boost::filesystem::path& file_full_path)
...
+  Json::StreamWriterBuilder json_bwriter;
+  json_bwriter["indentation"] = "\t";
+  std::unique_ptr<Json::StreamWriter> const json_writer(json_bwriter.newStreamWriter());
+
   boost::filesystem::create_directories(file_full_path.parent_path());
   std::ofstream json_file(file_full_path.string());
-  json_writer.write(json_file, root);
+  json_writer->write(root, &json_file);
   json_file.close();
 }
```
patrickvacek: Is anyone using this `dump` function? It doesn't look like it.

CMakeLists.txt
lbonn: Mainly meson https://github.com/open-source-parsers/jsoncpp/wiki/Building.

I've made another attempt by using CMake the "proper" way and mostly just `add_subdirectory()` but it was quite hard to control the compilation flags and I think I did not manage to remove the compilation of some internal test executable.

Just using the amalgamated source, like we were doing before, gives more flexibility.

Note that the amalgamation process is done at configure time and that cmake would have to be re-run after updating the submodule.

src/libaktualizr/CMakeLists.txt
lbonn: I will remove this, that's a relic of a previous version.

src/libaktualizr/utilities/utils_test.cc
lbonn: There were some uninitialized memory accesses until 1.8.2 (but not in the ancient version we were using).

See https://github.com/open-source-parsers/jsoncpp/pull/651 and https://github.com/open-source-parsers/jsoncpp/issues/578.

src/virtual_secondary/virtualsecondary.cc
lbonn: Seems to be used in tests: https://github.com/advancedtelematic/aktualizr/blob/dd73a67eb42e60727527b72cc2285170dd842c66/tests/uptane_test_common.h#L80

patrickvacek: Ah, there it is! Sorry, missed that somehow.

CMakeLists.txt
patrickvacek: Yeah, this seems fine, it's just a bit weird.

PR #1463, opened 2019-11-28
fix/ota-3865/Log connectivity restored after an interruption

Zee314159:
Draft code.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -948,6 +948,7 @@ bool SotaUptaneClient::putManifestSimple(const Json::Value &custom) {
     return false;
   }
 
+  static bool connected = true;
   auto manifest = AssembleManifest();
   if (custom != Json::nullValue) {
     manifest["custom"] = custom;
@@ -955,8 +956,14 @@ bool SotaUptaneClient::putManifestSimple(const Json::Value &custom) {
   auto signed_manifest = uptane_manifest.signManifest(manifest);
   HttpResponse response = http->put(config.uptane.director_server + "/manifest", signed_manifest);
   if (response.isOk()) {
+    if (!connected) {
+      LOG_INFO << "Connectivity is restored.";
+    }
+    connected = true;
     storage->clearInstallationResults();
     return true;
+  } else {
+    connected = false;
   }
 
   LOG_WARNING << "Put manifest request failed: " << response.getStatusStr();
--- a/src/libaktualizr/uptane/uptane_network_test.cc
+++ b/src/libaktualizr/uptane/uptane_network_test.cc
@@ -5,6 +5,7 @@
  * network issues.
  */
 #include <gtest/gtest.h>
+#include <gtest/internal/gtest-port.h>
 
 #include <boost/process.hpp>
 #include <fstream>
@@ -14,6 +15,7 @@
 #include <vector>
 
 #include "http/httpclient.h"
+#include "httpfake.h"
 #include "logging/logging.h"
 #include "primary/initializer.h"
 #include "primary/sotauptaneclient.h"
@@ -137,6 +139,72 @@ TEST(UptaneNetwork, DownloadFailure) {
 
   std::pair<bool, Uptane::Target> result = up->downloadImage(package_to_install);
   EXPECT_TRUE(result.first);
+}
+
+/*
+ * Output a log when connectivity is restored.
+ */
+class HttpUnstable : public HttpFake {
+ public:
+  HttpUnstable(const boost::filesystem::path &test_dir_in) : HttpFake(test_dir_in, "hasupdates") {}
+  HttpResponse get(const std::string &url, int64_t maxsize) override {
+    if (!connectSwitch) {
+      return HttpResponse({}, 503, CURLE_OK, "");
+    } else {
+      return HttpFake::get(url, maxsize);
+    }
+  }
+
+  HttpResponse put(const std::string &url, const Json::Value &data) override {
+    if (!connectSwitch) {
+      (void)data;
+      return HttpResponse(url, 503, CURLE_OK, "");
+    } else {
+      return HttpFake::put(url, data);
+    }
+  }
+
+  HttpResponse post(const std::string &url, const Json::Value &data) override {
+    if (!connectSwitch) {
+      (void)data;
+      return HttpResponse(url, 503, CURLE_OK, "");
+    } else {
+      return HttpFake::post(url, data);
+    }
+  }
+
+  bool connectSwitch = true;
+};
+
+TEST(UptaneNetwork, LogConnectivityRestored) {
+  TemporaryDirectory temp_dir;
+  auto http = std::make_shared<HttpUnstable>(temp_dir.Path());
+  Config config = UptaneTestCommon::makeTestConfig(temp_dir, http->tls_server);
+  config.uptane.director_server = http->tls_server + "director";
+  config.uptane.repo_server = http->tls_server + "repo";
+  config.pacman.type = PackageManager::kNone;
+  config.provision.primary_ecu_serial = "CA:FE:A6:D2:84:9D";
+  config.provision.primary_ecu_hardware_id = "primary_hw";
+  config.storage.path = temp_dir.Path();
+  config.tls.server = http->tls_server;
+  UptaneTestCommon::addDefaultSecondary(config, temp_dir, "secondary_ecu_serial", "secondary_hw");
+
+  auto storage = INvStorage::newStorage(config.storage);
+  auto up = std_::make_unique<UptaneTestCommon::TestUptaneClient>(config, storage, http);
+  EXPECT_NO_THROW(up->initialize());
+
+  result::UpdateCheck result = up->fetchMeta();
+  EXPECT_EQ(result.status, result::UpdateStatus::kUpdatesAvailable);
+
+  http->connectSwitch = false;
+  result = up->fetchMeta();
+  EXPECT_EQ(result.status, result::UpdateStatus::kError);
+
+  http->connectSwitch = true;
+  testing::internal::CaptureStdout();
+  result = up->fetchMeta();
+  EXPECT_EQ(result.status, result::UpdateStatus::kUpdatesAvailable);
+  EXPECT_NE(std::string::npos, testing::internal::GetCapturedStdout().find("Connectivity is restored."));
 }
 
 #ifndef __NO_MAIN__
--- a/src/libaktualizr/primary/sotauptaneclient.h
+++ b/src/libaktualizr/primary/sotauptaneclient.h
@@ -104,6 +104,7 @@ class SotaUptaneClient {
   FRIEND_TEST(UptaneCI, CheckKeys);
   FRIEND_TEST(UptaneKey, Check);  // Note hacky name
   FRIEND_TEST(UptaneNetwork, DownloadFailure);
+  FRIEND_TEST(UptaneNetwork, LogConnectivityRestored);
   FRIEND_TEST(UptaneVector, Test);
   FRIEND_TEST(aktualizr_secondary_uptane, credentialsPassing);
   friend class CheckForUpdate;       // for load tests
--- a/src/libaktualizr/uptane/CMakeLists.txt
+++ b/src/libaktualizr/uptane/CMakeLists.txt
@@ -51,7 +51,7 @@ add_dependencies(t_uptane_delegation uptane-generator)
 target_link_libraries(t_uptane_delegation virtual_secondary)
 set_tests_properties(test_uptane_delegation PROPERTIES LABELS "crypto")
 
-add_aktualizr_test(NAME uptane_network SOURCES uptane_network_test.cc PROJECT_WORKING_DIRECTORY)
+add_aktualizr_test(NAME uptane_network SOURCES uptane_network_test.cc PROJECT_WORKING_DIRECTORY LIBRARIES uptane_generator_lib)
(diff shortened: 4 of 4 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -948,15 +948,23 @@ bool SotaUptaneClient::putManifestSimple(const Json::Value &custom) {
...
   auto manifest = AssembleManifest();
   if (custom != Json::nullValue) {
     manifest["custom"] = custom;
   }
   auto signed_manifest = uptane_manifest.signManifest(manifest);
   HttpResponse response = http->put(config.uptane.director_server + "/manifest", signed_manifest);
   if (response.isOk()) {
+    if (!connect) {
+      LOG_INFO << "Connectivity is restored.";
+      sendEvent<event::LogConnectivityRestored>();
```
patrickvacek: I think just logging is enough. This is probably easier than we were thinking. Sending an event isn't a bad idea but for now we can skip it.

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -282,6 +282,44 @@ TEST(Uptane, PutManifestError) {
...
+
+  http = std::make_shared<HttpFake>(temp_dir.Path());
+  auto events_channel = std::make_shared<event::Channel>();
+  std::function<void(std::shared_ptr<event::BaseEvent> event)> f_cb = process_events_LogConnectivityRestored;
+  events_channel->connect(f_cb);
+  sota_client = std_::make_unique<UptaneTestCommon::TestUptaneClient>(conf, storage, http, events_channel);
+  EXPECT_NO_THROW(sota_client->initialize());
+  EXPECT_TRUE(sota_client->putManifestSimple());
+  EXPECT_EQ(num_events_LogConnectivityRestored, 1);
+}
```
patrickvacek: This general test format looks fine, although capturing logging output is slightly challenging, and I'd suggest putting this in `uptane_network_test.cc`, although to be honest this isn't really Uptane-specific.

Zee314159:
```

[2019-12-02 11:56:38.874602] [0x000000000404fcc0] [info]    Bootstrap empty SQL storage
4401
[2019-12-02 11:56:38.896849] [0x000000000404fcc0] [info]    Bootstraping DB to version 22
4402
[2019-12-02 11:56:39.130130] [0x000000000404fcc0] [trace]   meta with role root in repo director not present in db
4403
[2019-12-02 11:56:39.157554] [0x000000000404fcc0] [trace]   meta with role root in repo image not present in db
4404
[2019-12-02 11:56:39.214023] [0x000000000404fcc0] [trace]   device_id not present in db
4405
[2019-12-02 11:56:39.277134] [0x000000000404fcc0] [trace]   client_pkey not present in db
4406
[2019-12-02 11:56:39.316517] [0x000000000404fcc0] [trace]   client_cert not present in db
4407
[2019-12-02 11:56:39.347844] [0x000000000404fcc0] [trace]   ca_cert not present
4408
[2019-12-02 11:56:39.376881] [0x000000000404fcc0] [trace]   client_pkey not present in db
4409
[2019-12-02 11:56:39.837780] [0x000000000404fcc0] [trace]   post request body:{
4410
	"deviceId" : "tasty-raclette-116",
4411
	"ttl" : "logconnect"
4412
}
4413
[2019-12-02 11:56:39.858029] [0x000000000404fcc0] [trace]   response http code: 404
4414
[2019-12-02 11:56:39.858286] [0x000000000404fcc0] [trace]   response: 
4415
[2019-12-02 11:56:39.859687] [0x000000000404fcc0] [error]   Shared credential provisioning failed, response: 
4416
[2019-12-02 11:56:39.859988] [0x000000000404fcc0] [error]   Shared credential provisioning failed. Aborting initialization.
4417
../src/libaktualizr/uptane/uptane_network_test.cc:163: Failure
4418
Value of: result
4419
  Actual: false
4420
Expected: true
```

Any tips why my test failed at the first stage? I followed the `"noerrors"` test case to write this one. And `testing::internal::CaptureStdout` seems not like a right choice to capture LOG_INFO. Any other suggestions?

src/libaktualizr/uptane/uptane_network_test.cc
```diff
@@ -139,6 +140,44 @@ TEST(UptaneNetwork, DownloadFailure) {
   EXPECT_TRUE(result.first);
 }
 
+/*
+ * Output a log when connectivity is restored.
+ */
+TEST(UptaneNetwork, LogConnectivityRestored) {
+  //  testing::internal::CaptureStdout();
+  RecordProperty("zephyr_key", "OTA-991,TST-158");
```
patrickvacek: We can leave out this `RecordProperty` thing for now. Maybe someday that'll be important again, but for now, it isn't.

src/libaktualizr/uptane/uptane_network_test.cc
```diff
@@ -139,6 +140,44 @@ TEST(UptaneNetwork, DownloadFailure) {
...
+    KeyManager keys(store, conf.keymanagerConfig());
+    Initializer initializer(conf.provision, store, http, keys, {});
+    result = initializer.isSuccessful();
+  }
+  EXPECT_TRUE(result);
+  {
+    conf.provision.server = conf.provision.server.substr(conf.provision.server.size() - 2) + "11";
+    KeyManager keys(store, conf.keymanagerConfig());
+    Initializer initializer(conf.provision, store, http, keys, {});
+    result = initializer.isSuccessful();
```
patrickvacek: I like this idea of feigning a broken connection by changing the server URL. However, does it require re-initialization every time? I was hoping we could just test this with calls to `CheckUpdates`. At any rate, we almost certainly can avoid extra instantiations of `KeyManager`.

Zee314159: > At any rate, we almost certainly can avoid extra instantiations of KeyManager.

But `Initializer initializer(conf.provision, store, http, keys, {});` need `keys`. And from `doTestInit` it looks re-initialization is needed. I'm not sure though.

patrickvacek: I'm fairly confident that you can reuse the keys if not the entire initialization. To simulate a more realistic scenario, we should only initialize once and merely call `CheckUpdates` multiple times under different network conditions.

Zee314159: > we should only initialize once and merely call CheckUpdates multiple times under different network conditions.

How can I get under different network conditions if I don't use initialize? I didn't find the function I need.

src/libaktualizr/uptane/uptane_network_test.cc
```diff
@@ -139,6 +140,37 @@ TEST(UptaneNetwork, DownloadFailure) {
...
+  conf.provision.primary_ecu_serial = "noerrors";
+  const std::string good_url = conf.provision.server;
+
+  auto http = std::make_shared<HttpClient>();
+  auto store = INvStorage::newStorage(conf.storage);
+
+  auto sota_client = std_::make_unique<SotaUptaneClient>(conf, store, http);
+  EXPECT_NO_THROW(sota_client->initialize());
+  EXPECT_NO_THROW(sota_client->fetchMeta());
+  EXPECT_NO_THROW(sota_client->checkUpdates());
```
patrickvacek: If you're using the SotaUptaneClient functions directly, you probably want to just call `initialize` once and then `fetchMeta` each time. `fetchMeta` is actually what `Aktualizr::CheckUpdates` calls, and `fetchMeta` calls `SotaUptaneClient::checkUpdates`. It also is what calls `putManifestSimple`, which is where your change is located, so just calling `SotaUptaneClient::checkUpdates` won't trigger it.

patrickvacek: BTW if you make this test case a `FRIEND_TEST` of `SotaUptaneClient` (see `sotauptaneclient.h` for other examples) you can directly manipulate the `Config` object stored in `SotaUptaneClient` and then you can bypass needing to re-instantiate or re-initialize `SotaUptaneClient`.

src/libaktualizr/uptane/uptane_network_test.cc
```diff
@@ -139,6 +140,33 @@ TEST(UptaneNetwork, DownloadFailure) {
...
+  conf.storage.path = temp_dir.Path();
+  conf.provision.expiry_days = "noerrors";
+  conf.provision.primary_ecu_serial = "noerrors";
+  const std::string good_url = conf.provision.server;
+  auto http = std::make_shared<HttpClient>();
+  auto store = INvStorage::newStorage(conf.storage);
+
+  auto sota_client = std_::make_unique<SotaUptaneClient>(conf, store, http);
+  EXPECT_NO_THROW(sota_client->initialize());
+  result::UpdateCheck result = sota_client->fetchMeta();
```
Zee314159: Got this failure `HTTP/1.0 501 Unsupported method ('PUT')`. Is `auto http = std::make_shared<HttpClient>();` a susceptive target?

Zee314159: `HttpFake` seems to make no difference in local test.

lbonn:
It looks like it's failing right now because the "hasupdates" HttpFake should be used with a secondary. 

Adding this line makes the fetchMeta call succeed. It would even be better to use `makeTestConfig` from `uptane_test_common.h` instead of this shared object between all tests.

```
UptaneTestCommon::addDefaultSecondary(conf, temp_dir, "secondary_ecu_serial", "secondary_hw");
```

Zee314159:
> It looks like it's failing right now because the "hasupdates" HttpFake should be used with a secondary.
> Adding this line makes the fetchMeta call succeed. It would even be better to use makeTestConfig from uptane_test_common.h instead of this shared object between all tests.
> UptaneTestCommon::addDefaultSecondary(conf, temp_dir, "secondary_ecu_serial", "secondary_hw");

Thanks Laurent@lbonn ,yes, I tried your method, the fetchMeta call succeed, but when I try to simulate a broken connection, it seems still connected, and failed to directly manipulate the `Config object` stored in `SotaUptaneClient`. Any tips?

lbonn:
@Zee314159 Modifying the config object might work, or not if it has been copied before by an internal component.

The most straightforward thing that comes to my mind is to subclass HttpFake and add a new method to toggle a switch that causes all the requests to fail.

src/libaktualizr/uptane/uptane_network_test.cc
```diff
@@ -139,6 +141,85 @@ TEST(UptaneNetwork, DownloadFailure) {
...
+      return HttpResponse(url, 200, CURLE_OK, "");
+    }
+  }
+
+  HttpResponse post(const std::string &url, const Json::Value &data) override {
+    if (!connectSwitch) {
+      (void)data;
+      return HttpResponse(url, 503, CURLE_OK, "");
+    } else {
+      if (url.find("/devices") != std::string::npos || url.find("/director/ecus") != std::string::npos || url.empty()) {
```
lbonn: Is there a reason why we can't call `HttpFake::post()` here? Copying this logic here, would complicate maintenance.

src/libaktualizr/uptane/uptane_network_test.cc
```diff
@@ -139,6 +141,85 @@ TEST(UptaneNetwork, DownloadFailure) {
...
+      return HttpFake::get(url, maxsize);
+    }
+  }
+
+  HttpResponse put(const std::string &url, const Json::Value &data) override {
+    if (!connectSwitch) {
+      (void)data;
+      return HttpResponse(url, 504, CURLE_OK, "");
+    } else {
+      last_manifest = data;
```
lbonn: Can we also call `HttpFake::put()`?

lbonn:
Looking good apart from the two small comments. Also, before we merge, could you edit the commit message so that it describes the changes better?

src/libaktualizr/uptane/uptane_network_test.cc
Zee314159: I'm not sure, but if will fail if I use the original. I'll have a test.

Zee314159:
```
file served: "/tmp/aktualizr-4d8c-c7cb-fcae-9383/619b-47da-dir/repo/targets_hasupdates.json"
4429
[2019-12-09 12:59:48.635639] [0x000000000404fcc0] [info]    2 new updates found in Uptane metadata.
4430
[2019-12-09 12:59:48.659888] [0x000000000404fcc0] [info]    got UpdateCheckComplete event
4431
[2019-12-09 12:59:50.232787] [0x000000000404fcc0] [error]   Failed to update director metadata: 
4432
[2019-12-09 12:59:50.236768] [0x000000000404fcc0] [info]    got UpdateCheckComplete event
4433
../src/libaktualizr/uptane/uptane_network_test.cc:217: Failure
4434
Expected: (std::string::npos) != (testing::internal::GetCapturedStdout().find("Connectivity is restored.")), actual: 18446744073709551615 vs 1844674407370955161
```

lbonn: Well, you still need to handle the failure case. In the last version you've pushed, `put` and `post` are the same as in HttpFake.

If you follow the same pattern that what you did for `get()`, it should work.

e.g:

```
HttpResponse post(const std::string &url, const Json::Value &data) override {
 if (!connectSwitch) {
   (void)data;
   return HttpResponse(url, 503, CURLE_OK, "");
 } else {
   return HttpFake::post(url, data);
  }
}
```

Zee314159: Oh, I see, thank you!

lbonn:
It merged in master but the PR didn't get closed somehow :open_hands:

PR #1465, opened 2019-12-02
Setup a gitlab pipeline for tuf-test-vectors

xcheng-here:
Added a job for tuf-test-vectors, and added necessary
packages into docker images, updated submodule version.

### DIFF
--- a/docker/Dockerfile.ubuntu.bionic
+++ b/docker/Dockerfile.ubuntu.bionic
@@ -58,6 +58,7 @@ RUN apt-get update && apt-get -y install --no-install-suggests --no-install-reco
   python3-openssl \
   python3-pip \
   python3-venv \
+  python3-setuptools \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.ubuntu.bionic
```diff
@@ -74,4 +75,7 @@ RUN make install
 
 RUN useradd testuser
 
+RUN ln -s /usr/bin/pip3 /usr/bin/pip
+RUN pip install pip-tools virtualenv wheel pytest
```
patrickvacek: This looks dangerous (the `ln` part). Can we just call `pip3 install`?

xcheng-here: Ah, sorry for totally misunderstood the goal. I will fix it.

PR #1466, opened 2019-12-02
Fix/ostree logging

patrickvacek:


patrickvacek:
> I think that we maybe can do a little bit better with the naming getCurrent/getCurrentHash, but since we will likely need to revisit the logic there and naming is hard, I think it's good to go as it is.

Good point. I was mostly just being lazy but if we come through here again we can do better.

PR #1467, opened 2019-12-02
Fix a wrong scary log message

lbonn:
It was even appearing in the happy case

### DIFF
--- a/src/libaktualizr/primary/reportqueue.cc
+++ b/src/libaktualizr/primary/reportqueue.cc
@@ -54,10 +54,14 @@ void ReportQueue::flushQueue() {
 
   if (!report_array.empty()) {
     HttpResponse response = http->post(config.tls.server + "/events", report_array);
+
     // 404 implies the server does not support this feature. Nothing we can
     // do, just move along.
-    if (response.isOk() || response.http_status_code == 404) {
+    if (response.http_status_code == 404) {
       LOG_TRACE << "Server does not support event reports. Clearing report queue.";
+    }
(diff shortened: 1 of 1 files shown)


src/libaktualizr/primary/reportqueue.cc
```diff
@@ -57,7 +57,9 @@ void ReportQueue::flushQueue() {
     // 404 implies the server does not support this feature. Nothing we can
     // do, just move along.
     if (response.isOk() || response.http_status_code == 404) {
-      LOG_TRACE << "Server does not support event reports. Clearing report queue.";
+      if (!response.isOk()) {
```
eu-smirnov: I think `if (response.http_status_code == 404) {` would look clearer in this case and it doesn't need to be nested. Also, are we sure that we'll get 404 only in the case of endpoint absence and not in a sudden backend outage? Do we even want to support a backend without report queue?

patrickvacek: > are we sure that we'll get 404 only in the case of endpoint absence and not in a sudden backend outage?

No, but what else are we supposed to do? I think this solution is fine for now until we have a good reason to change it.

> Do we even want to support a backend without report queue?

Yes, in theory, although in practice it doesn't seem important. I think this was mostly written as a way to make tests work more easily.

lbonn: yes, I agree about moving the if upwards. Actually thought the same after pushing and going back home :).

PR #1470, opened 2019-12-03
Set codecov threshold to be more lenient

lbonn:


eu-smirnov:
Maybe it failed because it still takes the yml from master? Also is 0.3 supposed to be 0.3% or 30%? My impression that it will be 30%. And 0.3% seems a little bit to low...

lbonn:
From the docs (https://docs.codecov.io/docs/commit-status#section-threshold), it looks like it should be 0.3%.

But I forgot we already have codecov.yml and this PR added a new .codecov.yml (with a dot), so this might be the cause of the problem.

Let's see if it works better this time.

patrickvacek:
Can we move the coverage upload from Travis to gitlab? Was there a reason we couldn't?

lbonn:
> Looks fine to me, or did you want to test more?

Frankly, I'm not yet convinced this is actually working and it's rather annoying to test. But we could merge anyway and see what happens, as it's not really a dangerous change.

PR #1471, opened 2019-12-03
Misc HSM fixes

lbonn:
Found after trying the real thing.


CMakeLists.txt
```diff
@@ -125,12 +125,6 @@ if(BUILD_P11)
     add_definitions(-DBUILD_P11)
 endif(BUILD_P11)
 
-# Setup PKCS11
-if(TEST_PKCS11_MODULE_PATH)
-    add_definitions(-DTEST_PKCS11_MODULE_PATH="${TEST_PKCS11_MODULE_PATH}"
-                    -DTEST_PKCS11_ENGINE_PATH="${TEST_PKCS11_ENGINE_PATH}")
-endif(TEST_PKCS11_MODULE_PATH)
```
patrickvacek: Am I reading this right that you've basically moved these definitions to only get set in the source files that need them?

lbonn: Yes, that's the intent.

patrickvacek:
I recommend running the oe-selftests on this before merging since it seems likely to affect them. Looks good in general, though!

src/libaktualizr/crypto/p11engine.cc
```diff
@@ -127,20 +128,34 @@ P11Engine::P11Engine(P11Config config) : config_(std::move(config)), ctx_(config
...
-  const boost::filesystem::path openssl11_path = "/usr/lib/engines-1.1/pkcs11.so";
-  const boost::filesystem::path default_path = "/usr/lib/engines/pkcs11.so";
-  if (boost::filesystem::exists(custom_path)) {
-    return custom_path;
-  } else if (boost::filesystem::exists(openssl11_path)) {
-    return openssl11_path;
+  static const std::array<boost::filesystem::path, 3> engine_system_paths = {
+    "/usr/lib/engines-1.1/pkcs11.so",
+    "/usr/lib/engines/pkcs11.so",
+    "/usr/lib/x86_64-linux-gnu/engines-1.1/pkcs11.so"
```
ricardosalveti: x86_64-linux-gnu? This is not generic across architectures.

lbonn: Yes, this is a small sample of possible locations I've seen (in this case, Ubuntu Bionic x86-64).

The recommended way should be to define `PKCS11_ENGINE_PATH`, this is only a helpful fallback. But maybe this is more harmful than useful?

ricardosalveti: Problem is that if you want to cover other common locations this list will end up growing quite a bit as it depends on the distro and architecture used (e.g. bionic armhf/arm64/ppc64 would all have a different linux-gnu path).

Shouldn't necessarily be a problem, but having PKCS11_ENGINE_PATH is probably cleaner.

lbonn: Yes, I agree, that's why the list is not intended to be exhaustive, just be helpful for the common desktop testing scenario.

I will at least add some comments or maybe even do that through CMake in a safer way.

lbonn:
I've made some changes to remove the engine detection at runtime.
It's now done at configure time, only when compiling locally.

CMakeLists.txt
```diff
@@ -123,17 +123,32 @@ endif(BUILD_DOCKERAPP)
...
+            message(FATAL_ERROR "Could not auto-detect path of PKCS11 engine, please specify PKCS11_ENGINE_PATH")
+        endif()
+
+        set(PKCS11_ENGINE_PATH "${PKCS11_ENGINE_PATH}" CACHE STRING "Credentials.zip for tests involving the server")
+    endif()
+endif(BUILD_P11)
 
 if(BUILD_SOTA_TOOLS)
     find_package(GLIB2 REQUIRED)
-    find_program(STRACE NAMES strace)
```
patrickvacek: I thought I left a comment here but apparently not. `strace` is still used by the `garage-push-cacert-used` test. (I don't think that test actually runs in docker, though, because we never install strace.) Otherwise this PR looks fine.

lbonn: Yes, I missed that somehow...

I'd suggest we'd either nuke this test or make sure it's executed on CI. Any preference?

patrickvacek: Seems easy to just put it into CI; that'd be my preference. It is testing something otherwise uncovered.

PR #1473, opened 2019-12-04
OTA-4120: Test Uptane verification on Secondary for an ostree u…

mike-sul:

### DIFF
--- a/src/aktualizr_secondary/aktualizr_secondary.cc
+++ b/src/aktualizr_secondary/aktualizr_secondary.cc
@@ -86,24 +86,26 @@ bool AktualizrSecondary::putRootResp(const std::string& root, bool director) {
 }
 
 bool AktualizrSecondary::sendFirmwareResp(const std::shared_ptr<std::string>& firmware) {
-  auto targetsForThisEcu = director_repo_.getTargets(getSerial());
+  auto targetsForThisEcu = director_repo_.getTargets(getSerial(), getHardwareID());
 
   if (targetsForThisEcu.size() != 1) {
     LOG_ERROR << "Invalid number of targets (should be one): " << targetsForThisEcu.size();
     return false;
   }
-  auto targetToApply = targetsForThisEcu[0];
+  auto target_to_apply = targetsForThisEcu[0];
 
   std::string treehub_server;
+  std::size_t firmware_size = firmware->length();
 
-  if (targetToApply.IsOstree()) {
+  if (target_to_apply.IsOstree()) {
     // this is the ostree specific case
     try {
       std::string ca, cert, pkey, server_url;
-      extractCredentialsArchive(*firmware, &ca, &cert, &pkey, &treehub_server);
+      extractCredentialsArchive(*firmware, &ca, &cert, &pkey, &server_url);
       keys_.loadKeys(&ca, &cert, &pkey);
       boost::trim(server_url);
       treehub_server = server_url;
+      firmware_size = 0;
     } catch (std::runtime_error& exc) {
       LOG_ERROR << exc.what();
 
@@ -113,9 +115,9 @@ bool AktualizrSecondary::sendFirmwareResp(const std::shared_ptr<std::string>& fi
 
   data::InstallationResult install_res;
 
-  if (targetToApply.IsOstree()) {
+  if (target_to_apply.IsOstree()) {
 #ifdef BUILD_OSTREE
-    install_res = OstreeManager::pull(config_.pacman.sysroot, treehub_server, keys_, targetToApply);
+    install_res = OstreeManager::pull(config_.pacman.sysroot, treehub_server, keys_, target_to_apply);
 
     if (install_res.result_code.num_code != data::ResultCode::Numeric::kOk) {
       LOG_ERROR << "Could not pull from OSTree (" << install_res.result_code.toString()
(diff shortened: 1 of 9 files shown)


src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -128,23 +130,46 @@ bool AktualizrSecondary::sendFirmwareResp(const std::shared_ptr<std::string>& fi
...
 
-  // This check is not applicable for the ostree case, just for the binary file case(s)
-  if (targetToApply.length() != firmware->length()) {
+  if (targetToApply.length() != firmware_size) {
     LOG_ERROR << "The target image size specified in metadata " << targetToApply.length()
               << " does not match actual size " << firmware->length();
     return false;
   }
 
+  if (!targetToApply.IsOstree()) {
```
zabbal: That's odd - why do we have to brute-force hash type in here?

mike-sul: I am not sure if I understand the question.

zabbal: Why both sha256 and sha512 are computed simultaneously? We don't know which hash is actually used?

mike-sul: As a matter of fact, the targets.json sent from BE contains both type of hashes for each "BINARY" target format.
Also, primary supports these two types of hashes and can generate and verify any of them, although it verifies just the first hash found in json :). See the following for more details 

https://github.com/advancedtelematic/aktualizr/blob/6b1da3e473f3c9963a3221607dabc4d0cde06968/src/libaktualizr/package_manager/packagemanagerinterface.cc#L149

https://github.com/advancedtelematic/aktualizr/blob/6b1da3e473f3c9963a3221607dabc4d0cde06968/src/libaktualizr/package_manager/packagemanagerinterface.cc#L8

The questions are 
1) why there are two types of hashes present in targets.json ? 
2) why Primary verifies just the first found hash for a binary image instead of verifying each hash ?
3) what Secondary should do in order to support Full Uptane Verification ?

zabbal: That's weird indeed - I can't really picture situation when sha256 will match but sha512 wouldn't or other way around. @patrickvacek, @lbonn any ideas what's going on?

lbonn: Well, if one matches and the other doesn't, either somebody is doing a weird attack (and if they managed to sign the metadata, they can try more obvious things) or one of the hash function has been broken.

For 1., that's something that's allowed by Uptane, supposedly for hash algorithms agility, so I suppose that's "because we can". But then, should we really?..

For 2., I suppose it's technically a bug then.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -128,23 +130,46 @@ bool AktualizrSecondary::sendFirmwareResp(const std::shared_ptr<std::string>& fi
...
+    Uptane::Hash sha256_hash(Uptane::Hash::Type::kSha256,
+                             boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest(*firmware))));
+    Uptane::Hash sha512_hash(Uptane::Hash::Type::kSha512,
+                             boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha512digest(*firmware))));
+
+    if (!(targetToApply.MatchHash(sha256_hash) && targetToApply.MatchHash(sha512_hash))) {
+      LOG_ERROR << "The received image data hash doesn't match the hash specified in the target metadata";
+      return false;
+    }
```
zabbal: I think it's worth reporting which hash has failed in the error message. Will make code slightly bigger but will save headaches later on.

src/aktualizr_secondary/aktualizr_secondary.cc
zabbal: Makes sense - it's clearly a bug if we only check one hash instead of all available (and we should fail if any of them mismatch) as it defeats the purpose of having multiple hashes in a first place. Do we already have a ticket for this?

src/aktualizr_secondary/uptane_verification_test.cc
```diff
@@ -95,104 +195,209 @@ class SecondaryUptaneVerificationTest : public ::testing::Test {
...
+TEST_F(OstreeSecondaryUptaneVerificationTest, fullUptaneVerificationInvalidSerial) {
+  EXPECT_FALSE(_secondary->putMetadataResp(addTarget("", "", "invalid-serial-id")));
+}
+
+TEST_F(SecondaryUptaneVerificationTest, fullUptaneVerificationPositive) {
+  EXPECT_TRUE(_secondary->putMetadataResp(_uptane_repo.getCurrentMetadata()));
+  EXPECT_TRUE(_secondary->sendFirmwareResp(getImageData()));
 }
 
+TEST_P(SecondaryUptaneVerificationTestNegative, MalformedMetadaJson) {
```
lbonn: Neat! Could you group the TEST_P and INSTANTIATE_TEST_SUITE_P together with some explanatory comments, to make it clearer in further reviews?

Here, the lone `TEST_P` can be easily confused with the series of TEST_F before.

src/aktualizr_secondary/aktualizr_secondary.cc
lbonn: Realistically, they will both match or not. Logging the third imaginary case does not sound so important.

zabbal: I disagree: I think this case sould be explicitly tested for in tuf-test-vectors (at least) and hence logging will come in handy.

src/aktualizr_secondary/aktualizr_secondary.cc
mike-sul: > Do we already have a ticket for this?
No, I can create one so we can discuss a need in multiple hashes and their verification in scope of that ticket.

@zabbal  @lbonn What do you think I should do specifically in this pull request ?

Leave it as it is now
Mimic the current primary's logic, i.e. get the first found hash in the target,json, generate corresponding hash for received data and check whether the hashes matches

I/We can change the hash verification logic for both Primary and Secondary once we discuss and agree on it in the scope of another ticket and a pull request, just would like to unblock merging of this PR.

src/aktualizr_secondary/aktualizr_secondary.cc
mike-sul: No problem, will add more details into the log message.

src/aktualizr_secondary/uptane_verification_test.cc
mike-sul: ok.

src/aktualizr_secondary/aktualizr_secondary.cc
mike-sul: I've just double check this 'stuff' and it turned out that I was wrong when I was saying that there are two hashes in target.json for a binary image file. It turned out that the BE in the prod actually sends just one hash - sha256
 
`"targets": {
      "whatever_0.1": {
        "hashes": {
          "sha256": "30e14955ebf1352266dc2ff8067e68104607e750abb9d3b36582b8af909fcb58"
        },
`
The reason I was the original statement about the two hashes is the fact that 'uptane-genertor' actually generates two type of hashes

`"update_100.0": {
        "custom": {
          "ecuIdentifiers": {
            "some-serial": {
              "hardwareId": "secondary-ecu-hdw-ID-001"
            }
          },
          "targetFormat": "BINARY"
        },
        "hashes": {
          "sha256": "a866dd69dc84edc2991085f7a36bbd59f473ce4c8c18557369891bb5d4807934",
          "sha512": "d4f562d1d23282b70a0d67a25474b632f184321f5fd0abede9f4b0ae1ef2c260ef43ea4b06e83b28d05dcb527423156049247f99e978bd4633ffb6ab228fd736"
        },`


Therefore, I am going to apply to Secondary the same hash verification logic as in the case of Primary, which is check the first found hash in targets.json.

mike-sul:
@zabbal @lbonn Updated PR according to the comments.

PR #1474, opened 2019-12-05
Send config at boot

lbonn:


src/libaktualizr/http/httpclient.h
```diff
@@ -45,14 +45,6 @@ class HttpClient : public HttpInterface {
...
  private:
   FRIEND_TEST(GetTest, download_speed_limit);
-  /**
-   * These are here to catch a common programming error where a Json::Value is
-   * implicitly constructed from a std::string. By having an private overload
-   * that takes string (and with no implementation), this will fail during
-   * compilation.
-   */
-  HttpResponse post(const std::string &url, std::string data);
-  HttpResponse put(const std::string &url, std::string data);
```
patrickvacek: These are also in httpfake.h BTW.

src/libaktualizr/telemetry/telemetryconfig.h
```diff
@@ -8,9 +8,10 @@ struct TelemetryConfig {
...
   bool report_network{true};
+  bool report_config{true};
 
   void updateFromPropertyTree(const boost::property_tree::ptree& pt);
   void writeToStream(std::ostream& out_stream) const;
 };
 
-#endif  // TELEMETRY_TELEMETRY_CONFIG_H_
\ No newline at end of file
+#endif  // TELEMETRY_TELEMETRY_CONFIG_H_
```
zabbal: Not specific to this particular PR, just reminded about thing I wanted to ask: is there reason why we don't use "#pragma once"?

eu-smirnov: It's not in the standard, so why should we use it?

zabbal: It's so widely supported I though it's part of the standard already. Good to know.

lbonn: Yes, according to wikipedia it's basically supported everywhere.

But I suppose the main reason we don't use it, is just plain inertia: the benefits might or might not be offseted by the amount of work it takes to convert existing files and check that it works well for all our use cases.

src/libaktualizr/http/httpclient.h
lbonn: Thanks!

lbonn:
@eu-smirnov I've remove the default implementation of the json api in HttpInterface. Overloading in HttpClient and HttpFake is now more natural.

lbonn:
A very stupid copy-paste typo was failing the tests... It should be good by now.

PR #1480, opened 2019-12-10
Document HSM integration with Yocto

lbonn:
As it currently stands on meta-updater master.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/hsm-provisioning-example.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/hsm-provisioning-example.adoc
@@ -123,6 +123,10 @@ At this point, you can verify that your device contains the three objects by run
 +
 ----
 Using slot 1 with a present token (0x4)
+Public Key Object; RSA 2048 bits
+  label:      uptanekey
+  ID:         03
+  Usage:      encrypt, verify, wrap
 Public Key Object; EC  EC_POINT 256 bits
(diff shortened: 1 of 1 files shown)


docs/ota-client-guide/modules/ROOT/pages/hsm-provisioning-example.adoc
```diff
@@ -182,7 +183,24 @@ key_source = "pkcs11"
...
+----
+IMAGE_INSTALL_append = " opensc"
+SOTA_CLIENT_FEATURES = "hsm"
+SOTA_CLIENT_PROV = "aktualizr-device-prov-hsm"
+SOTA_DEPLOY_CREDENTIALS = "0"
+----
+
+Also, the configuration fragment `/usr/lib/sota/conf.d/20-sota-device-cred-hsm.toml` will also have to be modified, as detailed in the previous section, for example with a `.bbappend`.
+
+Note that for the moment, the gateway url and root certificate will still need to be copied manually to the device.
```
tkfu: Is this a hard limitation for some reason, or could we expect that someone using this in production would be able to write (or modify) a recipe to include these in the build?

lbonn: There is no technical obstacle as far as I know. And, indeed, integrators will have to choose how to pass that down to the devices. Putting it inside the image directly should also be fine as it's not secret and not device-specific.

PR #1481, opened 2019-12-10
curl_easy_init call was changed to thread-safe mode

mchekhovoi:
Signed-off-by: Chekhovoi, Mykola <ext-mykola.chekhovoi@here.com>

### DIFF
--- a/src/libaktualizr/http/httpclient.cc
+++ b/src/libaktualizr/http/httpclient.cc
@@ -35,8 +35,14 @@ static size_t writeString(void* contents, size_t size, size_t nmemb, void* userp
   return size * nmemb;
 }
 
+std::mutex HttpClient::curl_init_mutex;
+
 HttpClient::HttpClient(const std::vector<std::string>* extra_headers) {
-  curl = curl_easy_init();
+  {
+    std::lock_guard<std::mutex> guard(HttpClient::curl_init_mutex);
+    curl = curl_easy_init();
+  }
+
   if (curl == nullptr) {
     throw std::runtime_error("Could not initialize curl");
   }
--- a/src/libaktualizr/utilities/utils.cc
+++ b/src/libaktualizr/utilities/utils.cc
@@ -883,8 +883,13 @@ int Socket::connect() {
   return ::connect(socket_fd_, reinterpret_cast<const struct sockaddr *>(&sock_address_), sizeof(sock_address_));
 }
 
+std::mutex CurlEasyWrapper::curl_init_mutex;
+
 CurlEasyWrapper::CurlEasyWrapper() {
-  handle = curl_easy_init();
+  {
+    std::lock_guard<std::mutex> guard(CurlEasyWrapper::curl_init_mutex);
+    handle = curl_easy_init();
+  }
   if (handle == nullptr) {
     throw std::runtime_error("Could not initialize curl handle");
   }
--- a/src/libaktualizr/http/httpclient.h
+++ b/src/libaktualizr/http/httpclient.h
@@ -50,6 +50,7 @@ class HttpClient : public HttpInterface {
 
   static CurlGlobalInitWrapper manageCurlGlobalInit_;
   CURL *curl;
+  static std::mutex curl_init_mutex;
   curl_slist *headers;
   HttpResponse perform(CURL *curl_handler, int retry_times, int64_t size_limit);
   static curl_slist *curl_slist_dup(curl_slist *sl);
--- a/src/libaktualizr/utilities/utils.h
+++ b/src/libaktualizr/utilities/utils.h
@@ -3,6 +3,7 @@
 
 #include <boost/filesystem.hpp>
 #include <memory>
+#include <mutex>
 #include <string>
 
 #include <curl/curl.h>
@@ -156,6 +157,7 @@ class CurlEasyWrapper {
 
  private:
   CURL *handle;
+  static std::mutex curl_init_mutex;
 };
 
 template <typename... T>
--- a/src/sota_tools/ostree_object.cc
+++ b/src/sota_tools/ostree_object.cc
@@ -181,9 +181,15 @@ void OSTreeObject::QueryChildren(RequestPool &pool) {
 
 string OSTreeObject::Url() const { return "objects/" + object_name_; }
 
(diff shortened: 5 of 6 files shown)


src/libaktualizr/http/httpclient.cc
```diff
@@ -35,8 +35,14 @@ static size_t writeString(void* contents, size_t size, size_t nmemb, void* userp
   return size * nmemb;
 }
 
+std::mutex HttpClient::curl_init_mutex;
+
 HttpClient::HttpClient(const std::vector<std::string>* extra_headers) {
-  curl = curl_easy_init();
+  {
```
lbonn: Is it necessary here? If `curl_global_init()` was called before, `curl_easy_init()` should be thread safe: https://curl.haxx.se/libcurl/c/threadsafe.html

src/libaktualizr/utilities/utils.cc
```diff
@@ -883,8 +883,13 @@ int Socket::connect() {
   return ::connect(socket_fd_, reinterpret_cast<const struct sockaddr *>(&sock_address_), sizeof(sock_address_));
 }
 
+std::mutex CurlEasyWrapper::curl_init_mutex;
+
 CurlEasyWrapper::CurlEasyWrapper() {
-  handle = curl_easy_init();
+  {
```
lbonn: It could also be avoided if curl_global_init is called.
In the worst case, the mutex could also be removed from the class definition and just declared as static in the method.

eu-smirnov:
The problem is that we crash if libaktualizr instances are created from multiple threads.

I didn't notice that we already have this https://github.com/advancedtelematic/aktualizr/blob/7a68cd5877d0d205797d62357cea1c6e1680bf29/src/libaktualizr/http/httpclient.h#L20
presumably to avoid the problem.
This class shouldn't work, however, as there is no synchronization and curl_global_init() will be called concurrently. 

I would suggest changing this PR to fix the problem in CurlGlobalInitWrapper class.

lbonn:
> This class shouldn't work, however, as there is no synchronization and curl_global_init() will be called concurrently.
> 
> I would suggest changing this PR to fix the problem in CurlGlobalInitWrapper class.

But an instance of `CurlGlobalInitWrapper` is defined as a static member of `HttpClient`, so it should only be constructed once, at the beginning of the program.

```
(gdb) b CurlGlobalInitWrapper::CurlGlobalInitWrapper()
Breakpoint 1 at 0x149530: file ../src/libaktualizr/http/httpclient.h, line 20.
(gdb) r
Starting program: /home/lbonnans/ats/aktualizr/build/src/aktualizr_primary/aktualizr -c .
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Breakpoint 1, CurlGlobalInitWrapper::CurlGlobalInitWrapper (this=0x5555557a02f8 <HttpClient::manageCurlGlobalInit_>) at ../src/libaktualizr/http/httpclient.h:20
20        CurlGlobalInitWrapper() { curl_global_init(CURL_GLOBAL_DEFAULT); }
(gdb) bt
#0  0x000055555569d530 in CurlGlobalInitWrapper::CurlGlobalInitWrapper() (this=0x5555557a02f8 <HttpClient::manageCurlGlobalInit_>) at ../src/libaktualizr/http/httpclient.h:20
#1  0x000055555569d530 in __static_initialization_and_destruction_0(int, int) (__initialize_p=1, __priority=65535) at ../src/libaktualizr/http/httpclient.cc:71
#2  0x00005555556a40e0 in _GLOBAL__sub_I_httpclient.cc(void) () at ../src/libaktualizr/http/httpclient.cc:281
#3  0x0000555555728b6d in __libc_csu_init ()
#4  0x00007ffff6e3b16e in __libc_start_main (main=0x55555558f28a <main(int, char**)>, argc=3, argv=0x7fffffffd418, init=0x555555728b20 <__libc_csu_init>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffd408) at ../csu/libc-start.c:264
#5  0x000055555558e1ce in _start ()
```

eu-smirnov:
True, I mistakenly thought that it will be called on first use. Then it's not clear at all why this problem is there.

patrickvacek:
> The problem is that we crash if libaktualizr instances are created from multiple threads.

Do we actually (want to) support that?

lbonn:
@patrickvacek I suppose it would be used to run multi-threaded load tests

It looks fine to me for users of HttpClient at least, so libaktualizr. Code that use the CurlEasyWrapper (sota tools) might need another check but I suppose we won't need to support running multiple instances in threads there.

patrickvacek:
> @patrickvacek I suppose it would be used to run multi-threaded load tests

Fair enough. Just making sure it was worthwhile.

> sota tools might need another check but I suppose we won't need to support running multiple instances in threads there

We do use multiple threads to send requests to the server in garage-push/deploy, though.

eu-smirnov:
We won't go with multithreaded load tests for now, as another common resource that all instances share is logger and it requires significant amount of refactoring to get it right.
Since we've already encountered this bug, it would be nice to understand what is happening here, but we shouldn't spend too much time on that.

lbonn:
Ok, my guess is that the static objects are not properly initialized when using libaktualizr-c with rust. This would explain why `curl_global_init` is not call at startup and would cause thread safety issues in `curl_easy_init`.

But I don't think this PR is the correct way to fix this...

mchekhovoi:
> Ok, my guess is that the static objects are not properly initialized when using libaktualizr-c with rust. This would explain why `curl_global_init` is not call at startup and would cause thread safety issues in `curl_easy_init`.
> 
> But I don't think this PR is the correct way to fix this...

Yes, I have explained problems exclusively with calls from Rust, and not on C++ multithread.
But I have added logs locally and I see curl_global_init is called on start (with Rust). Only using of mutex lock near curl_easy_init helps to avoid crashes. So now the reason of problem is not clear for me.

patrickvacek:
Is this still needed and/or in progress, or can we close this PR?

mchekhovoi:
> Is this still needed and/or in progress, or can we close this PR?

Can be closed.
Thanks.

PR #1485, opened 2019-12-11
OLPSUP-8795: Import the initial hash from file for IP Secondary

mike-sul:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Our versioning scheme is `YEAR.N` where `N` is incremented whenever a new releas
 
 ## [??? (unreleased)]
 
-## [2019.11] - 2019-12-11
+## [2019.11] - 2019-12-12
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


lbonn:
Nice. Do you think adding a non-regression test would be possible here?

mike-sul:
> Nice. Do you think adding a non-regression test would be possible here?

Everything is possible :). The question is if it is reasonable. I am just not sure if the current approach to setting of the initial version is the right one. If not, then we have to rework it, so adding test to something that we are going to throw away does not make sense, otherwise I can add some test. Let's discuss it tomorrow during the meeting.

patrickvacek:
To summarize discussion: testing is of course a great idea, but since we think we want to redo this logic really soon anyway, perhaps a test for this quick-fix logic is not such a great use of our time.

mike-sul:
@lbonn @patrickvacek  please, check out if I updated the changelog correctly.

PR #1486, opened 2019-12-12
feat/ota-3196/Report aktualizr unit test results on gitlab

Zee314159:
Draft code. Just to trigger a CI test.

### DIFF
--- a/cmake-modules/CodeCoverage.cmake
+++ b/cmake-modules/CodeCoverage.cmake
@@ -156,7 +156,7 @@ function(SETUP_TARGET_FOR_COVERAGE_LCOV)
         COMMAND ${LCOV_PATH} --gcov-tool ${GCOV_PATH} -c -i -d . -o ${Coverage_NAME}.base
 
         # Run tests
-        COMMAND ${Coverage_EXECUTABLE}
+        COMMAND ${Coverage_EXECUTABLE} ${Coverage_EXECUTABLE_ARGS}
 
         # Capturing lcov counters and generating report
         COMMAND ${LCOV_PATH} --gcov-tool ${GCOV_PATH} --directory . --capture --output-file ${Coverage_NAME}.info
--- a/.ort.yml
+++ b/.ort.yml
@@ -12,6 +12,10 @@ excludes:
   - pattern: "third_party/HdrHistogram_c/**"
     reason: "TEST_TOOL_OF"
     comment: "This directory contains support code for tests which are not distributed."
+  - pattern: "third_party/junit/ctest2junit.xsl"
+    reason: "TEST_TOOL_OF"
+    comment: "This file contains support code for tests which are not distributed."
   - pattern: "thirdparty.spdx"
     reason: "TEST_TOOL_OF"
     comment: "This file contains metadata which are not distributed."
+
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -92,9 +92,12 @@ coverage:
   artifacts:
     paths:
       - build-coverage/coverage/
+    reports:
+      junit: build-coverage/report.xml
   script:
     - aws s3 cp s3://ota-gitlab-ci/hereotaconnect_prod.zip $CI_PROJECT_DIR/credentials.zip
     - ./scripts/test.sh
+    - xsltproc -o build-coverage/report.xml ./third_party/junit/ctest2junit.xsl build-coverage/Testing/**/Test.xml > /dev/null
 
 nop11:
   variables:
@@ -136,8 +139,11 @@ debian-build+static:
   artifacts:
     paths:
       - build-debian-testing/docs/doxygen/
+    reports:
+      junit: build-debian-testing/report.xml
   script:
     - ./scripts/test.sh
+    - xsltproc -o build-debian-testing/report.xml ./third_party/junit/ctest2junit.xsl build-debian-testing/Testing/**/Test.xml > /dev/null
 
 bionic-pkg:
   variables:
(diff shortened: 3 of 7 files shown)


Zee314159:
This failure `/aktualizr/CONTRIBUTING.md:41: error: unable to resolve reference to 'CHANGELOG.md' for \ref command (warning treated as error, aborting now)`  seems not caused by me. The need xml files have been generated, still waiting for GitLab administrator to enable the `junit_pipeline_view` feature, but I'm not sure if this is the right solution.

src/libaktualizr/uptane/uptane_network_test.cc
```diff
@@ -203,7 +203,7 @@ TEST(UptaneNetwork, LogConnectivityRestored) {
   http->connectSwitch = true;
   testing::internal::CaptureStdout();
   result = up->fetchMeta();
-  EXPECT_EQ(result.status, result::UpdateStatus::kUpdatesAvailable);
+  EXPECT_EQ(result.status, result::UpdateStatus::kError);
```
patrickvacek: This doesn't seem right. Was something going wrong here?

patrickvacek:
> This failure `/aktualizr/CONTRIBUTING.md:41: error: unable to resolve reference to 'CHANGELOG.md' for \ref command (warning treated as error, aborting now)` seems not caused by me.

Agreed. Perhaps a rebase will help, since I think this was a doxygen issue that Laurent recently fixed.

> The needed xml files have been generated, still waiting for GitLab administrator to enable the `junit_pipeline_view` feature, but I'm not sure if this is the right solution.

Looks like it could work! Do you have an example or a link to the generated output so we can take a look?

src/libaktualizr/uptane/uptane_network_test.cc
Zee314159: Nothing's wrong. Just make a failure case to test the GitLab report feature.

Zee314159:
> Do you have an example or a link to the generated output so we can take a look?

https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/jobs/1080837/artifacts/browse/build-coverage/Testing/20191219-0609/
`junit_pipeline_view` feature has been enabled, but the there's no report showed. I added `junit` in docker files, it's all the same, not sure what's going wrong.

Zee314159:
https://na-stage.gitlab.in.here.com/olrudenk/temp_traffic_mobile_application/pipelines/69/test_report 
This project works fine on gitlab-stg.

docker/Dockerfile.ubuntu.xenial
```diff
@@ -22,6 +22,7 @@ RUN apt-get update && apt-get -y install --no-install-suggests --no-install-reco
   gcc \
   git \
   jq \
+  junit \
```
patrickvacek: I think you should only need to add `junit` for the bionic image. I don't think it'll be used by the xenial or Debian testing images.

cmake-modules/CodeCoverage.cmake
```diff
@@ -156,7 +156,7 @@ function(SETUP_TARGET_FOR_COVERAGE_LCOV)
         COMMAND ${LCOV_PATH} --gcov-tool ${GCOV_PATH} -c -i -d . -o ${Coverage_NAME}.base
 
         # Run tests
-        COMMAND ${Coverage_EXECUTABLE}
+        COMMAND ${Coverage_EXECUTABLE} ${Coverage_EXECUTABLE_ARGS}
```
lbonn: Curious why is this required?

lbonn:
I think the problem is that the xml output from ctest is not compatible with junit and needs to be converted. See for example this cmake issue https://gitlab.kitware.com/cmake/cmake/issues/18654.

docker/Dockerfile.ubuntu.xenial
lbonn: Actually, do we need junit at all on the images?

Zee314159: Yes, it's a long shot... seems not needed, tried it anyway.

cmake-modules/CodeCoverage.cmake
Zee314159: `set(CTEST_EXTRA_ARGS -T Test --no-compress-output)`
`setup_target_for_coverage_lcov(NAME coverage EXECUTABLE ${CMAKE_CTEST_COMMAND} ${CTEST_EXTRA_ARGS} USES_TERMINAL)`
from `aktualizr/tests/CMakeLists.txt` the `Test.xml` is only generated when this argument is set.

Zee314159:
> I wish it didn't involve pulling in php, though.

Yes, It looks php didn't do anything to help. I'll remove it see if it still works.

Zee314159:
> I wish it didn't involve pulling in php, though.
> 
> Yes, It looks php didn't do anything to help. I'll remove it see if it still works.

Zee314159:
> I wish it didn't involve pulling in php, though.
> 
> Yes, It looks php didn't do anything to help. I'll remove it see if it still works.

Zee314159:
> I wish it didn't involve pulling in php, though.
> 
> Yes, It looks php didn't do anything to help. I'll remove it see if it still works.

Fixed.

Zee314159:
> Cool, glad to see this simplified. Where did the .xsl file come from? We need to be sure the copyright is friendly.

https://gitlab.kitware.com/cmake/cmake/issues/18654 
We need to add a description in `thirdparty.spdx` right?

patrickvacek:
> https://github.com/manticoresoftware/manticoresearch/tree/master/misc/junit
> We need to add a description in `thirdparty.spdx` right?

Yep. It's GPL, but since it's only used for CI, I think it's fine. You'll also need to add an entry to `.ort.yml` to exclude it.

PR #1491, opened 2019-12-16
Do not provision if primary times out while connecting to secon…

lbonn:


lbonn:
Unfortunately, an update of doxygen seems to have broken the docs generation. I'll merge anyway because this was already broken.

We'll have to fix that in another PR quite soon.

src/aktualizr_primary/main.cc
```diff
@@ -122,11 +122,14 @@ int main(int argc, char *argv[]) {
     conn = aktualizr.SetSignalHandler(f_cb);
 
     if (!config.uptane.secondary_config_file.empty()) {
-      if (boost::filesystem::exists(config.uptane.secondary_config_file)) {
+      try {
```
mike-sul: This try/catch was not necessary as this code is already wrapped by the outter try/catch.

src/aktualizr_primary/main.cc
```diff
@@ -122,11 +122,14 @@ int main(int argc, char *argv[]) {
...
         Primary::initSecondaries(aktualizr, config.uptane.secondary_config_file);
-      } else {
-        LOG_WARNING << "The specified secondary config file does not exist: " << config.uptane.secondary_config_file
-                    << "\nProceed further without secondary(ies)";
+      } catch (const std::exception &e) {
+        LOG_ERROR << "Secondary initialization failed";
```
mike-sul: I would add e.what() to the log message so we see more details of the secondary provisioning failure.

src/aktualizr_primary/main.cc
lbonn: The exception is ignored if `aktualizr.IsRegistered()`, I think we cannot rely on the outer try/catch in this case.

src/aktualizr_primary/main.cc
lbonn: In the current version, `e.what()` is logged in `initSecondaries()` but I agree this is maybe not really clear.
`initSecondaries()` could also log these errors and return a boolean, to simplify the call site in `main.cc`.

src/aktualizr_primary/main.cc
mike-sul: ok. I see, this is for the use case when Primary and Secondaries are trying to connect to each other after a device has been already provisioned, e.g. just a device and/or aktualizr reboot. I think, it might be not an optimal strategy to exit the connection procedure on a first failure if a device has been already registered, but, I suppose, it's not so important, adding secondaries dynamically is more important or not trying to connect to them on a startup at all if  a device already registered.

src/aktualizr_primary/main.cc
mike-sul: I see, makes sense, it's up to you whether to terminate exceptions in initSecondaries and return boolean or rethrow them.

src/aktualizr_primary/main.cc
lbonn: > I think, it might be not an optimal strategy to exit the connection procedure on a first failure if a device has been already registered, but, I suppose, it's not so important, adding secondaries dynamically is more important or not trying to connect to them on a startup at all if a device already registered.

That's a good point indeed... But I remember I chose this simple solution at first because right now, we can't really do better than what the IP secondaries config parser gives us and it handles all IP secondaries at once.

src/aktualizr_primary/secondary.cc
```diff
@@ -82,8 +81,10 @@ class SecondaryWaiter {
     timer_.async_wait([&](const boost::system::error_code& error_code) {
       if (!!error_code) {
         LOG_ERROR << "Wait for secondaries has failed: " << error_code;
+        throw std::runtime_error("Error while waiting for secondary");
```
mike-sul: As an alternative to these two 'throws' just one throw can be added at the end of createIPSecondaries().

PR #1492, opened 2019-12-17
Virtual Secondaries page added

ammaboateng:
Signed-off-by: amma <awuraaamma.okyere-boateng@here.com>

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
@@ -21,8 +21,7 @@ Virtual secondaries can be used as a way to update a configuration file on the p
 +
 You can use the provided `.deb` releases mentioned and the `virtualsec.json` file, which you can edit to suit your needs: such as removing one of the two entries and setting the paths and names to whatever you prefer.
 . Afterwards, you will be able to download the arbitrary files to the location specified in the `firmware_path` variable.
-. Upload the files for the virtual secondary to the https://connect.ota.here.com[OTA Connect Portal].
-. Now go to the *Software versions* tab and click *Add Software*. You should make sure the *ECU Types* field matches the value specified in the `ecu_hardware_id` variable of the `virtualsec.json` file.
+. Next, you can upload the files for the virtual secondary. To do so, go to the *Software versions* tab in the https://connect.ota.here.com[OTA Connect Portal] and click *Add Software*. You should make sure the *ECU Types* field matches the value specified in the `ecu_hardware_id` variable of the `virtualsec.json` file.
 +
 NOTE: `ecu_hardware_id` must be unique per each file type that you'll like to deliver, otherwise an update of one of them will rewrite the others.
 
@@ -33,9 +32,9 @@ To issue the update perform the following steps:
 
 . Go to your *Devices* tab and find your primary. (You can run `aktualizr-info` locally to get the name of the primary if you don't know it.)
 . Afterwards, select your virtual secondary and the file you uploaded to the *Software versions* tab.
-+
-(For future reference, you can use the *Campaigns* tab to send updates to multiple devices simultaneously.)
 . After your selection, the primary will begin to download the file.
++
+For future reference, you can use the *Campaigns* tab to send updates to multiple devices simultaneously.
 
 
 
--- a/docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
@@ -8,7 +8,7 @@ We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}
 endif::[]
 
 
-Virtual secondary devices can be used as a way to update a configuration file on the primary device (where the primary device is a standard OSTree image running an aktualizr). You can do this in two parts:
+Virtual secondaries can be used as a way to update a configuration file on the primary (where the primary is a standard OSTree image running an aktualizr). You can do this in two parts:
 
 * First you need a virtual secondary configured that can be used to update a configuration file.
 * Afterwards you can integrate it into your Yocto build.
@@ -20,36 +20,26 @@ Virtual secondary devices can be used as a way to update a configuration file on
 . You can create a virtual secondary to run locally by following the steps on how to xref:simulate-device-basic.adoc[simulate a device without building a disk image].
 +
 You can use the provided `.deb` releases mentioned and the `virtualsec.json` file, which you can edit to suit your needs: such as removing one of the two entries and setting the paths and names to whatever you prefer.
-. Afterwards, you will be able to download the arbitrary files to the location specified in the "firmware_path" variable, by uploading the files to the https://connect.ota.here.com[OTA Connect Portal].
-. Now go to the *Software versions* tab and click *Add Software*. You should make sure the *ECU Types* field matches the value specified in the "ecu_hardware_id" variable of the `virtualsec.json` file.
+. Afterwards, you will be able to download the arbitrary files to the location specified in the `firmware_path` variable.
+. Upload the files for the virtual secondary to the https://connect.ota.here.com[OTA Connect Portal].
+. Now go to the *Software versions* tab and click *Add Software*. You should make sure the *ECU Types* field matches the value specified in the `ecu_hardware_id` variable of the `virtualsec.json` file.
 +
-NOTE: "ecu_hardware_id" must be unique per each file type that you'll like to deliver, otherwise an update of one of them will rewrite the others.
+NOTE: `ecu_hardware_id` must be unique per each file type that you'll like to deliver, otherwise an update of one of them will rewrite the others.
 
 
 == Updating via Virtual Secondaries
 
 To issue the update perform the following steps:
 
-. Go to your *Devices* tab and find your primary device. (You can run 'aktualizr-info' locally to get the device name if you don't know it.)
+. Go to your *Devices* tab and find your primary. (You can run `aktualizr-info` locally to get the name of the primary if you don't know it.)
 . Afterwards, select your virtual secondary and the file you uploaded to the *Software versions* tab.
 +
 (For future reference, you can use the *Campaigns* tab to send updates to multiple devices simultaneously.)
+. After your selection, the primary will begin to download the file.
 
 
 
 == Integrating Virtual Secondaries into Yocto
 
-You can use the files in the repository https://github.com/advancedtelematic/meta-updater/pull/611[here] as an example to get the virtual secondaries to work in Yocto. You can use the files directly or you can copy them to one of your layers and modify them according to your needs.
-
-To use the example in the repository, you'll need to make sure that the relevant packages are added to your image. You can do so by adding this to your `local.conf`:
-----
-IMAGE_INSTALL_append += " aktualizr-virtualsec"
-----
-
-Currently, if you want to do this on any branch besides master, you'll need to also add the following to your `local.conf` to pull the latest version of aktualizr into your build:
-
-----
-require classes/sota_bleeding.inc
-----
-
+You can use the files in the repository https://github.com/advancedtelematic/meta-updater/blob/master/recipes-sota/config/aktualizr-virtualsec.bb[here] as an example to get the virtual secondaries to work in Yocto. You can use the files directly or you can copy them to one of your layers and modify them according to your needs.
 
--- a/docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
@@ -1,0 +1,55 @@
+= Use of Virtual Secondaries
+ifdef::env-github[]
+
+[NOTE]
+====
+We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}.html[view this article in our documentation portal]. Not all of our articles render correctly in GitHub.
+====
+endif::[]
+
+
+Virtual secondary devices can be used as a way to update a configuration file on the primary device (where the primary device is a standard OSTree image running an aktualizr). You can do this in two parts:
+
+* First you need a virtual secondary configured that can be used to update a configuration file.
+* Afterwards you can integrate it into your Yocto build.
+
+
+
+== Creating a Virtual Secondary
+
+. You can create a virtual secondary to run locally by following the steps on how to xref:simulate-device-basic.adoc[simulate a device without building a disk image].
++
+You can use the provided `.deb` releases mentioned and the `virtualsec.json` file, which you can edit to suit your needs: such as removing one of the two entries and setting the paths and names to whatever you prefer.
+. Afterwards, you will be able to download the arbitrary files to the location specified in the "firmware_path" variable, by uploading the files to the https://connect.ota.here.com[OTA Connect Portal].
+. Now go to the *Software versions* tab and click *Add Software*. You should make sure the *ECU Types* field matches the value specified in the "ecu_hardware_id" variable of the `virtualsec.json` file.
++
+NOTE: "ecu_hardware_id" must be unique per each file type that you'll like to deliver, otherwise an update of one of them will rewrite the others.
+
+
+== Updating via Virtual Secondaries
+
+To issue the update perform the following steps:
+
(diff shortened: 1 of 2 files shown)


mike-sul:
[1/1] Generating Doxygen Documentation

FAILED: docs/doxygen/CMakeFiles/doxygen 

cd /aktualizr/build-test/docs/doxygen && /usr/bin/doxygen Doxyfile

warning: Tag 'PERL_PATH' at line 2195 of file 'Doxyfile' has become obsolete.

         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"

warning: Tag 'MSCGEN_PATH' at line 2217 of file 'Doxyfile' has become obsolete.

         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"

/aktualizr/CONTRIBUTING.md:41: error: unable to resolve reference to 'CHANGELOG.md' for \ref command (warning treated as error, aborting now)

ninja: build stopped: subcommand failed.

https://travis-ci.org/advancedtelematic/aktualizr/jobs/626125506?utm_medium=notification&utm_source=github_status

lbonn:
@mike-sul this appeared when the version of doxygen changed on CI and I hope to fix these with https://github.com/advancedtelematic/aktualizr/pull/1493. I think we can ignore these errors in the meantime.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+To use the example in the repository, you'll need to make sure that the relevant packages are added to your image. You can do so by adding this to your `local.conf`:
+----
+IMAGE_INSTALL_append += " aktualizr-virtualsec"
+----
+
+Currently, if you want to do this on any branch besides master, you'll need to also add the following to your `local.conf` to pull the latest version of aktualizr into your build:
+
+----
+require classes/sota_bleeding.inc
```
mike-sul: AFAIK, the baking of a specific aktualizr git branch/revision requires more than adding the given line, see here https://github.com/advancedtelematic/aktualizr/blob/master/docs/ota-client-guide/modules/ROOT/pages/build-configuration.adoc#custom-aktualizr-versions for more details.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,55 @@
...
+. Go to your *Devices* tab and find your primary device. (You can run 'aktualizr-info' locally to get the device name if you don't know it.)
+. Afterwards, select your virtual secondary and the file you uploaded to the *Software versions* tab.
++
+(For future reference, you can use the *Campaigns* tab to send updates to multiple devices simultaneously.)
+
+
+
+== Integrating Virtual Secondaries into Yocto
+
+You can use the files in the repository https://github.com/advancedtelematic/meta-updater/pull/611[here] as an example to get the virtual secondaries to work in Yocto. You can use the files directly or you can copy them to one of your layers and modify them according to your needs.
```
mike-sul: @patrickvacek @xaviergully It doesn't look like a convenient way to add a custom update of a desired file by means of VirtualSecondary.

The current approach:

1. requires specifying non-important parameters, such as ecu_private_key, ecu_public_key, metadata_path, etc while a customer would be interested in just two of them in most cases, i.e. 'firmware_path' and 'ecu_hardware_id';

2. implies either modifying of the existing recipes or creating another layer and copy-pasting of the existing recipes.

Therefore, I'd suggest to do the following:

1. review and refactor the way Virtual Secondary are configured in order to minimize a number of parameters a customer needs to define;

2. review and refactor the virtual secondary recipes so a customer could specify just
     - a local.conf variable (e.g. VIRTUAL_SECONDARY_CONF) pointing to a customer's virtual secondary conf;
     - local.conf variables defining 'firmware_path' (and optionally ecu_hardware_id) and the virtual secondary recipes will do the rest of magic (e.g. auto-generate a virtual secondary configuration file).

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,55 @@
...
+* First you need a virtual secondary configured that can be used to update a configuration file.
+* Afterwards you can integrate it into your Yocto build.
+
+
+
+== Creating a Virtual Secondary
+
+. You can create a virtual secondary to run locally by following the steps on how to xref:simulate-device-basic.adoc[simulate a device without building a disk image].
++
+You can use the provided `.deb` releases mentioned and the `virtualsec.json` file, which you can edit to suit your needs: such as removing one of the two entries and setting the paths and names to whatever you prefer.
```
patrickvacek:
```suggestion
You can use the provided `.deb` releases of aktualizr as well as the `virtualsec.json` file, which you can edit to suit your needs: such as removing one of the two entries and setting the paths and names to whatever you prefer.
```

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,55 @@
...
+ifdef::env-github[]
+
+[NOTE]
+====
+We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}.html[view this article in our documentation portal]. Not all of our articles render correctly in GitHub.
+====
+endif::[]
+
+
+Virtual secondary devices can be used as a way to update a configuration file on the primary device (where the primary device is a standard OSTree image running an aktualizr). You can do this in two parts:
```
patrickvacek: "Device" is a word we have to be careful about, since usually we use the word to refer to an entire system, e.g. a car. I think we're trying to move away from referring to individual units as "ECUs" since that is quite automotive-specific, but I'm not sure what a better term is. You might be able to just say "the primary".

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,55 @@
...
+* Afterwards you can integrate it into your Yocto build.
+
+
+
+== Creating a Virtual Secondary
+
+. You can create a virtual secondary to run locally by following the steps on how to xref:simulate-device-basic.adoc[simulate a device without building a disk image].
++
+You can use the provided `.deb` releases mentioned and the `virtualsec.json` file, which you can edit to suit your needs: such as removing one of the two entries and setting the paths and names to whatever you prefer.
+. Afterwards, you will be able to download the arbitrary files to the location specified in the "firmware_path" variable, by uploading the files to the https://connect.ota.here.com[OTA Connect Portal].
```
patrickvacek:
```suggestion
. Afterwards, you will be able to download the arbitrary files to the location specified in the `firmware_path` variable, by uploading the files to the https://connect.ota.here.com[OTA Connect Portal].
```
Is this step necessary, though? As it is, we mention uploading twice, and this sentence seems just a bit confusing to me. Maybe it should be a higher-level comment.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,55 @@
...
+. Now go to the *Software versions* tab and click *Add Software*. You should make sure the *ECU Types* field matches the value specified in the "ecu_hardware_id" variable of the `virtualsec.json` file.
++
+NOTE: "ecu_hardware_id" must be unique per each file type that you'll like to deliver, otherwise an update of one of them will rewrite the others.
+
+
+== Updating via Virtual Secondaries
+
+To issue the update perform the following steps:
+
+. Go to your *Devices* tab and find your primary device. (You can run 'aktualizr-info' locally to get the device name if you don't know it.)
```
patrickvacek:
```suggestion
. Go to your *Devices* tab and find your primary device. (You can run `aktualizr-info` locally to get the device name if you don't know it.)
```

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,55 @@
...
+== Integrating Virtual Secondaries into Yocto
+
+You can use the files in the repository https://github.com/advancedtelematic/meta-updater/pull/611[here] as an example to get the virtual secondaries to work in Yocto. You can use the files directly or you can copy them to one of your layers and modify them according to your needs.
+
+To use the example in the repository, you'll need to make sure that the relevant packages are added to your image. You can do so by adding this to your `local.conf`:
+----
+IMAGE_INSTALL_append += " aktualizr-virtualsec"
+----
+
+Currently, if you want to do this on any branch besides master, you'll need to also add the following to your `local.conf` to pull the latest version of aktualizr into your build:
```
patrickvacek: This is no longer true; you can remove this line and the following bit.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
patrickvacek: It depends, but this can actually be removed now, since it's been backported into our other supported branches.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
patrickvacek: @mike-sul It's true, this isn't necessarily the easiest way to do this, and we can probably do better. For now, I think we should document how it works at present, and then we can come back later and iteratively improve what we have when we know more about what would be useful. Feel free to make a ticket if you'd like to get that conversation started.

@ammaboateng I think this is mostly fine as is, although I think we should change the link. It's currently pointing at a PR, which might be less helpful than pointing at the relevant files in the current repo instead.

ammaboateng: @patrickvacek can you please send me the link to the files that it should point to.

patrickvacek: Sure. The files changed in the PR are a pretty guide. The most important is `recipes-sota/config/aktualizr-virtualsec.bb`, and that links to the other two relevant files, so that's probably enough for now.

patrickvacek: Link: https://github.com/advancedtelematic/meta-updater/blob/master/recipes-sota/config/aktualizr-virtualsec.bb

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
ammaboateng: Should I remove this step about downloading files to that file path and only mention uploading to the OTA Connect portal?

patrickvacek: I think it's good to mention the part about downloading to that path, but I'd suggest moving it. I think it's just that mentioning the uploading and downloading in one sentence suggests that this happens simultaneously, when in fact they are two separate steps.

I see the steps of the process (at a high level) as something like this:
1. Configure the virtual secondary (which could include the bit about the download path).
2. Provision the device with the secondary.
3. Upload files for the secondary to the web UI.
4. Issuing the update/campaign.
5. Let the device download the file, which will probably happen automatically, but should be verified if desired or for demonstration purposes. (This might also include the bit about the path.)

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+
+== Creating a Virtual Secondary
+
+. You can create a virtual secondary to run locally by following the steps on how to xref:simulate-device-basic.adoc[simulate a device without building a disk image].
++
+You can use the provided `.deb` releases mentioned and the `virtualsec.json` file, which you can edit to suit your needs: such as removing one of the two entries and setting the paths and names to whatever you prefer.
+. Afterwards, you will be able to download the arbitrary files to the location specified in the `firmware_path` variable.
+. Upload the files for the virtual secondary to the https://connect.ota.here.com[OTA Connect Portal].
+. Now go to the *Software versions* tab and click *Add Software*. You should make sure the *ECU Types* field matches the value specified in the `ecu_hardware_id` variable of the `virtualsec.json` file.
```
patrickvacek: These two steps ("Upload the files..." and "Now go...") look to me like the same thing. The first one is just the high-level description and the second one has the details. They can probably be combined on one line with just minor changes to the transition words.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,45 @@
...
++
+NOTE: `ecu_hardware_id` must be unique per each file type that you'll like to deliver, otherwise an update of one of them will rewrite the others.
```
patrickvacek: I understand this, but I'm not sure people new to the system will. Do we have a explanation somewhere that we can link to that describes why it works this way?

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+
+== Updating via Virtual Secondaries
+
+To issue the update perform the following steps:
+
+. Go to your *Devices* tab and find your primary. (You can run `aktualizr-info` locally to get the name of the primary if you don't know it.)
+. Afterwards, select your virtual secondary and the file you uploaded to the *Software versions* tab.
++
+(For future reference, you can use the *Campaigns* tab to send updates to multiple devices simultaneously.)
```
patrickvacek: Should this "for future reference" line maybe get moved to after these steps? I think it might flow better to have the three steps together and then afterwards the separate line about campaigns, and then you won't need the parentheses, either.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
ammaboateng: @patrickvacek  good point. I will make the change.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
ammaboateng: @patrickvacek unfortunately we don't have extra information in our current documentation that explains this further. Maybe I could add the extra information here within the Note. 
Do you have any extra information on why it works this way which I could possibly add to the note to make it clearer for new users?

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
ammaboateng: @patrickvacek in that case should I make it a third point to read as: 
"You can use the *Campaigns* tab to send updates to multiple devices simultaneously" ?

patrickvacek: I don't think it counts as a point in the series of steps here, so to me the idea of it being a sidenote (as it is now) is fine. I just think it reads better to have it at the end instead of in the middle of these steps.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
patrickvacek: I think we can find a better place to describe that topic, and that can probably wait for another PR. I loosely understand conceptually why the system works that way, but in my opinion it is non-obvious, confusing, and possibly just plain wrong. I forget if the web UI at least says something about it now or not. In any case, you'd probably need to talk to a backend engineer or the product team to explain why it works that way.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
ammaboateng: @patrickvacek I get your point now. I will amend it.

docs/ota-client-guide/modules/ROOT/pages/virtual-secondaries.adoc
ammaboateng: @patrickvacek no problem. It was an extra point added by Mike within the description. 
Because this point is not yet fully explained in the doc, should I remove this note or do you think it is important to maintain the note for now and later write a more in-depth explanation on this?

patrickvacek: I think it's fine for now how you have it. Thanks!

PR #1494, opened 2019-12-18
Fix import path in tests

lbonn:
It should not use the default which reads in /var/sota/import.

ptest has been broken since 3515a2f3a984f7cd8cd68057f1113406686363d4, as
Aktualizr's constructor will call `importData` even if given a storage
object as an argument. It ended up importing the global installed
version running on the embedded system.

Before that, tests were side-stepping the import step, so the issue was
never revealed.


eu-smirnov:
Is it know why Travis is failing?

lbonn:
It looked transient but I can relaunch a test anyway

patrickvacek:
Nice, thanks! Have you confirmed that ptest works smoothly again?

lbonn:
@patrickvacek yes, I've done a manual test.

PR #1498, opened 2019-12-20
Pluggable primary package manager

lbonn:
This will allow external users of libaktualizr to add their own package managers without having to modify libaktualizr itself.

I went a bit fancy here, this is not a final version but more of a demo / food for thought.


doanac:
I really like this idea. I'm also a fan of using `__attribute__((constructor))` to help keep "registration" type things decoupled.

src/libaktualizr/package_manager/packagemanagerconfig.h
```diff
@@ -7,15 +7,20 @@
 
 #include "utilities/config_utils.h"
 
-enum class PackageManager { kNone = 0, kOstree, kDebian, kAndroid, kOstreeDockerApp };
-std::ostream& operator<<(std::ostream& os, PackageManager pm);
+#define PACKAGE_MANAGER_NONE "none"
+#define PACKAGE_MANAGER_OSTREE "ostree"
+#define PACKAGE_MANAGER_DEBIAN "debian"
+#define PACKAGE_MANAGER_ANDROID "android"
+#define PACKAGE_MANAGER_OSTREEDOCKERAPP "ostree+docker-app"
```
patrickvacek: This looks pretty cool! The only thing I don't like is that we still have to define the strings here directly in libaktualizr.

lbonn: Actually, that should not be required. It's just a convenience for the package manager that are bundled into libaktualizr (all of them right now). 

And from a quick look, even these could probably be moved to their corresponding headers.

src/libaktualizr/package_manager/packagemanagerfake.cc
```diff
@@ -1,7 +1,20 @@
...
 
 #include "utilities/fault_injection.h"
 
+extern "C" {
+__attribute__((constructor)) void packagemanagerfake_init(void);
+}
+
+void packagemanagerfake_init(void) {
+  PackageManagerFactory::registerPackageManager(
+      PACKAGE_MANAGER_NONE, [](PackageConfig pconfig, BootloaderConfig bconfig, std::shared_ptr<INvStorage> storage,
```
mike-sul: It imposes additional dependencies and policies on an user package manager as it has to depend on PackageConfig, BootloaderConfig, INvStorage and accept exactly these types of params in its ctor.
A registration of instances that share a common interface would be more flexible and loosely coupled.
Having said that, I don't think that this is really important here as PackageManagerInterface is not exactly an interface and it's already depends on the aforementioned stuff and actually requires those params in its ctor.

mike-sul:
I am not sure that a package manager's registry that can be populated by an user app solves any issues and fulfill any use-case & requirements that we've been discussed.
I thought, that we agreed to follow the approach #3 outlined here https://confluence.in.here.com/display/~mysul/Multi-type+update+support and this epic https://saeljira.it.here.com/browse/OTA-4070 actually implies it.

xcheng-here:
I really like this register mechanism. The only question I want to ask is that "constructor" and "destructor" are not standard attribute of C++, IIRC, and it seems it's the first time we use GNU extension  "__attribute__" in our source code. Is it already accepted by our coding style ?

lbonn:
> I am not sure that a package manager's registry that can be populated by an user app solves any issues and fulfill any use-case & requirements that we've been discussed.
> I thought, that we agreed to follow the approach #3 outlined here https://confluence.in.here.com/display/~mysul/Multi-type+update+support and this epic https://saeljira.it.here.com/browse/OTA-4070 actually implies it.

Well, nowhere it says that it would solve all of these issues and I've added in the description that this is more an unfinished PoC piece, so I do not know what you're trying to argue against. What this would allow in the short term is to move some of the existing package managers out of the main libaktualizr, other steps of the plan can be worked on before, after or in parallel depending on how we progress.

@xcheng-here 
> I really like this register mechanism. The only question I want to ask is that "constructor" and "destructor" are not standard attribute of C++, IIRC, and it seems it's the first time we use GNU extension "attribute" in our source code. Is it already accepted by our coding style ?

Indeed, that's a liberty that I took for this attempt. I preferred the non-standard C solution for now because it avoids name mangling issue of the required linker I had to use.

Alternatives I see are:

- pure C++ initialization in the constructor of a static object. The issues with name mangling could probably be avoided if package managers are moved out of libaktualizr itself.
- do not have this auto discovery mechanism and manually call `registerPackageManager` in application's initialization. Probably the simplest obvious solution, with the minor drawback that it will probably require a re-compilation when adding a new package manager instead of a simple link-time addition.

Also missing is the configuration handling. The package manager interface could have a method that accepts a "raw" key/value configuration fragment for its initialization. Right now, manager-specific fields are still leaking in libaktualizr's Config.

patrickvacek:
> > I am not sure that a package manager's registry that can be populated by an user app solves any issues and fulfill any use-case & requirements that we've been discussed.
> > I thought, that we agreed to follow the approach #3 outlined here https://confluence.in.here.com/display/~mysul/Multi-type+update+support and this epic https://saeljira.it.here.com/browse/OTA-4070 actually implies it.
> 
> Well, nowhere it says that it would solve all of these issues and I've added in the description that this is more an unfinished PoC piece, so I do not know what you're trying to argue against. What this would allow in the short term is to move some of the existing package managers out of the main libaktualizr, other steps of the plan can be worked on before, after or in parallel depending on how we progress.

I think this is a reasonable idea on how to solve one of the parts of the puzzle, specifically the subtask "Refactor package managers to allow custom implementations" (https://saeljira.it.here.com/browse/OTA-4072). It doesn't address most of the other issues, but we can add them on top of or around this. Or if the consensus really is that this isn't the right way to do this part, it can be redone as well.

mike-sul:
> I think this is a reasonable idea on how to solve one of the parts of the puzzle,

Yes, I think, that a registry will be needed at the end anyway, and the given registry design is good enough, IMHO The biggest piece of work around the puzzle is refactoring of sotauptaneclient, packagemanagerinterface and secondaries, this effort my have impact on the registry design so I thought that the registry itself should go last. But, I think, it's OK if it goes first, so it's kind of an iterative incremental development :)

doanac:
One thing that might be interesting for "pluggable" - I'm not totally sure its possible but if you look at my docker-app manager, it might be possible to implement it as a "ScriptPackageManager". The package manager could fork/exec a script to handle things. It could pass the script something like the target.json to install (as a file) and maybe the targets.json as a file as well. It could have sub commands like: `<script fetch>`, `<script verify>`, `<script install>`

I something generic like that might make it possible to get rid of the dockerapp, android, and deb package managers. I'm just not sure if there are places where the script might need to reach into libaktualizr to do things like interact with the bootloader? Anyway - something that might be worth thinking about.

lbonn:
Here is a new version that depends on the new changes, using dynamic libraries.

Also includes a way to add new configuration fields without having to modify the main `Config` object. This has been applied to DockerAppManager, as a PoC.

@doanac, does it look usable to you?

src/libaktualizr/package_manager/packagemanagerfactory.cc
```diff
@@ -1,55 +1,38 @@
...
-#endif
-    case PackageManager::kNone:
-      return std::make_shared<PackageManagerFake>(pconfig, bconfig, storage, http);
-    default:
-      LOG_ERROR << "Unrecognized package manager type: " << static_cast<int>(pconfig.type);
-      return std::shared_ptr<PackageManagerInterface>();  // NULL-equivalent
+    const PackageConfig &pconfig, const BootloaderConfig &bconfig, const std::shared_ptr<INvStorage> &storage,
+    const std::shared_ptr<HttpInterface> &http) {
+  for (const auto &b : *registered_pkgms_) {
+    if (b.first == pconfig.type.c_str()) {
```
doanac: I'm guessing there's something subtle with C++ here, but I'm not sure I see why you are call `.c_str()`?

lbonn: Haha no, it's just an artifact of a previous iteration. Good catch, thanks!

.clang-tidy
```diff
@@ -1,5 +1,5 @@
...
+Checks:          'clang-analyzer-*, performance-*,readability-*,-readability-else-after-return,-readability-identifier-naming,-readability-container-size-empty,-readability-redundant-declaration,modernize-*,-modernize-avoid-bind,-modernize-loop-convert,-modernize-use-using,-modernize-deprecated-headers,-modernize-pass-by-value,llvm-namespace-comment,cppcoreguidelines-pro-*,-cppcoreguidelines-pro-type-reinterpret-cast,-cppcoreguidelines-pro-type-vararg,-cppcoreguidelines-pro-type-const-cast,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-pro-bounds-constant-array-index,cert-*,-cert-err58-cpp,-cert-env33-c,-cert-err09-cpp,-cert-err60-cpp,-cert-err61-cpp, -cert-err61-cpp, bugprone-*, google-*,-google-runtime-references,-google-runtime-member-string-references, -google-explicit-constructor, -google-readability-todo, misc-*, -misc-unused-parameters, -misc-macro-parentheses, -misc-throw-by-value-catch-by-reference, -misc-suspicious-enum-usage'
```
patrickvacek: Not critical, but is it possible to split this over multiple lines to make it more reviewable?

src/libaktualizr/package_manager/packagemanagerconfig.cc
```diff
@@ -54,11 +32,10 @@ void PackageConfig::writeToStream(std::ostream& out_stream) const {
...
-  writeOption(out_stream, docker_app_params, "docker_app_params");
-  writeOption(out_stream, docker_app_bin, "docker_app_bin");
-  writeOption(out_stream, docker_compose_bin, "docker_compose_bin");
-#endif
+
+  // note that this is imperfect as it will not print default values deduced
+  // from users of `extra`
+  for (const auto& e : extra) {
+    writeOption(out_stream, e.first, e.second);
+  }
```
patrickvacek: What's the issue? If you have extra package manager config options, and they are left unset, they get overlooked by the config dump? That is slightly annoying but it seems like we ought to be able to sort that out.

And I definitely approve of putting implementation-specific options in their own module instead of in the general config.

src/libaktualizr/package_manager/packagemanagerconfig.h
patrickvacek: > And from a quick look, even these could probably be moved to their corresponding headers.

I realize it's convenient to have them grouped together, but conceptually, I think it makes sense to move them to their respective headers, since that's what actual custom implementations will have to do.

src/libaktualizr/package_manager/packagemanagerfactory.cc
```diff
@@ -1,55 +1,38 @@
...
+  // this trick is needed so that this object is constructed before any element
+  // is added to it
+  static std::list<std::pair<std::string, PackageManagerBuilder>> rpkgms;
 
-#include "logging/logging.h"
+  if (registered_pkgms_ == nullptr) {
+    registered_pkgms_ = &rpkgms;
+  }
+  registered_pkgms_->emplace_back(name, builder);
+  return true;
```
patrickvacek: Do we need to do something to prevent duplicate names?

src/libaktualizr/package_manager/packagemanagerfactory.cc
```diff
@@ -1,55 +1,38 @@
...
   }
+
+  LOG_ERROR << "Package manager type \"" << pconfig.type << "\" does not exist";
+  LOG_ERROR << "Available options are: " << []() {
+    std::stringstream ss;
+    for (const auto &b : *registered_pkgms_) {
+      ss << "\n" << b.first;
+    }
+    return ss.str();
+  }();
```
patrickvacek: Nice.

src/libaktualizr/package_manager/packagemanagerfactory.h
```diff
@@ -5,12 +5,33 @@
...
+    clsname##_PkgMRegister_() {                                                                              \
+      PackageManagerFactory::registerPackageManager(                                                         \
+          name, [](PackageConfig pconfig, BootloaderConfig bconfig, std::shared_ptr<INvStorage> storage,     \
+                   std::shared_ptr<HttpInterface> http) {                                                    \
+            return new clsname(std::move(pconfig), std::move(bconfig), std::move(storage), std::move(http)); \
+          });                                                                                                \
+    }                                                                                                        \
+  };                                                                                                         \
+  static clsname##_PkgMRegister_ clsname##_register_
+
```
patrickvacek: Oh no, not a macro! :)

Do you think this is possible with templating? I can live with the macro, but I want to resist them if possible.

tests/test_fixtures.py
```diff
@@ -310,7 +310,7 @@ def __init__(self, aktualizr_secondary_exe, id, port=9050, primary_port=9040):
     sqldb_path = "{db_path}"
 
     [pacman]
-    type = "fake"
+    type = "none"
```
patrickvacek: +1 for getting rid of "fake". Are there other uses of it? And is "none" still the right name for this? Should we call it "binary" or is that too misleading since it is so naïve?

lbonn:
Ok, it finally passed the tests and it's mostly done. Maybe we can wait a bit after the next release before merging that, but it can already be reviewed.

.clang-tidy
lbonn: yes, should be possible

src/libaktualizr/package_manager/packagemanagerconfig.cc
lbonn: I don't really know how to fix that here, since the `extra` won't even have been read at this point as `PackageConfig` knows nothing about its potential users.

tests/test_fixtures.py
lbonn: There was acutally another instance in `config/posix-secondary.toml`, I will fix it in another push to this branch.

We can discuss changing the name but we should be careful and be consistent. Plus we'd have several docs to change and I'm afraid it would be quite hard to remove all these uses of "fake" painlessly.

src/libaktualizr/package_manager/packagemanagerfactory.cc
lbonn: Yes, it's probably best to fail early in this case.

src/libaktualizr/package_manager/packagemanagerconfig.cc
patrickvacek: Yeah, sorry, didn't mean you have to fix it here and now! It might end up needing to be done elsewhere, but it seems minor enough to not worry about at present.

tests/test_fixtures.py
patrickvacek: > We can discuss changing the name but we should be careful and be consistent. Plus we'd have several docs to change and I'm afraid it would be quite hard to remove all these uses of "fake" painlessly.

Fair point. Might be worth consulting with tech writers or product team before we make bigger changes. No need to do all that work right now for this PR.

lbonn:
>>   And from a quick look, even these could probably be moved to their corresponding headers.

>I realize it's convenient to have them grouped together, but conceptually, I think it makes sense to move them to their respective headers, since that's what actual custom implementations will have to do.

After a quick attempt:

sotauptaneclient has a test:

`if (!target.IsOstree() && (config.pacman.type == PACKAGE_MANAGER_OSTREE || config.pacman.type == PACKAGE_MANAGER_OSTREEDOCKERAPP))`

So right now, even `sotauptaneclient.cc` would have to include `ostreemanager.h` and I'm not sure we would actually benefit from moving stuff arounds right now.

At least I can add a TODO before the list so that we remember.

patrickvacek:
> So right now, even `sotauptaneclient.cc` would have to include `ostreemanager.h` and I'm not sure we would actually benefit from moving stuff arounds right now.
> 
> At least I can add a TODO before the list so that we remember.

Fair point, we can punt on it. Perhaps just a comment that custom implementations should put their strings in their headers would be enough for now.

src/libaktualizr/package_manager/packagemanagerfactory.h
lbonn: Ok, I've tried:

```
template <typename T>
class AutoRegisterPkgm {
 public:
  AutoRegisterPkgm<T>() {
    PackageManagerFactory::registerPackageManager(
        T::regname, [](PackageConfig pconfig, BootloaderConfig bconfig, std::shared_ptr<INvStorage> storage,
                 std::shared_ptr<HttpInterface> http) {
          return new T(std::move(pconfig), std::move(bconfig), std::move(storage), std::move(http));
        });
  }
};
```

But every package manager would have to have this static `::regname` member, though it cannot be present in `PackageManagerInterface`. This is because I can't parametrize the template with string.

So I'd say it is a bit worse.

An alternative would be to switch back to gcc's `__attribute__(constructor)` like in the previous iteration.

doanac: I think the `__attribute__(constructor)` works with clang also, so while non-standard ... it sort of is.

patrickvacek: To summarize our discussion: all three options have pros and cons, but the current macro version is probably the best. Let's stick with that. Thanks for researching the options!

lbonn:
From our discussions: this is just missing tests and maybe an example before being considered ready.

doanac:
I assume this is on the path to get merged in its current form? (I have some upcoming changes and should base off this to avoid merge conflicts).

One interesting thing we could do as a follow-up on this: The registration logic could now detect at runtime if it was valid for the target. eg - if "dpkg" isn't available then maybe you don't register the DebianPackageManager.

lbonn:
@doanac yes, it shouldn't be too far now. It has been moved slightly on the side as we wanted to wait after the 2020.1 release and we had some other priorities.

But it should come around soon-ish.

Another thing to keep in mind is that we will probably continue to make changes in the package manager interfaces, one goal in particular is to try unifying primary and secondary installers as well as making them more flexible. It should lead to improvements but be advised that things are moving.

lbonn:
Rebased again on master. We can merge it soon or maybe wait after 2020.3 if we want to stay on the conservative side.

mike-sul:
> I think this is basically ready. A few thoughts, though:
> 
> * Can you update the changelog? I think we should start trying to be more consistent about updating that as changes are actually made.
> * Does it make sense to document this somewhere, or should we wait until we've finished all the changes we want to make to this process?
> * Is there more we can test? I don't see anything immediately obvious.
> * Can we have an example that isn't just baked in? We might need to debate what would be a good candidate for that.
> * Can we state somewhere what lingering issues remain, or what the next steps are?
> 
> None of that has to hold up merging this, though.

IMHO, it would be great to add some kind of "hello world" package manager along with a document that will provide a user with a guide and instructions on building and injecting its own package manager from an end-to-end perspective, starting from image-making and ending running it on a local host.

lbonn:
Yes, I agree with having a living example with a small doc. Though maybe not going up to baking an image for now, I would just start with something that can be run locally.

lbonn:
Ok after updating the changelog and discussing it a little, I think we can postpone the example/doc writing to another PR and maybe go forward with this one as it has been in this state for a long time.

PR #1501, opened 2019-12-23
Fix/ota 4119/ota 4132/root verification

patrickvacek:


src/libaktualizr/uptane/uptanerepository.cc
```diff
@@ -47,6 +47,45 @@ bool RepositoryCommon::verifyRoot(const std::string& root_raw) {
 
 void RepositoryCommon::resetRoot() { root = Root(Root::Policy::kAcceptAll); }
 
+bool RepositoryCommon::updateRoot(INvStorage& storage, const IMetadataFetcher& fetcher,
```
mike-sul: Cool, that exactly was I thought about when I saw the initial implementation - share as much code as possible between the director and imagerepo implementation.

Perhaps, it makes sense to add a reference(s) to the corresponding paragraphs of the uptane spec.

patrickvacek:
Ready for review but depends on https://github.com/advancedtelematic/tuf-test-vectors/pull/65 before merging.

patrickvacek:
Unfortunately, we can't merge this as is because the backend apparently expects that devices fetch the unnumbered `root.json` to qualify as "online". If that is difficult to change, as a temporary fix, I could reinstate fetching `root.json` the very first time a device provisions in order to satisfy that requirement, and then we could remove it once the backend no longer requires it.

mike-sul:
> I could reinstate fetching `root.json` the very first time a device provisions in order to satisfy that requirement, and then we could remove it once the backend no longer requires it.

I'd

mike-sul:
> If that is difficult to change, as a temporary fix, I could reinstate fetching `root.json` the very first time a device provisions in order to satisfy that requirement, and then we could remove it once the backend no longer requires it.

I vote for the temporary fix

patrickvacek:
> > If that is difficult to change, as a temporary fix, I could reinstate fetching `root.json` the very first time a device provisions in order to satisfy that requirement, and then we could remove it once the backend no longer requires it.
> 
> I vote for the temporary fix

Done. I also realized that I forgot to amend uptane-generator to stop creating an unnumbered `root.json`, but now I can't fix that until we can remove this hack.

src/libaktualizr/uptane/uptanerepository.cc
```diff
@@ -67,52 +75,28 @@ bool RepositoryCommon::updateRoot(INvStorage& storage, const IMetadataFetcher& f
...
 
-  // Update to latest Root metadata.
-  {
-    // According to the current design root.json without a number is guaranteed to be the latest version.
-    // Therefore we fetch the latest (root.json), and
-    // if it matches what we already have stored, we're good.
-    // If not, then we have to go fetch the missing ones by name/number until we catch up.
-
+  // 5.4.4.3.2. Update to the latest Root metadata file.
+  for (int version = rootVersion() + 1;; ++version) {
```
lbonn: Should this loop be upper-bounded? It seems a bit dangerous to never stop here if the server keeps supplying valid metadata.

patrickvacek: That's a really good point. Yes, it probably should be, but what's reasonable? 100? We should probably bring this up directly with the Uptane group. Or do you have an opinion?

lbonn: It sounds like another limit implementations should enforce, like the size of downloaded metadata artifacts.

I am thinking if we should have a limit for version jumps during an update or just a global upper bound (which we need anyway since we're using json integers and an `int` in this code).

The "jump" limit might complicate diagnostic of accidental trigger, since it would only affect a fraction of the devices.

patrickvacek: > It sounds like another limit implementations should enforce, like the size of downloaded metadata artifacts.

Could be, yes.

> I am thinking if we should have a limit for version jumps during an update or just a global upper bound (which we need anyway since we're using json integers and an `int` in this code).

Global might make more sense, because if you provision a new device, it'll have to catch up the whole way to the latest anyway. But I'm also a bit concerned that it might be unexpected to have such a limit.

I think we should either leave it is or make a quick decision now, and either way make a ticket to follow up on it and in the meantime move forward with getting this merged. Do you have a preference? If we want to just make a quick decision, I'd say 100 should be fine for now.

lbonn: Ok I see, then 100 or 1000 would be fine for now.

patrickvacek: Done (with 1000 for now), and made https://saeljira.it.here.com/browse/OTA-4273 to re-evaluate again.

PR #1502, opened 2019-12-24
sqlite utils fix an issue with reading blob-data

kbushgit:

Since a blob-data may contain null terminator, we have to explicitly specify the size of bytes sequence for the string.

### DIFF
--- a/src/libaktualizr/storage/sql_utils.h
+++ b/src/libaktualizr/storage/sql_utils.h
@@ -54,7 +54,8 @@ class SQLiteStatement {
     if (b == nullptr) {
       return boost::none;
     }
-    return std::string(b);
+    auto length = static_cast<size_t>(sqlite3_column_bytes(stmt_.get(), iCol));
+    return std::string(b, length);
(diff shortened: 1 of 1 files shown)


src/libaktualizr/storage/sql_utils.h
```diff
@@ -54,7 +54,8 @@ class SQLiteStatement {
     if (b == nullptr) {
       return boost::none;
     }
-    return std::string(b);
+    size_t length = static_cast<size_t>(sqlite3_column_bytes(stmt_.get(), iCol));
```
patrickvacek: This is a great catch, thanks! However, CI is complaining about static checks:

```
src/libaktualizr/storage/sql_utils.h:57:5: error: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto,-warnings-as-errors]
    size_t length = static_cast<size_t>(sqlite3_column_bytes(stmt_.get(), iCol));
    ^~~~~~
    auto
```

Issue #1446, opened 2019-11-13
Is "If-not-modified-since" possible?


doanac:
I've been thinking a little on how to make aktualizr/tuf-reposerver be a little less chatty. Currently you have to do 4 round trips to the server for root.json, timestamp.json, snapshot.json, and targets.json. I was wondering if there was a way to limit the total number of requests or just return `HTTP_304 Not Modified` by using the [`if-not-modified-since`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) header to limit the total amount of data transferred? I have a feeling the answer is "_no, it would create a security issue covered by TUF_". I just wanted to confirm.

doanac:
Another thing on this topic. I *think* we could add something like this to aktualizr:
~~~
 curl_easy_setopt(curl, CURLOPT_ACCEPT_ENCODING, "gzip");
~~~

And then something like this to the nginx proxy for the tuf-reposerver:
~~~
      gzip             on;
      gzip_comp_level  7;
      gzip_types       application/json;
~~~

and transfer a lot less data. My un-scientific test against a targets.json I have was:
~~~
 * targets.json  - 86924 bytes
 * targs.json.gz - 12266 bytes (~14% the size of the original)
~~~

And agents not specifying that header will still work properly and get the traditional un-compressed value.

patrickvacek:
> Currently you have to do 4 round trips to the server for root.json, timestamp.json, snapshot.json, and targets.json.

We've actually just been discussing this! TUF doesn't require downloading the next metadata file in the flow if the previous one has not indicated that something changed, and we realized that Uptane is not written to accommodate that (https://github.com/uptane/uptane-standard/issues/147). Aktualizr does not take that shortcut, but it could, so we will probably work on improving that soon.

> I was wondering if there was a way to limit the total number of requests or just return `HTTP_304 Not Modified` by using the [`if-not-modified-since`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) header to limit the total amount of data transferred?

Interesting idea, but shouldn't be necessary if we do the above correctly. Actually, maybe it'd still be useful for the root.json, since you always have to download that. @simao @koshelev how does that sound to you?

> Another thing on this topic. I _think_ we could add something like this to aktualizr:
> 
> ```
>  curl_easy_setopt(curl, CURLOPT_ACCEPT_ENCODING, "gzip");
> ```

[...]

PR #1503, opened 2019-12-30
Don't fetch Snapshot or Targets metadata if we already have the latest.

patrickvacek:
Also test that we don't fetch anything from the Image repo if the Director does not indicate that there are updates.

~~This should probably wait until https://github.com/advancedtelematic/aktualizr/pull/1501 is merged, since that will change how root metadata is fetched.~~

I did some manual testing to verify that the bandwidth savings are real. Note that my account has a rather large `targets.json` in the Image repo, so this demonstrates it well. I did a rather naive comparison by counting the bytes dumped to the log for each server response. (This is flawed for many reasons but should still give a reasonable idea of the benefit.) I did this three times: once for initialization (no updates, only Director metadata fetched), again for an update (all metadata fetched), and a third time for a second, pre-existing update (all metadata fetched on master, but with this PR, the Snapshot and Targets from the Image repo are skipped).

On current master:
```
$ grep "response:" dump_init.txt | wc -c
34
$ grep "response:" dump_first.txt | wc -c
190934
$ grep "response:" dump_second.txt | wc -c
184349
```

With these changes:
```
$ grep "response:" dump_init.txt | wc -c
34
$ grep "response:" dump_first.txt | wc -c
190934
$ grep "response:" dump_second.txt | wc -c
8419
```

FYI @doanac this hopefully helps address https://github.com/advancedtelematic/aktualizr/issues/1446.

### DIFF
--- a/src/libaktualizr/primary/CMakeLists.txt
+++ b/src/libaktualizr/primary/CMakeLists.txt
@@ -44,6 +44,9 @@ target_link_libraries(t_custom_url virtual_secondary)
 add_aktualizr_test(NAME target_mismatch SOURCES target_mismatch_test.cc PROJECT_WORKING_DIRECTORY
                    ARGS "$<TARGET_FILE:uptane-generator>" LIBRARIES uptane_generator_lib)
 target_link_libraries(t_target_mismatch virtual_secondary)
+add_aktualizr_test(NAME metadata_fetch SOURCES metadata_fetch_test.cc PROJECT_WORKING_DIRECTORY
+                   ARGS "$<TARGET_FILE:uptane-generator>" LIBRARIES uptane_generator_lib)
+target_link_libraries(t_metadata_fetch virtual_secondary)
 
 add_aktualizr_test(NAME device_cred_prov SOURCES device_cred_prov_test.cc PROJECT_WORKING_DIRECTORY LIBRARIES PUBLIC uptane_generator_lib)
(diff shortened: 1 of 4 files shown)


patrickvacek:
Rebased version here for reference: https://github.com/advancedtelematic/aktualizr/tree/feat/OTA-3995/dont-download-redundantly-rebased

doanac:
Thanks! I'm just getting back from holidays and have only glanced at this. One thing I was wondering - would it be a worth a LOG_DEBUG when skipping a remote fetch, or is this pretty easy to detect based on the current logging (ie no HTTP_GET is done, so it got skipped)?

patrickvacek:
FYI the standard has been updated accordingly to acknowledge (most of) these optimizations: https://github.com/uptane/uptane-standard/pull/151

patrickvacek:
Rebased on latest master.

> would it be a worth a LOG_DEBUG when skipping a remote fetch, or is this pretty easy to detect based on the current logging (ie no HTTP_GET is done, so it got skipped)?

@doanac I think it's easy enough to determine with loglevel 1 from watching the GETs (not) made.

lbonn:
Looking good! @doanac's comment is still valid though, adding a `LOG_DEBUG` should be pretty cheap.

patrickvacek:
> Looking good! @doanac's comment is still valid though, adding a `LOG_DEBUG` should be pretty cheap.

Two votes is enough to convince me. Done.

PR #1505, opened 2020-01-03
ISSUE-1504: Virtual Secondary config

mike-sul:
Signed-off-by: Mykhaylo Sul <myk.sul@gmail.com>

PR #1506, opened 2020-01-03
Use a bool type instead of a string in the virtual secondary config

patrickvacek:
Cherry-pick of https://github.com/advancedtelematic/aktualizr/pull/1505/, but for this branch, only update the example in the docs.


patrickvacek:
Travis is failing because of a change in Doxygen versions, but that isn't relevant to this branch and has been fixed on master. It can safely be ignored.

patrickvacek:
Also added a commit to address some inconsistencies in the docs about aktualizr requirements. I'd overlooked this list in the past. We really need to find a way to synchronize the list in the readme, this doc, and the dockerfiles.

PR #1507, opened 2020-01-03
Update code contribution guidelines.

patrickvacek:
Removed some redundant/obvious/irrelevant bits and added some more detail to reflect our current standards and preferences.

This is still up for discussion, so if you disagree with any of this or think something is missing, please speak up.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,30 +13,31 @@ Code quality and style
 
 All code should be developed according to the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html). In addition, the code should conform to the following guidelines:
 
-   * Code should be covered by tests
-      - Wherever possible, automated testing should be used
-      - Tests cases should at least exercise all documented requirements
-      - If automated testing is not possible, manual test cases should be described
-   * It must be easy for a developer checking out a project to run the test suite based on the information in the [readme](README.adoc)
-   * All code must pass all unit tests before a merge request is made
-      - Tests that don't pass should be marked pending (with justification) or should be fixed.
-   * All code must pass formatting and static linting tests
-   * Features should be developed in feature branches
-   * Only working code should go into the master branch.
-      - master should always be in a deployable state
-      - Undeployable code should stay on feature branches
-         - Code failing the active unit or integration tests for a project is undeployable
-         - Functionally incomplete code is not necessarily undeployable
-   * Feature branches should only contain single features
-      - Developers should not create large, month-long, multiple-feature branches
-      - Developers should try to have their code merged at least once every week
-   * All code must be reviewed before it is merged to a master branch.
-      - The reviewer can not be the code author
-      - All the reviewer's concerns must be resolved to reviewer's satisfaction
-      - Reviews should enforce the coding guidelines for the project
-   * Bugs reported against code on a master branch should be reproduced with a failing test case before they are resolved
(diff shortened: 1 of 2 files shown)


patrickvacek:
Although I appreciate the quick approval, I'm going to wait until some more of our team is back from vacation before merging, as I'd like to get a bit wider input on this.

CONTRIBUTING.md
```diff
@@ -13,30 +13,28 @@ Code quality and style
...
+   - Developers should not create large, month-long, multiple-feature branches.
+     - The larger the PR, the harder it is to review, the more likely it will need to be rebased, and the harder it is to rebase it.
+     - If a PR changes thousands of lines of code, please consider splitting it into smaller PRs.
+   - Developers should aim to have their code merged at least once every week.
+   - Multiple small fixes or refactorings can be grouped together in one PR, but each independent fix or refactoring should be a unique commit.
+* All code must be reviewed before it is merged to the master branch.
+   - The reviewer can not be the code author.
+   - All the reviewer's concerns must be resolved to reviewer's satisfaction.
+   - Reviews should enforce the coding guidelines for the project.
+* Bugs should be reproduced with a failing test case before they are resolved.
```
xaviergully: 👍

CONTRIBUTING.md
```diff
@@ -13,30 +13,28 @@ Code quality and style
...
+* Only working code that passes all tests in CI will be merged into the master branch.
+   - The master branch should always be in a deployable state.
+   - Functionally incomplete code is not necessarily undeployable.
+* Pull requests should only contain a single feature, fix, or refactoring.
+   - Developers should not create large, month-long, multiple-feature branches.
+     - The larger the PR, the harder it is to review, the more likely it will need to be rebased, and the harder it is to rebase it.
+     - If a PR changes thousands of lines of code, please consider splitting it into smaller PRs.
+   - Developers should aim to have their code merged at least once every week.
+   - Multiple small fixes or refactorings can be grouped together in one PR, but each independent fix or refactoring should be a unique commit.
+* All code must be reviewed before it is merged to the master branch.
```
eu-smirnov: I'd probably add something about commits in PR, like:
 - Make separate commits for logically separate changes within the PR.
 - Each commit should be buildable and ideally passing the tests, to simplify git bisect.
 - Short description should not exceed 50 chars.

patrickvacek: > Make separate commits for logically separate changes within the PR.

Agreed.

> Each commit should be buildable and ideally passing the tests, to simplify git bisect.

Generally agreed, but I'd consider an exception for the later line ("Bugs should be reproduced with a failing test case before they are resolved.").

> Short description should not exceed 50 chars.

Does it really have to be 50? That's so short! I'd prefer something around 70-80, which is closer to what we do in practice.

eu-smirnov: 50 is what usually recommended, e.g. https://gerrit.googlesource.com/plugins/commit-message-length-validator/+doc/master/src/main/resources/Documentation/config.md
But it can definitely break after 72 chars, so maybe we can put this limit here.

TBH, I'm not entirely sure about mentioning tests in p.2. I doubt that anybody will really check that.

patrickvacek: Updated (and rebased). There is some redundancy in the two lines about commits, but that's the only (minor) concern I have. I made the comment about tests passing for each commit more explicitly an ideal than a requirement, but I think it's fine to keep that as an aspiration.

PR #1509, opened 2020-01-07
Fix null pointer dereference.

patrickvacek:


eu-smirnov:
Looks like `targetsExpired` call on line 282 is also gonna segfault. Why do we want to use `shared_ptr` for `targets` at all?

eu-smirnov:
> Looks like `targetsExpired` call on line 282 is also gonna segfault.
No, sorry, it won't. Still, looks quite fragile to me.

patrickvacek:
> Why do we want to use `shared_ptr` for `targets` at all?

Looks like it came out of the delegation support work. To be honest, it doesn't seem entirely necessary anymore, but I just spent a while playing with it and I keep breaking the delegation vector tests, so I'd prefer to leave it as is unless there is a good reason to change it.

> Still, looks quite fragile to me.

Yeah, and I don't like the inconsistency, either. What exactly is fragile to you, though?

My vote is to merge the quick fix soon and worry about refactoring later.

PR #1510, opened 2020-01-08
Fix/bad metadata handling

patrickvacek:
This fixes a bug I recently reported and that I believe @kbushgit may have also encountered.


patrickvacek:
Just realized there's a small change I'd like to make to the test I wrote. (No point verifying nothing was installed if we never install anything.)

PR #1512, opened 2020-01-08
Switch to dynamic libraries

lbonn:
Besides the obvious reuse we have between aktualizr and aktualizr-info, it might be time to ship libaktualizr as a standalone library.
For now, `aktualizr_static_lib` still exists, we might remove it at a later point.

This will also allow some simplifications in #1498.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,7 +387,7 @@ if(hasParent)
     set (AKTUALIZR_EXTERNAL_LIBS ${AKTUALIZR_EXTERNAL_LIBS} PARENT_SCOPE)
 endif()
 
-set (TEST_LIBS ${AKTUALIZR_EXTERNAL_LIBS} gtest gmock)
+set (TEST_LIBS gtest gmock)
 if(BUILD_WITH_CODE_COVERAGE)
     set(COVERAGE_LCOV_EXCLUDES '/usr/include/*' ${CMAKE_BINARY_DIR}'*' ${CMAKE_SOURCE_DIR}'/third_party/*' ${CMAKE_SOURCE_DIR}'/tests/*' '*_test.cc')
     include(CodeCoverage)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,7 +387,7 @@ if(hasParent)
     set (AKTUALIZR_EXTERNAL_LIBS ${AKTUALIZR_EXTERNAL_LIBS} PARENT_SCOPE)
 endif()
 
-set (TEST_LIBS gtest gmock)
+set (TEST_LIBS gtest gmock aktualizr_lib)
 if(BUILD_WITH_CODE_COVERAGE)
     set(COVERAGE_LCOV_EXCLUDES '/usr/include/*' ${CMAKE_BINARY_DIR}'*' ${CMAKE_SOURCE_DIR}'/third_party/*' ${CMAKE_SOURCE_DIR}'/tests/*' '*_test.cc')
     include(CodeCoverage)
--- a/src/libaktualizr/CMakeLists.txt
+++ b/src/libaktualizr/CMakeLists.txt
@@ -18,7 +18,10 @@ if(BUILD_ISOTP)
     add_subdirectory("isotp_conn")
 endif(BUILD_ISOTP)
 
+# deprecated, we recommend using aktualizr_lib
 add_library(aktualizr_static_lib STATIC
+    $<TARGET_OBJECTS:asn1>
+    $<TARGET_OBJECTS:asn1_lib>
     $<TARGET_OBJECTS:bootloader>
     $<TARGET_OBJECTS:bootstrap>
     $<TARGET_OBJECTS:campaign>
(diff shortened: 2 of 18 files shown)


src/libaktualizr/CMakeLists.txt
```diff
@@ -33,9 +36,29 @@ add_library(aktualizr_static_lib STATIC
...
+
 if (BUILD_ISOTP)
     target_sources(aktualizr_static_lib PRIVATE $<TARGET_OBJECTS:isotp_conn>)
+    target_sources(aktualizr_lib PRIVATE $<TARGET_OBJECTS:isotp_conn>)
 endif (BUILD_ISOTP)
 
-target_include_directories(aktualizr_static_lib PUBLIC
-    $<TARGET_PROPERTY:package_manager,INCLUDE_DIRECTORIES>)
+target_include_directories(aktualizr_static_lib PUBLIC $<TARGET_PROPERTY:package_manager,INCLUDE_DIRECTORIES>)
+target_include_directories(aktualizr_lib PUBLIC $<TARGET_PROPERTY:package_manager,INCLUDE_DIRECTORIES>)
```
patrickvacek: Why are these special bits for `package_manager` required?

src/sota_tools/CMakeLists.txt
```diff
@@ -15,33 +15,35 @@ set(SOTA_TOOLS_LIB_SRC
...
-    ${Boost_LIBRARIES}
-    ${JSONCPP_LIBRARIES}
-    ${CMAKE_THREAD_LIBS_INIT}
-    ${LibArchive_LIBRARIES}
-    ${CURL_LIBRARIES}
-    ${OPENSSL_LIBRARIES}
-    ${sodium_LIBRARY_RELEASE}
-    ${GLIB2_LIBRARIES})
-
 if (BUILD_SOTA_TOOLS)
```
patrickvacek: The `endif` and `if` here are redundant now.

CMakeLists.txt
```diff
@@ -213,6 +213,9 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)")
     endif ()
 endif()
 
+# helpful for debugging installed executables, see `man ld-linux.so`
+set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib")
```
patrickvacek: Do we really want this? I've had bad experiences with rpaths.

src/libaktualizr/CMakeLists.txt
lbonn: To have access to `ostree.h`. I can double check that but it seems necessary.

CMakeLists.txt
lbonn: It should be for comfort only, I'll try without.

src/libaktualizr/CMakeLists.txt
lbonn: Ok, actually it doesn't look like it's making a difference... Removed in the last version.

CMakeLists.txt
lbonn: Specifically, this is the difference:

With:

```
$ DESTDIR=/tmp/test ninja install
$ /tmp/test/usr/local/bin/aktualizr --version
# works
```

without:

```
$ DESTDIR=/tmp/test ninja install
$ /tmp/test/usr/local/bin/aktualizr --version
# does not work
$ LD_LIBRARY_PATH=/tmp/test/usr/local/bin /tmp/test/usr/local/bin/aktualizr --version
# works
```

But then, it might be bad practice and cause other issues. Do you remember some details about the issues you had?

patrickvacek: The problem with using rpath is that you're forcing your path structure on your users. It means that when dynamically linking, everything in `"$ORIGIN/../lib"` will be taken over anything in the user's `$LD_LIBRARY_PATH`, which can have unintended side effects.

There are two points of view on this depending on what you think one's users prefer. One is that rpath makes it easier to find your library, but the downside is messing around with link path searching. The other is that rpath breaks traditional Linux norms and you should only really install things into default locations or change your `PATH` and/or `LD_LIBRARY_PATH` accordingly.

My preference is the latter, but I'm not set on that. Our circumstances are quite different than my past experience. If this makes things easier, it's fine. It does seem it'd make testing locally a bit easier.

eu-smirnov: My view is that if a user installs it to a non-standard path, it is up to them to adjust LD_LIBRARY_PATH accordingly.

src/sota_tools/CMakeLists.txt
```diff
@@ -15,33 +15,36 @@ set(SOTA_TOOLS_LIB_SRC
...
 
-if (BUILD_SOTA_TOOLS)
-    set(GARAGE_TOOLS_VERSION "${AKTUALIZR_VERSION}")
-    set_property(SOURCE garage_tools_version.cc PROPERTY COMPILE_DEFINITIONS GARAGE_TOOLS_VERSION="${GARAGE_TOOLS_VERSION}")
-    add_library(sota_tools_static_lib STATIC ${SOTA_TOOLS_LIB_SRC})
-    target_include_directories(sota_tools_static_lib PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${PROJECT_SOURCE_DIR} ${GLIB2_INCLUDE_DIRS})
-endif (BUILD_SOTA_TOOLS)
-
+# do not link tests with libaktualizr
+list(REMOVE_ITEM TEST_LIBS aktualizr_lib)
```
patrickvacek: I thought that the garage tools used some of functionality from libaktualizr, like Utils and such. Is that linked in still somehow, or how does that work now?

CMakeLists.txt
lbonn: > The problem with using rpath is that you're forcing your path structure on your users. It means that when dynamically linking, everything in `"$ORIGIN/../lib"` will be taken over anything in the user's `$LD_LIBRARY_PATH`, which can have unintended side effects.

According to the ld-linux.so's manual page, and a quick test, this doesn't seem to be the case. What CMake calls `RPATH` is actually `RUNPATH` which has a lower priority than `LD_LIBRARY_PATH`:

```
readelf -a usr/local/bin/aktualizr | grep PATH
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/../lib]
```

That said, the risks of confusion might indeed outweigh the benefits, so I've removed it for now.

src/sota_tools/CMakeLists.txt
lbonn: It's now included in `sota_tools_lib` which links statically with libaktualizr. I've pushed some new changes to make it clearer.

CMakeLists.txt
patrickvacek: > What CMake calls `RPATH` is actually `RUNPATH`

Wow, that is actually the most confusing part of all of this! If it's actually just setting `RUNPATH`, I don't really mind it. For now, though, if you've already removed it, let's see how it goes without it, and we can add it later if we think it'd be helpful.

PR #1513, opened 2020-01-09
OTA-4174: IP-Secondary: match Director's and Imagerepo's targets

patrickvacek:
Cherry-pick of https://github.com/advancedtelematic/aktualizr/pull/1500/commits/fc6eafdf59fe040e676e4d91d42446fbc2a4e57b from https://github.com/advancedtelematic/aktualizr/pull/1500 to address a possible bug. Still requires some further investigation and review, but I think we might need this sooner than the rest of that PR.

- Each target listed in Director's target file should have corresponding match in ImageRepo's one
- Test for the use-case if IP Secondary Director's target file and ImageRepo's one have different number of targets
- Separate the ostree specific tests from non-ostree one in the IP Secondary Verification test suite

### DIFF
--- a/src/aktualizr_secondary/CMakeLists.txt
+++ b/src/aktualizr_secondary/CMakeLists.txt
@@ -60,7 +60,7 @@ include(AddAktualizrTest)
 list(INSERT TEST_LIBS 0 aktualizr_secondary_lib)
 
 add_aktualizr_test(NAME aktualizr_secondary_config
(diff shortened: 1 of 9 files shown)


patrickvacek:
Unfortunately, my first attempt at testing this failed failed with these errors on the Secondary when I sent it an update:
```
Failed to update director metadata:
Invalid number of targets (should be one): 0
```

Also seeing this repeatedly on the Primary: `Getting the manifest key of a secondary`, which is annoying.

patrickvacek:
Doesn't work and seems unnecessary after all.

PR #1514, opened 2020-01-10
Don't copy CA certificates to client certificate data

eu-smirnov:
I don't see any reason why we would want to have CA certs in the client cert data.


lbonn:
Seems to be fine. Should we need to test that on meta-updater as well (ie provisioning oe-selftest)

config/sota-local.toml
```diff
@@ -10,6 +10,3 @@ path = "storage"
 
 [pacman]
 type = "none"
-
-[uptane]
-secondary_config_file = "virtualsec.json"
```
patrickvacek: Why remove this? I think it's just used for testing and as an example.

eu-smirnov: Since recently the initialization fails if this file doesn't exist. It seems like a correct behavior, but makes this config unusable without a secondary config. I'd really like to keep this config as a minimal example. Maybe we can create another one which will include a secondary.

eu-smirnov:
It might be worth to test it in oe-selftest, but the pipeline seems to be broken.

patrickvacek:
> It might be worth to test it in oe-selftest, but the pipeline seems to be broken.

Yeah, seems to be unrelated, but you might be able to work around it. Or just test manually/locally. Or try zeus, which seems fine.

eu-smirnov:
Tested on meta-updater master, specifying TEST_AKTUALIZR_REMOTE=y, TEST_AKTUALIZR_BRANCH=fix/cert and TEST_AKTUALIZR_REV=bacb528d9ed5cb086ce3048cdebd7d98d463ee75 in pipeline variables: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/pipelines/303873

zabbal:
> I also can't tell why we were doing this before.

How's the function result is actually used? I mean in theory we might need CA cert to verify the cert which was signed by that CA but I'm not familiar enough with the code to figure out how it's used quickly.

eu-smirnov:
@zabbal We are still using them to verify server cert. We just stored them not only in `CURLOPT_CAINFO` but also in `CURLOPT_SSLCERT` file. This PR fixes that.

PR #1515, opened 2020-01-13
Run ldconfig at the end of Dockerfile.aktualizr

lbonn:
Otherwise, /usr/local/lib doesn't end up being used for some reason...


lbonn:
Ok great. I reproduced the problem locally and this fixed it so I think we can merge.

PR #1520, opened 2020-01-15
Check disk space before downloading files.

patrickvacek:
There is no reason to download a file that we don't have space for on disk. I've set 1MB as reserved so that we can still (hopefully) do the database operations to let libaktualizr send a manifest to the server.

This should help with the problems observed with non-OSTree images (see https://github.com/advancedtelematic/meta-updater/pull/637) when additional overhead isn't provided to the qemu image.

### DIFF
--- a/src/libaktualizr/storage/invstorage.h
+++ b/src/libaktualizr/storage/invstorage.h
@@ -184,6 +184,7 @@ class INvStorage {
   virtual void saveEcuReportCounter(const Uptane::EcuSerial& ecu_serial, int64_t counter) = 0;
   virtual bool loadEcuReportCounter(std::vector<std::pair<Uptane::EcuSerial, int64_t>>* results) = 0;
 
+  virtual bool checkAvailableDiskSpace(uint64_t required_bytes) const = 0;
   virtual boost::optional<std::pair<size_t, std::string>> checkTargetFile(const Uptane::Target& target) const = 0;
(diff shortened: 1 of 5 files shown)


src/libaktualizr/storage/invstorage.h
```diff
@@ -184,6 +184,7 @@ class INvStorage {
   virtual void saveEcuReportCounter(const Uptane::EcuSerial& ecu_serial, int64_t counter) = 0;
   virtual bool loadEcuReportCounter(std::vector<std::pair<Uptane::EcuSerial, int64_t>>* results) = 0;
 
+  virtual bool checkAvailableDiskSpace(uint64_t required_bytes) const = 0;
```
kbushgit: one minor remark, regarding the "checkAvailableDiskSpace" function, is it really necessary to make this function pure virtual in the "invstorage" interface? 

I suspect this function will be called quite often, does it make sense to increase virtual table?

patrickvacek: I'm not sure I have a choice here. We only have one fully functional storage implementation at present, but I think this depends on the specific implementation. For SQL, we store the files on disk in the same directory as the database, so I know I can check the filesystem that all that lives on. But a different implementation might do something else, like use multiple partitions, and then my current logic would have to be adapted to that.

PR #1522, opened 2020-01-16
Some linking fixes

lbonn:
I was running tests after setting up clang sanitizers:

```CC=clang-9 CXX=clang++-9 cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Asan```

... and it worked surprisingly well!

It just found an issue where we were linking with an object twice and there was a weird python error that was fixed by removing an `LD_PRELOAD` use.

Will maybe uncover more errors when used with more features, but it looks like it's close to a state where we could use it on CI.


mchekhovoi:
Using of `add_aktualizr_test` instead of `add_test` for `c_api` looks very good.

lbonn:
@mchekhovoi glad to hear that!
I didn't know why `SET(CMAKE_SKIP_RPATH TRUE)` was there and now it works without.

Also interesting that the leak in the test was detected by valgrind and gcc's sanitizers but not clang's... Maybe something from my setup.

PR #1525, opened 2020-01-17
Merge pull request #1525 from advancedtelematic/fix/secondary-root-version-bump

patrickvacek:
aktualizr_secondary: Accept initial Root metadata with version > 1.

### DIFF
--- a/src/uptane_generator/repo_test.cc
+++ b/src/uptane_generator/repo_test.cc
@@ -497,6 +497,126 @@ TEST(uptane_generator, generateCampaigns) {
   EXPECT_EQ(campaigns["campaigns"][0]["metadata"][2]["value"], "20");
 }
 
+/*
+ * Bump the version of the Director Root metadata.
+ */
+TEST(uptane_generator, refreshDirectorRoot) {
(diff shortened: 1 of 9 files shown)


src/uptane_generator/repo_test.cc
```diff
@@ -497,6 +497,126 @@ TEST(uptane_generator, generateCampaigns) {
   EXPECT_EQ(campaigns["campaigns"][0]["metadata"][2]["value"], "20");
 }
 
+/*
+ * Bump the version of the Director Root metadata.
+ */
+TEST(uptane_generator, refreshDirectorRoot) {
```
lbonn: These tests looks really similar.
Would be cool to make most of their body common, with `TEST_P` or one test.

Might be impractical or just as ugly though, no big deal.

patrickvacek: Yeah, these are pretty repetitive. I haven't gotten into parameterized testing much yet. Would be cool to do, just haven't had the time yet.

lbonn: Ok, in this case it's probably not worth holding up this PR then.

PR #1526, opened 2020-01-17
Creation of path for testing with root certificate

ammaboateng:
I created a path under "Integrate OTA Connect" and placed the pages for testing with a root certificate using device credentials, as I noticed they didn't have a path within the documentation and were a little difficult to locate.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/nav.adoc
+++ b/docs/ota-client-guide/modules/ROOT/nav.adoc
@@ -62,6 +62,11 @@ ifndef::env-github[:pageroot:]
 ** xref:{pageroot}hsm-provisioning-example.adoc[Generate a device certificate using an HSM]
 ** xref:{pageroot}enable-device-cred-provisioning.adoc[Enable and install device certificates]
 
+* xref:{pageroot}test-root-cert.adoc[Testing with a root certificate]
+** xref:{pageroot}provide-testroot-cert.adoc[Register your test root certificate]
+** xref:{pageroot}generatetest-devicecert.adoc[Generate a test device certificate]
+** xref:{pageroot}enable-device-cred-provtest.adoc[Enable and install test device certificates]
+
 * xref:{pageroot}secure-software-updates.adoc[Secure your software repository]
 ** xref:{pageroot}install-garage-sign-deploy.adoc[Install the garage-deploy tool]
 ** xref:{pageroot}rotating-signing-keys.adoc[Manage keys for software metadata]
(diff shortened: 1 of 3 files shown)


lbonn:
The errors should be fixed by merging https://github.com/advancedtelematic/aktualizr/pull/1493/, because doxygen had been updated in debian testing.

But then that's a bit annoying that the stable docs branch breaks because the doxygen run depends on an unstable container... We should maybe think about using something more reliable here.

docs/ota-client-guide/modules/ROOT/pages/test-root-cert.adoc
```diff
@@ -0,0 +1,30 @@
...
+ifdef::env-github[]
+
+[NOTE]
+====
+We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}.html[view this article in our documentation portal]. Not all of our articles render correctly in GitHub.
+====
+endif::[]
+
+
+//MC: This is intended for the "test" use case of device credentials provisioning.
```
patrickvacek: Is this comment still necessary? I think that was just left over from Merlin.

docs/ota-client-guide/modules/ROOT/nav.adoc
```diff
@@ -62,6 +62,11 @@ ifndef::env-github[:pageroot:]
 ** xref:{pageroot}hsm-provisioning-example.adoc[Generate a device certificate using an HSM]
 ** xref:{pageroot}enable-device-cred-provisioning.adoc[Enable and install device certificates]
 
+* xref:{pageroot}test-root-cert.adoc[Testing with a root certificate]
+** xref:{pageroot}provide-testroot-cert.adoc[Register your test root certificate]
+** xref:{pageroot}generatetest-devicecert.adoc[Generate a test device certificate]
+** xref:{pageroot}enable-device-cred-provtest.adoc[Enable and install test device certificates]
```
patrickvacek: I didn't realize all these pages were hidden until now!

ammaboateng: @patrickvacek I realized they were hidden after going through the developer portal doc. I'm not sure if all the information is still up to date but I tried to make these easier to find on the doc portal.

patrickvacek:
FYI I'm in the middle of creating release 2020.1 right now, and I just made the 2020.1-docs branch. It might make sense to move this PR to point to that.

tkfu:
Yeah, this doesn't make sense. It's the exact same content as the "Provision Devices" section. No idea why these files exist in the first place, but let's get rid of them.

PR #1527, opened 2020-01-17
Revert "Don't copy CA certificates to client certificate data"

eu-smirnov:
This reverts commit c6c354363fb35e9c0d9aa959089c831266944982,
which seems to break things when curl is compiled with GnuTLS backend.


src/libaktualizr/crypto/crypto.cc
```diff
@@ -297,6 +297,7 @@ bool Crypto::parseP12(BIO *p12_bio, const std::string &p12_password, std::string
   for (int i = 0; i < sk_X509_num(ca_certs.get()); i++) {  // NOLINT
     ca_cert = sk_X509_value(ca_certs.get(), i);            // NOLINT
     PEM_write_bio_X509(ca_sink.get(), ca_cert);
+    PEM_write_bio_X509(cert_sink.get(), ca_cert);
```
lbonn: Could we add a comment here to note that it should not be necessary, works with OpenSSL but not gnutls?

eu-smirnov: This is just a quick revert, I'd like to find the root cause before adding any comments. I'll definitely follow up with other PR.

PR #1528, opened 2020-01-21
Docs/removing ecu events doc

ammaboateng:
Hiding `ecu_events.adoc` page on the documentation portal. The page will be updated later and added to the API documentation in the doc portal, which is yet to be created.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/ecu_events.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/ecu_events.adoc
@@ -7,7 +7,7 @@ We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}
 ====
 endif::[]
 
-// MC: This page is currently hidden in the developer guide portal and will be updated later to be connected to the API documentation.
+// AB: This page is currently out of date and will be temporarily hidden in the developer guide portal to be updated later and connected to the API documentation.
 
 :aktualizr-github-url: https://github.com/advancedtelematic/aktualizr/tree/master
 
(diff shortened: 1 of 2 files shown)


patrickvacek:
I think you might want to rebase this branch (probably on 2020.1-docs). I see several commits that don't seem to belong (and should already be there).

ammaboateng:
> Why do we want to hide this information? It's already just in the reference section; it's mostly just intended for internal use. Why not just update it once it is ready to be updated?

patrickvacek:
> temporarily hiding it from customers on the doc portal

But why? It's still completely public and visible in github, and this is inconsistent with all of our other "internal" aktualizr docs that live inside the aktualizr repo. This particular document is for developer use and was written by a developer. I would like to treat all of these dev docs the same way: either all of them should be on the docs portal or none of them.

ammaboateng:
> 
> I would like to treat all of these dev docs the same way: either all of them should be on the docs portal or none of them.

In that case would you suggest removing the page for now or rather updating it?

patrickvacek:
> In that case would you suggest removing the page for now or rather updating it?

Just discussed with Jon. I didn't realize this doc is rather out of date and not very useful anymore. "Hiding" it like this is an acceptable intermediate solution in that case, although deletion or adding a specific comment that it is obsolete and will be replaced are also options. I retract my objections.

docs/ota-client-guide/modules/ROOT/pages/ecu_events.adoc
```diff
@@ -7,6 +7,8 @@ We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}
 ====
 endif::[]
 
+// MC: This page is currently hidden in the developer guide portal and will be updated later to be connected to the API documentation.
```
tkfu: @ammaboateng: "MC" are the initials of a previous tech writer. If you want to leave a note like this, put your own initials on it.

PR #1529, opened 2020-01-21
Secondary implementation improvement

mike-sul:

### DIFF
--- a/src/aktualizr_secondary/update_agent_ostree.cc
+++ b/src/aktualizr_secondary/update_agent_ostree.cc
@@ -12,11 +12,22 @@ bool OstreeUpdateAgent::isTargetSupported(const Uptane::Target& target) const { 
 bool OstreeUpdateAgent::getInstalledImageInfo(Uptane::InstalledImageInfo& installed_image_info) const {
   bool result = false;
   try {
-    installed_image_info.hash = _ostreePackMan->getCurrentHash();
-    // This is the policy on a target image name in case of ostree
-    // The policy in followed and implied in meta-updater (garage-sign/push) and the backend
-    installed_image_info.name = _targetname_prefix + "-" + installed_image_info.hash;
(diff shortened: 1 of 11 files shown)


src/aktualizr_secondary/update_agent_ostree.cc
```diff
@@ -12,11 +12,22 @@ bool OstreeUpdateAgent::isTargetSupported(const Uptane::Target& target) const {
...
-    installed_image_info.hash = _ostreePackMan->getCurrentHash();
-    // This is the policy on a target image name in case of ostree
-    // The policy in followed and implied in meta-updater (garage-sign/push) and the backend
-    installed_image_info.name = _targetname_prefix + "-" + installed_image_info.hash;
     installed_image_info.len = 0;
+    installed_image_info.hash = _ostreePackMan->getCurrentHash();
+
+    // TODO: consider more elegant way of storing currently installed target name
+    // usage of the SQLStorage and ostree implementions aimed for Primary is
+    // a quite overhead for Secondary
```
patrickvacek: What would you propose? We can make a ticket to reconsider this, but if we have to use the same storage and OSTree manager classes, it doesn't seem like much of an extra burden here.

mike-sul: Yeah, I have a few ideas, better to talk/discuss it during a call/meeting.

mike-sul:
> Looks good but CI ain't happy.

Yeah, I am working on it. I tested this fix on QEMU and it does work for me. Also, I am going to add a fix for the second minor issue into this PR (2) Secondary outputs an incorrect hash value just after applying the pending hash/version into logs, but it does not break anything)

mike-sul:
@patrickvacek I verified this version of QEMU multiple times with different value of OSTREE_BRANCHNAME and different reboot sequence, looks like it works and fixes the Unu issue.

PR #1531, opened 2020-01-23
Don't copy all CA certs to client cert

eu-smirnov:
We need to copy only the SubCA for device registration to support curl with GnuTLS backend, which expects us to provide a full client certificate chain.

There are also some unclear things about the certificates we get from the backend:
autoprov_credentials.p12 contains two identical copies of the server RootCA:
```
        Issuer: CN = e237937a-74ef-4d64-8d47-3df9c56eb79e
        Validity
            Not Before: Sep 13 09:04:13 2018 GMT
            Not After : Sep 10 09:04:43 2028 GMT
        Subject: CN = e237937a-74ef-4d64-8d47-3df9c56eb79e
```
Why do we need two of those?

P12 archive that we get after the device registration, again contains two copies of the server RootCA.
Why two? Do we even need to receive the same certificate we already use to establish connection to device gateway for registration?

P12 archive that we get after the device registration also contains the following RootCA:
```
        Issuer: C = IE, O = Baltimore, OU = CyberTrust, CN = Baltimore CyberTrust Root
        Validity
            Not Before: May 12 18:46:00 2000 GMT
            Not After : May 12 23:59:00 2025 GMT
        Subject: C = IE, O = Baltimore, OU = CyberTrust, CN = Baltimore CyberTrust Root
```
What is this certificate and where is it used?

Maybe somebody from the backend team can shed some light on this questions. @koshelev?

### DIFF
--- a/src/libaktualizr/crypto/crypto.cc
+++ b/src/libaktualizr/crypto/crypto.cc
@@ -234,8 +234,14 @@ bool Crypto::ED25519Verify(const std::string &public_key, const std::string &sig
                                      reinterpret_cast<const unsigned char *>(public_key.c_str())) == 0;
 }
 
-bool Crypto::parseP12(BIO *p12_bio, const std::string &p12_password, std::string *out_pkey, std::string *out_cert,
-                      std::string *out_ca) {
+std::string Crypto::bioToString(BIO *b) {
+  char *buf;
+  auto len = BIO_get_mem_data(b, &buf);  // NOLINT(cppcoreguidelines-pro-type-cstyle-cast)
+  return std::string(buf, static_cast<size_t>(len));
+}
+
+bool Crypto::parseP12(BIO *p12_bio, const std::string &p12_password, std::string &out_pkey, std::string &out_cert,
+                      std::vector<std::string> &out_ca) {
 #if AKTUALIZR_OPENSSL_PRE_11
   SSLeay_add_all_algorithms();
 #endif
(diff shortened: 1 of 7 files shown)


eu-smirnov:
Also, a nice article on how curl behaves differently depending on the TLS library: https://wiki.nikhef.nl/grid/Funny_Curly_things

src/libaktualizr/crypto/crypto.cc
```diff
@@ -234,8 +234,14 @@ bool Crypto::ED25519Verify(const std::string &public_key, const std::string &sig
...
 
-bool Crypto::parseP12(BIO *p12_bio, const std::string &p12_password, std::string *out_pkey, std::string *out_cert,
-                      std::string *out_ca) {
+std::string Crypto::bioToString(BIO *b) {
+  char *buf;
+  auto len = BIO_get_mem_data(b, &buf);
+  return std::string(buf, static_cast<size_t>(len));
+}
+
+bool Crypto::parseP12(BIO *p12_bio, const std::string &p12_password, std::string &out_pkey, std::string &out_cert,
```
lbonn: These were pointers because it's recommended by the google style guide: https://google.github.io/styleguide/cppguide.html#Output_Parameters.

We follow it pretty loosely though.

zabbal: This probably can (should?) be split into separate commit as it's orthogonal to CA bundling question.

src/libaktualizr/primary/initializer.cc
```diff
@@ -122,13 +122,25 @@ InitRetCode Initializer::initTlsCreds() {
...
     LOG_ERROR << "Received a malformed P12 package from the server";
     return InitRetCode::kBadP12;
   }
+  // The last cert in ca_certs is a subCA for client certificate,
+  // which server expects us to send together with the client cert.
+  // Depending on the curl backend we need to keep subCA cert in different places:
+  // OpenSSL expects only the client cert to be passed as `SSLCERT` and will
+  // build client certificate chain searching `CAINFO` for intermediate certs.
+  // GnuTLS expect complete certificate chain in `SSLCERT` and doesn't look at `CAINFO`.
+  cert += ca_certs.back();
```
zabbal: Does this mean that we rely on a certain order of certificates? Is there guarantees that the order would always be the same?

eu-smirnov:
Talked to Vladimir, things are clearer now, will update the PR tomorrow.

patrickvacek:
@eu-smirnov What's the status of this ticket? Is this still something you want to improve?

eu-smirnov:
Needs more work, probably also from the backend side, so I'm closing this. Described the possible further steps in https://saeljira.it.here.com/browse/OTA-5162

PR #1533, opened 2020-01-23
Wait for secondaries before installing

lbonn:
This is quite big and will need more tests.

### DIFF
--- a/config/sql/migration/migrate.23.sql
+++ b/config/sql/migration/migrate.23.sql
@@ -2,7 +2,7 @@
 SAVEPOINT MIGRATION;
 
 ALTER TABLE ecu_serials RENAME TO ecus;
-CREATE TABLE secondary_ecus(serial TEXT PRIMARY KEY, sec_type TEXT, public_key_type TEXT, public_key TEXT, extra TEXT);
+CREATE TABLE secondary_ecus(serial TEXT PRIMARY KEY, sec_type TEXT, public_key_type TEXT, public_key TEXT, extra TEXT, manifest TEXT);
 
 DELETE FROM version;
 INSERT INTO version VALUES(23);
--- a/config/sql/migration/migrate.23.sql
+++ b/config/sql/migration/migrate.23.sql
@@ -1,0 +1,10 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+ALTER TABLE ecu_serials RENAME TO ecus;
+CREATE TABLE secondary_ecus(serial TEXT PRIMARY KEY, sec_type TEXT, public_key_type TEXT, public_key TEXT, extra TEXT);
+
+DELETE FROM version;
+INSERT INTO version VALUES(23);
+
+RELEASE MIGRATION;
--- a/src/aktualizr_primary/CMakeLists.txt
+++ b/src/aktualizr_primary/CMakeLists.txt
@@ -1,11 +1,15 @@
-set(SOURCES main.cc secondary_config.h secondary_config.cc secondary.h secondary.cc)
+set(SOURCES main.cc secondary_config.cc secondary.cc)
+set(HEADERS secondary_config.h secondary.h)
 
 add_executable(aktualizr ${SOURCES})
 target_link_libraries(aktualizr aktualizr_lib virtual_secondary aktualizr-posix)
 target_include_directories(aktualizr PUBLIC ${PROJECT_SOURCE_DIR}/third_party)
 
-aktualizr_source_file_checks(${SOURCES})
-
 install(TARGETS aktualizr RUNTIME DESTINATION bin COMPONENT aktualizr)
+
+add_aktualizr_test(NAME primary_secondary_registration SOURCES primary_secondary_registration_test.cc secondary.cc secondary_config.cc PROJECT_WORKING_DIRECTORY ARGS ${PROJECT_BINARY_DIR}/uptane_repos LIBRARIES PUBLIC aktualizr-posix virtual_secondary uptane_generator_lib)
+target_include_directories(t_primary_secondary_registration PUBLIC ${PROJECT_SOURCE_DIR}/src/libaktualizr-posix)
+
+aktualizr_source_file_checks(${SOURCES} ${HEADERS} ${TEST_SOURCES})
 
 # vim: set tabstop=4 shiftwidth=4 expandtab:
--- a/src/aktualizr_primary/CMakeLists.txt
+++ b/src/aktualizr_primary/CMakeLists.txt
@@ -1,11 +1,8 @@
-set(TARGET aktualizr)
 set(SOURCES main.cc secondary_config.h secondary_config.cc secondary.h secondary.cc)
 
-add_executable(${TARGET} ${SOURCES} $<TARGET_OBJECTS:aktualizr-posix>)
-target_link_libraries(${TARGET} aktualizr_lib virtual_secondary)
-target_include_directories(${TARGET} PUBLIC
-  ${PROJECT_SOURCE_DIR}/src/libaktualizr-posix
-  ${PROJECT_SOURCE_DIR}/third_party)
+add_executable(aktualizr ${SOURCES})
+target_link_libraries(aktualizr aktualizr_lib virtual_secondary aktualizr-posix)
+target_include_directories(aktualizr PUBLIC ${PROJECT_SOURCE_DIR}/third_party)
 
 aktualizr_source_file_checks(${SOURCES})
 
--- a/src/aktualizr_primary/main.cc
+++ b/src/aktualizr_primary/main.cc
@@ -130,11 +130,9 @@ int main(int argc, char *argv[]) {
       try {
         Primary::initSecondaries(aktualizr, config.uptane.secondary_config_file);
       } catch (const std::exception &e) {
-        LOG_ERROR << "Secondary initialization failed: " << e.what();
-        if (!aktualizr.IsRegistered()) {
-          LOG_ERROR << "Cannot provision without all secondaries present, exiting...";
-          return EXIT_FAILURE;
-        }
+        LOG_ERROR << "Failed to init secondaries :" << e.what();
+        LOG_ERROR << "Exiting...";
+        return EXIT_FAILURE;
       }
     }
 
--- a/src/aktualizr_primary/secondary.cc
+++ b/src/aktualizr_primary/secondary.cc
@@ -4,6 +4,7 @@
 #include <boost/asio/placeholders.hpp>
 #include <boost/bind.hpp>
 
+#include <algorithm>
 #include <unordered_map>
 #include <unordered_set>
 
@@ -14,18 +15,19 @@
 namespace Primary {
 
 using Secondaries = std::vector<std::shared_ptr<Uptane::SecondaryInterface>>;
-using SecondaryFactoryRegistry = std::unordered_map<std::string, std::function<Secondaries(const SecondaryConfig&)>>;
+using SecondaryFactoryRegistry =
+    std::unordered_map<std::string, std::function<Secondaries(const SecondaryConfig&, Aktualizr& aktualizr)>>;
 
-static Secondaries createIPSecondaries(const IPSecondariesConfig& config);
+static Secondaries createIPSecondaries(const IPSecondariesConfig& config, Aktualizr& aktualizr);
 
 static SecondaryFactoryRegistry sec_factory_registry = {
     {IPSecondariesConfig::Type,
-     [](const SecondaryConfig& config) {
+     [](const SecondaryConfig& config, Aktualizr& aktualizr) {
        auto ip_sec_cgf = dynamic_cast<const IPSecondariesConfig&>(config);
-       return createIPSecondaries(ip_sec_cgf);
+       return createIPSecondaries(ip_sec_cgf, aktualizr);
      }},
     {VirtualSecondaryConfig::Type,
-     [](const SecondaryConfig& config) {
+     [](const SecondaryConfig& config, Aktualizr& /* unused */) {
        auto virtual_sec_cgf = dynamic_cast<const VirtualSecondaryConfig&>(config);
        return Secondaries({std::make_shared<VirtualSecondary>(virtual_sec_cgf)});
      }},
@@ -34,8 +36,8 @@ static SecondaryFactoryRegistry sec_factory_registry = {
     //  }
 };
 
-static Secondaries createSecondaries(const SecondaryConfig& config) {
-  return (sec_factory_registry.at(config.type()))(config);
+static Secondaries createSecondaries(const SecondaryConfig& config, Aktualizr& aktualizr) {
+  return (sec_factory_registry.at(config.type()))(config, aktualizr);
 }
 
 void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config_file) {
@@ -48,7 +50,7 @@ void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config
   for (auto& config : secondary_configs) {
     try {
       LOG_INFO << "Creating " << config->type() << " secondaries...";
-      Secondaries secondaries = createSecondaries(*config);
+      Secondaries secondaries = createSecondaries(*config, aktualizr);
 
       for (const auto& secondary : secondaries) {
         LOG_INFO << "Adding Secondary with ECU serial: " << secondary->getSerial()
@@ -141,20 +143,86 @@ class SecondaryWaiter {
   std::unordered_set<std::string> secondaries_to_wait_for_;
 };
 
-static Secondaries createIPSecondaries(const IPSecondariesConfig& config) {
+static Secondaries createIPSecondaries(const IPSecondariesConfig& config, Aktualizr& aktualizr) {
   Secondaries result;
-  SecondaryWaiter sec_waiter{config.secondaries_wait_port, config.secondaries_timeout_s, result};
+  const bool provision = !aktualizr.IsRegistered();
 
-  for (auto& ip_sec_cfg : config.secondaries_cfg) {
-    auto secondary = Uptane::IpUptaneSecondary::connectAndCreate(ip_sec_cfg.ip, ip_sec_cfg.port);
-    if (secondary) {
-      result.push_back(secondary);
-    } else {
-      sec_waiter.addSecondary(ip_sec_cfg.ip, ip_sec_cfg.port);
+  if (provision) {
+    SecondaryWaiter sec_waiter{config.secondaries_wait_port, config.secondaries_timeout_s, result};
+
+    for (const auto& ip_sec_cfg : config.secondaries_cfg) {
+      auto secondary = Uptane::IpUptaneSecondary::connectAndCreate(ip_sec_cfg.ip, ip_sec_cfg.port);
+      if (secondary) {
+        result.push_back(secondary);
+      } else {
+        sec_waiter.addSecondary(ip_sec_cfg.ip, ip_sec_cfg.port);
+      }
+    }
+
+    sec_waiter.wait();
+
+    // set ip/port in the db so that we can match everything later
+    for (size_t k = 0; k < config.secondaries_cfg.size(); k++) {
+      const auto cfg = config.secondaries_cfg[k];
+      const auto sec = result[k];
+      Json::Value d;
+      d["ip"] = cfg.ip;
+      d["port"] = cfg.port;
+      aktualizr.SetSecondaryData(sec->getSerial(), Utils::jsonToCanonicalStr(d));
+    }
+  } else {
+    auto secondaries_info = aktualizr.GetSecondaries();
+
+    for (const auto& cfg : config.secondaries_cfg) {
+      Uptane::SecondaryInterface::Ptr secondary;
+      const SecondaryInfo* info = nullptr;
+
+      auto f = std::find_if(secondaries_info.cbegin(), secondaries_info.cend(), [&cfg](const SecondaryInfo& i) {
+        Json::Value d = Utils::parseJSON(i.extra);
+        return d["ip"] == cfg.ip && d["port"] == cfg.port;
+      });
+
+      if (f == secondaries_info.cend() && config.secondaries_cfg.size() == 1 && secondaries_info.size() == 1) {
+        // /!\ backward compatibility: handle the case with one secondary, but
+        // store the info for later anyway
+        info = &secondaries_info[0];
+        Json::Value d;
+        d["ip"] = cfg.ip;
+        d["port"] = cfg.port;
+        aktualizr.SetSecondaryData(info->serial, Utils::jsonToCanonicalStr(d));
+        LOG_INFO << "Migrated single IP secondary to new storage format";
+      } else if (f == secondaries_info.cend()) {
+        // Match the other way if we can
+        secondary = Uptane::IpUptaneSecondary::connectAndCreate(cfg.ip, cfg.port);
+        if (secondary == nullptr) {
+          LOG_ERROR << "Could not instantiate secondary " << cfg.ip << ":" << cfg.port;
+          continue;
+        }
+        auto f_serial =
+            std::find_if(secondaries_info.cbegin(), secondaries_info.cend(),
+                         [&secondary](const SecondaryInfo& i) { return i.serial == secondary->getSerial(); });
+        if (f_serial == secondaries_info.cend()) {
+          LOG_ERROR << "Could not instantiate secondary " << cfg.ip << ":" << cfg.port;
+          continue;
+        }
+        info = &(*f_serial);
+      } else {
+        info = &(*f);
+      }
+
+      if (secondary == nullptr) {
+        secondary =
+            Uptane::IpUptaneSecondary::connectAndCheck(cfg.ip, cfg.port, info->serial, info->hw_id, info->pub_key);
+      }
+
+      if (secondary != nullptr) {
+        result.push_back(secondary);
+      } else {
+        LOG_ERROR << "Could not instantiate secondary " << info->serial;
+      }
     }
   }
 
-  sec_waiter.wait();
   return result;
 }
 
--- a/src/aktualizr_secondary/aktualizr_secondary.h
+++ b/src/aktualizr_secondary/aktualizr_secondary.h
@@ -21,6 +21,7 @@ class AktualizrSecondary : public Uptane::SecondaryInterface {
   AktualizrSecondary(AktualizrSecondaryConfig config, std::shared_ptr<INvStorage> storage,
                      std::shared_ptr<KeyManager> key_mngr, std::shared_ptr<UpdateAgent> update_agent);
 
+  std::string Type() const override { return ""; }
   Uptane::EcuSerial getSerial() const override;
   Uptane::HardwareIdentifier getHwId() const override;
   PublicKey getPublicKey() const override;
--- a/src/aktualizr_secondary/aktualizr_secondary.h
+++ b/src/aktualizr_secondary/aktualizr_secondary.h
@@ -27,6 +27,7 @@ class AktualizrSecondary : public Uptane::SecondaryInterface {
   PublicKey getPublicKey() const override;
 
   Uptane::Manifest getManifest() const override;
+  bool ping() const override { return true; }
   bool putMetadata(const Uptane::RawMetaPack& meta_pack) override { return putMetadata(Metadata(meta_pack)); }
   int32_t getRootVersion(bool director) const override;
   bool putRoot(const std::string& root, bool director) override;
--- a/src/libaktualizr-posix/ipuptanesecondary.cc
+++ b/src/libaktualizr-posix/ipuptanesecondary.cc
@@ -72,9 +72,12 @@ SecondaryInterface::Ptr IpUptaneSecondary::connectAndCheck(const std::string& ad
         return nullptr;
       }
       auto p = sec->getPublicKey();
-      if (p != pub_key) {
-        LOG_WARNING << "Mismatch between public keys " << p.Value() << " and " << pub_key.Value() << " for secondary "
-                    << serial;
+      if (pub_key.Type() == KeyType::kUnknown) {
+        LOG_INFO << "Secondary " << s << " do not have a known public key";
+      } else if (p != pub_key) {
+        LOG_ERROR << "Mismatch between public keys " << p.Value() << " and " << pub_key.Value() << " for secondary "
+                  << serial;
+        return nullptr;
       }
       return sec;
     }
--- a/src/libaktualizr-posix/ipuptanesecondary.cc
+++ b/src/libaktualizr-posix/ipuptanesecondary.cc
@@ -15,13 +15,13 @@ Uptane::SecondaryInterface::Ptr IpUptaneSecondary::connectAndCreate(const std::s
 
   ConnectionSocket con_sock{address, port};
 
-  if (con_sock.connect() != 0) {
-    LOG_ERROR << "Failed to connect to a secondary: " << std::strerror(errno);
-    return std::shared_ptr<Uptane::SecondaryInterface>();
+  if (con_sock.connect() == 0) {
+    LOG_INFO << "Connected to IP Secondary: "
+             << "(" << address << ":" << port << ")";
+  } else {
+    LOG_WARNING << "Failed to connect to a secondary: " << std::strerror(errno);
+    return nullptr;
   }
-
-  LOG_INFO << "Connected to IP Secondary: "
-           << "(" << address << ":" << port << ")";
 
   return create(address, port, *con_sock);
 }
@@ -50,6 +50,40 @@ Uptane::SecondaryInterface::Ptr IpUptaneSecondary::create(const std::string& add
            << "hw-ID: " << hw_id << " serial: " << serial;
 
   return std::make_shared<IpUptaneSecondary>(address, port, serial, hw_id, pub_key);
+}
+
+SecondaryInterface::Ptr IpUptaneSecondary::connectAndCheck(const std::string& address, unsigned short port,
+                                                           EcuSerial serial, HardwareIdentifier hw_id,
+                                                           PublicKey pub_key) {
+  // try to connect:
+  // - if it succeeds compare with what we expect
+  // - otherwise, keep using what we know
+  try {
+    auto sec = IpUptaneSecondary::connectAndCreate(address, port);
+    if (sec != nullptr) {
+      auto s = sec->getSerial();
+      if (s != serial) {
+        LOG_ERROR << "Mismatch between secondary serials " << s << " and " << serial;
+        return nullptr;
+      }
+      auto h = sec->getHwId();
+      if (h != hw_id) {
+        LOG_ERROR << "Mismatch between hardware ids " << h << " and " << hw_id;
+        return nullptr;
+      }
+      auto p = sec->getPublicKey();
+      if (p != pub_key) {
+        LOG_WARNING << "Mismatch between public keys " << p.Value() << " and " << pub_key.Value() << " for secondary "
+                    << serial;
+      }
+      return sec;
+    }
+  } catch (std::exception& e) {
+    LOG_WARNING << "Could not connect to secondary " << serial << " at " << address << ":" << port
+                << ", using previously known registration data";
+  }
+
+  return std::make_shared<IpUptaneSecondary>(address, port, std::move(serial), std::move(hw_id), std::move(pub_key));
 }
 
 IpUptaneSecondary::IpUptaneSecondary(const std::string& address, unsigned short port, EcuSerial serial,
--- a/src/libaktualizr-posix/ipuptanesecondary.cc
+++ b/src/libaktualizr-posix/ipuptanesecondary.cc
@@ -181,4 +181,16 @@ Manifest IpUptaneSecondary::getManifest() const {
   std::string manifest = ToString(r->manifest.choice.json);  // NOLINT
   return Utils::parseJSON(manifest);
 }
+
+bool IpUptaneSecondary::ping() const {
+  Asn1Message::Ptr req(Asn1Message::Empty());
+  req->present(AKIpUptaneMes_PR_getInfoReq);
+
+  auto m = req->getInfoReq();
+
+  auto resp = Asn1Rpc(req, getAddr());
+
+  return resp->present() == AKIpUptaneMes_PR_getInfoResp;
+}
+
 }  // namespace Uptane
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -891,8 +891,18 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
     }
   }
 
+  // wait some time for secondaries to come up
+  // note: this fail after a time out but will be retried at the next install
+  // phase if the targets have not been changed. This is done to avoid being
+  // stuck in an unrecoverable state here
+  if (!waitSecondariesReachable(updates)) {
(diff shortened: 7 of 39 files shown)


config/sql/migration/migrate.23.sql
```diff
@@ -0,0 +1,12 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+ALTER TABLE ecu_serials ADD COLUMN sec_type TEXT NOT NULL DEFAULT "";
```
mike-sul: The following comment is not exactly related to this PR and is not blocker for its merging, just simply general comment to the overall data model of aktualizr.

I think, it makes sense to have a notion of ECU in the data model (not ecu_serial) and corresponding table in the DB (ECU table), so `manifest`, `public_key`, etc are attributes of ECU not ecu_serials entity.

src/aktualizr_primary/main.cc
```diff
@@ -131,11 +131,9 @@ int main(int argc, char *argv[]) {
       try {
         Primary::initSecondaries(aktualizr, config.uptane.secondary_config_file);
       } catch (const std::exception &e) {
-        LOG_ERROR << "Secondary initialization failed: " << e.what();
-        if (!aktualizr.IsRegistered()) {
-          LOG_ERROR << "Cannot provision without all secondaries present, exiting...";
-          return EXIT_FAILURE;
-        }
+        LOG_ERROR << e.what();
```
mike-sul: Just minor comment. Perhaps, it makes sense to add some message before `e.what` or combine it with the next LOG_ERROR.

src/aktualizr_primary/secondary.cc
```diff
@@ -141,20 +147,53 @@ class SecondaryWaiter {
   std::unordered_set<std::string> secondaries_to_wait_for_;
 };
 
-static Secondaries createIPSecondaries(const IPSecondariesConfig& config) {
+static Secondaries createIPSecondaries(const IPSecondariesConfig& config, bool provision, Aktualizr& aktualizr) {
```
mike-sul: This dependency of the IP Secondary factory on Aktualizr is questionable. Actually, it's me who implicitly introduced this dependency by putting createIPSecondaries() implementation in this file :). Anyway, this is not exactly comment to this PR, just thing that we might need to consider at  some point.

src/aktualizr_primary/secondary.cc
```diff
@@ -141,20 +147,53 @@ class SecondaryWaiter {
...
+      if (secondary) {
+        result.push_back(secondary);
+      } else {
+        sec_waiter.addSecondary(ip_sec_cfg.ip, ip_sec_cfg.port);
+      }
+    }
+
+    sec_waiter.wait();
+  } else {
+    auto secondaries_info = aktualizr.GetRegisteredSecondaries();
```
mike-sul: Can the secondaries that have been added to the aktualizr instance be 'not registered', if not then perhaps renaming of GetRegisteredSecondaries() to just GetSecondaries() would make sense. Or, as an option, `storage_` can be passed to createIPSecondaries instead of `aktualizr` so we can call `storage_->loadSecondaryInfo(&info);` here and replace dependency on Aktualizr with dependency on Storage which looks better since it reduces cross-dependencies, IMHO.

src/aktualizr_primary/secondary.cc
```diff
@@ -141,20 +147,53 @@ class SecondaryWaiter {
...
+    }
+
+    if (secondaries_info.size() != config.secondaries_cfg.size()) {
+      std::stringstream err_ss;
+      err_ss << "Expected " << config.secondaries_cfg.size() << " IP secondaries, but device was provisioned with "
+             << secondaries_info.size();
+      throw std::runtime_error(err_ss.str());
+    }
+
+    // note: order in configuration and storage should be the same
```
mike-sul: perhaps, IP and port could be stored in the DB too, so an order that secondaries are listed in the conf won't matter?

src/aktualizr_primary/secondary.cc
```diff
@@ -141,20 +147,53 @@ class SecondaryWaiter {
...
+    // matches, without matching on their type
+    if (secondaries_info.size() != config.secondaries_cfg.size()) {
+      secondaries_info.erase(
+          std::remove_if(secondaries_info.begin(), secondaries_info.end(),
+                         [](const SecondaryInfo& info) -> bool { return info.type != IPSecondariesConfig::Type; }));
+    }
+
+    if (secondaries_info.size() != config.secondaries_cfg.size()) {
+      std::stringstream err_ss;
+      err_ss << "Expected " << config.secondaries_cfg.size() << " IP secondaries, but device was provisioned with "
```
mike-sul: In this case we have to make sure that device is not provisioned if one or more of the secondaries listed in the conf is not available during the provisioning phase.

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -52,6 +52,36 @@ Uptane::SecondaryInterface::Ptr IpUptaneSecondary::create(const std::string& add
...
+SecondaryInterface::Ptr IpUptaneSecondary::connectAndCheck(const std::string& address, unsigned short port,
+                                                           EcuSerial serial, HardwareIdentifier hw_id,
+                                                           PublicKey pub_key) {
+  // try to connect:
+  // - if it succeeds compare with what we expect
+  // - otherwise, keep using what we know
+  try {
+    auto sec = IpUptaneSecondary::connectAndCreate(address, port);
+    if (sec != nullptr) {
+      auto s = sec->getSerial();
```
mike-sul: Mismatch of serial and hardware ID looks like critical to me, I am not sure if anything will work after such error at all?

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -52,6 +52,36 @@ Uptane::SecondaryInterface::Ptr IpUptaneSecondary::create(const std::string& add
...
+        LOG_ERROR << "Mismatch between hardware ids " << h << " and " << hw_id;
+      }
+      auto p = sec->getPublicKey();
+      if (p != pub_key) {
+        LOG_ERROR << "Mismatch between public keys " << p.Value() << " and " << pub_key.Value();
+      }
+      // note: we still trust what the device gave us for now
+      return sec;
+    }
+  } catch (std::exception& e) {
```
mike-sul: I think, an error or debug message would be helpful here.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1060,6 +1085,49 @@ void SotaUptaneClient::verifySecondaries() {
...
+  LOG_INFO << "Waiting for secondaries to connect to start installation...";
+
+  // tries for 10 minutes
+  for (unsigned int tries = 1; tries < 600; tries++) {
+    if (targeted_secondaries.empty()) {
+      return true;
+    }
+
+    for (auto sec_it = targeted_secondaries.begin(); sec_it != targeted_secondaries.end();) {
+      if (sec_it->second->getManifest() != Json::Value()) {
```
mike-sul: sec_it->second->getManifest() sets up a TCP connection with Secondary, a timeout of the setting up TCP connection really depends on multiple factors and configuration of TCP stack at a kernel level. The same is true for write/read to/from socket, so it won't be 10 minutes for sure and actually will significantly differs in a case of a local host and a non-local one (http://willbryant.net/overriding_the_default_linux_kernel_20_second_tcp_socket_connect_timeout)

src/libaktualizr/storage/invstorage.h
```diff
@@ -150,6 +161,9 @@ class INvStorage {
   virtual bool loadEcuSerials(EcuSerials* serials) = 0;
```
mike-sul: Since we are extending the ecu_serials table with the additional attributes/columns why not extend the corresponding write/read functions, i.e loadEcuSerials/storeEcuSerials/clearEcuSerials instead of introducing the new methods.

config/sql/migration/migrate.23.sql
lbonn: Yes, I've also wondered if it should just be another table.

As we've added fields to `ecu_serials` it has effectively become what you describe and would probably more aptly named `ecus`.

src/aktualizr_primary/secondary.cc
lbonn: `GetSecondaries()` would probably work.

My first version was also passing the storage around, the problem I had is that call originates from `src/aktualizr_primary/main.cc` directly where the underlying storage object is not accessible.

Passing `std::vector<SecondaryInfo>` might be fine or maybe too specific? Also, because of the previous point, the aktualizr method would still be needed in the current state.

src/aktualizr_primary/secondary.cc
lbonn: In the same vein as the previous comment, it is made harder by the fact we've abstracted the ip secondary concept from SotaUptaneClient and the storage.

One version I've explored was to add another free form string column to the `ecu_serials` table and `SecondaryInfo` that might have to be re-parsed.

src/aktualizr_primary/secondary.cc
lbonn: Indeed! This should be covered by what we added in #1491.

src/libaktualizr-posix/ipuptanesecondary.cc
lbonn: Changed in latest version

src/libaktualizr-posix/ipuptanesecondary.cc
lbonn: Added in latest version

lbonn:
Looks like I've overlooked something, https://github.com/advancedtelematic/aktualizr/pull/1533/commits/e2e62c6120c8f5b97e66eae3d23a3af37014048b breaks Uptane.UptaneSecondaryMisconfigured

patrickvacek:
> Looks like I've overlooked something, [e2e62c6](https://github.com/advancedtelematic/aktualizr/commit/e2e62c6120c8f5b97e66eae3d23a3af37014048b) breaks Uptane.UptaneSecondaryMisconfigured

Not necessarily unexpected, right? That test might need some reworking to still be meaningful.

src/aktualizr_primary/secondary.cc
patrickvacek: We'll want to avoid tables with secondary type-specific columns, so the free-form string might be the best option. Tedious but might be the most future-proof.

mike-sul: I agree that adding secondary specific columns into a generic table (ecu_serials here) is a bad idea. I see a few options here:

1) extend the existing ecu_serials table with an opaque column (e.g. sec_data, sec_info, etc) intended for serialized secondary specific data required for its instantiation;

2) don't alter the existing ecu_serials table. Introduce another table `ecu` with the generic fields + the column containing serialized secondary specific data;

3) Introduce another table ip_secondary that contains the ip secondary specific info. This table is managed solely by the IP secondary factory object/method/function. In this case we can avoid dependency on Aktualizr object at all, what is needed just a path to the DB if we would like to store this table in the same DB file.

4) Introduce two new tables ECU containing generic columns + the ip_secondary specific table.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -218,10 +216,27 @@ Json::Value SotaUptaneClient::AssembleManifest() {
...
 
   for (auto it = secondaries.begin(); it != secondaries.end(); it++) {
+    const Uptane::EcuSerial &ecu_serial = it->first;
     Uptane::Manifest secmanifest = it->second->getManifest();
 
+    bool from_cache = false;
+    if (secmanifest == Json::Value()) {
+      // could not get the secondary manifest, a cached value will be provided
+      std::string cached;
+      if (storage->loadCachedEcuManifest(ecu_serial, &cached)) {
```
mike-sul: This 'manifest caching' functionality looks like IP Secondary specific functionality, so I would 'hide' it behind the secondary interface and  implemented it in `IpUptaneSecondary::getManifest()`.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -871,8 +886,18 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
     }
   }
 
+  // wait some time for secondaries to come up
+  // note: this fail after a time out but will be retried at the next install
+  // phase if the targets have not been changed. This is done to avoid being
+  // stuck in an unrecoverable state here
+  if (!waitSecondariesReachable(updates)) {
```
mike-sul: I am not sure that this functionality belongs here, IMHO, this is secondary specific functionality and should not be in a generic code. Why we wait just here before the sending metatada but not wait before sending actual images, secondary can be available at the first step but becomes not available before the second (installation) step ?
This rhetorical question is another argument for having this `wait` or `retry policy` withing the IP secondary implementation, specifically, within IpUptaneSecondary implementation.

src/libaktualizr/uptane/secondaryinterface.h
```diff
@@ -14,6 +14,7 @@ class SecondaryInterface {
   using Ptr = std::shared_ptr<SecondaryInterface>;
 
  public:
+  virtual std::string Type() const = 0;
```
mike-sul: I think, this `Type()` as well as the logic of passing around of the `registered` secondaries can be eliminated if we persist required information about IP secondaries within the IP secondary builder/factory, e.g. in `createIPSecondaries` under `if (provision) {` branch. Therefore, we won't need aktualizr instance here and in its `else` we won't need to `secondaries_info.erase(`, we will just load the info from the DB.

mike-sul:
Regardless of the solution/PR design, I think, it makes sense to add test(s) into `ipsecondary_test.py` testing the given use-case.

src/libaktualizr/storage/invstorage.h
lbonn: To answer generally, the fact that this is a column shouldn't really drive the storage api changes, in my opinion. I see it as two independent choices: what makes the api more usable and where does it make sense to store this additional data.

Including it in the existing functions should be possible but I'm afraid it would require a lot of changes in different places, for example the tests.

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: Yes that's true, I will rework something.

src/libaktualizr/storage/invstorage.h
mike-sul: At the moment, I think, that it would be better not to touch this table `ecu_serials` nor the API, just introduce another table.

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: I would like that also, but there should then be a way to access the storage from there. To evaluate

src/libaktualizr/storage/invstorage.h
lbonn: Do you imply that your original suggestion to add these to load/store/clear EcuSerials is now obsolete?

mike-sul: > Do you imply that your original suggestion to add these to load/store/clear EcuSerials is now obsolete?

Yes, I am pretty sure that it's wrong direction now.

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: BTW, why not just do certain amount of retries within getManifest() (I mean Asn1Rpc call) or keep retrying until certain time if we need to wait at least some period of time.

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: Maybe dedicated table in sql.db or even DB that is initialized in IP secondary factory (preferable) or ctor.

src/libaktualizr/uptane/secondaryinterface.h
lbonn: Passing the `provision` around was indeed needlessly complicated, the last version is a bit more simple.

I'll continue the investigations about the `Type()` method and dedicated storage.

lbonn:
Here is another version with two tables `ecus` and `secondary_ecus`, it uses some more complicated sql.

lbonn:
The version currently pushed only caches the manifest in memory and it is indeed simpler. But I am gonna revert to put it in the storage somehow because it doesn't fit the bill: if the secondary never gets up between the primary start and the attempted install, the backend marks the install as failed before it can start as the manifest is incomplete. And we clearly want to support this case.

tests/ipsecondary_test.py
```diff
@@ -168,6 +170,42 @@ def test_primary_timeout_during_first_run(uptane_repo, secondary, aktualizr, **k
...
+    end_ev = threading.Event()
+    def run_secondary():
+        time.sleep(10)
+        with secondary:
+            end_ev.wait()
+
+    sec_thread = threading.Thread(target=run_secondary)
+
+    with aktualizr:
+        sec_thread.start()
```
mike-sul: why it cannot be 
`
with aktualizr:
   time.sleep(10)
   with secondary:
       aktualizr.wait_for_completion()

config/sql/migration/migrate.23.sql
```diff
@@ -0,0 +1,10 @@
+-- Don't modify this! Create a new migration instead--see docs/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+ALTER TABLE ecu_serials RENAME TO ecus;
+CREATE TABLE secondary_ecus(serial TEXT PRIMARY KEY, sec_type TEXT NOT NULL, public_key_type TEXT NOT NULL, public_key TEXT NOT NULL, FOREIGN KEY(serial) REFERENCES ecus ON DELETE CASCADE ON UPDATE CASCADE);
```
mike-sul: IMHO, these two tables should not know anything about each other. libaktualizr should not know details of any secondary implementation. In ideal world the IP secondary should have its own dedicated persistent layer.

lbonn: Well the *real* world is all about compromises.

tests/ipsecondary_test.py
lbonn: yes, that's true. thanks!

lbonn:
It should be close to the state I want to reach. I'll just want to resolve the TODO I've added in `secondary.cc` before merging.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -151,6 +151,162 @@ void SQLStorage::clearPrimaryKeys() {
...
+void SQLStorage::saveSecondaryData(const Uptane::EcuSerial& ecu_serial, const std::string& data) {
+  SQLite3Guard db = dbConnection();
+
+  if (!db.beginTransaction()) {
+    LOG_ERROR << "Can't start transaction: " << db.errmsg();
+    return;
+  }
+
+  auto statement =
+      db.prepareStatement<std::string>("SELECT count(*) FROM secondary_ecus WHERE serial = ?;", ecu_serial.ToString());
```
patrickvacek: This too.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: I think it's fine. We're likely to need a generic way to wait for Secondaries regardless of protocol. Obviously for something like virtual Secondaries it can just return true. And if we can improve this later, that's fine too.

src/aktualizr_secondary/aktualizr_secondary.h
```diff
@@ -21,11 +21,13 @@ class AktualizrSecondary : public Uptane::SecondaryInterface {
   AktualizrSecondary(AktualizrSecondaryConfig config, std::shared_ptr<INvStorage> storage,
                      std::shared_ptr<KeyManager> key_mngr, std::shared_ptr<UpdateAgent> update_agent);
 
+  std::string Type() const override { return ""; }
```
patrickvacek: What is the significance of leaving this empty by default?

src/aktualizr_primary/secondary.cc
```diff
@@ -14,18 +15,19 @@
...
-static Secondaries createIPSecondaries(const IPSecondariesConfig& config);
+static Secondaries createIPSecondaries(const IPSecondariesConfig& config, Aktualizr& aktualizr);
 
 static SecondaryFactoryRegistry sec_factory_registry = {
     {IPSecondariesConfig::Type,
-     [](const SecondaryConfig& config) {
+     [](const SecondaryConfig& config, Aktualizr& aktualizr) {
        auto ip_sec_cgf = dynamic_cast<const IPSecondariesConfig&>(config);
-       return createIPSecondaries(ip_sec_cgf);
+       return createIPSecondaries(ip_sec_cgf, aktualizr);
```
patrickvacek: Passing in the Aktualizr object here seems like a big change, but I don't see any immediate concerns. Are there other ways to do this that would simplify this?

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -52,6 +52,40 @@ Uptane::SecondaryInterface::Ptr IpUptaneSecondary::create(const std::string& add
...
+      }
+      auto h = sec->getHwId();
+      if (h != hw_id) {
+        LOG_ERROR << "Mismatch between hardware ids " << h << " and " << hw_id;
+        return nullptr;
+      }
+      auto p = sec->getPublicKey();
+      if (p != pub_key) {
+        LOG_WARNING << "Mismatch between public keys " << p.Value() << " and " << pub_key.Value() << " for secondary "
+                    << serial;
```
patrickvacek: Should we store the new key in this case? Or is that already covered by `SotaUptaneClient::addSecondary`?

tests/ipsecondary_test.py
```diff
@@ -201,8 +229,7 @@ def test_primary_timeout_after_device_is_registered(uptane_repo, secondary, aktu
     with aktualizr:
         aktualizr.wait_for_completion()
 
-    return (aktualizr.get_current_primary_image_info() == primary_image_hash)\
-            and not aktualizr.is_ecu_registered(secondary.id)
+    return aktualizr.get_current_primary_image_info() == primary_image_hash
```
patrickvacek: What necessitated simplifying this condition?

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -151,6 +151,162 @@ void SQLStorage::clearPrimaryKeys() {
...
+  key_type_str = key_type_ss.str();
+  key_type_str.erase(std::remove(key_type_str.begin(), key_type_str.end(), '"'), key_type_str.end());
+
+  if (!db.beginTransaction()) {
+    LOG_ERROR << "Can't start transaction: " << db.errmsg();
+    return;
+  }
+
+  auto statement =
+      db.prepareStatement<std::string>("SELECT count(*) FROM secondary_ecus WHERE serial = ?;", ecu_serial.ToString());
```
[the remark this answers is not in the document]
lbonn: `serial` is a primary key of the table, so this is already enforced.

src/aktualizr_secondary/aktualizr_secondary.h
lbonn: This is used on the secondary side and the `Type()` is not used there. But maybe setting it to  `ip` here would be more accurate.

patrickvacek: Sounds like it's fine as is if it isn't actually used. Is it just there because of the interface, then?

At some point, `aktualizr-secondary` was intended to be reusable by multiple communication protocols. How explicitly tied to the IP Secondary logic is it now?

src/aktualizr_primary/secondary.cc
lbonn: Maybe the new methods to Aktualizr is the biggest change in this context. We're coming from `initSecondaries` that already takes an `Aktualizr` instance. But these are the two sides of the same problem

The alternatives that I know of would be:
- to pass the storage directly in some way (but it's opaque from here, as it's in Aktualizr). Having a `Aktualizr::GetStorage()` would be quite ugly.
- create another storage instance from the storage path and use it there. Probably not SQLStorage but something that derives from SQLStorageBase. I am would be worried about the added boilerplate.
- use another dedicated storage mechanism (?)

src/aktualizr_secondary/aktualizr_secondary.h
lbonn: Ah I forgot about this, yes the method is here because of the interface.
I'll keep it as empty then.

src/aktualizr_primary/secondary.cc
mike-sul: Back then when I introduced this 'Primary::initSecondaries' my intention/plan was and still is to consider it as an initial version of generic Abstract Factory that is part of Secondary Registry mechanism that we might introduce in the future. So, I consider 'Primary::initSecondaries' as something that belongs to Primary/main namespace and it's OK to pass an instance of Aktualizr to it. Having said that, I think that it's not OK to pass this instance to a specific secondary factory method, like createIPSecondaries() as it should belong IpUptaneSecondary namespace. Also,  createIPSecondaries() should be under IpUptaneSecondary namespace too, I did not put it there in the first place because we didn't want to spend much time to implementation of some secondary registration mechanism (e.g. something similar to what Laurent proposes for the package manager registry).

patrickvacek: It's probably fine as is, but my concern is that the Aktualizr class is mostly meant to be an API for library users, and I think we're extending that a bit here. I fear that something similar is happening to what has happened to SotaUptaneClient. I'm not sure how to fix that, and it might just take some time to step back and figure out a good refactoring strategy. Or maybe I'm worrying too much about things that don't matter.

lbonn: > It's probably fine as is, but my concern is that the Aktualizr class is mostly meant to be an API for library users, and I think we're extending that a bit here.

I share your concern but in this case, adding a secondary implementation is expected by our library users, so this might be in the scope. But it's indeed not clear cut.

mike-sul: IMHO, both Aktualizr::GetStorage() and std::vector<SecondaryInfo> Aktualizr::GetSecondaries() along with Aktualizr::SetSecondaryData() are quite ugly. Also, I have the same feeling about adding of SotaUptaneClient::waitSecondariesReachable() into SotaUptaneClient. 
IMHO, both of them are consequences of trying to use business logic specific to one secondary (IP secondary in our case) or a few secondaries (but not all, probable future case, e.g. IP Secondary, CAN Secondary but not Virtual Secondary or Docker Secondary) in an object/objects that operate over array of secondaries (Aktualizr, SotaUptaneClient).

Even, if some business logic,e.g. this "wait" stuff, is applicable/relevant to some of the secondaries it SHOULD NOT be injected into Aktualizr/SotaUptaneClient but just simply reused in the relevant secondaries' implementation by means of composition (e.g. the way we reuse the uptane verification logic) or inheritance (base class that include the wait logic). In any case, this logic should be "hidden" behind the secondary interface.

That's why I strongly against of Aktualizr::GetStorage(), Aktualizr::GetSecondaries(), Aktualizr::SetSecondaryData(), SotaUptaneClient::waitSecondariesReachable() as well as this really ugly SecondaryInterface::ping() which is BTW also another consequances of injecting secondary specific business logic into SotaUptaneClient.

I think if we just reuse SQLStorageBase() in createIPSecondaries then we would not need all the aforementioned functions at all and overall implementation will be actually simpler compared to the current implementation.

In the long run, I think it absolutely makes sense to reconsider the storage/persistence engine for secondary because of the usage of a SQL engine for storing a relatively small number of records in a single table seam like way overhead. BTW, I think, the same is true for libaktualizr, why do we need SQL DB for it? Why not use some simple a key-value storage, e.g. https://github.com/google/leveldb (this is just example).

lbonn: > In the long run, I think it absolutely makes sense to reconsider the storage/persistence engine for secondary because of the usage of a SQL engine for storing a relatively small number of records in a single table seam like way overhead. BTW, I think, the same is true for libaktualizr, why do we need SQL DB for it? Why not use some simple a key-value storage, e.g. [google/leveldb](https://github.com/google/leveldb) (this is just example).

This is another discussion entirely and frankly it sounds like change for the sake of change to me. I believe the main advantages we got from using sqlite was to have atomic transactions and corruption resistance (leveldb seems to have some type of transactions FWIW) and it's also a very largely deployed and tested piece of code with few dependencies. We haven't had a problem with the supposed overhead that we pay now but we'd definitely have issues migrating to something else.

patrickvacek: > Even, if some business logic,e.g. this "wait" stuff, is applicable/relevant to some of the secondaries it SHOULD NOT be injected into Aktualizr/SotaUptaneClient but just simply reused in the relevant secondaries' implementation by means of composition (e.g. the way we reuse the uptane verification logic) or inheritance (base class that include the wait logic). In any case, this logic should be "hidden" behind the secondary interface.

While I agree that the composition or inheritance options are viable, I disagree that the `wait` or `ping` logic is unique to IP Secondaries. These are methods that we will probably need for almost every Secondary that sits on a separate system, which is the default expectation. Virtual and Docker Secondaries are an exception to the rule, and arguably still not the definitive solution to the problem we are using them to solve. I'd rather treat the expected use case as the default and optimize for that. We can revise the current solution later, but I think it's acceptable as is.

> In the long run, I think it absolutely makes sense to reconsider the storage/persistence engine for secondary because of the usage of a SQL engine for storing a relatively small number of records in a single table seam like way overhead.

I agree, we probably need to rework this sooner or later. Feel free to make a ticket to track that work.

> the same is true for libaktualizr, why do we need SQL DB for it? Why not use some simple a key-value storage, e.g. https://github.com/google/leveldb (this is just example).

SQLite is pretty small and has some nice atomicity guarantees. Does leveldb really provide the same features as SQLite? How easy would it be to support migration, table joins, typed columns, etc.? I'm open to alternative storage backends but I don't want to invest the effort without a substantial reason to believe that it would be a significant improvement.

mike-sul: > I disagree that the `wait` or `ping` logic is unique to IP Secondaries.

It's not exactly what I say. I mean, I am not saying that `wait` or `ping` logic is unique to IP Secondaries, I think and agree that it can be relevant for more than one secondary. What I am trying to say is that this logic can be reused/applied in another way, e.g. just include it into implementations of secondaries that need it, what IMHO, simpler, cleaner,  compared to adding it to sotauptaneclient/aktualizr and polluting interface.

mike-sul: > a very largely deployed and tested piece of code with few dependencies. We haven't had a problem with the supposed overhead that we pay now but we'd definitely have issues migrating to something else

Yes, I agree, sqlite has been proven by the time and is quite reliable piece of software and there is no particular reason for replacing it, at the moment. Yet, SQL DB usage in case of aktualizr's use-case is quite overhead and dealing with the stuff that Patrick mentioned would be much simpler.

patrickvacek: > What I am trying to say is that this logic can be reused/applied in another way, e.g. just include it into implementations of secondaries that need it, what IMHO, simpler, cleaner, compared to adding it to sotauptaneclient/aktualizr and polluting interface.

I think I see your point now. Do you agree that we need some abstract notion of `wait` and `ping`? The implementations can indeed live in the subclasses, but my point was that we need some way to ask if a Secondary is available.

tests/ipsecondary_test.py
lbonn: In this PR, even if a secondary is not connected when aktualizr starts, it is still considered as present.

`is_ecu_registered()` only returns True if the primary abandoned the idea of running the secondary after the original time out and would not send updates its way.

patrickvacek:
This is still labeled "do-not-merge-until-ready". Is that accurate? I think there is still room for improvement, as the ongoing discussions have indicated, but I think we could take this as a good first step and work from there.

src/aktualizr_primary/secondary.cc
lbonn: @patrickvacek Mike suggested that if we want to wait for secondaries, we could so in the implementation of `install()` for example. I don't think he advocates having abstract notions of `wait()` and `ping()` as virtual methods (@mike-sul please correct me if I'm wrong).

In principle I think it's a good design and I also agree that it's dangerous to make this interface grow too much but there are things that would be harder to implement this way:

- waiting a fixed time for all secondaries globally and not waiting N times
- do not attempt to update the primary if secondaries are not available and the updates target them. We can still rollback and that's maybe something we should think about anyway (since secondary install can still fail even we found the secondaries)

We might rethink these things or just decide that we don't need it but I think that the clean solution is also less powerful.

src/libaktualizr-posix/ipuptanesecondary.cc
lbonn: The storage part is indeed covered in `SotaUptaneClient` but I've added another commit to try to cover more cases.

lbonn:
@patrickvacek I've removed the label but I intended to add one or two more tests.

src/aktualizr_primary/secondary.cc
mike-sul: > I don't think he advocates having abstract notions of `wait()` and `ping()` as virtual methods (@mike-sul please correct me if I'm wrong).

that's correct. I strongly against it :).

> Do you agree that we need some abstract notion of wait and ping? The implementations can indeed live in the subclasses, but my point was that we need some way to ask if a Secondary is available.

No, I disagree. IMHO, we already have functionality that kind of "waits" for IP Secondary, it's all software stack running under Asn1Rpc(). What we need is just extend/wrap this functionality in order to add additional policy(ies) into it, e.g. desired timeout/wait time as in this specific case  and then use parametrized Asn1Rpc() from IpUptaneSecondary methods. For example, in this specific case, we could call Asn1Rpc(timeout=10m) in IpUptaneSecondary::install() and Asn1Rpc(timeout=60s) in another (but this is also questionable since we might wanna wait "longer" at the stage of sending metadata too).

lbonn: The solution you're describing just doesn't have the same behavior as what we've agreed on when we took on this task. The two points I've raised are what we want to satisfy right now. 

I sympathize with you as you want to keep away the bloat and coupling and I see that this change can seem to go in the wrong direction and might make things more difficult to untangle later.

However after discussing with @patrickvacek , we still think the best way forward right now is to try to get this merged first. The two main reasons are:

- to implement what we agreed on right now, we believe that overriding the `install()` step is not sufficient. As we are giving more fine-grained control on the installation process, we will inevitably need more fine-grained interfaces. It does not necessarily have to be something low-level like `ping()`, we could also imagine having a `globalPreInstall()` step in `SecondaryInterface()` but that would maybe require more careful design, that we haven't yet started
- we do not claim this is a perfect solution that will be set in stone. Though, moving the piece of code later when we have a more satisfying design should be relatively easy. There is a balance to be found between a hacky solution and an all-encompassing plan that is never implemented in practice. And as we are trying to work in an iterative fashion, the current version feels like it's just good enough right now.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -28,11 +28,14 @@ void SotaUptaneClient::addSecondary(const std::shared_ptr<Uptane::SecondaryInter
...
   SecondaryInfo info;
-  if (!storage->loadSecondaryInfo(serial, &info) || info.type == "") {
+  if (!storage->loadSecondaryInfo(serial, &info) || info.type == "" || info.pub_key.Type() == KeyType::kUnknown) {
     info.serial = serial;
     info.hw_id = sec->getHwId();
     info.type = sec->Type();
-    info.pub_key = sec->getPublicKey();
+    const PublicKey &p = sec->getPublicKey();
+    if (p.Type() != KeyType::kUnknown) {
+      info.pub_key = sec->getPublicKey();
```
patrickvacek:
```suggestion
      info.pub_key = p;
```
Oder?

tests/ipsecondary_test.py
```diff
@@ -232,6 +232,36 @@ def test_primary_timeout_after_device_is_registered(uptane_repo, secondary, aktu
...
 
+@with_uptane_backend()
+@with_secondary(start=False)
+@with_secondary(start=False, arg_name='secondary2')
+@with_aktualizr(start=False, output_logs=True)
+def test_primary_multiple_secondaries(uptane_repo, secondary, secondary2, aktualizr, **kwargs):
+    '''Test Aktualizr with multiple ip secondaries'''
+
+    with aktualizr, secondary, secondary2:
+        aktualizr.wait_for_completion()
```
patrickvacek: This is all happy-path, with everything more or less starting simultaneously, right? Can we extended this for the cases when one or more Secondaries don't come up immediately?

src/aktualizr_primary/secondary.cc
mike-sul: > just doesn't have the same behavior as what we've agreed on when we took on this task

It won't be exactly the same internally, but effect/outcome from an external standpoint of view will be the same.

> overriding the install() step is not sufficient.

I don't think we need to override install() step, we were mistaken referring to install() method at all, it should be rather sendFirmware() or putMetadata() or both.

> we could also imagine having a globalPreInstall() step in SecondaryInterface() but that would maybe require more careful design, that we haven't yet started

I am not sure if I understand if we need to have it at all. I think this idea about "waiting" for secondaries prior to sending firmware to it is conceptually flawed if to look into it from a broader perspective and take into account other aspects such as layered nature on networking stack and potential use-case of "dynamic" secondary support.

> There is a balance to be found between a hacky solution and an all-encompassing plan that is never implemented in practice. And as we are trying to work in an iterative fashion, the current version feels like it's just good enough right now.

That's fair enough and makes sense in many cases but in this particular case the implementation option that goes in line with an all-encompassing plan (proper design) is simpler and easier compared to the given implementation.

@lbonn @patrickvacek There is no need to persuade me :). We've had a great and constructive discussion around this PR and it's absolutely OK that different people have different opinions. I fully understand that discussions cannot last forever, and a decision should be made. What worked the best for me at my previous companies if there was no full consensus among peers is to let a tech lead/architect (or owner of software under discussion) to make a final decision and everyone else should accept and complain to it.

So, that's absolutely fine with me if this change is merged :) as long as this change fulfill the original goal/issue and the CI jobs are happy.

patrickvacek: I appreciate the robust discussion; I think that leads us to a better solution in the end and helps keep us aware of the trade-offs and alternatives. My suggestion for now is to move forward with this PR as is to satisfy our immediate product goals, and then have a followup discussion (perhaps at the sprint review or the next relevant grooming) to consider the longer-term product goals and then decide where to go from here.

src/aktualizr_primary/CMakeLists.txt
```diff
@@ -1,14 +1,15 @@
...
-
-aktualizr_source_file_checks(${SOURCES})
+add_executable(aktualizr ${SOURCES})
+target_link_libraries(aktualizr aktualizr_lib virtual_secondary aktualizr-posix)
+target_include_directories(aktualizr PUBLIC ${PROJECT_SOURCE_DIR}/third_party)
 
 install(TARGETS aktualizr RUNTIME DESTINATION bin COMPONENT aktualizr)
 
+add_aktualizr_test(NAME primary_secondary_registration SOURCES primary_secondary_registration_test.cc secondary.cc secondary_config.cc PROJECT_WORKING_DIRECTORY ARGS ${PROJECT_BINARY_DIR}/uptane_repos LIBRARIES PUBLIC $<TARGET_OBJECTS:aktualizr-posix> virtual_secondary uptane_generator_lib)
+target_include_directories(t_primary_secondary_registration PUBLIC ${PROJECT_SOURCE_DIR}/src/libaktualizr-posix)
```
patrickvacek: Missing the actual test file?

lbonn: Thanks! It went away during a probably-too-involved rebase.

PR #1534, opened 2020-01-24
OTA-4411 Update image references to use new location

tkfu:
This changes the docs pages to use the new repository for images over at https://github.com/advancedtelematic/ota-connect-docs-images.


tkfu:
Should be doable, but I'd like to do it in a separate PR, just so we can get the structural change closed off without too much trouble.

PR #1538, opened 2020-01-27
aktualizr-lite: Include primary ECU for target download

doanac:
With the introduction of commit
 fead283be7813688f9f69037d2c879a9857157c5

SotaUptaneClient::downloadImage now requires the Target to have
EcuSerial information that would normally come from Uptane. When running
aktualizr-lite, the Target is taken directly from the TUF targets and
does not include this information.

This change allows aktualizr-lite to:

1) Set primary ECU information in the SotaUptaneClient
This change isn't too bad (splitting logic into smaller pieces).
However, it would make more sense if this was a private method.

2) Set Ecu information on the Uptane::Target object.

### DIFF
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -108,7 +108,7 @@ static std::unique_ptr<Uptane::Target> find_target(const std::shared_ptr<SotaUpt
 }
 
 static int do_update(LiteClient &client, Uptane::Target &target) {
-  target.InsertEcu(client.primary_ecu.first, client.primary_ecu.second);
+  target.InsertEcu(client.primary_ecu);
   if (!client.primary->downloadImage(target).first) {
     return 1;
   }
--- a/src/aktualizr_lite/main.cc
+++ b/src/aktualizr_lite/main.cc
@@ -108,6 +108,7 @@ static std::unique_ptr<Uptane::Target> find_target(const std::shared_ptr<SotaUpt
 }
 
 static int do_update(LiteClient &client, Uptane::Target &target) {
+  target.InsertEcu(client.primary_ecu.first, client.primary_ecu.second);
   if (!client.primary->downloadImage(target).first) {
     return 1;
   }
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -266,6 +266,16 @@ Json::Value SotaUptaneClient::AssembleManifest() {
 
 bool SotaUptaneClient::hasPendingUpdates() const { return storage->hasPendingInstall(); }
 
+void SotaUptaneClient::initializePrimaryEcu() {
+  EcuSerials serials;
+  if (!storage->loadEcuSerials(&serials) || serials.size() == 0) {
+    throw std::runtime_error("Unable to load ECU serials after device registration.");
+  }
+
+  primary_ecu_serial_ = serials[0].first;
+  hw_ids.insert(serials[0]);
+}
+
 void SotaUptaneClient::initialize() {
   LOG_DEBUG << "Checking if device is provisioned...";
   auto keys = std::make_shared<KeyManager>(storage, config.keymanagerConfig());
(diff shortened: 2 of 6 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -266,6 +266,16 @@ Json::Value SotaUptaneClient::AssembleManifest() {
...
 bool SotaUptaneClient::hasPendingUpdates() const { return storage->hasPendingInstall(); }
 
+void SotaUptaneClient::initializePrimaryEcu() {
+  EcuSerials serials;
+  if (!storage->loadEcuSerials(&serials) || serials.size() == 0) {
+    throw std::runtime_error("Unable to load ECU serials after device registration.");
+  }
+
+  primary_ecu_serial_ = serials[0].first;
+  hw_ids.insert(serials[0]);
```
mike-sul: As an alternative, `primary_ecu_serial_` and `hw_ids` could be initialized in SotaUptaneClient ctor if the additional params EcuSerial and HardwareIdentifier are added to it that equal to Uptane::EcuSerial::Unknown() and HardwareIdentifier::Unknown() respectively. In this case we won't this additional method and aktualizr-lite won't need to do additional post-initialization after SotaUptaneClient object is constructed.

Actually, I think we should get rid of  SotaUptaneClient::initialize() in aktualizr too.

src/aktualizr_lite/main.cc
```diff
@@ -108,6 +108,7 @@ static std::unique_ptr<Uptane::Target> find_target(const std::shared_ptr<SotaUpt
 }
 
 static int do_update(LiteClient &client, Uptane::Target &target) {
+  target.InsertEcu(client.primary_ecu.first, client.primary_ecu.second);
```
mike-sul: It means that aktualizr-lite supports updating just a single ECU. If an update of multiple ECUs is required then ECUs (or hardware IDs) has to be specified in the targets.json and Target object respectively.

src/libaktualizr/uptane/tuf.h
```diff
@@ -262,6 +262,10 @@ class Target {
   void setUri(std::string uri) { uri_ = std::move(uri); };
   bool MatchHash(const Hash &hash) const;
 
+  void InsertEcu(const EcuSerial &ecuIdentifier, const HardwareIdentifier &hwid) {
```
mike-sul: Why not 'insert' the pair? It's declared as the pair in the lite client context `std::pair<Uptane::EcuSerial, Uptane::HardwareIdentifier> primary_ecu;` so passing it around as the pair would be a bit more convenient, IMHO.

src/aktualizr_lite/main.cc
doanac: yep - this is currently limited to a single primary ECU. I think we could untangle this logic to do secondaries, but i just haven't had anyone ask me for it.

src/libaktualizr/uptane/tuf.h
doanac: Good point.  Fixed

src/libaktualizr/primary/sotauptaneclient.cc
doanac: I took a fairly conservative approach to doing this and left the current initialize function in tact.

I think things are cleaner now. Thing are in 3 commit so you can see your comments being addressed. However, I think you'll want to squash them into 1 if you decide this is ready to be merged.

src/aktualizr_lite/main.cc
mike-sul: ok

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: thx, it looks great now.

mike-sul:
@lbonn This PR looks good to me, but I don't have permissions to merge it.

lbonn:
Yes, the changes look good. Can we maybe squash the commits into one?

lbonn:
Actually used the "squash and merge" button :)

mike-sul:
> Actually used the "squash and merge" button :)

Nice button, I'd like to have such button too :)

lbonn:
@mike-sul I think you actually could merge this PR. The problem was that nobody approved it when you asked and you can also approve it yourself.

mike-sul:
> @mike-sul I think you actually could merge this PR. The problem was that nobody approved it when you asked and you can also approve it yourself.
> (I also had this "administrator required" cross before I approved)

I don't think so, currently, I am authorized to merge into master just for meta-updater :) for aktualizr I see the message that I am not authorized to merge even after PR is approved by a reviewer(s)

lbonn:
Ah you're right, you have write rights but the master branch is "protected" and only maintainers can merge there.

PR #1539, opened 2020-01-27
Improved step-by-step instructions for device credential provisioning

tkfu:
This is a complete overhaul of the step-by-step instructions for device credential provisioning. The previous instructions were even more opaque than necessary, and just directed the user to paste a whole bunch of openssl commands. When I looked at them in detail, we were having people generate a server chain of trust, giving files confusing or non-descriptive names (like, multiple different `ca.crt` files only distinguished by location in the directory tree, and just generally not being remotely helpful at explaining what was really going on.

This new version isn't exactly x.509 for dummies, but it should at least be a lot clearer about what we're doing and why we're doing it.


tkfu:
I should add, this also still preserves the ability to ignore the instructions, blindly copy-paste everything inside a code block into a terminal, and still have it work.

lbonn:
I haven't reproduced the steps manually (the copy paste exercise), but the instructions are indeed more clear.

tkfu:
Thanks! I'm just gonna rebase on 2020.2-docs before I merge.

PR #1542, opened 2020-01-29
Update the changelog and versions for 2020.2

mike-sul:


lbonn:
I've pushed a small tweak: we don't need the entry about the docs change.

PR #1544, opened 2020-01-29
aktualizr-lite: notify if not configured

ldts:
notify the user if aktualizr-lite is not configured to be built.

Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>

### DIFF
--- a/src/aktualizr_lite/CMakeLists.txt
+++ b/src/aktualizr_lite/CMakeLists.txt
@@ -23,7 +23,8 @@ add_aktualizr_test(NAME lite-helpers SOURCES helpers.cc helpers_test.cc
                    ARGS ${PROJECT_BINARY_DIR}/ostree_repo)
 set_tests_properties(test_lite-helpers PROPERTIES
         ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:t_lite-mock> LABELS "noptest")
-
+else(BUILD_OSTREE)
+	message(STATUS "aktualizr-lite not configured, skipping")
 endif(BUILD_OSTREE)
 
 aktualizr_source_file_checks(main.cc ${AKTUALIZR_LITE_SRC} ${AKTUALIZR_LITE_HEADERS} helpers_test.cc ostree_mock.cc)


ldts:
use case:

what is: aktualzr-lite is not being installed after a successful build.
reason: BUILD_OSTREE is not being defined.
what (I think) should be: when the required set of configs is not being requested, inform if some of the binaries will not be build and installed

lbonn:
We have other executables gated by a config option, like garage-push/BUILD_SOTA_TOOLS and we've kept the false case silent so far.
It can be useful for troubleshooting when starting on the project but one usually quickly learn the options they're interested in.

That said, it's not that verbose and we're already output a lot of junk during cmake configure. So maybe that would be fine but we'll probably want to keep that consistent with the other executables concerned.

src/aktualizr_lite/CMakeLists.txt
```diff
@@ -23,7 +23,8 @@ add_aktualizr_test(NAME lite-helpers SOURCES helpers.cc helpers_test.cc
                    ARGS ${PROJECT_BINARY_DIR}/ostree_repo)
 set_tests_properties(test_lite-helpers PROPERTIES
         ENVIRONMENT LD_PRELOAD=$<TARGET_FILE:t_lite-mock> LABELS "noptest")
-
+else(BUILD_OSTREE)
+	message(STATUS "acktualizer-lite not configured, skipping")
```
lbonn:
```suggestion
	message(STATUS "aktualizr-lite not configured, skipping")
```

patrickvacek:
I think I agree with Laurent that this makes this a bit of an outlier. There are several features that require cmake variables set to build them correctly. I'm not convinced that this deserves a special exception. It's also quite easy to lose it in the mess of generated output.

That said, I'm open to considering how we can make these things clearer in the documentation. Is there something we could improve there that would make things clearer for users that might not get what they expect?

Or we could go all the way and print out a list of features enabled and disabled at configure time. That could also be useful.

PR #1548, opened 2020-02-03
OLPSUP-9046: Handle expired Director's target metadata

mike-sul:
- Apply pending update even if its metadata are expired provided that it
was installed before the expiration
- Add tests for the target metadata expiration use-case

### DIFF
--- a/src/aktualizr_secondary/update_agent_ostree.cc
+++ b/src/aktualizr_secondary/update_agent_ostree.cc
@@ -78,18 +78,7 @@ data::ResultCode::Numeric OstreeUpdateAgent::install(const Uptane::Target& targe
 }
 
 data::InstallationResult OstreeUpdateAgent::applyPendingInstall(const Uptane::Target& target) {
-  if (!_ostreePackMan->rebootDetected()) {
-    // it should be removed from here, once we refactor the ostree package manager,
-    // e.g. _ostreePackMan->finalizeInstall() does this check/if
-    return data::InstallationResult(data::ResultCode::Numeric::kNeedCompletion,
-                                    "Reboot is required for the pending update application");
-  }
-
-  data::InstallationResult install_result = _ostreePackMan->finalizeInstall(target);
-  // it should be removed from here, once we refactor the ostree package manager,
-  // e.g. the pacman will reset/clear the flag by itself
-  _ostreePackMan->rebootFlagClear();
-  return install_result;
+  return _ostreePackMan->finalizeInstall(target);
 }
 
 void extractCredentialsArchive(const std::string& archive, std::string* ca, std::string* cert, std::string* pkey,
--- a/src/libaktualizr/package_manager/ostreemanager.cc
+++ b/src/libaktualizr/package_manager/ostreemanager.cc
@@ -215,24 +215,20 @@ data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &ta
                                     "Reboot is required for the pending update application");
   }
 
-  data::InstallationResult install_result = applyInstall(target);
-  // it should be removed from here, once we refactor the ostree package manager,
-  // e.g. the pacman will reset/clear the flag by itself
-  bootloader_->rebootFlagClear();
-  return install_result;
-}
-
-data::InstallationResult OstreeManager::applyInstall(const Uptane::Target &target) {
   LOG_INFO << "Checking installation of new OSTree sysroot";
   const std::string current_hash = getCurrentHash();
+
+  data::InstallationResult install_result =
+      data::InstallationResult(data::ResultCode::Numeric::kOk, "Successfully booted on new version");
 
   if (current_hash != target.sha256Hash()) {
     LOG_ERROR << "Expected to boot " << target.sha256Hash() << " but found " << current_hash
               << ". The system may have been rolled back.";
-    return data::InstallationResult(data::ResultCode::Numeric::kInstallFailed, "Wrong version booted");
+    install_result = data::InstallationResult(data::ResultCode::Numeric::kInstallFailed, "Wrong version booted");
   }
 
-  return data::InstallationResult(data::ResultCode::Numeric::kOk, "Successfully booted on new version");
+  bootloader_->rebootFlagClear();
+  return install_result;
 }
 
 OstreeManager::OstreeManager(PackageConfig pconfig, BootloaderConfig bconfig, std::shared_ptr<INvStorage> storage,
--- a/src/libaktualizr/package_manager/ostreemanager.cc
+++ b/src/libaktualizr/package_manager/ostreemanager.cc
@@ -209,7 +209,20 @@ void OstreeManager::completeInstall() const {
   bootloader_->reboot();
 }
 
-data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &target) const {
+data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &target) {
+  if (!bootloader_->rebootDetected()) {
+    return data::InstallationResult(data::ResultCode::Numeric::kNeedCompletion,
+                                    "Reboot is required for the pending update application");
+  }
+
+  data::InstallationResult install_result = applyInstall(target);
+  // it should be removed from here, once we refactor the ostree package manager,
+  // e.g. the pacman will reset/clear the flag by itself
+  bootloader_->rebootFlagClear();
+  return install_result;
+}
+
+data::InstallationResult OstreeManager::applyInstall(const Uptane::Target &target) {
   LOG_INFO << "Checking installation of new OSTree sysroot";
   const std::string current_hash = getCurrentHash();
 
--- a/src/libaktualizr/package_manager/ostreemanager.h
+++ b/src/libaktualizr/package_manager/ostreemanager.h
@@ -65,7 +65,6 @@ class OstreeManager : public PackageManagerInterface {
 
  private:
   TargetStatus verifyTargetInternal(const Uptane::Target &target) const;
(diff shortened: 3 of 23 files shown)


mike-sul:
@lbonn Please, see if it makes sense to you, I am just not sure if this is the right fix for https://saeljira.it.here.com/browse/OLPSUP-9046, as an alternative we could "cancel" a pending update if its metadata are expired.

src/libaktualizr/uptane/directorrepository.cc
```diff
@@ -63,6 +63,7 @@ bool DirectorRepository::checkMetaOffline(INvStorage& storage) {
     }
 
     if (targetsExpired()) {
+      last_exception = Uptane::ExpiredMetadata(type.toString(), Role::Targets().ToString());
```
lbonn: :+1:

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -76,54 +76,52 @@ data::InstallationResult SotaUptaneClient::PackageInstall(const Uptane::Target &
...
-    // nothing to do
+  // TODO: consider bringing checkAndUpdatePendingSecondaries and the following functionality
+  // to the common denominator
+  if (!hasPendingUpdates()) {
+    LOG_INFO << "No any pending update, continue with initialization";
     return;
   }
 
-  LOG_INFO << "Device has been rebooted after an update";
+  LOG_INFO << "Check if there is any pending update for Primary ECU to be applied.";
```
lbonn: This text reads like a suggestion to the user, whereas it's a description of the current operation.

```suggestion
  LOG_INFO << "Checking for pending update for primary ECU to apply";
```

src/libaktualizr/package_manager/packagemanagerinterface.h
```diff
@@ -35,47 +35,24 @@ enum class TargetStatus {
 
 class PackageManagerInterface {
  public:
-  PackageManagerInterface(PackageConfig pconfig, BootloaderConfig bconfig, std::shared_ptr<INvStorage> storage,
```
lbonn: As you mentioned in your comment, this would break #1498. Do you then have an alternative design to propose? Also, the exception in this PR is the Debian package manager which we don't really care about.

Maybe an intermediate solution would just be to pass a const reference to the full config here, so that implementations can pick what's needed.

In any case, I would prefer if we kept this change in another PR since it's independent from the other changes.

mike-sul: It's not independent, the interface cleanup depends on the fix implementation.

I thought that this is what Eugene intended eventually to do with moving the bootloader instance under the ostree package manager.

I will remove the pack man interface cleanup from this PR and create another PR, although the later will depend on this PR.

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: ok, will update accordingly

src/aktualizr_secondary/update_agent_ostree.cc
```diff
@@ -78,18 +78,7 @@ data::ResultCode::Numeric OstreeUpdateAgent::install(const Uptane::Target& targe
...
-    return data::InstallationResult(data::ResultCode::Numeric::kNeedCompletion,
-                                    "Reboot is required for the pending update application");
-  }
-
-  data::InstallationResult install_result = _ostreePackMan->finalizeInstall(target);
-  // it should be removed from here, once we refactor the ostree package manager,
-  // e.g. the pacman will reset/clear the flag by itself
-  _ostreePackMan->rebootFlagClear();
-  return install_result;
+  return _ostreePackMan->finalizeInstall(target);
```
lbonn: Not related to this PR really but according to good sources, identifiers starting with underscores should be avoided as they are reserved for "C++ implementations": https://stackoverflow.com/a/20496955/1931271

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -209,7 +209,20 @@ void OstreeManager::completeInstall() const {
...
 
-data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &target) const {
+data::InstallationResult OstreeManager::finalizeInstall(const Uptane::Target &target) {
+  if (!bootloader_->rebootDetected()) {
+    return data::InstallationResult(data::ResultCode::Numeric::kNeedCompletion,
+                                    "Reboot is required for the pending update application");
+  }
+
+  data::InstallationResult install_result = applyInstall(target);
+  // it should be removed from here, once we refactor the ostree package manager,
```
lbonn: Should this comment be removed?

src/libaktualizr/package_manager/ostreemanager.h
```diff
@@ -65,6 +65,7 @@ class OstreeManager : public PackageManagerInterface {
 
  private:
   TargetStatus verifyTargetInternal(const Uptane::Target &target) const;
+  data::InstallationResult applyInstall(const Uptane::Target &target);
```
lbonn: Do we need this new method? We would have `install`, `completeInstall`, `finalizeInstall` and `applyInstall`. This is starting to get a bit confusing. Or maybe it's missing some docs and/or   comments?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -76,54 +76,52 @@ data::InstallationResult SotaUptaneClient::PackageInstall(const Uptane::Target &
...
-    LOG_ERROR << "Invalid Uptane metadata in storage.";
+    // finalize failed, unset pending flag so that the rest of the uptane process can go forward again
+    storage->saveInstalledVersion(primary_ecu_serial.ToString(), *pending_target, InstalledVersionUpdateMode::kNone);
+    report_queue->enqueue(std_::make_unique<EcuInstallationCompletedReport>(primary_ecu_serial, correlation_id, false));
   }
 
-  package_manager_->rebootFlagClear();
+  director_repo.dropTargets(*storage);  // fix for OTA-2587, listen to backend again after end of install
+  computeDeviceInstallationResult(nullptr, correlation_id);
+  // Do we really need to send a device manifest here at the initialization phase
```
lbonn: The uptane cycle does not necessarily need to happen at regular intervals. The state of the installation has changed, so we try to send it as soon as it happens.

lbonn:
Looks good, just had some minor comments.
The tests are also nice :+1:.

src/aktualizr_secondary/update_agent_ostree.cc
mike-sul: It actually contradicts the google coding guidelines that we supposed to follow, I found out about it just after this change was merged. I will change this underscore naming to make it consistent with our coding style in another PR. 

But personally I prefer/like this code guideline https://www.appinf.com/download/CppCodingStyleGuide.pdf which says the following in regard of underscore usage:

"
```
Rule 30
Global names must not begin with one or two underscores (“_” or “__”).
This is forbidden by the C++ standard as such names are reserved for use by
the compiler.

Rule 31
With the exceptions of constants and enumeration values, there are no
underscores in names (after the first character of the name). 2
"

Rule 33
The name of a private or protected member variable begins with a single
underscore.3
 This is the only underscore in such a name (see Rule 31).


```

src/libaktualizr/package_manager/ostreemanager.cc
mike-sul: yes

src/libaktualizr/package_manager/ostreemanager.h
mike-sul: It's a private helper method and is not part of the class interface as the other listed methods are, but it, indeed is not really useful now and can be removed

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: ok

src/aktualizr_secondary/update_agent_ostree.cc
lbonn: Ah cool, this guide might have interesting insights as well.

In this case, however, even the C++ standard has something to say (2.10):

> In addition, some identifiers are reserved for use by C++implementations and shall not be used otherwise;no diagnostic is required.
— Each identifier that contains a double underscore__or begins with an underscore followed by an uppercase letter is reserved to the implementation for any use.
— Each identifier that begins with an underscore is reserved to the implementation for use as a name int he global namespace.

It doesn't exactly conflict as these are not in the global namespace and do not follow the underscore by an uppercase letter but that's quite close.

src/libaktualizr/package_manager/ostreemanager.h
mike-sul: updated

mike-sul:
@lbonn Updated

simao:
I am not sure we should merge this. The backend is returning expired metadata, which is wrong, there is  fix in sit right now for this

mike-sul:
> I am not sure we should merge this. The backend is returning expired metadata, which is wrong, there is fix in sit right now for this

Without this fix the issue will persists even if the backend returns correctly refreshed metadata (not expired date stamp and incremented version). I emulated the correct back-end behavior by using uptane-generator utility and could reproduce the issue, changes in this PR fixes this issue.

lbonn:
This is a different issue than the backend one.

The scenario is:

- aktualizr deploys a tree from valid metadata but does not reboot
- targets metadata used for the install expire
- the device reboots, see that some install was ongoing, check the metadata stored on the device (no interaction with backend here). Since the metadata is expired, aktualizr panics.

simao:
@lbonn ah ok, my mistake.

PR #1555, opened 2020-02-12
Feat/better logging

patrickvacek:
Requested by @koshelev to speed up debugging, particularly when all we have is a log with the default loglevel. This is approximately how it will look:

```
Adding Secondary with ECU serial: 2c1644ed995acb07bebc4f0b1d0d3ad4865a7c3cb8eeb6de6f58bb543cd4c188 with hardware ID: virtualsec-demo
Primary ECU serial: 6d0a57cc949f6eaa1f0fb9e08984cdbc9a7328c66990290a7e5b5408061bce70 with hardware ID: local-fake
Device ID: grand-apfelkuchen-701
Device Gateway URL: https://3c877e1a-c969-4b71-9ba9-16b909a88c44.device-gateway.ota.api.here.com:443
Certificate subject: CN=grand-apfelkuchen-701, OU=Devices
Certificate issuer: CN=Device Registration
Certificate valid from: Jan  9 09:51:35 2020 GMT until Dec  4 09:38:36 2028 GMT
```

### DIFF
--- a/src/aktualizr_primary/secondary.cc
+++ b/src/aktualizr_primary/secondary.cc
@@ -51,8 +51,8 @@ void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config
       Secondaries secondaries = createSecondaries(*config);
 
       for (const auto& secondary : secondaries) {
-        LOG_INFO << "Adding Secondary to Aktualizr."
-                 << "HW_ID: " << secondary->getHwId() << " Serial: " << secondary->getSerial();
+        LOG_INFO << "Adding Secondary with ECU serial: " << secondary->getSerial()
+                 << " with hardware ID: " << secondary->getHwId();
         aktualizr.AddSecondary(secondary);
       }
     } catch (const std::exception& exc) {
--- a/src/libaktualizr/crypto/keymanager.cc
+++ b/src/libaktualizr/crypto/keymanager.cc
@@ -192,8 +192,64 @@ std::string KeyManager::getCN() const {
   }
   boost::scoped_array<char> buf(new char[len + 1]);
   X509_NAME_get_text_by_NID(X509_get_subject_name(x.get()), NID_commonName, buf.get(), len + 1);
-  std::string cn(buf.get());
+  const std::string cn(buf.get());
   return cn;
+}
+
+void KeyManager::getCertInfo(std::string *subject, std::string *issuer, std::string *not_before,
+                             std::string *not_after) const {
+  std::string not_found_cert_message = "Certificate is not found, can't extract device certificate";
+  std::string cert;
+  if (config_.tls_cert_source == CryptoSource::kFile || config_.tls_cert_source == CryptoSource::kAndroid) {
+    if (!backend_->loadTlsCert(&cert)) {
+      throw std::runtime_error(not_found_cert_message);
+    }
+  } else {  // CryptoSource::kPkcs11
+    if (!built_with_p11) {
+      throw std::runtime_error("Aktualizr was built without PKCS#11 support, can't extract device certificate");
+    }
+    if (!(*p11_)->readTlsCert(&cert)) {
+      throw std::runtime_error(not_found_cert_message);
+    }
+  }
+
+  StructGuard<BIO> bio(BIO_new_mem_buf(const_cast<char *>(cert.c_str()), static_cast<int>(cert.size())), BIO_vfree);
+  StructGuard<X509> x(PEM_read_bio_X509(bio.get(), nullptr, nullptr, nullptr), X509_free);
+  if (x == nullptr) {
+    throw std::runtime_error("Could not parse certificate");
+  }
+
+  StructGuard<BIO> subj_bio(BIO_new(BIO_s_mem()), BIO_vfree);
+  X509_NAME_print_ex(subj_bio.get(), X509_get_subject_name(x.get()), 1, 0);
+  char *subj_buf = nullptr;
+  auto subj_len = BIO_get_mem_data(subj_bio.get(), &subj_buf);  // NOLINT(cppcoreguidelines-pro-type-cstyle-cast)
+  if (subj_buf == nullptr) {
+    throw std::runtime_error("Could not parse certificate subject");
+  }
+  *subject = std::string(subj_buf, static_cast<size_t>(subj_len));
+
+  StructGuard<BIO> issuer_bio(BIO_new(BIO_s_mem()), BIO_vfree);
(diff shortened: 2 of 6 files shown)


src/libaktualizr/crypto/keymanager.cc
```diff
@@ -192,10 +192,60 @@ std::string KeyManager::getCN() const {
...
   boost::scoped_array<char> buf(new char[len + 1]);
   X509_NAME_get_text_by_NID(X509_get_subject_name(x.get()), NID_commonName, buf.get(), len + 1);
-  std::string cn(buf.get());
+  const std::string cn(buf.get());
   return cn;
 }
 
+void KeyManager::getCertInfo(std::string *subject, std::string *issuer, std::string *not_before,
+                             std::string *not_after) const {
+  std::string not_found_cert_message = "Certificate is not found, can't extract device_id";
```
lbonn:
```suggestion
  std::string not_found_cert_message = "Certificate is not found, can't extract device certificate";
```

src/libaktualizr/crypto/keymanager.cc
```diff
@@ -192,10 +192,60 @@ std::string KeyManager::getCN() const {
...
+  std::string cert;
+  if (config_.tls_cert_source == CryptoSource::kFile || config_.tls_cert_source == CryptoSource::kAndroid) {
+    if (!backend_->loadTlsCert(&cert)) {
+      throw std::runtime_error(not_found_cert_message);
+    }
+  } else {  // CryptoSource::kPkcs11
+    if (!built_with_p11) {
+      throw std::runtime_error("Aktualizr was built without PKCS#11 support, can't extract device_id");
```
lbonn:
```suggestion
      throw std::runtime_error("Aktualizr was built without PKCS#11 support, can't extract device certificate");
```

src/libaktualizr/crypto/keymanager.cc
```diff
@@ -192,10 +192,60 @@ std::string KeyManager::getCN() const {
...
+  }
+
+  StructGuard<BIO> bio(BIO_new_mem_buf(const_cast<char *>(cert.c_str()), static_cast<int>(cert.size())), BIO_vfree);
+  StructGuard<X509> x(PEM_read_bio_X509(bio.get(), nullptr, nullptr, nullptr), X509_free);
+  if (x == nullptr) {
+    throw std::runtime_error("Could not parse certificate");
+  }
+
+  StructGuard<BIO> subj_bio(BIO_new(BIO_s_mem()), BIO_vfree);
+  X509_NAME_print_ex(subj_bio.get(), X509_get_subject_name(x.get()), 1, 0);
```
lbonn: The docs I've found are not really helpful but are we sure that this call or the next (`BIO_get_mem_data`) can't fail and we end up creating a string from an uninitialized pointer (`subj_buf`).

Maybe a simple precaution:

```
char *subj_buf = nullptr;
...
if (subj_buf == nullptr) {
    throw xxx;
}
*subject = std::string(subj_buf, ...);
```

lbonn:
Looks like it's gonna actually be helpful in the future!

src/libaktualizr/crypto/keymanager.cc
patrickvacek: Yeah I was annoyed that the SSL docs don't even mention the return value of `X509_NAME_print_ex`. However, https://manpages.debian.org/unstable/libssl-doc/X509_NAME_print_ex.3ssl.en.html says that "X509_NAME_print_ex() and X509_NAME_print_ex_fp() return 1 on success or 0 on error if the XN_FLAG_COMPAT is set, which is the same as X509_NAME_print(). Otherwise, it returns -1 on error or other values on success." Lovely. I'll play around with this a bit and try to be a bit more careful.

patrickvacek: I'm a bit scared of the return value mess so I ended up just taking your advice literally.

patrickvacek:
Judging by the previous CI run (https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/jobs/1577872) it looks like I may have some openssl version compatibility issues to confront. Cool...

mike-sul:
> Judging by the previous CI run (https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/jobs/1577872) it looks like I may have some openssl version compatibility issues to confront. Cool...

looks quite similar to what I've seen on macOS, I think, the problem is that our cmake wants 1.0.2 while our code actually relies/requires on 1.1.x since switching from 1.0.2t version to 1.1.1t on the mac helped me to resolve it.

src/aktualizr_primary/secondary.cc
```diff
@@ -51,8 +51,8 @@ void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config
       Secondaries secondaries = createSecondaries(*config);
 
       for (const auto& secondary : secondaries) {
-        LOG_INFO << "Adding Secondary to Aktualizr."
-                 << "HW_ID: " << secondary->getHwId() << " Serial: " << secondary->getSerial();
+        LOG_INFO << "Adding Secondary with ECU serial: " << secondary->getSerial()
```
mike-sul: mixing metadata/values with message text might not be optimal as you need to search for values between text. The best practice here is to separate a static part of a log message from its variable part. For example,
`
Adding Secondary:
    serial: <serial>
    hardware ID: <hwid>
`

but it's not critical here as there are just two variable elements here, it's rather important when there are more of them

mike-sul:
good start, I think, it makes sense to review and improve logging in other places too.

src/aktualizr_primary/secondary.cc
patrickvacek: I'm not quite sure I follow. All I did was rephrase it and swap the serial and hardware ID.

patrickvacek:
> > Judging by the previous CI run (https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/jobs/1577872) it looks like I may have some openssl version compatibility issues to confront. Cool...
> 
> looks quite similar to what I've seen on macOS, I think, the problem is that our cmake wants 1.0.2 while our code actually relies/requires on 1.1.x since switching from 1.0.2t version to 1.1.1t on the mac helped me to resolve it.

I think we generally still want to support 1.0.2, so I have to work around that. The reason I didn't already, though, was that the [website](https://www.openssl.org/docs/man1.1.0/man3/X509_get0_notBefore.html) claims that "These functions are available in all versions of OpenSSL." Obviously that's not true.

patrickvacek:
Fixed the openssl compatibility issue and incorporated all of @lbonn's suggestions. Your reviews were auto-dismissed but I'm ready for another round now.

PR #1557, opened 2020-02-13
dockerapp: Include support for upcoming version 1.0

doanac:
The 1.0 version of docker-app (technically speaking even everything >0.8) has some important improvements. However, it does change functionality in a way that makes things hard to handle in a single implementation. This change makes the current docker-app manager class a pass-through to point at the correct implementation as defined at run-time. With this approach we can do seamless migrations assuming people create a proper targets.json that has both the old format and the new pinned "uri" attribute.

Its my hope, that over the next couple of months we can remove the old implementation and only worry about the new style.


src/libaktualizr/package_manager/dockerapp_bundles.cc
```diff
@@ -0,0 +1,169 @@
...
+struct AppBundle {
+  AppBundle(std::string app_name, const DockerAppManagerConfig &config)
+      : name(std::move(app_name)),
+        app_root(config.docker_apps_root / name),
+        app_params(config.docker_app_params),
+        compose_bin(boost::filesystem::canonical(config.docker_compose_bin).string()) {}
+
+  // Utils::shell isn't interactive. The docker app commands can take a few
+  // seconds to run, so we use std::system to stream it to stdout/sterr
+  bool cmd_streaming(const std::string &cmd) { return std::system(cmd.c_str()) == 0; }
```
patrickvacek: Is the problem just that the output of `Utils::shell` is only dumped once the call returns? Or why do you need an interactive shell?

No big deal at any rate, this is fine if it does what you need.

src/libaktualizr/package_manager/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ aktualizr_source_file_checks(ostreemanager.cc ostreereposync.cc
                              ostreemanager.h ostreereposync.h)
 
 aktualizr_source_file_checks(androidmanager.cc androidmanager.h)
-aktualizr_source_file_checks(dockerappmanager.cc dockerappmanager.h dockerappmanager_test.cc)
+aktualizr_source_file_checks(dockerappmanager.cc dockerappmanager.h dockerapp_standalone.cc dockerapp_bundles.cc dockerappmanager_test.cc)
```
mike-sul: This change https://github.com/advancedtelematic/aktualizr/commit/d06aa4fb06e408b03df39b1fe0216e3ef05d8e97 provides a capability to build the dockerapp pacman out the libaktualizr's cmakes/build system as well as register it without a need in libaktualizr code modification.

doanac: I'm quite interested in the pluggable package managers. I could you all eventually wanting me to maintain this out of tree. For now, I'm hoping to keep in tree because of all the advantages of being in the upstream. However, I can understand if over time this just isn't worth the effort for you to worry about (ie merge conflict type stuff).

src/libaktualizr/package_manager/dockerapp_bundles.cc
doanac: Its to stream output as the command runs. In the original implementation, I'd copy and pasted that logic around and had to keep that comment. So for this new one I decided to put it in one place so it wasn't as confusing.

src/libaktualizr/package_manager/CMakeLists.txt
mike-sul: The dockerapp package manager actually consists of two package managers, ostree and dockerapp. It potentially can be split into two package managers, but libaktualizr doesn't support multiple package managers. Another alternative here is to use the secondary mechanism to extend Primary's default ostree update with the dockeapp one. In this case, the dockerapp package manager will be represented as Secondary in libaktualizr, which allows you to build it and register at libaktualizr without modifying libaktualizr code.

doanac:
> How does the application decide which of the implementations to use? I think I'm missing how you can still end up with the old version.
This logic in package manager decides at run-time:
~~~
    auto standalone = DockerAppManagerConfig(pconfig).docker_app_bin;
    if (boost::filesystem::exists(standalone)) {
~~~

> Would it make sense to have a test to cover the upgrade scenario?

That's a good idea. What I can do is have the test Target for each version have both types of app specs. eg:
~~~
 "docker_apps": {
  "test-app": {
    "uri": "should get used by new version",
    "filename": "should get used by old version"
  }
 }
~~~

That's actually how I'm doing things for our devices now. I create TUF Targets with both items in it. This allows a device to upgrade from the standalone docker-app to the new version seamlessly. And it can also rollback if needed.

src/libaktualizr/package_manager/CMakeLists.txt
patrickvacek: > I'm quite interested in the pluggable package managers. I could you all eventually wanting me to maintain this out of tree. For now, I'm hoping to keep in tree because of all the advantages of being in the upstream. However, I can understand if over time this just isn't worth the effort for you to worry about (ie merge conflict type stuff).

That's still an open topic of discussion. For now, I think it's fine, especially since it's pretty tightly coupled due to the lack of a proper TUF API. If we had something like that, though, and we got it to a point that it was reasonably stable, then maybe it would be time to separate it. But for now, it might even be beneficial to have it directly in this project so that we know what the expectations are for the use cases that make the DockerApp package manager unique.

Part of the point of the pluggable package manger, though, is to make the barrier of entry to people making their own lower.

Both of Mike's point are quite salient as well. Supporting multiple package managers is another long-term goal of ours, and again, we'll want to make sure that whatever solution we end up with works for your use case. Treating the DockerApps as independent Secondaries is logically closer to what we've recommended to other users in the meantime until we have proper support for multiple package managers (and targets) on one ECU, but I don't know if it's worth the effort for you to rework the existing system to fit that mold.

patrickvacek:
> auto standalone = DockerAppManagerConfig(pconfig).docker_app_bin;
>     if (boost::filesystem::exists(standalone)) {

Looked right past that. Thanks!> That's a good idea. What I can do is have the test Target for each version have both types of app specs. eg:


> > Would it make sense to have a test to cover the upgrade scenario?
>
> That's a good idea. What I can do is have the test Target for each version have both types of app specs. eg:

Wow, that was easier that I expected. Cool.

I'm going to run the full CI on this. Let me know if there's anything else you want to add before we merge.

src/libaktualizr/package_manager/CMakeLists.txt
doanac: I've got a TODO for understanding secondaries more deeply. One thing we like about the current approach is that an update is a single immutable Target. However, it comes at a cost. The other thing I'd have to figure out is migration path(s) for existing devices.

PR #1558, opened 2020-02-14
Fix/todo cleanup

patrickvacek:
Pretty minor, just reviewing TODOs. Ironically I just had a discussion about cleaning up old lingering Jira tickets that we'll never get around to, and I just made a few new tickets for TODOs I didn't want to lose sight of.

### DIFF
--- a/src/aktualizr_secondary/CMakeLists.txt
+++ b/src/aktualizr_secondary/CMakeLists.txt
@@ -12,7 +12,7 @@ set(AKTUALIZR_SECONDARY_LIB_SRC
 # do not link tests with libaktualizr
 list(REMOVE_ITEM TEST_LIBS aktualizr_lib)
 
-# TODO: a lot of these should be removed
+# TODO(OTA-4544): a lot of these should be removed
 add_library(aktualizr_secondary_lib SHARED
     ${AKTUALIZR_SECONDARY_LIB_SRC}
     $<TARGET_OBJECTS:aktualizr-posix>
--- a/src/libaktualizr/storage/invstorage.h
+++ b/src/libaktualizr/storage/invstorage.h
@@ -87,7 +87,6 @@ class StorageTargetRHandle {
     file.close();
   }
 
-  // TODO: this function loads the whole image to the memory
   friend std::ostream& operator<<(std::ostream& os, StorageTargetRHandle& handle) {
     std::array<uint8_t, 256> arr{};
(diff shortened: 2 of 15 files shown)


src/aktualizr_secondary/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ set(AKTUALIZR_SECONDARY_LIB_SRC
 # do not link tests with libaktualizr
 list(REMOVE_ITEM TEST_LIBS aktualizr_lib)
 
-# TODO: a lot of these should be removed
+# TODO(OTA-4544): a lot of these should be removed
```
mike-sul: I am not sure if anything can be removed here, but still worth checking.

src/libaktualizr/storage/invstorage.h
```diff
@@ -87,7 +87,7 @@ class StorageTargetRHandle {
     file.close();
   }
 
-  // FIXME this function loads the whole image to the memory
+  // TODO: this function loads the whole image to the memory
```
lbonn: This comment doesn't look accurate actually, this function works buffers in blocks of 256 bytes. Or did we miss something (the comment is more recent than the code!)?

But I think @mike-sul mentioned that we still have places where this could happen, if that's the case we should also have a ticket for it and solve it quite soon.

src/libaktualizr/utilities/utils.cc
```diff
@@ -457,7 +457,7 @@ std::string Utils::randomUuid() {
   return boost::uuids::to_string(uuid_gen());
 }
 
-// FIXME Doesn't work with broken symlinks
+// TODO: Doesn't work with broken symlinks
```
lbonn: Maybe we do not need to have that as a `TODO` if we don't need it, but we definitely should document it in a comment somewhere.

src/libaktualizr/storage/invstorage.h
patrickvacek: Good point. I'm removing this TODO. If anyone finds somewhere else where the supposed problem is occurring, please leave a comment and/or make a ticket.

src/libaktualizr/utilities/utils.cc
patrickvacek: Good point. I'll remove the TODO and leave the comment. No need to make it sound so scary and important.

PR #1559, opened 2020-02-17
Store reports in the database

xcheng-here:
It's for OTA-3637. Store reports in database when events generated
and delete them after sent them to server.

### DIFF
--- a/src/libaktualizr/primary/reportqueue.cc
+++ b/src/libaktualizr/primary/reportqueue.cc
@@ -2,8 +2,9 @@
 
 #include <chrono>
 
-ReportQueue::ReportQueue(const Config& config_in, std::shared_ptr<HttpInterface> http_client)
-    : config(config_in), http(std::move(http_client)) {
+ReportQueue::ReportQueue(const Config& config_in, std::shared_ptr<HttpInterface> http_client,
+                         std::shared_ptr<INvStorage> storage_in)
+    : config(config_in), http(std::move(http_client)), storage(std::move(storage_in)) {
   thread_ = std::thread(std::bind(&ReportQueue::run, this));
 }
 
@@ -33,16 +34,15 @@ void ReportQueue::run() {
 void ReportQueue::enqueue(std::unique_ptr<ReportEvent> event) {
   {
     std::lock_guard<std::mutex> lock(m_);
-    report_queue_.push(std::move(event));
+    storage->saveReportEvent(event->toJson());
   }
   cv_.notify_all();
 }
 
 void ReportQueue::flushQueue() {
-  while (!report_queue_.empty()) {
-    report_array.append(report_queue_.front()->toJson());
-    report_queue_.pop();
-  }
+  int64_t max_id = 0;
+  Json::Value report_array{Json::arrayValue};
+  storage->loadReportEvents(&report_array, &max_id);
 
   if (config.tls.server.empty()) {
     // Prevent a lot of unnecessary garbage output in uptane vector tests.
@@ -61,6 +61,7 @@ void ReportQueue::flushQueue() {
 
     if (response.isOk() || response.http_status_code == 404) {
       report_array.clear();
+      storage->deleteReportEvents(max_id);
     }
   }
 }
--- a/src/libaktualizr/primary/reportqueue_test.cc
+++ b/src/libaktualizr/primary/reportqueue_test.cc
@@ -10,6 +10,8 @@
 #include "config/config.h"
 #include "httpfake.h"
 #include "reportqueue.h"
+#include "storage/invstorage.h"
+#include "storage/sqlstorage.h"
 #include "utilities/types.h"  // TimeStamp
 #include "utilities/utils.h"
 
@@ -51,6 +53,15 @@ class HttpFakeRq : public HttpFake {
         }
         return HttpResponse("", 200, CURLE_OK, "");
       }
+    } else if (url.find("reportqueue/StoreEvents") == 0) {
+      for (int i = 0; i < static_cast<int>(data.size()); ++i) {
+        EXPECT_EQ(data[i]["eventType"]["id"], "EcuDownloadCompleted");
+        EXPECT_EQ(data[i]["event"]["ecu"], "StoreEvents" + std::to_string(events_seen++));
+      }
+      if (events_seen == expected_events_) {
+        expected_events_received.set_value(true);
+      }
+      return HttpResponse("", 200, CURLE_OK, "");
     }
     LOG_ERROR << "Unexpected event: " << data;
     return HttpResponse("", 400, CURLE_OK, "");
@@ -70,7 +81,8 @@ TEST(ReportQueue, SingleEvent) {
 
   size_t num_events = 1;
   auto http = std::make_shared<HttpFakeRq>(temp_dir.Path(), num_events);
-  ReportQueue report_queue(config, http);
+  auto sql_storage = std::make_shared<SQLStorage>(config.storage, false);
+  ReportQueue report_queue(config, http, sql_storage);
 
   report_queue.enqueue(std_::make_unique<EcuDownloadCompletedReport>(Uptane::EcuSerial("SingleEvent"), "", true));
 
@@ -88,7 +100,8 @@ TEST(ReportQueue, MultipleEvents) {
 
   size_t num_events = 10;
   auto http = std::make_shared<HttpFakeRq>(temp_dir.Path(), num_events);
-  ReportQueue report_queue(config, http);
+  auto sql_storage = std::make_shared<SQLStorage>(config.storage, false);
+  ReportQueue report_queue(config, http, sql_storage);
 
   for (int i = 0; i < 10; ++i) {
     report_queue.enqueue(std_::make_unique<EcuDownloadCompletedReport>(
@@ -110,16 +123,54 @@ TEST(ReportQueue, FailureRecovery) {
 
   size_t num_events = 10;
   auto http = std::make_shared<HttpFakeRq>(temp_dir.Path(), num_events);
-  ReportQueue report_queue(config, http);
+  auto sql_storage = std::make_shared<SQLStorage>(config.storage, false);
+  ReportQueue report_queue(config, http, sql_storage);
 
-  for (int i = 0; i < 10; ++i) {
+  for (size_t i = 0; i < num_events; ++i) {
     report_queue.enqueue(std_::make_unique<EcuDownloadCompletedReport>(
         Uptane::EcuSerial("FailureRecovery" + std::to_string(i)), "", true));
   }
 
-  // Wait at most 30 seconds for the messages to get processed.
+  // Wait at most 20 seconds for the messages to get processed.
   http->expected_events_received.get_future().wait_for(std::chrono::seconds(20));
   EXPECT_EQ(http->events_seen, num_events);
+}
+
+/* Test persistent storage of unsent events in the database across
+ * ReportQueue instantiations. */
+TEST(ReportQueue, StoreEvents) {
+  TemporaryDirectory temp_dir;
+  Config config;
+  config.storage.path = temp_dir.Path();
+  config.tls.server = "";
+
+  auto sql_storage = std::make_shared<SQLStorage>(config.storage, false);
+  size_t num_events = 10;
+  auto check_sql = [sql_storage](size_t count) {
+    int64_t max_id = 0;
+    Json::Value report_array{Json::arrayValue};
+    sql_storage->loadReportEvents(&report_array, &max_id);
+    EXPECT_EQ(max_id, count);
+  };
+
+  {
+    auto http = std::make_shared<HttpFakeRq>(temp_dir.Path(), num_events);
+    ReportQueue report_queue(config, http, sql_storage);
+    for (size_t i = 0; i < num_events; ++i) {
+      report_queue.enqueue(std_::make_unique<EcuDownloadCompletedReport>(
+          Uptane::EcuSerial("StoreEvents" + std::to_string(i)), "", true));
+    }
+    check_sql(num_events);
+  }
+
+  config.tls.server = "reportqueue/StoreEvents";
+  auto http = std::make_shared<HttpFakeRq>(temp_dir.Path(), num_events);
+  ReportQueue report_queue(config, http, sql_storage);
+  // Wait at most 20 seconds for the messages to get processed.
+  http->expected_events_received.get_future().wait_for(std::chrono::seconds(20));
+  EXPECT_EQ(http->events_seen, num_events);
+  sleep(1);
+  check_sql(0);
 }
 
 #ifndef __NO_MAIN__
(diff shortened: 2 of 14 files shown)


src/libaktualizr/primary/reportqueue.cc
```diff
@@ -32,17 +33,15 @@ void ReportQueue::run() {
 
 void ReportQueue::enqueue(std::unique_ptr<ReportEvent> event) {
   {
-    std::lock_guard<std::mutex> lock(m_);
-    report_queue_.push(std::move(event));
+    //std::lock_guard<std::mutex> lock(m_);
```
patrickvacek: Is the lock really no longer needed here?

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1610,6 +1610,82 @@ bool SQLStorage::loadEcuReportCounter(std::vector<std::pair<Uptane::EcuSerial, i
   return true;
 }
 
+void SQLStorage::saveReportEvent(const Json::Value& json_value) {
+  std::string json_string = Utils::jsonToCanonicalStr(json_value);
+  SQLite3Guard db = dbConnection();
+  int64_t id = 0;
+  {
+    auto statement = db.prepareStatement("SELECT MAX(id) FROM report_events;");
```
lbonn: The two should be in a transaction.

Or better, it can be done in one step:

```
db.prepareStatement<std::string>("INSERT INTO report_events SELECT MAX(id) + 1, ? FROM report_events", json_string);
```

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1610,6 +1610,82 @@ bool SQLStorage::loadEcuReportCounter(std::vector<std::pair<Uptane::EcuSerial, i
...
+  if (statement.step() != SQLITE_DONE) {
+    LOG_ERROR << "Can't set ecu counter: " << db.errmsg();
+    return;
+  }
+}
+
+bool SQLStorage::loadReportEvents(Json::Value* report_array, int64_t* id_max) {
+  SQLite3Guard db = dbConnection();
+  std::vector<std::pair<Uptane::EcuSerial, int64_t>> ecu_cnt;
+  // keep the same order as in ecu_serials (start with primary)
```
lbonn: It looks like this comment was copy-pasted by mistake?

src/libaktualizr/primary/reportqueue.cc
```diff
@@ -32,17 +33,15 @@ void ReportQueue::run() {
...
   cv_.notify_all();
 }
 
 void ReportQueue::flushQueue() {
-  while (!report_queue_.empty()) {
-    report_array.append(report_queue_.front()->toJson());
-    report_queue_.pop();
-  }
+  int64_t max_id = 0;
+  storage->loadReportEvents(&report_array, &max_id);
```
lbonn: If you use load everything from storage here, you do not need to keep `report_array` as a class member because the state is in the database.

It looks like it can be a local variable in this function and calls to `report_array.clear()` would change to `storage->deleteReportEvents(...)`.

IMO, one of the important thing here is to try to minimize double-sent events. It looks like the backend should be able to discard duplicates by looking at the random uuid we send, though we should probably check with them.

src/libaktualizr/primary/reportqueue.cc
xcheng-here: Added it back.

src/libaktualizr/storage/sqlstorage.cc
xcheng-here: Thank you, modified.

src/libaktualizr/storage/sqlstorage.cc
xcheng-here: Yes, removed.

src/libaktualizr/primary/reportqueue.cc
xcheng-here: I modified the report_array. But I'm still thinking about how to prevent double-sent. Do we have a mechanism to check it with backend ? ( And does a check better than just send it ? )

xcheng-here:
@patrickvacek I didn't modify the branch name yet because I'm not sure whether it will make the comments lost.

src/libaktualizr/storage/sqlstorage.cc
lbonn: I don't see the changes yet?

patrickvacek:
> I didn't modify the branch name yet because I'm not sure whether it will make the comments lost.

No worries, looks like it's not possible to change it once the PR has already been made without making a new PR, which isn't worth it. Anyway, you put a description in the first comment of the PR, so that should help.

src/libaktualizr/storage/sqlstorage.cc
xcheng-here: I made a mistake in the last push. It should be updated now.

src/libaktualizr/primary/reportqueue.cc
lbonn: Checking before sending sounds complicated and not worth the overhead anyway. I'm trying to find out with the backend team if they can live with occasional resends.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -12,6 +12,8 @@
 #include "sql_utils.h"
 #include "utilities/utils.h"
 
+std::mutex SQLite3Guard::m_;
```
patrickvacek: I'm a bit confused here. This treats this mutex like a global instead of as a member variable. Is that really what we have to do? If so, then can you add a comment that explains why it has to be that way?

xcheng-here: If the mutex is a member of SQLite3Guard, it will be destroyed with the SQLite3Guard variable. And currently the way of our code using SQLite3Guard is that define it in the stack of functions which will access database, so it will be destroyed when the function returned. IMO, it means every function using a new mutex, so it will not really lock. Please correct me if I made something wrong.

patrickvacek: > currently the way of our code using SQLite3Guard is that define it in the stack of functions which will access database, so it will be destroyed when the function returned. IMO, it means every function using a new mutex, so it will not really lock.

Isn't that what we want? We should only lock the database when we are writing to it, and then immediately release. Or do we really want to hold on to the mutex for the duration of the Storage object? I don't really see what the point of that is if the Storage is basically a singleton.

Also, I just noticed that the `extern std::mutex sql_mutex` declared in sql_utils.h appears to be unused and an obsolete relic of previous rounds of locking logic.

patrickvacek: Sorry, just realized what you meant now. Yes, the mutex does have to live a higher level to be meaningful. But wouldn't a member variable of SqlStorage still be preferable to this?

xcheng-here: If it's a member of SqlStorage, I think I will have two options to implement the lock. The first one is that add a lock_guard to every function of SqlStorage, which will lock/unlock this mutex, I don't think it's more graceful than using SQLite3Guard, which is already used in every function. The second option is let the mutex be member of SqlStorage, but still use SQLite3Guard to lock it. I think it's also not graceful. So I choose to use a static member of SQLite3Guard.

lbonn: This was mentioned on Tuesday I think.
It's also my opinion that we'd better have the lock live in some instance instead of being static if we can. Obvious drawback is that we'd have separate instances of Storage locking each others out, even if they pertain to different files.

The SQLite3Guard can be associated to a storage-specific lock in `SQLStorageBase::dbConnection()`.

xcheng-here: Ah, SQLStorageBase::dbConnection() is a good place to do it for most cases, but there is still some problems in class SQLTargetWHandle. If I keep a pointer/reference of the mutex of Storage in this WHandle class, it will be a little weird (and maybe dangerous). Actually I'm a little confused why we need the class SQLTargetWHandle, IMHO, it's function might be combine into Storage.

xcheng-here: Oh, it's also possible to keep a pointer of storage in that WHandle class, I think it will be better than keep a pointer of mutex, I will try it.

xcheng-here: Another problem is that now we could init SQLite3Guard with the path of database, that means we didn't associate a Storage object to it. If we want the mutex be a member of Storage, maybe we'd better avoid init it in this way.

lbonn: Keeping a pointer to storage in WHandle seems fine.

And you want to pass a mutex in `SQLite3Guard`'s constructor, is that right?
As long as the other calls to the constructors are in existing tests which don't try to access the db in parallel with SQLStorage, that's maybe fine if the lock is made optional.

We can discuss again on Friday to clarify the question.

xcheng-here: Yes, I want to pass a mutex into SQLite3Guard's constructor, and use a shared_ptr of mutex to keep it. But when I start coding, I found I still feel not very good for this design, it still make a pointer to a member variable of another class.
And yes, I also think make the lock be optional will make the existing test code work.

xcheng-here: Looks it's not a problem to use a pointer of mutex SQLite3Guard, because we usually only use SQLite3Guard as local variable. And I haven't found another way which is graceful enough and will not need to restruct all of the storage code. So I think I will still use this solution.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1700,7 +1760,8 @@ boost::optional<std::pair<uintmax_t, std::string>> SQLStorage::checkTargetFile(c
 class SQLTargetWHandle : public StorageTargetWHandle {
  public:
   SQLTargetWHandle(const SQLStorage& storage, Uptane::Target target)
-      : db_(storage.dbPath()), target_(std::move(target)) {
+      : db_path_(storage.dbPath()), target_(std::move(target)) {
+    storage_ = &storage;
```
patrickvacek: `db_path` looks unused now, can it be removed? And I think `storage_` can be set in the initializer list.

xcheng-here: Moved storage_ to initializer list. The db_path_ is still using, it's mostly for existing test code.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1756,11 +1817,25 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
-    if (sqlite3_changes(db_.get()) > 0) {
-      auto statement =
-          db_.prepareStatement<std::string>("DELETE FROM target_images WHERE targetname=?;", target_.filename());
-      if (statement.step() != SQLITE_DONE) {
-        LOG_ERROR << "could not delete " << target_.filename() << " from sql storage";
+
+    if (storage_ != nullptr) {
+      SQLite3Guard db = storage_->dbConnection();
+      ;
+      if (sqlite3_changes(db.get()) > 0) {
```
lbonn: It looks like `sqlite3_changes` will always return 0 in this case, as the database connection has just been created: https://www.sqlite.org/c3ref/changes.html.

We have to find another way to track that here, maybe another attribute would be enough.

xcheng-here: Ah, yes, it's a good point. And I found the "insert/replace" is in one of the constructors of WHandle class, so in this case we could run "delete" anyway in abort() function. And in another constructor which is basically for test code, there is nothing to do with the table in database, so we don't need to run "delete" in this case. Please correct me if I made something wrong.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1756,11 +1817,15 @@ class SQLTargetWHandle : public StorageTargetWHandle {
...
     }
-    if (sqlite3_changes(db_.get()) > 0) {
-      auto statement =
-          db_.prepareStatement<std::string>("DELETE FROM target_images WHERE targetname=?;", target_.filename());
-      if (statement.step() != SQLITE_DONE) {
-        LOG_ERROR << "could not delete " << target_.filename() << " from sql storage";
+
+    if (storage_ != nullptr) {
+      SQLite3Guard db = storage_->dbConnection();
+      if (sqlite3_changes(db.get()) > 0) {
```
lbonn: There is another instance here.

xcheng-here: Aha, I'm so careless. Will modify it now.

lbonn:
Looks good to me, even better if we add a simple ReportQueue test that demonstrates that events are non volatile.

xcheng-here:
I think it means we need to restart aktualizr in the test, just after we store several events into database, but I haven't find a way which could do it and will not break the whole test process yet. And if we don't restart aktualizr, I think this change is transparent to caller of ReportQueue interfaces, because the details are hided in "enqueue" and "flush_queue".

patrickvacek:
> I think it means we need to restart aktualizr in the test, just after we store several events into database, but I haven't find a way which could do it and will not break the whole test process yet.

We have various ways of simulating reboots, but I don't think we even need to do that. In this case, can't we test this just be creating a ReportQueue directly, preventing it from sending, destroying it, then recreating it with the same storage database, and then verifying that it sends as expected then?

xcheng-here:
Yes, good idea. I added a test case.

src/libaktualizr/primary/reportqueue_test.cc
```diff
@@ -122,6 +136,32 @@ TEST(ReportQueue, FailureRecovery) {
...
+          Uptane::EcuSerial("StoreEvents" + std::to_string(i)), "", true));
+    }
+  }
+
+  config.tls.server = "reportqueue/StoreEvents";
+  auto http = std::make_shared<HttpFakeRq>(temp_dir.Path(), num_events);
+  ReportQueue report_queue(config, http, sql_storage);
+  // Wait at most 30 seconds for the messages to get processed.
+  http->expected_events_received.get_future().wait_for(std::chrono::seconds(20));
+  EXPECT_EQ(http->events_seen, num_events);
```
patrickvacek: This looks really good! My only concern is that we should check that there is actually something in the database before restarting the `ReportQueue`. We aren't testing anything interesting if it turns out that the first `ReportQueue` was actually able to send out all of the events before it got shut down. Basically I just want to make sure the test is testing what we think it is.

xcheng-here: Ah, because the server url which is used for the first ReportQueue is empty, it could not send any events out. But it's a good idea to check the events number in sql, I added it to the case, and I also checked whether the number was set to 0 after I send them out.

src/libaktualizr/primary/reportqueue_test.cc
```diff
@@ -110,16 +123,52 @@ TEST(ReportQueue, FailureRecovery) {
...
         Uptane::EcuSerial("FailureRecovery" + std::to_string(i)), "", true));
   }
 
-  // Wait at most 30 seconds for the messages to get processed.
+  // Wait at most 20 seconds for the messages to get processed.
+  http->expected_events_received.get_future().wait_for(std::chrono::seconds(20));
+  EXPECT_EQ(http->events_seen, num_events);
+}
+
+/* Test event store. */
```
patrickvacek: This test looks really good now! My last suggestion is to make the comment here more descriptive. Maybe something like "Test persistent storage of unsent events in the database across ReportQueue instantiations."

PR #1561, opened 2020-02-17
Minor fixes for tests

eu-smirnov:


eu-smirnov:
@eu-smirnov We dynamic_cast the event and access fields that aren't there

eu-smirnov:
@patrickvacek Same in the other commit, we access vector element that doesn't exist. It's better to use ASSERT for such cases.

lbonn:
I also agree with having ASSERTs in this case, it's usually fine to let the tests continue if they can but here it would just lead to undefined scenarios.
We could have a macro that does the assert and cast at once but maybe overkill.

PR #1562, opened 2020-02-17
Add configuration paramater for wait-for-secondary

lbonn:
As we want to track this value on the backend, it needs to show up in
our main configuration object which gets sent when aktualizr starts.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -917,6 +917,8 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
   // stuck in an unrecoverable state here
   if (!waitSecondariesReachable(updates)) {
     result.dev_report = {false, data::ResultCode::Numeric::kInternalError, ""};
+    storage->storeDeviceInstallationResult(result.dev_report, "Secondaries were not available", correlation_id);
+    sendEvent<event::AllInstallsComplete>(result);
     return result;
   }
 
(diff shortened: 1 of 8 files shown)


lbonn:
@patrickvacek yes I think we can write a test with a short timeout and unavailable ip secondary and check that it returns the error we expect.

lbonn:
@patrickvacek yes, I think this time we should wait for the test to merge actually. Sorry, should have marked it as not ready.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -873,116 +878,127 @@ result::UpdateStatus SotaUptaneClient::checkUpdatesOffline(const std::vector<Upt
...
-    } else {
-      result.dev_report = {false, data::ResultCode::Numeric::kInternalError, ""};
+  // put most of the logic in a lambda so that we can take care of common
+  // post-operations
+  result::Install r;
+  std::string raw_report;
+  bool drop_targets = false;
+
+  std::tie(r, raw_report, drop_targets) = [this, &updates,
+                                           &correlation_id]() -> std::tuple<result::Install, std::string, bool> {
```
patrickvacek: This is a great idea and should really make things easier to deal with. Took me a minute to follow it, though. I think part of it is that the lambda is rather large. Would it perhaps make it easier to just make it a proper function? Or would that be unnecessary? I realize most of this function *is* the lambda now, so maybe that wouldn't really help things...

lbonn: I tend to agree actually but my bad excuse is that I don't really know how to name this function, it's really doing the bulk of the work...

The `do { ... break; } while(0);` idiom was considered but a return is in a nested loop, so it would not work.

Another note is that `return std::make_tuple(...)` can be changed to `return {...}` from c++14 http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4387.html. The first version was written in this style but it tripped on xenial.

PR #1563, opened 2020-02-18
Docs/minor updates

patrickvacek:


lbonn:
I think the tests did not run on gitlab because the branch name starts with `docs/`

patrickvacek:
> I think the tests did not run on gitlab because the branch name starts with `docs/`

Correct. No need to run them here, right?

lbonn:
@patrickvacek no it's fine, but I was not sure it was intended?

patrickvacek:
> @patrickvacek no it's fine, but I was not sure it was intended?

In this case, yes.

PR #1564, opened 2020-02-18
Give better names to shared libraries

lbonn:
Use set_target_properties so that we don't confuse the name of library
and executable targets.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/libaktualizr-getstarted.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/libaktualizr-getstarted.adoc
@@ -48,7 +48,7 @@ make
 
 NOTE: The `--recursive` flag in the `git clone` command is needed to recursively clone all the *git submodules*.
 
-After the build is finished, you'll find the client library at the following location: `$\{PROJECT_ROOT}/build/src/libaktualizr/libaktualizr_lib.so`.
+After the build is finished, you'll find the client library at the following location: `$\{PROJECT_ROOT}/build/src/libaktualizr/libaktualizr.so`.
 
 == Integrating libaktualizr into your application
 
--- a/src/aktualizr_secondary/CMakeLists.txt
+++ b/src/aktualizr_secondary/CMakeLists.txt
@@ -28,6 +28,7 @@ add_library(aktualizr_secondary_lib SHARED
     $<TARGET_OBJECTS:logging>
     $<TARGET_OBJECTS:uptane>
     $<TARGET_OBJECTS:campaign>)
+set_target_properties(aktualizr_secondary_lib PROPERTIES LIBRARY_OUTPUT_NAME aktualizr_secondary)
 target_link_libraries(aktualizr_secondary_lib aktualizr-posix ${AKTUALIZR_EXTERNAL_LIBS})
 
 if (BUILD_ISOTP)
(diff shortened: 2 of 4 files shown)


lbonn:
Note: it will require an update on meta-updater

mike-sul:
It requires adjusting of this https://github.com/advancedtelematic/homebrew-otaconnect/blob/a4b13aec79ff2dd8896973a935155185d226eb66/aktualizr.rb#L34

lbonn:
> Does this change the linking at all, or is cmake smart enough to sort it out automatically?

What do you mean by "change the linking"?

We now have:

```
$ ninja -v aktualizr
...
[2/2] : && /usr/bin/c++ ...  src/libaktualizr/libaktualizr.so ...
```

So:

```
$ ldd src/aktualizr_primary/aktualizr
        linux-vdso.so.1 (0x00007fff5a10c000)
        libaktualizr.so => .../build/src/libaktualizr/libaktualizr.so (0x00007f7fbb5aa000)
```

lbonn:
@mike-sul 
> It requires adjusting of this [advancedtelematic/homebrew-otaconnect:aktualizr.rb@`a4b13ae`#L34](https://github.com/advancedtelematic/homebrew-otaconnect/blob/a4b13aec79ff2dd8896973a935155185d226eb66/aktualizr.rb#L34)

Can we use `make install` in these steps or does it bring some complications?

patrickvacek:
> What do you mean by "change the linking"?

You actually answered exactly what I was looking for in your comment ("Note: it will require an update on meta-updater") but I didn't see it somehow. :)

I assume https://github.com/advancedtelematic/libaktualizr-demo-app will also need an update.

mike-sul:
> @mike-sul
> 
> > It requires adjusting of this [advancedtelematic/homebrew-otaconnect:aktualizr.rb@`a4b13ae`#L34](https://github.com/advancedtelematic/homebrew-otaconnect/blob/a4b13aec79ff2dd8896973a935155185d226eb66/aktualizr.rb#L34)
> 
> Can we use `make install` in these steps or does it bring some complications?

Yes, I think it will be the correct way to do it in the formula and I'll try to do it. In any case feel free to merge this change since it won't impact the formula anyway since it builds & installs 2020.2 version.

lbonn:
Ok great, then I think it's close to be ready. Just one question remains IMO is whether we should also care about versions in .so names, like what is done on a typical distribution (ie: `libaktualizr.so.0`, `libaktualizr.so.2019.2`...).

As a system typically contains only one application (or two including aktualizr-info) linking with these .so, I would say it's probably fine in the current state?

mike-sul:
> Ok great, then I think it's close to be ready. Just one question remains IMO is whether we should also care about versions in .so names, like what is done on a typical distribution (ie: `libaktualizr.so.0`, `libaktualizr.so.2019.2`...).
> 
> As a system typically contains only one application (or two including aktualizr-info) linking with these .so, I would say it's probably fine in the current state?

Effectively, customers do not have a mean to update just aktualizr/libaktualizr, they are updated via updating an overall ostree-based rootfs, so the packaging unit is at an ostree revision which means that this library versioning does not bring much value. On a flip side, versioning of libaktualizr could clearly designate what version it is actually is.

BTW, perhaps it makes sense to consider updating of aktualizr without a need in bitbaking an overall rootfs for each device, so customers will have much simpler way to update it.

PR #1565, opened 2020-02-19
Specify branch name in ostree_pull

eu-smirnov:
This makes branchname visible in the target system.

### DIFF
--- a/src/libaktualizr/package_manager/ostreemanager.cc
+++ b/src/libaktualizr/package_manager/ostreemanager.cc
@@ -70,6 +70,8 @@ data::InstallationResult OstreeManager::pull(const boost::filesystem::path &sysr
                                              OstreeProgressCb progress_cb) {
   const std::string refhash = target.sha256Hash();
   const char *const commit_ids[] = {refhash.c_str()};
+  const std::string branchname = target.filename().substr(0, target.filename().rfind("-"));
+  const char *const branchnames[] = {branchname.c_str()};
   GError *error = nullptr;
   GVariantBuilder builder;
   GVariant *options;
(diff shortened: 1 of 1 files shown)


eu-smirnov:
I tested an update from version n to n+1 and then an update to version n-1. Seems to work as expected.

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -104,8 +106,8 @@ data::InstallationResult OstreeManager::pull(const boost::filesystem::path &sysr
 
   g_variant_builder_init(&builder, G_VARIANT_TYPE("a{sv}"));
   g_variant_builder_add(&builder, "{s@v}", "flags", g_variant_new_variant(g_variant_new_int32(0)));
-
-  g_variant_builder_add(&builder, "{s@v}", "refs", g_variant_new_variant(g_variant_new_strv(commit_ids, 1)));
+  g_variant_builder_add(&builder, "{s@v}", "refs", g_variant_new_variant(g_variant_new_strv(branchnames, 1)));
```
patrickvacek: What happens if `branchnames` is empty?

doanac:
I use offline credentials.zip files and when garage-push runs I get:
~~~
Credentials contain offline signing keys, not pushing root ref
~~~

So my treehub(s) don't ever have anything under `refs/heads/`. Will this change cause me problems?

I also was wondering about how you determine the branch name by rfind("-") on the target name. I assumed the format of target names was just convention. Is this format an actual hard requirement?

mike-sul:
> I assumed the format of target names was just convention. Is this format an actual hard requirement?

This is hard requirement imposed by OTA Connect backend while aktualizr/libaktualizr should be agnostic of it.

src/libaktualizr/package_manager/ostreemanager.cc
```diff
@@ -70,6 +70,8 @@ data::InstallationResult OstreeManager::pull(const boost::filesystem::path &sysr
                                              OstreeProgressCb progress_cb) {
   const std::string refhash = target.sha256Hash();
   const char *const commit_ids[] = {refhash.c_str()};
+  const std::string branchname = target.filename().substr(0, target.filename().rfind("-"));
+  const char *const branchnames[] = {branchname.c_str()};
```
doanac: rfind can return `string::npos`. Should we detect that and not try and set a branch if that's the case?

eu-smirnov:
Neither this PR nor the one for meta-updater isn't meant to be merged in the current state. I did this to quickly test if we can pass the OSTREE_BRANCHNAME to the target system.

The end goal here is to be able to reconstruct target name based on the data reported by OSTree, without writing any additional files to the filesystem or database.

Branchname looks like a good candidate for that, we could also probably use os_name parameter.

In any case, while I don't think the current scheme for target name is a hard requirement, we need to have some convention for naming targets to achieve this goal.

eu-smirnov:
I'll close it, for now, we can continue the discussion on other channels.

patrickvacek:
> So my treehub(s) don't ever have anything under `refs/heads/`. Will this change cause me problems?

BTW @doanac I'm looking into this. In general, it doesn't seem to matter, but I think the fact that we weren't pushing refs was an oversight.

PR #1567, opened 2020-02-25
OTA-4594: add a job to the CI to build aktualizr on OSX

mike-sul:
The least intrusive way to add the brew build job into the aktualizr CI pipeline.
The job is included in the pipeline only if executed as 'scheduled` at `docker` stage.


ci/gitlab/.gitlab-ci.yml
```diff
@@ -408,3 +408,18 @@ trigger-docsite-build:
...
+    - brew uninstall -f aktualizr-dev.rb
+  script:
+    - brew tap advancedtelematic/otaconnect
+    - brew install --HEAD -v aktualizr-dev
+    - brew link -f aktualizr-dev
+    - aktualizr --version
+  tags:
+    - osx
+  only:
+    - schedules
```
lbonn: We have another weekly schedule for Veracode scans, we probably don't need to run the osx-build in that case.

Using `only: variables: - $OSX_BUILD` and defining the variable in the schedule would solve that.

Also, it makes it easier to trigger the job manually without having to go through a schedule.

mike-sul: The scheduling mechanism allows triggering a job manually too.
Anyway, I agree with you, it makes sense

mike-sul:
BTW, if I trigger the pipeline from the schedules https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipeline_schedules that pipeline run/instance is attached to PR :)

patrickvacek:
Looks cool!

mike-sul:
@patrickvacek @lbonn Please, see this option of implementation - triggering an external project CI pipeline (https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/pipelines/402012)

In addition to triggering of the aktualizr' brewing job/pipeline I could also add triggering of the aktualizr release version brewing and adding it to the release artifacts if `github-release` is triggered

patrickvacek:
> Please, see this option of implementation - triggering an external project CI pipeline

Looks good!

> In addition to triggering of the aktualizr' brewing job/pipeline I could also add triggering of the aktualizr release version brewing and adding it to the release artifacts if `github-release` is triggered

That would be the actual replacement for the github action that we were using to do that, right? Seems like the right thing to do.

mike-sul:
> That would be the actual replacement for the github action that we were using to do that, right? Seems like the right thing to do.

Replacement or addition if we would like to support both mojave (the mac in the lab) and catalina (macOS virtual machine provided by github)?

patrickvacek:
> Replacement or addition if we would like to support both mojave (the mac in the lab) and catalina (macOS virtual machine provided by github)?

I think @tkfu suggested that it wasn't worth maintaining the github action, and I'm inclined to agree.

mike-sul:
> I think @tkfu suggested that it wasn't worth maintaining the github action, and I'm inclined to agree.

OK. Agreed.

patrickvacek:
> It's still itching a bit to see the step under the docker stage but we can think about another organization later.

What's your concern with that?

@mike-sul this has to wait until the homebrew recipe is merged so that you can update the branchname that's hardcoded here, right?

mike-sul:
> @mike-sul this has to wait until the homebrew PR is merged so that you can update the branchname that's hardcoded here, right?

Yes, that's correct + I would like to add also the job to build the release version as well as trying to improve the existing two formulae so we have just one instead of two.

mike-sul:
> It's still itching a bit to see the step under the docker stage but we can think about another organization later.

I don't like it too, but the reorganization of the CI stages will take some time and it's not primary goal of this specific task/PR (Frankly speaking I expected to see something like build, test, deploy, end-to-end test, etc)

lbonn:
> > It's still itching a bit to see the step under the docker stage but we can think about another organization later.
> 
> What's your concern with that?

It has nothing to do with docker, so it's confusing. We could add a comment in the meantime, or a new stage ("early-trigger"?) before "docker".

mike-sul:
> It has nothing to do with docker, so it's confusing. We could add a comment in the meantime, or a new stage ("early-trigger"?) before "docker".

or it can be part of `test` stage?

lbonn:
@mike-sul I thought the reason you put it in docker was to execute it as soon as possible. If it goes in "test", then why not "trigger" which fires up at the end?

mike-sul:
> @mike-sul I thought the reason you put it in docker was to execute it as soon as possible. If it goes in "test", then why not "trigger" which fires up at the end?

Yes, that was my intention to trigger it ASAP, but the rationale behind it just to see results ASAP while I am working on it. I open for suggestions regarding a stage at which the job should be triggered.

patrickvacek:
> It has nothing to do with docker, so it's confusing.

Oh, I see. I misunderstood the connection there. I agree now but don't have a strong feeling about where it ends up.

mike-sul:
I think this is ready for a merge, just made the 2020.3 OSX release with the CI job defined in this PR.
https://github.com/advancedtelematic/aktualizr/releases
https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/jobs/1776855

ci/gitlab/.gitlab-ci.yml
```diff
@@ -408,3 +408,44 @@ trigger-docsite-build:
...
+    - brew install ghr
+  script:
+    - brew tap advancedtelematic/otaconnect
+    - brew install -v --build-bottle aktualizr
+    - aktualizr --version
+    - brew bottle --json --no-rebuild --force-core-tap --root-url=${RELEASE_BASE_URL}/${VERSION} aktualizr
+    - brew bottle --merge --write --no-commit ./aktualizr--${VERSION}.mojave.bottle.json
+    - brew cat aktualizr | tee aktualizr.rb
+    - mv aktualizr--${VERSION}.mojave.bottle.tar.gz aktualizr-${VERSION}.mojave.bottle.tar.gz
+    - ls -alrt aktualizr-${VERSION}.mojave.bottle.tar.gz
```
patrickvacek: `ls`?

ci/gitlab/.gitlab-ci.yml
```diff
@@ -408,3 +408,44 @@ trigger-docsite-build:
...
     project: olp/edge/ota/documentation/ota-connect-docs
     branch: master
+
+trigger-osx-build:
+  stage: trigger
+  trigger:
+    project: olp/edge/ota/connect/client/homebrew-otaconnect
+    branch: master 
+  rules:
+    - if: $OSX_BUILD
```
patrickvacek: I didn't realize you could do this. Cool!

ci/gitlab/.gitlab-ci.yml
mike-sul: just for testing/debugging purposes - it does not cause any issue while might be helpful for troubleshooting if something goes wrong.

mike-sul:
@patrickvacek So, I am merging it?

PR #1570, opened 2020-02-26
Throw when trying to build a non-existing package manager

lbonn:
Also fix pacman factory tests when compiled without deb or ostree.

The old behaviour was different in the case of imaginary pacman names
and existing but not compiled pacman. Now, it throws an exception in any
case.

### DIFF
--- a/src/libaktualizr/package_manager/packagemanagerfactory_test.cc
+++ b/src/libaktualizr/package_manager/packagemanagerfactory_test.cc
@@ -11,7 +11,7 @@
 
 boost::filesystem::path sysroot;
 
-/* Support OSTree as a package manager. */
+#ifdef BUILD_OSTREE
 TEST(PackageManagerFactory, Ostree) {
   Config config;
   config.pacman.type = PACKAGE_MANAGER_OSTREE;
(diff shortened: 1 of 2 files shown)


src/libaktualizr/package_manager/packagemanagerfactory_test.cc
```diff
@@ -11,41 +11,32 @@
...
-#ifdef BUILD_OSTREE
   std::shared_ptr<PackageManagerInterface> pacman =
       PackageManagerFactory::makePackageManager(config.pacman, config.bootloader, storage, nullptr);
   EXPECT_TRUE(pacman);
-#else
-  EXPECT_THROW(std::shared_ptr<PackageManagerInterface> pacman =
-                   PackageManagerFactory::makePackageManager(config.pacman, config.bootloader, storage, nullptr),
-               std::runtime_error);
 #endif
 }
```
patrickvacek: `#endif` probably has to come after the closing brace.

lbonn: ! thanks

PR #1571, opened 2020-02-26
Fix for binary download resume

eu-smirnov:
This can be considered a hotfix for the issue with download resume after a nongraceful shutdown (OLPSUP-9401).
I plan to follow up this PR with another one, removing usage of target_images table for binary downloads and possibly moving this functionality to (Binary) package manager.

### DIFF
--- a/src/libaktualizr/package_manager/fetcher_death_test.cc
+++ b/src/libaktualizr/package_manager/fetcher_death_test.cc
@@ -16,8 +16,8 @@
 #include "test_utils.h"
 #include "uptane/tuf.h"
 
-static const int die_after = 50;       // percent
-static const int pause_duration = 20;  // seconds
+static const int die_after = 50;      // percent
+static const int pause_duration = 2;  // seconds
 
 std::string server;
 
@@ -56,7 +56,7 @@ void resume(const Uptane::Target& target) {
   EXPECT_TRUE(res);
 }
 
-void pause_and_die(const Uptane::Target& target) {
+void try_and_die(const Uptane::Target& target, bool graceful) {
   std::shared_ptr<INvStorage> storage(new SQLStorage(config.storage, false));
   auto http = std::make_shared<HttpClient>();
   api::FlowControlToken token;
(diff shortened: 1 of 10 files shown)


src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -9,7 +9,7 @@ struct DownloadMetaStruct {
         target{std::move(target_in)},
         token{token_in},
         progress_cb{std::move(progress_cb_in)} {}
-  uint64_t downloaded_length{0};
+  uintmax_t downloaded_length{0};
```
patrickvacek: I get why `size_t` was wrong, but is `uintmax_t` really necessary (compared to `uint64_t`)?

src/libaktualizr/package_manager/fetcher_death_test.cc
```diff
@@ -56,7 +56,7 @@ void resume(const Uptane::Target& target) {
   EXPECT_TRUE(res);
 }
 
-void pause_and_die(const Uptane::Target& target) {
+void try_and_die(const Uptane::Target& target, bool graceful) {
```
patrickvacek: Hilarious name. :)

src/libaktualizr/package_manager/packagemanagerinterface.cc
lbonn: I had the same question, it seems that the reason is that the c++ filesystem api (also boost) returns this type for file sizes: https://en.cppreference.com/w/cpp/filesystem/file_size.

eu-smirnov: It should be not necessary, just wanted to be consistent with return types from boost.

patrickvacek: Fair enough. Seems fine.

lbonn:
Looks fine to me, we can merge it after fixing the minor format nit.

eu-smirnov:
It's weird, but the order of commits on the branch doesn't match the order on github. Commit 864bcf0 goes before f4db090, so it's easier to test that the following commits are fixing the failing test. On github 864bcf0 is shown as the last commit.

lbonn:
Maybe this "feature" https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-in-the-wrong-order

PR #1574, opened 2020-02-26
Feat/prep 2020.3

patrickvacek:
~~I probably want to wait until https://github.com/advancedtelematic/aktualizr/pull/1571 is merged, but I at least got this started.~~


patrickvacek:
This is ready for review now, and it's docs-only, so CI is kinda irrelevant.

docs/ota-client-guide/modules/ROOT/pages/release-process.adoc
```diff
@@ -33,6 +33,13 @@ This is also a good time to review the docs in general and to consider whether a
 
 The docs published as https://docs.ota.here.com/ota-client/latest/index.html[latest] in the OTA Connect Developer Guide are built from the most recent release's docs branch (`\{version}-docs`). There will very likely be changes from there that have not been pulled into master yet. Open up a PR to merge the previous release's docs into master, resolving any merge conflicts as needed. Once that PR is merged, you can move on to the next step.
 
+The cleanest way to do this (especially if there were multiple changes to the docs branch) is to merge the docs branch locally and then rebase on master to remove the merge commits:
```
lbonn: Do we want to actually rebase in this case? I haven't followed the discussions on this workflow too closely but having different doc commits in two branches might cause problem.

@tkfu have you an opinion on this?

lbonn:
Looks good apart from the workflow question.

docs/ota-client-guide/modules/ROOT/pages/release-process.adoc
tkfu: I don't expect it to be a problem: 
* It's a relatively rare occurrence for us to update docs branches other than the latest release, and when we do, we're generally doing it with a cherry-pick anyway
* I think the procedure @patrickvacek is suggesting here is only for when you merge the latest docs branch into master _for immediate creation of a new release_, so I wouldn't really worry about different doc commits between a non-current docs branch and master; the new docs branch will have whatever commits you end up with after rebase, and those are the ones that you may end up merging back in later

I'm fine with it, and don't expect it to cause problems. As far as I can tell, though, the only actual reason to do a rebase is to make the commit history a little cleaner, so I don't really have any strong opinions here. If you'd prefer to leave in the merge commits that's cool too.

patrickvacek: Yeah, actually the ugliest part was that if you viewed the git history with `git log --graph`, you'd see these merge lines from previous releases. Not the worst, but it seemed unintuitive to me.

PR #1577, opened 2020-02-28
Fix some edge cases in the release scripts

lbonn:
* more retries when using the github api
* do not trip if the doc release has no work to do

### DIFF
--- a/scripts/publish_github_docs.sh
+++ b/scripts/publish_github_docs.sh
@@ -49,7 +49,11 @@ fi
 find . \( -regex './[^/]*' -and -type f -and -not -path ./.git \) -or \( -path './search/*' \) -exec git rm -r {} +
 cp -ar "$DOX_DOCS/." .
 git add .
-gitcommit -m "Update docs to latest ($DESCRIBE)"
+if git diff --cached --quiet; then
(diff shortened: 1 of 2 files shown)


scripts/publish_github_docs.sh
```diff
@@ -49,7 +49,11 @@ fi
 find . \( -regex './[^/]*' -and -type f -and -not -path ./.git \) -or \( -path './search/*' \) -exec git rm -r {} +
 cp -ar "$DOX_DOCS/." .
 git add .
-gitcommit -m "Update docs to latest ($DESCRIBE)"
+if git diff --cached --quiet; then
+    echo "Docs already updated to the latest version, skipping..."
+else
+    gitcommit -m "Update docs to latest ($DESCRIBE)"
+fi
```
patrickvacek: Wow, so the implication is that between the 2020.2 and 2020.3 releases we didn't change anything that affected the doxygen? Weird.

lbonn: Ah no, the reality is weirder. Somehow the doc publishing step was successful once (moved to 2020.3) and as the script was re-run, doxygen generated the same output twice, hence no changes.

PR #1578, opened 2020-02-28
Aktualizr secondary auto reboot

lbonn:
Tested manually on qemu.

### DIFF
--- a/src/aktualizr_secondary/main.cc
+++ b/src/aktualizr_secondary/main.cc
@@ -88,7 +88,14 @@ int main(int argc, char *argv[]) {
     LOG_DEBUG << "Current directory: " << boost::filesystem::current_path().string();
 
     auto secondary = AktualizrSecondaryFactory::create(config);
-    SecondaryTcpServer(*secondary, config.network.primary_ip, config.network.primary_port, config.network.port).run();
+    SecondaryTcpServer tcp_server(*secondary, config.network.primary_ip, config.network.primary_port,
+                                  config.network.port, config.uptane.force_install_completion);
+
+    tcp_server.run();
+
+    if (tcp_server.exit_reason() == SecondaryTcpServer::ExitReason::kRebootNeeded) {
+      secondary->completeInstall();
+    }
 
   } catch (std::runtime_error &exc) {
     LOG_ERROR << "Error: " << exc.what();
--- a/src/aktualizr_secondary/secondary_tcp_server.cc
+++ b/src/aktualizr_secondary/secondary_tcp_server.cc
@@ -10,7 +10,11 @@
 
 SecondaryTcpServer::SecondaryTcpServer(Uptane::SecondaryInterface &secondary, const std::string &primary_ip,
                                        in_port_t primary_port, in_port_t port)
-    : SecondaryTcpServer(secondary, port) {
+    : keep_running_(true), impl_(secondary), listen_socket_(port) {
+  if (primary_ip.empty()) {
+    return;
+  }
+
   ConnectionSocket conn_socket(primary_ip, primary_port, listen_socket_.port());
   if (conn_socket.connect() == 0) {
     LOG_INFO << "Connected to Primary, sending info about this secondary...";
--- a/src/aktualizr_secondary/secondary_tcp_server.cc
+++ b/src/aktualizr_secondary/secondary_tcp_server.cc
@@ -31,17 +31,18 @@ void SecondaryTcpServer::run() {
   LOG_INFO << "Secondary TCP server listens on " << listen_socket_.toString();
 
   while (keep_running_.load()) {
-    int con_fd;
     sockaddr_storage peer_sa{};
     socklen_t peer_sa_size = sizeof(sockaddr_storage);
 
     LOG_DEBUG << "Waiting for connection from client...";
-    if ((con_fd = accept(*listen_socket_, reinterpret_cast<sockaddr *>(&peer_sa), &peer_sa_size)) == -1) {
+    int con_fd = accept(*listen_socket_, reinterpret_cast<sockaddr *>(&peer_sa), &peer_sa_size);
+    if (con_fd == -1) {
       LOG_INFO << "Socket accept failed. aborting";
       break;
     }
+    Socket con_socket(con_fd);
     LOG_DEBUG << "Connected...";
-    bool continue_serving = HandleOneConnection(con_fd);
+    bool continue_serving = HandleOneConnection(*con_socket);
     LOG_DEBUG << "Client disconnected";
     if (!continue_serving) {
       break;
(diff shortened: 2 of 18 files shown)


src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -164,6 +175,12 @@ void SecondaryTcpServer::HandleOneConnection(int socket) {
     } else {
       LOG_DEBUG << "Not sending a response to message " << msg->present();
     }
+
```
mike-sul: I don't think that there is much reason in calling `stop` here as it's mainly intended for the server stooping from another thread. Perhaps just returning a bool/int from HandleOneConnection() and handling in the while loop would be slightly better (you could set keep_running_ = false; or just `break` directly there).
Also, the name `may_reboot` a bit confusing, it's rather `want reboot` or `require reboot`, just minor :)

src/aktualizr_secondary/main.cc
```diff
@@ -88,7 +88,14 @@ int main(int argc, char *argv[]) {
...
 
     auto secondary = AktualizrSecondaryFactory::create(config);
-    SecondaryTcpServer(*secondary, config.network.primary_ip, config.network.primary_port, config.network.port).run();
+    SecondaryTcpServer tcp_server(*secondary, config.network.primary_ip, config.network.primary_port,
+                                  config.network.port, config.uptane.force_install_completion);
+
+    tcp_server.run();
+
+    if (tcp_server.exit_reason() == SecondaryTcpServer::ExitReason::kRebootNeeded) {
+      secondary->completeInstall();
```
mike-sul: How can we guarantee here that the TCP/IP stack has sent all data from its internal buffers before the reboot is called?
It looks like there is missing connection shutdown in HandleOneConnection() and connection socket closing in run().

src/aktualizr_secondary/secondary_tcp_server.h
```diff
@@ -14,11 +14,14 @@ class SecondaryInterface;
  */
 class SecondaryTcpServer {
  public:
-  SecondaryTcpServer(Uptane::SecondaryInterface& secondary, const std::string& primary_ip, in_port_t primary_port,
-                     in_port_t port = 0);
+  enum class ExitReason {
+    kNA,
```
patrickvacek: "NA" for "not applicable"? Minor, but maybe worth spelling out.

lbonn:
> How can we test this?

There should be something to do with ipsecondary_test.py.

src/aktualizr_secondary/secondary_tcp_server.h
lbonn: I've also renamed Other to Unknown, more consistent with other enums we have.

src/aktualizr_secondary/main.cc
lbonn: I've added a RAII socket in `run()`.

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -157,15 +172,22 @@ void SecondaryTcpServer::HandleOneConnection(int socket) {
...
       }
       optval = 1;
       setsockopt(socket, IPPROTO_TCP, TCP_NODELAY, &optval, sizeof(int));
     } else {
       LOG_DEBUG << "Not sending a response to message " << msg->present();
     }
+
+    if (need_reboot && reboot_after_install_) {
+      exit_reason_ = ExitReason::kRebootNeeded;
+      return false;
```
mike-sul: As an option `exit_reason_` can be returned from HandleOneConnection to run() and run() in turn can return the exit status to a caller/client so no class member is required to store an exit reason and no additional call for a client to get this exit reason. But it' really minor and not important just an alternative.

lbonn:
Now comes with a test.

tests/ipsecondary_test.py
```diff
@@ -145,6 +145,40 @@ def test_secondary_ostree_update(uptane_repo, secondary, aktualizr, treehub, sys
     return True
 
 
+@with_treehub()
+@with_uptane_backend()
+@with_director()
+@with_sysroot()
+@with_secondary(start=False, output_logs=False, force_reboot=True)
+@with_aktualizr(start=False, run_mode='once', output_logs=True)
+def test_secondary_ostree_reboot(uptane_repo, secondary, aktualizr, treehub, sysroot, director, **kwargs):
```
patrickvacek: Do we have a negative test for this? (Same goes for the Primary, actually.) Less important, but just curious.

tests/ipsecondary_test.py
```diff
@@ -145,6 +145,40 @@ def test_secondary_ostree_update(uptane_repo, secondary, aktualizr, treehub, sys
...
+        logger.error("Pending version {} != the target one {}".format(pending_rev, target_rev))
+        return False
+
+    sysroot.update_revision(pending_rev)
+
+    with secondary:
+        with aktualizr:
+            aktualizr.wait_for_completion()
+
+    if not director.get_install_result():
```
mike-sul: I think it makes sense to add checking if what is currently installed on Secondary is what we expect
something like this `if target_rev != aktualizr.get_current_image_info(secondary.id):

tests/CMakeLists.txt
```diff
@@ -186,9 +186,12 @@ add_test(NAME test_log_negative
 set_tests_properties(test_log_negative
                      PROPERTIES PASS_REGULAR_EXPRESSION "Invalid log level")
 
+if (BUILD_OSTREE)
+    set(TEST_IPSEC_ARGS "--ostree")
+endif ()
 add_test(NAME test_ip_secondary
          COMMAND ${PROJECT_SOURCE_DIR}/tests/ipsecondary_test.py
-         --build-dir ${PROJECT_BINARY_DIR} --src-dir ${PROJECT_SOURCE_DIR} --ostree ${BUILD_OSTREE})
+         --build-dir ${PROJECT_BINARY_DIR} --src-dir ${PROJECT_SOURCE_DIR} ${TEST_IPSEC_ARGS})
```
mike-sul: What `${TEST_IPSEC_ARGS}` is for?

lbonn: It's here to set or not set `--ostree` as an argument to the script, as using `${BUILD_OSTREE}` in the python script would force us to reimplement CMake boolean logic:

> True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number. False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND. Named boolean constants are case-insensitive. If the argument is not one of these specific constants, it is treated as a variable or string and the following signature is used.
https://cmake.org/cmake/help/latest/command/if.html#command:if

I had the problem that the tests would not run because I configured my build directory with `-DBUILD_OSTREE=on` instead of `-DBUILD_OSTREE=ON`.

mike-sul: OK, I got it.

PR #1582, opened 2020-03-02
OTA-4601 Add the Additional environments subsection

tkfu:
In a separate section, describe that, instead of creating multiple
accounts, users can create additional env using only
one account. Add a beta icon to that feature.

Relates-to: OTA-4601

Note: this is being re-opened because the previous PR was merged prematurely. @houcros will merge this once the corresponding server feature is live.


houcros:
This PR only contains changes to the docs, so I assume it's safe to consider the Travis failure a false-positive.

patrickvacek:
> This PR only contains changes to the docs, so I assume it's safe to consider the Travis failure a false-positive.

Correct. We don't even bother running our primary CI (gitlab) on docs-only PRs, but we let Travis do its thing because we haven't bothered to tell it not to.

PR #1583, opened 2020-03-03
fix: disable trying fetch target in case of abort or pause

kbushgit:
Skip trying to fetch(download) target if the ```FlowControlToken``` state is set to ```pause``` or ```abort```.
So we try to fetch target against only in case if the ```PackageManagerInterface::fetchTarget(...)``` returns FALSE and the ```FlowControlToken STATE == running```.


patrickvacek:
Looks good except for CI.

patrickvacek:
Looks like the CI failure is this:
```
 /builds/olp/edge/ota/connect/client/aktualizr/src/libaktualizr/primary/sotauptaneclient.cc:651:23: error: implicit conversion 'const api::FlowControlToken *' -> bool [readability-implicit-bool-conversion,-warnings-as-errors]
      if (success || (token && !token->canContinue(false))) {
                      ^~~~~
                      ( != nullptr)
```

PR #1587, opened 2020-03-04
Add OSTree commands for pruning objects.

patrickvacek:


docs/ota-client-guide/modules/ROOT/pages/ostree-usage.adoc
```diff
@@ -42,3 +42,20 @@ The -R option is supported, for recursive file listing.
 
 Refs can be branch names or commit hashes, much like in git. Note that this does not show the contents of the diff; it just shows files added, deleted, and modified.
 
+== Pruning unused/unwanted objects
+
+First you will need to find which commits and refs are not currently deployed and no longer needed. You can use these commands to help make that determination:
```
ammaboateng:
```suggestion
First, you will need to find which commits and refs are currently not deployed and are no longer needed. You can use these commands to help make that determination:
```

patrickvacek:
@ammaboateng I've incorporated your suggestions and squashed the commit.

doanac:
This is orthogonal, but: Do any such commands exist that can be run in TreeHub? ie - I could give it a list of refs known to my targets.json and then basically it could clean up anything else not reachable

patrickvacek:
> This is orthogonal, but: Do any such commands exist that can be run in TreeHub? ie - I could give it a list of refs known to my targets.json and then basically it could clean up anything else not reachable

@simao would you have any insight here?

simao:
No, we don't have anything like this on the BE side.

patrickvacek:
> I don't quite understand the context for this, and I think it would be much better with a one- or two-sentence explanation of why you'd potentially want or need to do this.

We had a user ask if it was possible to garbage-collect old commits/objects after deploying a new update. The use case was if the previous commit had some sort of flaw and you knew that it would be unsafe to rollback to that commit. The question was if we could prevent that from happening (as an attack or accidentally).

> Normally there shouldn't be any junky old objects or commits lying around on a device's OSTree repo, because we garbage-collect any objects that aren't in the currently deployed image or the previous one, right?

That is my understanding, although I don't recall off the top of my head how the mechanism actually works.

> So is the use case for this just troubleshooting if your device's OSTree repo is all messed up somehow? Or would this be for cleaning up a bloated repo on a build machine?

Those weren't the intentions but I expect this could help with both of those use cases as well.

I can add some context to the doc about why this might be useful.

PR #1588, opened 2020-03-04
Fix available space check for 32-bit systems

eu-smirnov:
The issue was reported by a customer. This function:
https://github.com/advancedtelematic/aktualizr/blob/3c09519c97f8a40d9a84c62d02559dac18416c93/src/libaktualizr/storage/sqlstorage.cc#L1633
gives incorrect results on 32-bit systems.

The root cause is that both `f_bsize` and `f_bavail` members of statvfs struct are 4 bytes long. Hence, the result of this expression
https://github.com/advancedtelematic/aktualizr/blob/3c09519c97f8a40d9a84c62d02559dac18416c93/src/libaktualizr/storage/sqlstorage.cc#L1640
is also 4 byte long and overflows on 4GB.

This fix makes aktualizr use 64-bit filesystem interface. Member `f_bavail` becomes 8 bytes long and the expression evaluates correctly.
This has no effect on 64-bit architectures.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,10 @@ unset(AKTUALIZR_CHECKED_SRCS CACHE)
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+# To ensure better support of large files on 32-bit systems.
+# See https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html
+add_definitions("-D_FILE_OFFSET_BITS=64")
+
 if (CCACHE)
     find_program(CCACHE_PROGRAM ccache)
     if (CCACHE_PROGRAM)
--- a/src/libaktualizr/storage/sqlstorage.cc
+++ b/src/libaktualizr/storage/sqlstorage.cc
@@ -1637,7 +1637,7 @@ bool SQLStorage::checkAvailableDiskSpace(const uint64_t required_bytes) const {
     LOG_WARNING << "Unable to read filesystem statistics: error code " << stat_res;
     return true;
   }
-  const uint64_t available_bytes = (stvfsbuf.f_bsize * stvfsbuf.f_bavail);
+  const uint64_t available_bytes = (static_cast<uint64_t>(stvfsbuf.f_bsize) * stvfsbuf.f_bavail);
   const uint64_t reserved_bytes = 1 << 20;
 
   if (required_bytes + reserved_bytes < available_bytes) {


lbonn:
Defining the macro is probably a good idea for preventing future issues but I would also add some manual casts in the parts we know are problematic.

Here, casting `stvfsbuf.f_bsize` to uint64_t should not hurt.

eu-smirnov:
This has nothing to do with gcc, but it rather depends on C standard library implementation. So far I can see that it's implemented in the same way in glibc and ulibc.

kbushgit:
> This has nothing to do with gcc, but it rather depends on C standard library implementation. So far I can see that it's implemented in the same way in glibc and ulibc.

kbushgit:
> This fix makes aktualizr use 64-bit filesystem interface. Member `f_bavail` becomes 8 bytes long 

I've tried simple example and I do not see this transformation, or my test is incorrect? 

sudo apt-get install g++-multilib

```
#include <sys/statvfs.h>
#include <iostream>

int main() {
  struct statvfs stvfsbuf {};
  const int stat_res = statvfs("/", &stvfsbuf);

  const auto available_bytes = (stvfsbuf.f_bsize * stvfsbuf.f_bavail);

  std::cout << "sizeof stvfsbuf.f_bsize -> " << sizeof(stvfsbuf.f_bsize) << std::endl;
  std::cout << "sizeof stvfsbuf.f_bsize -> " << sizeof(stvfsbuf.f_bsize) << std::endl;
  std::cout << "sizeof avaiulable -> " << sizeof(available_bytes) << std::endl;

  return 0;
}
```

```
g++ -m32 -std=c++11 main.cc
```
result:
```
sizeof stvfsbuf.f_bsize -> 4
sizeof stvfsbuf.f_bsize -> 4
sizeof avaiulable -> 4
```
with flag:
```
g++ -m32 -D_FILE_OFFSET_BITS=64 -std=c++11 main.cc
```
result:
```
sizeof stvfsbuf.f_bsize -> 4
sizeof stvfsbuf.f_bsize -> 4
sizeof avaiulable -> 8
```

lbonn:
@kbushgit there is indeed an error in your program, it prints `sizeof(stvfsbuf.f_bsize)` twice.

After the fix:

```
$ ./statvfs
sizeof stvfsbuf.f_bsize -> 4
sizeof stvfsbuf.f_bavail -> 8
sizeof available -> 8
```

patrickvacek:
Can we take Kosta's test app and turn it into a proper unit test?

eu-smirnov:
> Defining the macro is probably a good idea for preventing future issues but I would also add some manual casts in the parts we know are problematic.
> 
> Here, casting `stvfsbuf.f_bsize` to uint64_t should not hurt.

I think you are right. I wanted to avoid explicit castings, but we still might end up with u32 * u32, which is wrong.

eu-smirnov:
> Can we take the idea behind Kosta's test app and turn it into a proper unit test?

A proper way to do it would be to have something similar to this : https://github.com/libarchive/libarchive/blob/master/build/cmake/CheckFileOffsetBits.cmake
But I don't think it's worth an effort, tbh.

PR #1589, opened 2020-03-05
ci: trigger the osx build job unconditionally

mike-sul:


mikesul-vd:
@patrickvacek Can we merge this so we can see an effect through the weekend?

PR #1592, opened 2020-03-06
Fix/renaming

patrickvacek:
The Uptane standard refers to an "Image repository" but never an "images repository". I've caught most trivial references and corrected them. I've also searched for other common examples of deviation from the naming norms of Uptane, mostly in terms of capitalization. This might seem trivial but should improve clarity when trying to read our code and follow the standard. For guidance on this subject, see: https://github.com/uptane/uptane-standard#style-guide

PR #1593, opened 2020-03-09
Fix/logging

patrickvacek:
Some seemingly minor changes that should make things a bit clearer in the logs. It might be nice to merge https://github.com/advancedtelematic/aktualizr/pull/1592 before this and then rebase if necessary (rather than the other way around).


doanac:
thanks for taking time to fix things like this!

patrickvacek:
> thanks for taking time to fix things like this!

You're welcome, and glad to hear that it resonates. There should be some improvements to the Secondary side coming soon as well, but I think that might not interest you so much. If there are other messages you see room to improve, please let us know or make a PR!

PR #1594, opened 2020-03-09
Only return hasPendingInstall() for primary

lbonn:
Because of this, primary was rebooting needlessly.

Bug found here: https://github.com/advancedtelematic/meta-updater/pull/696

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -1033,8 +1033,14 @@ void SotaUptaneClient::campaignPostpone(const std::string &campaign_id) {
 }
 
 bool SotaUptaneClient::isInstallCompletionRequired() const {
-  bool force_install_completion = (hasPendingUpdates() && config.uptane.force_install_completion);
-  return force_install_completion;
+  std::vector<std::pair<Uptane::EcuSerial, Uptane::Hash>> pending_ecus;
+  storage->getPendingEcus(&pending_ecus);
+  bool pending_for_ecu = std::find_if(pending_ecus.begin(), pending_ecus.end(),
+                                      [this](const std::pair<Uptane::EcuSerial, Uptane::Hash> &ecu) -> bool {
+                                        return ecu.first == primary_ecu_serial_;
+                                      }) != pending_ecus.end();
+
+  return pending_for_ecu && config.uptane.force_install_completion;
 }
 
 void SotaUptaneClient::completeInstall() const {
--- a/src/libaktualizr/storage/storage_common_test.cc
+++ b/src/libaktualizr/storage/storage_common_test.cc
@@ -321,6 +321,7 @@ TEST(storage, load_store_installed_versions) {
   {
     boost::optional<Uptane::Target> current;
     EXPECT_TRUE(storage->loadInstalledVersions("primary", &current, nullptr));
+    EXPECT_FALSE(storage->hasPendingInstall());
     EXPECT_TRUE(!!current);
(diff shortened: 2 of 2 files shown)


src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1397,7 +1397,10 @@ bool SQLStorage::loadInstalledVersions(const std::string& ecu_serial, boost::opt
 bool SQLStorage::hasPendingInstall() {
   SQLite3Guard db = dbConnection();
 
-  auto statement = db.prepareStatement("SELECT count(*) FROM installed_versions where is_pending = 1");
+  auto statement = db.prepareStatement(
```
mikesul-vd: The business logic in SotaUptaneClient::fetchMeta() expects that hasPendingInstall/hasPendingUpdates returns true if there is at least one pending update regardless if it's Primary or Secondary.

lbonn:
@patrickvacek actually `is_primary` is 1 in storage for secondary ecu!

But Mike's remark is indeed valid, this will need more work.

mike-sul:
> @patrickvacek actually `is_primary` is 1 in storage for secondary ecu!
> 
> But Mike's remark is indeed valid, this will need more work.

Yeah, this change won't impact Secondary as Secondary is treated as a primary ECU in Secondary's DB :), but it will impact Primary. I think the usage of `storage->getPendingEcus(&pending_ecus); and  `if (primaryEcuSerial() == pending_ecu.first)` can be helpful here.

patrickvacek:
> @patrickvacek actually `is_primary` is 1 in storage for secondary ecu!

Ha! Okay, sorta makes sense, but that deserves a comment somewhere if it's not already there.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1394,10 +1394,21 @@ bool SQLStorage::loadInstalledVersions(const std::string& ecu_serial, boost::opt
   return true;
 }
 
-bool SQLStorage::hasPendingInstall() {
+bool SQLStorage::hasPendingInstall(bool ecu_only) {
```
mike-sul: if an ECU serial is a parameter of this method with a default value EcuSerial::Unknown() or an empty string then the sql query can be simpler SELECT count(*) FROM installed_versions where is_pending = 1 AND serial = %serial%

mike-sul:
IMHO, there is a simpler and less intrusive option to fix it - just add third condition here https://github.com/advancedtelematic/aktualizr/blob/b06f2c10c8d77b4cf492297479f2eb96aa6962d4/src/libaktualizr/primary/sotauptaneclient.cc#L1036 (e.g. storage->getPendingEcus(&pending_ecus); if (primaryEcuSerial() == pending_ecu.first)` this logix is used in a few other places so can be encapsulated in a helper function)
or hasPendingInstall() accept an ecu serial as parameter

lbonn:
> IMHO, there is a simpler and less intrusive option to fix it - just add third condition here
> 
> https://github.com/advancedtelematic/aktualizr/blob/b06f2c10c8d77b4cf492297479f2eb96aa6962d4/src/libaktualizr/primary/sotauptaneclient.cc#L1036
> (e.g. storage->getPendingEcus(&pending_ecus); if (primaryEcuSerial() == pending_ecu.first)` this logix is used in a few other places so can be encapsulated in a helper function)
> or hasPendingInstall() accept an ecu serial as parameter

You're probably right, I'll do something in this spirit.

lbonn:
@patrickvacek thanks!

No that't it right now, I've tried manually on qemu.

PR #1595, opened 2020-03-10
Fix/more capitalization

patrickvacek:
The fun just doesn't stop.


patrickvacek:
> Looks like you are enjoying it :)

Yes and no, it's a bit like an obsessive compulsion. Speaking of which, just found another!

PR #1596, opened 2020-03-11
Fix/misc init

patrickvacek:
Very minor improvements found while debugging/investigating other topics.


src/aktualizr_info/main.cc
```diff
@@ -96,6 +96,7 @@ int main(int argc, char **argv) {
       exit(EXIT_FAILURE);
     }
 
+    logger_init();
```
mike-sul: good catch!

PR #1597, opened 2020-03-12
GitHub Actions for CI

lbonn:


lbonn:
Caching should already work (after a merge for master for docker).

Right now, codecov has been scraped, I'll see if it's easy to restore.

lbonn:
@patrickvacek 

> What is the goal of the docker workflow? Are we not already updating those images in gitlab?

In gitlab CI, we store the images in gitlab's docker repository which is only accessible from the company network.
Here, we have to store them somewhere else and github conveniently has a space for that: https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-docker-for-use-with-github-packages
 
> Are these two log messages particularly concerning?
> 
> `WARNING! Your password will be stored unencrypted in /home/runner/.docker/config.json.`

No, not really that's the recommended way (see link above).

More info on the token here: https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token.

I suspect that actions triggered from external forks will actually won't be able to use this token and won't benefit from caching. That should probably be tested as well.

Tagging @zabbal, just in case :)

> 
> And:
> 
> `Error response from daemon: repository docker.pkg.github.com/advancedtelematic/aktualizr/aktualizr-ci not found: name unknown: docker package "aktualizr-ci" does not exist under owner "advancedtelematic/aktualizr"`

Right now, no image has been pushed but this should go away as soon as we merge this PR to master.

> 
> It's also weird that I don't see the output of the tests in the coverage run. I guess if it succeeds, it's fine.

I've also noticed that it sometimes gives out logs very slowly or not at all. But the "View raw logs" from the menu in the corner has worked for me so far.

> 
> Actually uploading coverage would be good, but can't we move that to gitlab as well? Might be better.

I'll look into it today :).

zabbal:
Yepp, looks fine to me. Github have access to their token anyway so when it's stored in their CI runner it makes no difference from security PoV

Out of curiosity, what's the advantage of gh-actions compared to travis?

patrickvacek:
> I've also noticed that it sometimes gives out logs very slowly or not at all. But the "View raw logs" from the menu in the corner has worked for me so far.

Thanks for that tip!

And everything else looks/sounds fine.

lbonn:
> I suspect that actions triggered from external forks will actually won't be able to use this token and won't benefit from caching. That should probably be tested as well.

I was wrong actually (https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow):

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

But the security is here: https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token

However, this also means that codecov won't work for forked PRs if I add it here, as it uses a codecov token as a secret.

@zabbal thanks for the quick look. Travis and GitHub Actions are both free for public repositories. The advantage is that GitHub actions seems much more supported right now than Travis which looks like it's maintained on life support. It's been slower over time and it has the hard time out limit of 50 minutes that has caused problems in the past.

lbonn:
test_imagerepo_failure just failed three times in a  row...

mike-sul:
> test_imagerepo_failure just failed three times in a row...

Maybe, it indicates that there is an issue(s) with handling edge cases of Images repo in aktualizr...
I'll look into it on Monday.

mike-sul:
Well the following message speaks for itself
`ERROR:test_fixtures:None
ERROR:test_fixtures:Failed to get aktualizr info/status`

lbonn:
> Well, the following message speaks for itself
> `ERROR:test_fixtures:None ERROR:test_fixtures:Failed to get aktualizr info/status`

Does it? What do you mean?

lbonn:
>     1. What changed since the last round of errors?

The bad kind of "fix", pressed retry :). I've looked a little bit in the test but I'm not sure what could make it more reliable.

>     2. Is it better to have the codecov upload here or in gitlab? I'm fine either way, so there's no problem with this as it is.

Opinion:

gitlab:
\+ it is our canonical reference CI, more logical to put it here
\- won't run for external PRs
\- already stores the html coverage report in html format, though it's quite impractical to use (a zip archive)

github:
\+ already done haha
\- it's our backup CI
\- won't run for external PRs in the current state (because it uses `GITHUB_TOKEN`)
\~ it looks like it could work for external contributions using the official action https://github.com/marketplace/actions/codecov. I've considered it but I'm not sure all the options we use are accessible this way right now

But switching it to gitlab should be easy in the current state anyway, just a matter of adding the token there.

lbonn:
I forgot a `|| true` in the docker caching steps, so they are blocked right now. I'll push something manually which should fix it, then open a new small PR.

PR #1598, opened 2020-03-12
fix: configurable number of attempts to connect to secondary

mike-sul:
- parameterize a number of connection attempts to secondary
- try to connect to secondary for three times during manifest fetching
- try to connect to secondary for four times at provisioning state

### DIFF
--- a/src/aktualizr_secondary/secondary_tcp_server.cc
+++ b/src/aktualizr_secondary/secondary_tcp_server.cc
@@ -20,7 +20,7 @@ SecondaryTcpServer::SecondaryTcpServer(Uptane::SecondaryInterface &secondary, co
     LOG_INFO << "Connected to Primary, sending info about this secondary...";
     HandleOneConnection(*conn_socket);
   } else {
-    LOG_INFO << "Failed to connect to Primary";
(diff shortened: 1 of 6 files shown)


mike-sul:
I've just realized that this is a useless and questionable workaround.

patrickvacek:
> I've just realized that this is a useless and questionable workaround.

That sounds a little extreme, but if https://github.com/advancedtelematic/meta-updater/pull/697 works as well as it sounds, then let's just do that instead of this!

mike-sul:
> > I've just realized that this is a useless and questionable workaround.
> 
> That sounds a little extreme, but if [advancedtelematic/meta-updater#697](https://github.com/advancedtelematic/meta-updater/pull/697) works as well as it sounds, then let's just do that instead of this!

Actually, I don't think that [advancedtelematic/meta-updater#697] will fix the ptest issue, I think, the test itself should be slightly restructured to emulate the real-life use case properly, the use-case when Secondary's ostree is updated and Primary can spin through a few uptane cycles before Secondary actually is rebooted and reports the applied version to Primary. Also, there is a bug in the code - aktualizr don't check if getManifest() RPC call was actually successful and goes straight to verification of the manifest signature.

PR #1600, opened 2020-03-13
Feat/ota 4118/director targets sanity

patrickvacek:
Depends on https://github.com/advancedtelematic/tuf-test-vectors/pull/69.

### DIFF
--- a/src/libaktualizr/uptane/directorrepository.cc
+++ b/src/libaktualizr/uptane/directorrepository.cc
@@ -8,7 +8,13 @@ void DirectorRepository::resetMeta() {
   latest_targets = Targets();
 }
 
-bool DirectorRepository::targetsExpired() const { return latest_targets.isExpired(TimeStamp::Now()); }
+bool DirectorRepository::targetsExpired() {
+  if (latest_targets.isExpired(TimeStamp::Now())) {
+    last_exception = Uptane::ExpiredMetadata(type.toString(), Role::Targets().ToString());
+    return true;
+  }
+  return false;
+}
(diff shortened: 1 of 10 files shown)


src/libaktualizr/uptane/directorrepository.cc
```diff
@@ -8,7 +8,37 @@ void DirectorRepository::resetMeta() {
...
+  for (auto target_it = targets.targets.begin(); target_it != targets.targets.end(); ++target_it) {
+    for (auto ecu_it = target_it->ecus().begin(); ecu_it != target_it->ecus().end(); ++ecu_it) {
+      if (ecu_ids.find(ecu_it->first) == ecu_ids.end()) {
+        ecu_ids.insert(ecu_it->first);
+      } else {
+        last_exception = Uptane::InvalidMetadata(type.toString(), Role::Targets().ToString(), "Found repeated ECU ID.");
+        return false;
+      }
+    }
+  }
```
lbonn:
```suggestion
  for (const auto &target: targets.targets) {
    for (const auto &ecu: target.ecus()) {
      if (ecu_ids.find(ecu.first) == ecu_ids.end()) {
        ecu_ids.insert(ecu.first);
      } else {
        last_exception = Uptane::InvalidMetadata(type.toString(), Role::Targets().ToString(), "Found repeated ECU ID.");
        return false;
      }
    }
  }
```

patrickvacek: Incorporated and rebased.

PR #1602, opened 2020-03-15
libaktualizr: dockerapp: dont prune label aktualizr-no-prune

mike-scott:
Let's create a way of preserving images which are not currently
being used by marking them with the following label:
aktualizr-no-prune=1

An example use-case for this would be keeping around an image used
for manual debugging (it isn't always running).

Signed-off-by: Michael Scott <mike@foundries.io>


mike-scott:
@doanac please take a look

PR #1603, opened 2020-03-16
Move the default branch to zeus.

patrickvacek:
Thud in poky has been moved to community support only. We are still supporting it in our repos for now, but new users should use a newer release branch.

We should consider putting the branchname in a common place as a variable as we've done for some other things.


Raigi:
Hmm.. I see the change for raspberrypi, is qemu is under same change?

patrickvacek:
> Hmm.. I see the change for raspberrypi, is qemu is under same change?

Yep, the qemu page reuses most of the same content from the rpi source.

Raigi:
> > Hmm.. I see the change for raspberrypi, is qemu is under same change?
> 
> Yep, the qemu page reuses most of the same content from the rpi source.

:+1:

PR #1606, opened 2020-03-17
Some clang-tidy fixes

lbonn:
Some general lint fixes.

Haven't bumped clang-tidy yet because I have been dealing with a lot of false positives. This should at least improve the situation in the meantime.

### DIFF
--- a/src/libaktualizr/utilities/dequeue_buffer.cc
+++ b/src/libaktualizr/utilities/dequeue_buffer.cc
@@ -8,7 +8,7 @@ char* DequeueBuffer::Head() {
   return buffer_.data();
 }
 
-size_t DequeueBuffer::Size() {
+size_t DequeueBuffer::Size() const {
   assert(sentinel_ == kSentinel);
   return written_bytes_;
 }
--- a/src/libaktualizr/utilities/dequeue_buffer.cc
+++ b/src/libaktualizr/utilities/dequeue_buffer.cc
@@ -23,8 +23,8 @@ void DequeueBuffer::Consume(size_t bytes) {
     throw std::logic_error("Attempt to DequeueBuffer::Consume() more bytes than are valid");
   }
   // Shuffle up the buffer
-  auto next_unconsumed_byte = buffer_.begin() + bytes;
-  auto end_of_written_area = buffer_.begin() + written_bytes_;
+  auto* next_unconsumed_byte = buffer_.begin() + bytes;
+  auto* end_of_written_area = buffer_.begin() + written_bytes_;
   std::copy(next_unconsumed_byte, end_of_written_area, buffer_.begin());
   written_bytes_ -= bytes;
 }
--- a/src/libaktualizr/utilities/sig_handler.cc
+++ b/src/libaktualizr/utilities/sig_handler.cc
@@ -12,14 +12,17 @@ SigHandler& SigHandler::get() {
 }
 
 SigHandler::~SigHandler() {
-  {
-    std::lock_guard<std::mutex> g(exit_m_);
-    exit_flag_ = true;
-  }
-  exit_cv_.notify_all();
+  try {
+    {
+      std::lock_guard<std::mutex> g(exit_m_);
+      exit_flag_ = true;
+    }
+    exit_cv_.notify_all();
 
-  if (polling_thread_.joinable()) {
-    polling_thread_.join();
+    if (polling_thread_.joinable()) {
+      polling_thread_.join();
+    }
+  } catch (...) {
   }
 }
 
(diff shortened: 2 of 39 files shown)


src/libaktualizr/utilities/utils.h
```diff
@@ -191,12 +191,12 @@ struct _Unique_if {
...
 template <class T>
-struct _Unique_if<T[]> {
-  using _Unknown_bound = std::unique_ptr<T[]>;
+struct _Unique_if<T[]> {                        // NOLINT
+  using _Unknown_bound = std::unique_ptr<T[]>;  // NOLINT
 };
 
 template <class T, size_t N>
-struct _Unique_if<T[N]> {
+struct _Unique_if<T[N]> {  // NOLINT
```
patrickvacek: Can you specify exactly what is being NOLINTed here?

src/libaktualizr/utilities/dequeue_buffer.cc
```diff
@@ -23,8 +23,8 @@ void DequeueBuffer::Consume(size_t bytes) {
     throw std::logic_error("Attempt to DequeueBuffer::Consume() more bytes than are valid");
   }
   // Shuffle up the buffer
-  auto next_unconsumed_byte = buffer_.begin() + bytes;
-  auto end_of_written_area = buffer_.begin() + written_bytes_;
+  auto* next_unconsumed_byte = buffer_.begin() + bytes;
+  auto* end_of_written_area = buffer_.begin() + written_bytes_;
```
patrickvacek: Is this to prevent the variables from becoming arrays or to prevent them from being the wrong kind of const?

src/libaktualizr/utilities/utils.cc
```diff
@@ -52,134 +53,134 @@ const char *adverbs[] = {
...
-                       "Tilsit",
-                       "Toastbrot",
-                       "Topfenstrudel",
-                       "Vollkornbrot",
-                       "Wibele",
-                       "Wickelkloesse",
-                       "Zimtwaffeln",
-                       "Zwetschkenroester",
-                       "Zwiebelkuchen"};
+static const std::array<const char *, 128> names = {"Allerlei",
```
patrickvacek: Really annoying that you have to specify the length. Is this another reasonable candidate for a NOLINT?

src/libaktualizr/utilities/sig_handler.cc
```diff
@@ -12,14 +12,17 @@ SigHandler& SigHandler::get() {
...
+      exit_flag_ = true;
+    }
+    exit_cv_.notify_all();
 
-  if (polling_thread_.joinable()) {
-    polling_thread_.join();
+    if (polling_thread_.joinable()) {
+      polling_thread_.join();
+    }
+  } catch (...) {
```
patrickvacek: What necessitated this try-catch?

lbonn: Exceptions should not be thrown in destructors https://clang.llvm.org/extra/clang-tidy/checks/bugprone-exception-escape.html.

src/libaktualizr/utilities/dequeue_buffer.cc
lbonn: It's only for readability: https://clang.llvm.org/extra/clang-tidy/checks/readability-qualified-auto.html

src/libaktualizr/utilities/utils.cc
lbonn: Yes kinda annoying but I don't think we'll change these very often so it shouldn't be a big deal. And just calling `.size()` rather than the sizeof dance is already beneficial.

patrickvacek:
I'm satisfied except for those unrestrained NOLINTs. I think any new NOLINTS should specify what it is they are ignoring.

src/libaktualizr/utilities/utils.h
lbonn: updated

lbonn:
I've added a couple more fixes, now that I've re-run clang-tidy 10.

src/libaktualizr/utilities/utils.cc
```diff
@@ -798,7 +799,7 @@ TemporaryFile::TemporaryFile(const std::string &hint)
 
 TemporaryFile::~TemporaryFile() { boost::filesystem::remove(tmp_name_); }
 
-void TemporaryFile::PutContents(const std::string &contents) {
+void TemporaryFile::PutContents(const std::string &contents) const {
```
patrickvacek: There's something unintuitive about a function that makes system calls (`write` no less!) being `const`, but I won't complain about more `const`s!

lbonn:
rebased

PR #1607, opened 2020-03-18
Improve logging in aktualizr-secondary

xcheng-here:
Changed log in aktualizr-secondary and aktualizr-info

In aktualizr-secondary, print manifest, meta, and status like download success/fail,
installation success/fail.
Added a parameter "--secondary" for aktualizr-info to make the log for secondary
database looks better.

### DIFF
--- a/src/aktualizr_info/main.cc
+++ b/src/aktualizr_info/main.cc
@@ -103,6 +103,8 @@ int main(int argc, char **argv) {
 
     AktualizrInfoConfig config(vm);
 
+    bool secondary_db = false;
+
     bool readonly = true;
     if (vm.count("allow-migrate") != 0u) {
       readonly = false;
@@ -137,9 +139,9 @@ int main(int argc, char **argv) {
       storage = INvStorage::newStorage(config.storage, readonly);
     }
 
-    if (!storage->loadDeviceId(&device_id)) {
-      std::cout << "Couldn't load device ID" << std::endl;
-    } else {
+    bool deviceid_loaded = false;
+    if (storage->loadDeviceId(&device_id)) {
+      deviceid_loaded = true;
       // Early return if only printing device ID.
       if (vm.count("name-only") != 0u) {
         std::cout << device_id << std::endl;
@@ -150,17 +152,23 @@ int main(int argc, char **argv) {
     registered = registered || storage->loadEcuRegistered();
     has_metadata = has_metadata || storage->loadLatestRoot(&director_root, Uptane::RepositoryType::Director());
 
-    // TLS credentials
-    if (vm.count("tls-creds") != 0u) {
+    bool tlscred_loaded = false;
+    {
       std::string ca;
       std::string cert;
       std::string pkey;
 
       storage->loadTlsCreds(&ca, &cert, &pkey);
-      std::cout << "Root CA certificate:" << std::endl << ca << std::endl;
-      std::cout << "Client certificate:" << std::endl << cert << std::endl;
-      std::cout << "Client private key:" << std::endl << pkey << std::endl;
-      cmd_trigger = true;
+      if (!ca.empty() || !cert.empty() || !pkey.empty()) {
+        tlscred_loaded = true;
+      }
+      // TLS credentials
+      if (vm.count("tls-creds") != 0u) {
+        std::cout << "Root CA certificate:" << std::endl << ca << std::endl;
+        std::cout << "Client certificate:" << std::endl << cert << std::endl;
+        std::cout << "Client private key:" << std::endl << pkey << std::endl;
+        cmd_trigger = true;
+      }
     }
 
     if (vm.count("tls-root-ca") != 0u) {
@@ -185,11 +193,15 @@ int main(int argc, char **argv) {
     }
 
     // ECU credentials
-    if (vm.count("ecu-keys") != 0u) {
-      std::string priv;
-      std::string pub;
+    bool ecukeys_loaded = false;
+    std::string priv;
+    std::string pub;
 
-      storage->loadPrimaryKeys(&pub, &priv);
+    storage->loadPrimaryKeys(&pub, &priv);
+    if (!pub.empty() && !priv.empty()) {
+      ecukeys_loaded = true;
+    }
+    if (vm.count("ecu-keys") != 0u) {
       std::cout << "Public key:" << std::endl << pub << std::endl;
       std::cout << "Private key:" << std::endl << priv << std::endl;
       cmd_trigger = true;
@@ -289,16 +301,28 @@ int main(int argc, char **argv) {
       return EXIT_SUCCESS;
     }
 
+    if (!deviceid_loaded && !tlscred_loaded && ecukeys_loaded) {
+      secondary_db = true;
+    }
+
     // Print general information if user does not provide any argument.
-    std::cout << "Device ID: " << device_id << std::endl;
+    if (!secondary_db) {
+      if (!deviceid_loaded) {
+        std::cout << "Couldn't load device ID" << std::endl;
+      } else {
+        std::cout << "Device ID: " << device_id << std::endl;
+      }
+    }
+
+    std::string ecu_name = secondary_db ? "Secondary" : "Primary";
     EcuSerials serials;
     if (!storage->loadEcuSerials(&serials)) {
       std::cout << "Couldn't load ECU serials" << std::endl;
     } else if (serials.size() == 0) {
-      std::cout << "Primary serial is not found" << std::endl;
+      std::cout << ecu_name << " serial is not found" << std::endl;
     } else {
-      std::cout << "Primary ecu serial ID: " << serials[0].first << std::endl;
-      std::cout << "Primary ecu hardware ID: " << serials[0].second << std::endl;
+      std::cout << ecu_name << " ECU serial ID: " << serials[0].first << std::endl;
+      std::cout << ecu_name << " ECU hardware ID: " << serials[0].second << std::endl;
     }
 
     if (serials.size() > 1) {
@@ -333,7 +357,7 @@ int main(int argc, char **argv) {
     std::vector<MisconfiguredEcu> misconfigured_ecus;
     storage->loadMisconfiguredEcus(&misconfigured_ecus);
     if (misconfigured_ecus.size() != 0u) {
-      std::cout << "Removed or not registered ecus:" << std::endl;
+      std::cout << "Removed or not registered ECUs:" << std::endl;
       std::vector<MisconfiguredEcu>::const_iterator it;
       for (it = misconfigured_ecus.begin(); it != misconfigured_ecus.end(); ++it) {
         std::cout << "   '" << it->serial << "' with hardware_id '" << it->hardware_id << "' "
@@ -341,7 +365,9 @@ int main(int argc, char **argv) {
       }
     }
 
-    std::cout << "Provisioned on server: " << (registered ? "yes" : "no") << std::endl;
+    if (!secondary_db) {
+      std::cout << "Provisioned on server: " << (registered ? "yes" : "no") << std::endl;
+    }
     std::cout << "Fetched metadata: " << (has_metadata ? "yes" : "no") << std::endl;
 
     auto pacman = PackageManagerFactory::makePackageManager(config.pacman, config.bootloader, storage, nullptr);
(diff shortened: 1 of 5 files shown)


src/aktualizr_info/main.cc
```diff
@@ -79,7 +79,9 @@ int main(int argc, char **argv) {
     ("images-target",  "Outputs targets.json from Image repo")
     ("images-targets",  "Outputs targets.json from Image repo")
     ("image-target",  "Outputs targets.json from Image repo")
-    ("director-target",  "Outputs targets.json from Director repo");
+    ("director-target",  "Outputs targets.json from Director repo")
+  // Identify the database is for secondary ECU
+    ("secondary",  "the database is for secondary ECU");
```
xcheng-here: I added a parameter for aktualizr-info, it means if you want to make the log for secondary database looks better, we need to add this parameter to command line when the database is for secondary. Does it make sense for all of our scenario ?

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -110,6 +110,7 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
         auto r = resp->manifestResp();
         r->manifest.present = manifest_PR_json;
         SetString(&r->manifest.choice.json, manifest);  // NOLINT
+        LOG_TRACE << "Manifest : \n" << manifest;
```
xcheng-here: The manifest was print frequently in my test because we always keep connect/disconnect, maybe annoying. Is it make sense to remove this line ? Or maybe only print it once until next reboot or installation ?

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -139,13 +140,20 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
         resp->present(AKIpUptaneMes_PR_putMetaResp);
         auto r = resp->putMetaResp();
         r->result = ok ? AKInstallationResult_success : AKInstallationResult_failure;
+        if (ok) {
+          LOG_TRACE << "Image root :\n" << meta_pack.image_root;
+          LOG_TRACE << "Image targets :\n" << meta_pack.image_targets;
+          LOG_TRACE << "Image snapshot :\n" << meta_pack.image_snapshot;
+          LOG_TRACE << "Image timestamp :\n" << meta_pack.image_timestamp;
+        }
```
patrickvacek: I think these should be at DEBUG level, and they should be printed in the format "Received Image repo Root metadata: ".

I suggest moving these up to right after they are parsed (line 121) and also adding the output statements for the Director metadata.

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -139,13 +140,20 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
...
       } break;
       case AKIpUptaneMes_PR_sendFirmwareReq: {
         auto fw = msg->sendFirmwareReq();
         auto send_firmware_result = impl_.sendFirmware(ToString(fw->firmware));
         resp->present(AKIpUptaneMes_PR_sendFirmwareResp);
         auto r = resp->sendFirmwareResp();
         r->result = send_firmware_result ? AKInstallationResult_success : AKInstallationResult_failure;
+        LOG_TRACE << "Download : " << (send_firmware_result ? "sucess" : "failed");
```
patrickvacek:
```suggestion
        LOG_INFO << "Download " << (send_firmware_result ? "successful" : "failed") << ".";
```
Can we also print something about the Target, like the name and version?

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -159,6 +167,7 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
         if (install_result == data::ResultCode::Numeric::kNeedCompletion) {
           need_reboot = true;
         }
+        LOG_TRACE << "Install complete " << (need_reboot ? "but need reboot" : "!");
```
patrickvacek:
```suggestion
        LOG_INFO << "Installation complete" << (need_reboot ? ", but needs reboot to complete." : "!");
```
And again, can we also print something about the Target, like the name and version?

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -77,8 +77,8 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
 
     do {
       received = recv(socket, buffer.Tail(), buffer.TailSpace(), 0);
-      LOG_TRACE << "Got " << received << " bytes "
-                << Utils::toBase64(std::string(buffer.Tail(), static_cast<size_t>(received)));
+      // LOG_TRACE << "Got " << received << " bytes "
+      //           << Utils::toBase64(std::string(buffer.Tail(), static_cast<size_t>(received)));
```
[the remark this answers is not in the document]
xcheng-here: I think the most apparent difference in database is the "device_info" table for secondary is empty. So maybe I can check it to know whether this is primary or secondary.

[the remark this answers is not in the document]
patrickvacek: > I think the most apparent difference in database is the "device_info" table for secondary is empty. So maybe I can check it to know whether this is primary or secondary.

I think if that's missing, it could also mean that provisioning has failed. Maybe if you check if that is absent but something else is present that would suffice?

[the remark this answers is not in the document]
xcheng-here: Ah, yes. I think for now, we can check "tls_creds", it's only used in primary application. Do we have any plan to change this behaviour ?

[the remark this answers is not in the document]
patrickvacek: `tls_creds` might be even better than `device_info`, true. I don't expect that behavior will change. But I think we still might need to check if one or the other is empty *and* something else is non-empty to determine if it's a Secondary.

src/aktualizr_secondary/secondary_tcp_server.cc
mike-sul: I am not sure if it belongs here, IMHO it should be in AktualizrSecondary context, e.g. AktualizrSecondary::putMetadata.

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -139,13 +140,20 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
...
+          LOG_TRACE << "Image timestamp :\n" << meta_pack.image_timestamp;
+        }
       } break;
       case AKIpUptaneMes_PR_sendFirmwareReq: {
         auto fw = msg->sendFirmwareReq();
         auto send_firmware_result = impl_.sendFirmware(ToString(fw->firmware));
         resp->present(AKIpUptaneMes_PR_sendFirmwareResp);
         auto r = resp->sendFirmwareResp();
         r->result = send_firmware_result ? AKInstallationResult_success : AKInstallationResult_failure;
+        LOG_INFO << "Download " << (send_firmware_result ? "successful" : "failed") << ".";
```
mike-sul: The same here, IMHO it should be in AktualizrSecondary::sendFirmware

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -159,6 +167,7 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
         if (install_result == data::ResultCode::Numeric::kNeedCompletion) {
           need_reboot = true;
         }
+        LOG_INFO << "Installation complete" << (need_reboot ? ", but needs reboot to complete." : "!");
```
mike-sul: What's wrong with logs in  'AktualizrSecondary::install', this log just duplicates what is in the install() method?

xcheng-here: Ah, yes, this is duplicated with existed log. I will remove it.

src/aktualizr_secondary/secondary_tcp_server.cc
xcheng-here: Emm, I think logging here is more like showing what's came from network, logging in AktualizrSecondary is more like showing what's will be put into database. They should be same theoretically. Which is better ? I was a little bit more like the log be close to network, but I'm also open if you think it's better to be close to database.

patrickvacek: My original suggestion was indeed to print the messages just after decoding them. Currently, we are printing the encoded ASN.1 output, which is pretty useless, and my idea was that we could do a lot better by printing the decoded messages instead.

In fact, for the download and install commands, I'd make a similar suggestion that we should print something as soon as we realize we have a valid message/order from the Primary, and then let the internals of aktualizr-secondary handle printing anything about success or failure.

mike-sul: I don't have a strong opinion here, in overall, in the scope of the given design it doesn't really matter. I suppose in an ideal world logging should be in both places, in a message/request dispatcher, and in a message handler.

src/aktualizr_info/main.cc
```diff
@@ -289,16 +302,28 @@ int main(int argc, char **argv) {
       return EXIT_SUCCESS;
     }
 
+    if (!deviceid_loaded && !tlscred_loaded && ecukeys_loaded) {
```
mike-sul: A quite interesting criterion to determine if it's Secondary's DB :)

tests/test_fixtures.py
```diff
@@ -158,7 +158,7 @@ def get_info(self, retry=30):
                                           timeout=60, stdout=subprocess.PIPE, env=self._run_env)
             if info_exe_res.returncode == 0 and \
                     str(info_exe_res.stdout).find('Provisioned on server: yes') != -1 and \
-                    str(info_exe_res.stdout).find('Current primary ecu running version:') != -1:
+                    str(info_exe_res.stdout).find('Current Primary ECU running version:') != -1:
```
mike-sul: That's another proof that it's an antipattern to use raw string output as an input for another software component. aktualizr-info should be able to output the status info in a structured way - json, xml, yaml, etc. 
Most likely this change will require adjustment in OTF tests, too.

src/aktualizr_info/main.cc
xcheng-here: The idea here is that if there is no device id and no tls creds, it means it's for secondary or a database not init successfully, and if the ecu keys existing, it means it's not caused by failure at init time. Yeah, it's a little confused, so, if there is any other options, I like to modify it.

tests/test_fixtures.py
patrickvacek: Sorry, how would reformatting into xml or json have changed this in any way? Unless the strings are stored as constants in a shared place, you still have to change the string you are matching in both places.

That said, I'm not opposed to having all aktualizr-info output in json or another format, albeit probably with an optional flag. There might even be a ticket for it somewhere.

src/aktualizr_info/main.cc
```diff
@@ -103,6 +103,9 @@ int main(int argc, char **argv) {
 
     AktualizrInfoConfig config(vm);
 
+    bool secondary_db = false;
+    ;
```
patrickvacek:
```suggestion
```

src/aktualizr_info/main.cc
patrickvacek: It's a bit hacky but it might be the best we can do for now. I'm open to other suggestions.

tests/test_fixtures.py
mike-sul: > Sorry, how would reformatting into xml or json have changed this in any way?

I am not talking about just wrapping the text currently output by aktualizr-info into json or xml. First of all, it's about defining a data model and then corresponding representation of it in json. For example, in yaml it would be something like the following. I understand, that you might say that the yaml key names also can be changed but log messages are something that changes much more often compared to a key definition.

`
- device
    id: <>
    provisioned: true
    version: <>

    - ecu
         serial: <serial>
         hw-id: <>
         type: primary

   - ecu
        serial: <serial>
        hw-id: <>
        type: secondary
`

src/aktualizr_secondary/secondary_tcp_server.cc
patrickvacek: I think this is fine for now at trace level. That's expected to be very verbose, and the manifest is useful.

PR #1608, opened 2020-03-18
fix keymanager: use ed25519 for to sign manifest

kbushgit:

### DIFF
--- a/src/libaktualizr/crypto/keymanager.cc
+++ b/src/libaktualizr/crypto/keymanager.cc
@@ -298,7 +298,18 @@ Json::Value KeyManager::signTuf(const Json::Value &in_data) const {
   }
 
   Json::Value signature;
-  signature["method"] = "rsassa-pss";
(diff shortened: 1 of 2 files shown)


kbushgit:
> Looks fine, but can we have a unit test that exercises this for both key algorithms?

Yes sure, but it still not clear for me how the different methods are processed on the BE, so far we have only one definition for the RSA algorithm ```rsassa-pss``` ,  it's not related to the ```ed25519``` but anyway...

patrickvacek:
This looks fine. How does the BE react to the manifest?

kbushgit:
> This looks fine. How does the BE react to the manifest?

Without this fix, the server response with the following error
```
response: {"code":"signature_not_valid","description":"The given signature is not valid","cause":null,"errorId":"5cc595df-2e8e-4ed0-83f2-9b76e3309666"}
Put manifest request failed: 0  HTTP 400

```

PR #1609, opened 2020-03-20
Adjust some log in aktualizr_secondary

xcheng-here:


xcheng-here:
I haven't modify below 3 lines log:
Could not find primary ecu serial, defaulting to empty serial: unknown error
Could not find hardware_id for serial : unknown error
No valid and pending Director targets to be applied

The reason is the first 2 lines are from libaktualizr, describing the status of database, so if we want it don't output for secondary, it means maybe we need to modify some interfaces of libaktualizr. I think maybe we don't want to modify so much to remove those 2 lines log.
The third one is from aktualizr_secondary, but the problem is similar as above, it's describing the status of database, I'm not sure whether we really want to add code to check if it's the first time launching.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -101,6 +101,7 @@ bool AktualizrSecondary::sendFirmware(const std::string& firmware) {
     return false;
   }
 
+  LOG_INFO << "Download successful.";
```
patrickvacek: Can you add the name of the target?

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -198,6 +199,7 @@ bool AktualizrSecondary::doFullVerification(const Metadata& metadata) {
 
   pending_target_ = targetsForThisEcu[0];
 
+  LOG_DEBUG << "Meta verification passed.";
```
patrickvacek:
```suggestion
  LOG_DEBUG << "Metadata verified, new update found.";
```

xcheng-here:
> How hard is it to fix this (and the other two)? If it requires a bunch of work, it might not be worth it, but it seems like we ought to be able to do better here.

IMHO, it's doable but maybe not worthy because those log also indicating some invalid status for both of primary and secondary. If I don't want to put some ugly code into our source, I will need to change some interfaces and maybe add something into database.

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -146,14 +146,15 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
         r->result = ok ? AKInstallationResult_success : AKInstallationResult_failure;
       } break;
       case AKIpUptaneMes_PR_sendFirmwareReq: {
+        LOG_DEBUG << "Received sendFirmwareReq from Primary.";
```
patrickvacek: Actually, this should probably be INFO level.

src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -146,14 +146,15 @@ bool SecondaryTcpServer::HandleOneConnection(int socket) {
...
         auto fw = msg->sendFirmwareReq();
         auto send_firmware_result = impl_.sendFirmware(ToString(fw->firmware));
         resp->present(AKIpUptaneMes_PR_sendFirmwareResp);
         auto r = resp->sendFirmwareResp();
         r->result = send_firmware_result ? AKInstallationResult_success : AKInstallationResult_failure;
-        LOG_INFO << "Download " << (send_firmware_result ? "successful" : "failed") << ".";
       } break;
       case AKIpUptaneMes_PR_installReq: {
+        LOG_DEBUG << "Received installReq from Primary.";
```
patrickvacek: This should also be INFO.

patrickvacek:
>  No valid and pending Director targets to be applied

@mike-sul Can you actually explain why this message is printed? I also note that after an update has been installed with aktualizr-secondary, I see "There is a valid and pending Director target to be applied" even after rebooting. I think there's a problem with how we are checking the Director metadata here. We are doing the right thing, but the logging is wrong. `director_repo_.checkMetaOffline` should only fail if there is no metadata, and then we have to do additional work to determine if there are pending Targets to apply.

mikesul-vd:
> Can you actually explain why this message is printed?

To inform that there is no target (or target metadata) verification and installation of which is currently in progress. 

> director_repo_.checkMetaOffline should only fail if there is no metadata, and then we have to do additional work to determine if there are pending Targets to apply.

I agree with this statement and I think the code is inline with it. Just after `initPendingTargetIfAny()` the following is called `if (hasPendingUpdate()) {`.

patrickvacek:
> > Can you actually explain why this message is printed?
> 
> To inform that there is no target (or target metadata) verification and installation of which is currently in progress.
> 
> > director_repo_.checkMetaOffline should only fail if there is no metadata, and then we have to do additional work to determine if there are pending Targets to apply.
> 
> I agree with this statement and I think the code is inline with it. Just after `initPendingTargetIfAny()` the following is called `if (hasPendingUpdate()) {`.

The code works correctly, but I think the logging is misleading. First, `director_repo_.checkMetaOffline` failing or succeeding does not tell us anything about whether there are updates to install or apply. It only tells us whether we have valid metadata in the database. Second, the message "There is a valid and pending Director target to be applied" appears every time the Secondary is started even after the target has been applied.

mikesul-vd:
> > > Can you actually explain why this message is printed?
> > 
> > 
> > To inform that there is no target (or target metadata) verification and installation of which is currently in progress.
> > > director_repo_.checkMetaOffline should only fail if there is no metadata, and then we have to do additional work to determine if there are pending Targets to apply.
> > 
> > 
> > I agree with this statement and I think the code is inline with it. Just after `initPendingTargetIfAny()` the following is called `if (hasPendingUpdate()) {`.
> 
> The code works correctly, but I think the logging is misleading. First, `director_repo_.checkMetaOffline` failing or succeeding does not tell us anything about whether there are updates to install or apply. It only tells us whether we have valid metadata in the database. Second, the message "There is a valid and pending Director target to be applied" appears every time the Secondary is started even after the target has been applied.

Oh, now I understand your point. Yes, the first log message is misleading for sure, it should be changed or removed at all.

The second one should be OK provided that the metadata are removed from DB just after successful installation, at least that was my plan. So, if the metadata are removed just after successful installation then this `if (!director_repo_.checkMetaOffline(*storage_))` should be true and `LOG_INFO << "There is a valid and pending Director target to be applied";` is never called.

src/aktualizr_secondary/aktualizr_secondary.cc
```diff
@@ -246,7 +248,7 @@ void AktualizrSecondary::uptaneInitialize() {
 
 void AktualizrSecondary::initPendingTargetIfAny() {
   if (!director_repo_.checkMetaOffline(*storage_)) {
-    LOG_INFO << "No valid and pending Director targets to be applied";
+    LOG_INFO << "No valid meta.";
```
patrickvacek:
```suggestion
    LOG_INFO << "No valid metadata found in storage.";
```

patrickvacek:
> > IMHO, it's doable but maybe not worthy because those log also indicating some invalid status for both of primary and secondary. If I don't want to put some ugly code into our source, I will need to change some interfaces and maybe add something into database.
> 
> Fair enough. I'll look into it some more and think about it, but it doesn't have to block the rest of this.

I've addressed these two issues in https://github.com/advancedtelematic/aktualizr/pull/1611.

patrickvacek:
> The second one should be OK provided that the metadata are removed from DB just after successful installation, at least that was my plan.

The metadata is not removed from the database, and it should not be. I've pushed another commit that address this issue as well as a few other minor things that were bugging me.

mike-sul:
> The metadata is not removed from the database, and it should not be

What for do we need to keep it in the DB after a target update has been installed and applied?

mike-sul:
> > The metadata is not removed from the database, and it should not be
> 
> What for do we need to keep it in the DB after a target update has been installed and applied?

Got already answer to it :)

PR #1612, opened 2020-03-23
Secondary metadata error

lbonn:
The test is done with a simple fiu_fail

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -1082,6 +1082,48 @@ TEST(Aktualizr, InstallationFailure) {
   }
 }
 
+/*
+ * Verifies that updates fail after metadata verification failure reported by secondaries
+ */
+TEST(Aktualizr, SecondaryMetaFailure) {
+  TemporaryDirectory temp_dir;
+  auto http_server_mock = std::make_shared<HttpFake>(temp_dir.Path(), "hasupdates", fake_meta_dir);
+  Config conf = UptaneTestCommon::makeTestConfig(temp_dir, http_server_mock->tls_server);
+  auto storage = INvStorage::newStorage(conf.storage);
+
+  fault_injection_init();
+  fiu_enable("secondary_putmetadata", 1, nullptr, 0);
+
+  UptaneTestCommon::TestAktualizr aktualizr(conf, storage, http_server_mock);
+
+  struct {
+    result::Install result;
+    std::promise<void> promise;
+  } ev_state;
+
+  auto f_cb = [&ev_state](const std::shared_ptr<event::BaseEvent>& event) {
+    if (event->variant != "AllInstallsComplete") {
+      return;
+    }
+    const auto installs_complete = dynamic_cast<event::AllInstallsComplete*>(event.get());
+    ev_state.result = installs_complete->result;
+
+    ev_state.promise.set_value();
+  };
+  boost::signals2::connection conn = aktualizr.SetSignalHandler(f_cb);
+  aktualizr.Initialize();
+  aktualizr.UptaneCycle();
+
+  auto status = ev_state.promise.get_future().wait_for(std::chrono::seconds(20));
+  if (status != std::future_status::ready) {
+    FAIL() << "Timed out waiting for installation to complete.";
+  }
+
+  ASSERT_FALSE(ev_state.result.dev_report.isSuccess());
+
+  fiu_disable("secondary_putmetadata");
+}
+
 #endif  // FIU_ENABLE
 
 /*
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -940,7 +940,10 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
     std::vector<Uptane::Target> primary_updates = findForEcu(updates, primary_ecu_serial);
 
     //   6 - send metadata to all the ECUs
-    sendMetadataToEcus(updates);
+    if (!sendMetadataToEcus(updates)) {
+      result.dev_report = {false, data::ResultCode::Numeric::kInternalError, "Metadata verification failed"};
+      return std::make_tuple(result, "Secondary metadata verification failed", true);
+    }
 
     //   7 - send images to ECUs (deploy for OSTree)
     if (primary_updates.size() != 0u) {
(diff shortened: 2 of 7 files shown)


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1225,51 +1228,54 @@ void SotaUptaneClient::rotateSecondaryRoot(Uptane::RepositoryType repo, Uptane::
...
+      if (sec == secondaries.end()) {
+        continue;
+      }
+
+      /* Root rotation if necessary */
+      rotateSecondaryRoot(Uptane::RepositoryType::Director(), *(sec->second));
+      rotateSecondaryRoot(Uptane::RepositoryType::Image(), *(sec->second));
+      if (!sec->second->putMetadata(meta)) {
+        LOG_ERROR << "Sending metadata to " << sec->first << " failed";
+        put_meta_succeed = false;
```
patrickvacek: Should we abort if any installation fails? Or is that a separate problem to solve since we don't handle atomicity well at present and that's a bigger topic?

src/virtual_secondary/managedsecondary.cc
```diff
@@ -72,6 +73,10 @@ void ManagedSecondary::rawToMeta() {
 }
 
 bool ManagedSecondary::putMetadata(const Uptane::RawMetaPack &meta_pack) {
+  if (fiu_fail("secondary_putmetadata") != 0) {
+    return false;
+  }
+
```
patrickvacek: Can this go in VirtualSecondary instead of ManagedSecondary?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -940,7 +940,10 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
     std::vector<Uptane::Target> primary_updates = findForEcu(updates, primary_ecu_serial);
 
     //   6 - send metadata to all the ECUs
-    sendMetadataToEcus(updates);
+    if (!sendMetadataToEcus(updates)) {
+      result.dev_report = {false, data::ResultCode::Numeric::kInternalError, "Metadata verification failed"};
+      return std::make_tuple(result, "Secondary metadata verification failed", false);
+    }
```
patrickvacek: Just confirming: if this fails, does the backend recognize it and abort the update, or does it stay pending and cause aktualizr to retry it?

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -1082,6 +1082,47 @@ TEST(Aktualizr, InstallationFailure) {
...
+  auto status = ev_state.future.wait_for(std::chrono::seconds(20));
+  if (status != std::future_status::ready) {
+    FAIL() << "Timed out waiting for installation to complete.";
+  }
+
+  ASSERT_FALSE(ev_state.result.dev_report.isSuccess());
+
+  fiu_disable("secondary_putmetadata");
+}
+
```
patrickvacek: This test looks good, but is it possible to do it at a lower level? Maybe not; this is probably fine. Either way, I think a brief comment explaining what exactly this test is covering would be a good idea.

lbonn: Yes, I will add some explanatory comments.
But what do you mean by lower level? Something not using lifiu?

patrickvacek: libfiu is fine, I just meant without having to use the libaktualizr API and all. But don't sweat it, this is already fine.

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: That's a good question. After further thinking, it might be worse than ignoring the check as it allows an hostile secondary to block a primary (and other secondaries) update, even if the metadata checks out from the PoV of the primary.

src/virtual_secondary/managedsecondary.cc
lbonn: Yes, though it needs an overriding of the method there but no big deal.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Interesting point. I would think that would be okay, but we need to properly decide what to do in these cases and this is decidedly out of the scope of this work.

src/libaktualizr/primary/aktualizr_test.cc
```diff
@@ -1082,6 +1082,47 @@ TEST(Aktualizr, InstallationFailure) {
...
+    const auto installs_complete = dynamic_cast<event::AllInstallsComplete*>(event.get());
+    ev_state.result = installs_complete->result;
+
+    ev_state.promise.set_value();
+  };
+  boost::signals2::connection conn = aktualizr.SetSignalHandler(f_cb);
+  aktualizr.Initialize();
+  aktualizr.UptaneCycle();
+
+  auto status = ev_state.future.wait_for(std::chrono::seconds(20));
```
mike-sul: it's really minor, just IMHO it would look slightly more readable if ev_state.future field is not defined at all and just `ev_state.promise.get_future().wait_for()` is used.

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: It's actually a good question. It's out of the scope of the given PR nevertheless I think it makes sense to return something more meaningful to `putMetadata`, not just bool like we do for `install`. So, the follow-up behavior can depend on an error type, e.g. if it's some network-related error then it makes sense to retry if Uptane verification of metadata actually fails then there is no point to retry.

I suppose proper testing of the actualize behavior after putMetadata failure requires adding test(s) to the python-based tests and OTF tests in order to see how it behaves if the uptane cycle is running.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1225,51 +1228,54 @@ void SotaUptaneClient::rotateSecondaryRoot(Uptane::RepositoryType repo, Uptane::
...
-    return;
+    return false;
   }
 
-  for (auto targets_it = targets.cbegin(); targets_it != targets.cend(); ++targets_it) {
-    for (auto ecus_it = targets_it->ecus().cbegin(); ecus_it != targets_it->ecus().cend(); ++ecus_it) {
-      const Uptane::EcuSerial ecu_serial = ecus_it->first;
-
+  bool put_meta_succeed = true;
+  for (const auto &target : targets) {
```
mike-sul: +

src/libaktualizr/primary/sotauptaneclient.cc
mike-sul: My understanding is that even if we don't interrupt on the first putMetadata error anyway sendFirmware/install won't be called even for those Secondaries for which putMetdata was successful.
I agree with Patrick that it has to be discussed with a wider audience including product owners/managers, personally, I think aktualizr should carry on "metadata putting" and installation/update process if there is at least one "successful" case and then retry if there were non-verification/validation errors and then report a result to the backend...

src/virtual_secondary/managedsecondary.cc
```diff
@@ -72,6 +73,10 @@ void ManagedSecondary::rawToMeta() {
 }
 
 bool ManagedSecondary::putMetadata(const Uptane::RawMetaPack &meta_pack) {
+  if (fiu_fail("secondary_putmetadata") != 0) {
```
mike-sul: I suppose it's fine in the given context to inject into the "production" code some test code but IMHO, implementation of SecondaryInterface that realizes specific test case and is instantiated within a corresponding test and then just aktualizr->AddSecondary would do more robust and flexible. Otherwise, adding a new test case means adding another `if (fiu_fail("") != 0)` into the "production" code.

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: ?

mike-sul: I meant this +1: :)

src/virtual_secondary/managedsecondary.cc
lbonn: Yes that's a valid point. I've chosen the fiu solution one part because it's simpler and it's also easier to trigger the error against the real backend by scheduling an update and launching aktualizr with fiu.

For now, the code clutter seem minimal enough to me but I understand that's subjective.

mike-sul: +1:

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: @patrickvacek I just had a tried and the installation shows as failed on the UI, with nothing pending.

However, if I re-run `aktualizr check` after the failure, it still shows 1 new update to install, which it will do if `aktualizr once` is called without the injected error.

So it's not ideal but it sounds like we can't really fix that on the client?

patrickvacek: I think this actually might require a client-side fix. I'm wondering if we need to drop the Director Targets metadata in this case (and all the similar returns for the metadata checks and such above this in this function). I want to avoid more of these cases where we get stuck in a loop, and as long as the empty targets optimization is in place, we have to work around it.

lbonn: Ah you're right, it doesn't persist if we drop the metadata. That's what I pushed in the last version.

patrickvacek: Glad to know that fixed it, but annoying that we have to keep expanding this target dropping business. I suspect all of the errors in that lambda function should return true for `drop_targets`.

lbonn:
Added a test on secondary side, using mocks.

Is there still something missing in your opinion?

src/virtual_secondary/managedsecondary.cc
patrickvacek: Can you get rid of this `fiu_fail` bit now that it's in `VirtualSecondary`?

lbonn: Oops I thought I did. Thanks!

PR #1614, opened 2020-03-24
fix: update the formula with a new version before bottling a new release

mike-sul:


mike-sul:
@patrickvacek This should fix the homebrew release job

patrickvacek:
> This should fix the homebrew release job

Thanks! How do we re-trigger it?

PR #1619, opened 2020-03-27
get_garage_sign.py: Point at new AWS bucket.

patrickvacek:


lbonn:
There is also a size field, we could at least check it. And because it goes through TLS it's maybe good enough for integrity (detects partial downloads).

patrickvacek:
New bucket is ready, I think this can be safely merged now if we want.

zabbal:
On a related note, is there a place where we document the URL? Some stable page like docs.ota.connect or smth like that which we can refer to in the help message so users can manually check and fix things (provided we expose hardcoded value as command line option) if things go south again?

lbonn:
@zabbal 
> Also, the rename commit seems to be entirely orthogonal and can be sent as a separate PR, but that's just minor nitpicking from my side.

We prefer not to open too many small PRs if not necessary. Typically, for a related trivial change like this one, having it in a separate commit is enough for our purpose. We would split if there were some big contentious changes and a smaller commit that is more urgent and would need to be fast-tracked for example.

In this case, having two PRs would also create unnecessary headaches with merge conflicts and rebasing.

patrickvacek:
> why hardcode it at all, why not add another command-line option with default value of hardcoded URL?

It's an option, but I never considered it'd be necessary before. I'll consider that.

> why bother with "LastModified" checks - can we ask to add fixed tuf-cli-latest.tgz which would always link to the newest tuf-cli-v1.2.3.tgz every time the new version is uploaded?

Yes, that could also be an option, but that's outside of aktualizr. :)

> is there a place where we document the URL?

Nope.

Merging for now, happy to think about how to further future-proof this.

patrickvacek:
> > why hardcode it at all, why not add another command-line option with default value of hardcoded URL?
> 
> It's an option, but I never considered it'd be necessary before. I'll consider that.

@zabbal BTW not the same as what you suggested, but there's already support for downloading the archive yourself and providing the path to the script/CMake.

zabbal:
> there's already support for downloading the archive yourself and providing the path to the script/CMake

Cool, where can I read more about it?

patrickvacek:
> Cool, where can I read more about it?

https://github.com/advancedtelematic/aktualizr/blob/master/src/sota_tools/CMakeLists.txt#L78
https://github.com/advancedtelematic/aktualizr/blob/master/scripts/get-garage-sign.py#L20

PR #1623, opened 2020-03-31
Move isotp out of libaktualizr

xcheng-here:


patrickvacek:
Looks fine, but I think I'd prefer if the two new subdirectories were merged into one. I don't think there's enough code to justify separating them.

patrickvacek:
The LGTM check is failing. I thought it might be bogus, but when I try to build with `-DBUILD_ISOTP=off` I get the same error:

```
CMake Error at CMakeLists.txt:436 (message):
  src/libaktualizr-isotpconn/isotp_allocate.cc not checked
```

xcheng-here:
> Looks fine, but I think I'd prefer if the two new subdirectories were merged into one. I don't think there's enough code to justify separating them.

OK. I put them together.

CMakeLists.txt
```diff
@@ -423,19 +423,21 @@ add_subdirectory("docs/doxygen")
...
 # Note: does not check `tests` directory which depends too much on conditional
 # compilation
-file(GLOB_RECURSE ALL_SOURCE_FILES RELATIVE ${CMAKE_SOURCE_DIR}
-    src/*.cc src/*.c src/*.h)
-foreach(FILE ${ALL_SOURCE_FILES})
-    string (FIND ${FILE} "/isotp_conn/" EXCLUDE_DIR_FOUND)
-    if (NOT ${EXCLUDE_DIR_FOUND} EQUAL -1)
-      continue()
-    endif()
+if(BUILD_ISOTP)
```
lbonn, line 426: This is not really satisfactory, this check should run even if `BUILD_ISOTP` is off.

The solution is to put the `aktualizr_source_file_checks` outside the check, so that it's run unconditionally.

xcheng-here: Yes, I think you are right. I will modify it.

CMakeLists.txt
```diff
@@ -426,9 +426,9 @@ add_subdirectory("docs/doxygen")
 file(GLOB_RECURSE ALL_SOURCE_FILES RELATIVE ${CMAKE_SOURCE_DIR}
     src/*.cc src/*.c src/*.h)
 foreach(FILE ${ALL_SOURCE_FILES})
-    string (FIND ${FILE} "/isotp_conn/" EXCLUDE_DIR_FOUND)
+    string (FIND ${FILE} "/libaktualizr-isotpconn/" EXCLUDE_DIR_FOUND)
```
patrickvacek, line 429: This should be `libaktualizr-socketcan` in order to match the actual directory, right?

But on second thought, should the subdirectory actually be `libaktualizr-isotp`? I think that's the more specific, higher-level protocol being used here, and this is code is less specific to socketcan. But I'm open to debate on that.

xcheng-here: Yes, libaktualizr-isotp is more close to the name of our source file, I changed it.

PR #1624, opened 2020-03-31
Feat/ota 4782/ip secondary interface

mike-sul:


src/aktualizr_secondary/secondary_tcp_server.cc
```diff
@@ -30,50 +35,68 @@ void SecondaryTcpServer::run() {
...
+  }
+
   while (keep_running_.load()) {
     sockaddr_storage peer_sa{};
     socklen_t peer_sa_size = sizeof(sockaddr_storage);
 
     LOG_DEBUG << "Waiting for connection from Primary...";
     int con_fd = accept(*listen_socket_, reinterpret_cast<sockaddr *>(&peer_sa), &peer_sa_size);
     if (con_fd == -1) {
+      // why are we aborting here?
```
patrickvacek, line 51: What would you do instead?

src/aktualizr_secondary/secondary_rpc_test.cc
```diff
@@ -0,0 +1,227 @@
...
+ protected:
+  MsgDispatcher msg_dispatcher_;
+  SecondaryTcpServer secondary_server_;
+  std::thread secondary_server_thread_;
+};
+
+// The given test fails because the Secondary TCP server implementation
+// is a single-threaded, synchronous and blocking hence it cannot accept any new connections
+// until the current one is closed. Therefore, if a client/Primary does not close its socket for some reason then
+// Secondary becomes "unavailable"
```
patrickvacek, line 168: Do you expect that we can fix or work around this issue and then restore this test?

src/aktualizr_secondary/aktualizr_secondary_interface.h
```diff
@@ -0,0 +1,29 @@
+#ifndef AKTUALIZR_SECONDARY_INTERFACE_H
+#define AKTUALIZR_SECONDARY_INTERFACE_H
+
+#include "uptane/manifest.h"
+
+class IAktualizrSecondary {
```
patrickvacek, line 6: What's the benefit of having an interface that only the Secondary uses? If it's just for the mock, couldn't the mock inherit directly from the AktualizrSecondary class?

src/aktualizr_secondary/secondary_tcp_server.cc
mike-sul: Just continue accepting/listening for a new connection, otherwise Secondary becomes unavailable. On the other hand, the secondary daemon is managed by systemd which is supposed to restart it so it starts listening for a new connection again.

src/aktualizr_secondary/secondary_rpc_test.cc
mike-sul: In general yes. I am not sure if it makes sense to fix it in the given implementation in the light of potential switch to the mqtt based solution which won't have this issue, but the test I guess is valid for both implementations so can be reused.

src/aktualizr_secondary/aktualizr_secondary_interface.h
mike-sul: Yes, at the moment it's just for the mock/test. Yes, the mock can inherit from AktualizrSecondary. I've considered a few options including the one you mentioned. The drawback of this approach is a circular reference between AktualirzSecondary and AktualizrSecondaryMsgDispatcher (which can be mitigated by a forward declaration)  and the test becomes depending on AktualizrSecondary implementation internals, which can be mitigated by applying PIMPL to AktualizrSecondary. Frankly speaking, I don't have a strong opinion here, the inheritance also works for me as long as a proper forward declaration is applied and MsgDispatcher and AktualizrSecondaryMsgDispatcher are declared in different header files so SecondaryTcpServer does not depend on AktualizrSecondaryMsgDispatcher.

src/aktualizr_secondary/secondary_tcp_server.cc
mike-sul: > t I'd prefer to remove the comment to prevent future confusion, unless you either want to make it an actual explanation of why it is aborting or a TODO with a specific improvement proposal.

I'll add an explanation and leave the `break` and let systemd do the work here.

src/aktualizr_secondary/secondary_rpc_test.cc
mike-sul: > That's good to hear. I don't think I understand why it would fail now but not before this PR, though.

Because the server is a single-threaded and synchronous so it does not listen/accept a new connection until a current one is closed. This test fails now and will fail for any new changes until the server concurrency model is changed.

src/aktualizr_secondary/aktualizr_secondary_interface.h
mike-sul: OK, I will leave it unchanged in this PR, anyway it looks like I will have to rework it in order to implement the new install method (my current task).

mike-sul:
Now, I see the logs from last week :)

src/aktualizr_secondary/secondary_rpc_test.cc
patrickvacek: But was it multithreaded before? I still don't quite understand why it succeeded before this PR.

mike-sul: No, it never succeeded before, this is a new test.

patrickvacek: D'oh. Sorry, I thought it was copied over. Carry on!

PR #1626, opened 2020-04-02
clean the cached aktualizr repo before brewing

mike-sul:
At first run, this release job https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/jobs/2203603 has failed because I left some changes (2020.5 tag) in the locally cached aktualizr repo what led to the repo update failure (the local tag and the new remote tag pointed to different revisions/commits).
So, I added an additional step to the CI job - remove the locally caches aktualizr repo to make sure that we have "clean" repo before starting the brewing process.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -461,6 +461,6 @@ build-osx-release:
     - if: $CI_COMMIT_TAG =~ /^20\d\d\.\d\d?-docs$/
       when: never
     - if: $OSX_RELEASE && $CI_COMMIT_TAG =~ /^\d\d\d\d\.\d+(-\w+)?$/
-      when: always
+      when: on_success
   tags:
   - osx
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -434,6 +434,7 @@ build-osx-release:
   before_script:
     - brew uninstall -f aktualizr
     - brew untap advancedtelematic/otaconnect
+    - rm -rf $(brew --cache)/aktualizr--git
     - brew install ghr
   script:
     # clone a repo that contains the aktualizr formula


ci/gitlab/.gitlab-ci.yml
```diff
@@ -434,6 +434,7 @@ build-osx-release:
   before_script:
     - brew uninstall -f aktualizr
     - brew untap advancedtelematic/otaconnect
+    - rm -rf /Users/gitlab/Library/Caches/Homebrew/aktualizr--git
```
patrickvacek, line 437: Slightly ugly that we have to both hardcode a path like that and use `rm -Rf`. Both of those make me uncomfortable. Are there other options?

mikesul-vd: yeah, I don't like it too, I tried `brew cleanup -s` and `brew install -v -f --build-bottle aktualizr` ('f' option) but they do not do the job, looks like it's another brew issue according to what I read...

patrickvacek: Anyone else see any options here? Or if it's what we have to do, then so be it...

mike-sul: This part of the path `/Users/gitlab/Library/Caches/Homebrew` can be obtained from `$(brew --cache)`, let me see if I can find anything for `aktualizr--git`

patrickvacek: > This part of the path `/Users/gitlab/Library/Caches/Homebrew` can be obtained from `$(brew --cache)`

Ah, interesting. The `rm` is still scary, but maybe less so.

> let me see if I can find anything for `aktualizr--git`

That's the part I'm least worried about. :)

mike-sul: > Ah, interesting. The rm is still scary, but maybe less so.

the maximum what can be removed is the stuff that `gitlab` user is permitted to remove without sudo and it's not system critical stuff so the follow-up brew command should restore its required state.

patrickvacek: Fair point. And if the `aktualizr--git` part is hardcoded we can be fairly sure it's not going to do anything too dangerous.

PR #1627, opened 2020-04-03
Update the Antora docs to keep track of the latest version.

houcros:
The docs can't be published because of the conflicting versions between `2020.4-docs` and `2020.5-docs` branches. Since the latter seems to now be the latest version, the former needs to be "downgraded".

### DIFF
--- a/docs/ota-client-guide/antora.yml
+++ b/docs/ota-client-guide/antora.yml
@@ -1,6 +1,6 @@
 name: ota-client
 title: OTA Connect Developer Guide
-version: latest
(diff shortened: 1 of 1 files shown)


patrickvacek:
Looks identical to my PR from yesterday (https://github.com/advancedtelematic/aktualizr/pull/1625) except that yours has quotes. Are those necessary?

houcros:
Ups, didn't see that 😄
Don't know if the quotes are strictly necessary. But the couple other branches I checked do have quotes, so I figured they were necessary to avoid parsing as a float or something...

PR #1628, opened 2020-04-04
Added API to direct update raw_roport in the installation result

kbushgit:

This API call should be used after installation complete and before clear of installation result ```clearInstallationResults``` in the storage.
If the device installation result does not exist in the storage the invocation of this API won't have any negative effect it just returns a negative result of the execution.


patrickvacek:
Can you rebase now that we have a fix for the CI problem?

patrickvacek:
How can we convince CI to refresh the docker image instead of pulling from the cache?

PR #1629, opened 2020-04-06
Fix broken yocto link

tkfu:
Normally I wouldn't bother with a PR for a minor fix like this, but I wanted to ping @ammaboateng and @HalynaDumych because this was caused by an asciidoc behaviour that might not be obvious: this page included content from another page that had the `{yocto-version}` attribute substitution in it, but didn't include the definition of the attribute, causing a broken link.

Note that this will be fixable in a cleaner way when Antora 2.3 is released, because support for component-scoped attributes was added: https://gitlab.com/antora/antora/issues/251

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/build-qemu.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/build-qemu.adoc
@@ -8,6 +8,7 @@ We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}
 ====
 endif::[]
 
+include::_partials/aktualizr-version.adoc[]
 
 :page-layout: page
 :page-categories: [quickstarts]


patrickvacek:
Ah sorry, that was my doing! I also didn't realize that would be necessary. Good catch!

houcros:
I still don't have a good understanding of the docs publishing process, but is it possible this needs to be applied to other docs branches as well? The error in [this pipeline](https://main.gitlab.in.here.com/olp/edge/ota/documentation/ota-connect-docs/-/jobs/2253929) make me think that it's missing in the `2020.2-docs` branch, namely the errors:
```
Parent URL file:///builds/olp/edge/ota/documentation/ota-connect-docs/site/ota-client/2020.2/build-qemu.html, line 646, col 26
294 Real URL   https://www.yoctoproject.org/docs/%7Byocto-version%7D/ref-manual/ref-manual.html#detailed-supported-distros
```

tkfu:
There's something a bit strange going on...I'm investigating now. I think it has to do with how xrefs without explicitly-specified versions get resolved, though.

PR #1632, opened 2020-04-06
Drop targets when timing out because of secondaries

lbonn:


src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -933,7 +933,7 @@ result::Install SotaUptaneClient::uptaneInstall(const std::vector<Uptane::Target
     // stuck in an unrecoverable state here
     if (!waitSecondariesReachable(updates)) {
       result.dev_report = {false, data::ResultCode::Numeric::kInternalError, "Unreachable secondary"};
-      return std::make_tuple(result, "Secondaries were not available", false);
+      return std::make_tuple(result, "Secondaries were not available", true);
```
patrickvacek, line 936: I'm in favor of this, but also in favor of changing all the failures to `true` for `drop_targets`.

lbonn: I've first tried a similar approach but it broke some assumption in test_uptane:InstallFakeBad, so I opted for a minimal change for now.

PR #1633, opened 2020-04-06
Fix include syntax for partials

tkfu:
This was a tricky one, even though the actual fix is fairly small. I've been troubleshooting it for most of the day, though, so I want to make sure it's explained in a PR (and again, I'm tagging @ammaboateng and @HalynaDumych).

It's possible to include content in Antora using the asciidoc `include::[]` directive in two different ways: by [resource ID](https://docs.antora.org/antora/2.1/page/resource-id/), or by (relative) path. In general, we should always be including by resource ID, because inheriting by resource ID means that we retain the context from the page where the include occurs.

To include a page via resource ID, there's a small syntax difference from the xref syntax: for disambiguation purposes, you need to include the **family**--includes could be pages, partials, examples, etc., whereas xrefs can only be pages. This is what it looks like in practice:
```
include::page$build-raspberry.adoc[] // for a page
include::ota-client::page$build-raspberry.adoc[] // for a page, but specifying the component explicitly
include::partial$aktualizr-version.adoc[] // for a partial
include::example$sota-local.toml[] // for an example
```

There were several places where we weren't using resource IDs, and that caused some unexpected bugginess. Most of them were actually quite subtle: the include would work, but would grab a version from the wrong branch, for example. This was never caught because the only places it cause incorrect rendering was in older branches; include-by-relative-path happened to generate correct docs for the latest branch in all the places it was used (as far as I can tell).

My initial (erroneous) fix for the problem (#1629) actually ended up surfacing those bugs by causing broken links, which the link-checker CI job caught. Figuring out why that happened is what led to this fix. I'm also pushing similar fix commits to all previous docs branches, but won't be opening a PR for every one of those.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/add-meta-updater-to-vendors-sdk.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/add-meta-updater-to-vendors-sdk.adoc
@@ -7,7 +7,7 @@ We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}
 ====
 endif::[]
 
-include::_partials/aktualizr-version.adoc[]
+include::partial$aktualizr-version.adoc[]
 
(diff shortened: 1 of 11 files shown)


HalynaDumych:
Looks really tricky! Thanks for figuring that out :)

PR #1635, opened 2020-04-08
ref: disable dropping director's targets in the install's call

kbushgit:
Signed-off-by: Kostiantyn Bushko <kostiantyn.bushko@yahoo.com>

By moving dropping director's targets into the ```SotaUptaneClient:::AssembleManifest()```,  means that any failure in the  ```SotaUptaneClient::uptaneInstall(...)``` call, will lead to drop the director's targets. 
Any installation result code, except ```kOk``` is considered as a state when we must drop the director's target.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -268,6 +268,10 @@ Json::Value SotaUptaneClient::AssembleManifest() {
   std::string correlation_id;
   bool has_results = storage->loadDeviceInstallationResult(&dev_result, &raw_report, &correlation_id);
   if (has_results) {
+    if (!(dev_result.isSuccess() || dev_result.needCompletion())) {
+      director_repo.dropTargets(*storage);  // fix for OTA-2587, listen to backend again after end of install
+    }
+
     installation_report["result"] = dev_result.toJson();
     installation_report["raw_report"] = raw_report;
     installation_report["correlation_id"] = correlation_id;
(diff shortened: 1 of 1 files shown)


patrickvacek:
Looks promising! I'm glad none of the tests broke, but I'm still trying to make sure this won't change anything unexpected. Probably not. :)

kbushgit:
We already have a test which cowers case when we process each update(```Uptane::Target```) separately:

https://github.com/advancedtelematic/aktualizr/blob/f4591933a345795305d69842349c99d5e17dcc15/src/libaktualizr/primary/aktualizr_test.cc#L416

But this test covers the case only with two ECU and with the scenario when both ECU updated successfully, this scenario works fine since we dropping director target only in case of installation failing or completion is needed.

An undefined behavior can occurs in the next scenario:
- we have 3 or more updates(```Uptane::Target```)
- we process each update one by one (download/install)
- one ECU reports installation failing (important not last one ECU, the first or the second one)

Expected behavior is: the updating process must continue even if one ECU reports failing during installation process.

patrickvacek:
> But this test covers the case only with two ECU and with the scenario when both ECU updated successfully, this scenario works fine since we dropping director target only in case of installation failing or completion is needed.
> 
> An undefined behavior can occurs in the next scenario:
> 
>     * we have 3 or more updates(`Uptane::Target`)
> 
>     * we process each update one by one (download/install)
> 
>     * one ECU reports installation failing (important not last one ECU, the first or the second one)

Yep, I agree, we don't test what happens if one of them fails. I never had considered that a requirement before.

> Expected behavior is: the updating process must continue even if one ECU reports failing during installation process.

I thought for this use case we actually wanted to abort and possibly even rollback the updates that had succeeded. Are you suggesting we should be able to continue updating ECUs after one reports a failure? I think we might need a new test case that demonstrates the functionality you are expecting.

patrickvacek:
Just realized this fixes a bug I ran into (because of targets not getting dropped in all failure cases). @kbushgit can you rebase so we can merge this? I'd still like to see the test case we'd discussed here, but it doesn't have to be part of this PR.

PR #1642, opened 2020-04-15
Refact/ota 4814/install on secondary

mike-sul:
verified the given PR's aktualizr-primary against aktualizr-secondary of the given PR, the current master and 2020.5, both types of an update were verified, ostree and binary/file (qemu).

Also, a basic test to verify the backward compatibility is added to this test suite https://github.com/advancedtelematic/aktualizr/blob/7c6ad014c8ad5dc5a5b5ff456befb0ecfcb9de36/src/aktualizr_secondary/secondary_rpc_test.cc#L364.


src/aktualizr_secondary/update_agent.h
```diff
@@ -11,8 +11,11 @@ class UpdateAgent {
   virtual bool isTargetSupported(const Uptane::Target& target) const = 0;
   virtual bool getInstalledImageInfo(Uptane::InstalledImageInfo& installed_image_info) const = 0;
 
-  virtual bool download(const Uptane::Target& target, const std::string& data) = 0;
+  //  virtual bool download(const Uptane::Target& target, const std::string& data) = 0;
+  //  virtual data::ResultCode::Numeric receiveData(const Uptane::Target& target, const uint8_t* data, size_t size) = 0;
```
patrickvacek, line 15: Are there plans here or can these be removed?

src/aktualizr_secondary/aktualizr_secondary.h
```diff
@@ -3,62 +3,76 @@
...
+ protected:
+  AktualizrSecondary(AktualizrSecondaryConfig config, std::shared_ptr<INvStorage> storage);
+
+  // protected interface to be defined by child classes, i.e. a specific IP secondary type (e.g. OSTree, File, etc)
+  virtual bool getInstalledImageInfo(Uptane::InstalledImageInfo& installed_image_info) const = 0;
+  virtual bool isTargetSupported(const Uptane::Target& target) const = 0;
+  virtual data::ResultCode::Numeric installPendingTarget(const Uptane::Target& target) = 0;
+  virtual data::InstallationResult applyPendingInstall(const Uptane::Target& target) = 0;
+
+ protected:
```
patrickvacek, line 41: Redundant `protected`.

src/aktualizr_secondary/aktualizr_secondary.h
```diff
@@ -3,62 +3,76 @@
...
+  // protected interface to be used by child classes
+  Uptane::Target& pendingTarget() { return pending_target_; }
+  INvStorage& storage() { return *storage_; }
+  std::shared_ptr<INvStorage>& storagePtr() { return storage_; }
```
patrickvacek, line 45: Do we really need both? Just curious.

src/aktualizr_secondary/aktualizr_secondary_interface.h
```diff
@@ -13,7 +13,10 @@ class IAktualizrSecondary {
   virtual std::tuple<Uptane::EcuSerial, Uptane::HardwareIdentifier, PublicKey> getInfo() const = 0;
   virtual Uptane::Manifest getManifest() const = 0;
   virtual bool putMetadata(const Uptane::RawMetaPack& meta_pack) = 0;
+
   virtual bool sendFirmware(const std::string& firmware) = 0;
+
+  virtual data::ResultCode::Numeric sendFirmware(const uint8_t* data, size_t size) = 0;
```
patrickvacek, line 19: This class (`IAktualizrSecondary`) appears unused now. Can we get rid of it?

src/libaktualizr/uptane/secondaryinterface.h
```diff
@@ -26,15 +33,16 @@ class SecondaryInterface {
...
 
-  virtual bool sendFirmware(const std::string& data) = 0;
-  virtual data::ResultCode::Numeric install(const std::string& target_name) = 0;
+  virtual data::ResultCode::Numeric install(const Target& target) = 0;
 
   virtual ~SecondaryInterface() = default;
 
  public:
-  // make children non-copyable
+  // make children non-copyable and non-movable
```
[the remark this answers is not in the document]
mike-sul: I think I've just put all that 'non-copyable' and 'non-movable' stuff there because that's what is usually is required for a base class to avoid the slicing problem. But in this case, it's an abstract class, so it doesn't make sense to keep all that stuff in the interface declaration and let specific non-abstract children decide whether to make a class non-copyable/non-movable or not in their specific case. So I am just removing all that stuff.

src/virtual_secondary/managedsecondary.cc
```diff
@@ -139,44 +142,56 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
...
-    }
-  }
-  detected_attack = "";
-  const bool result = storeFirmware(expected_target_name, data);
-  return result;
-}
-
-data::ResultCode::Numeric ManagedSecondary::install(const std::string &target_name) {
-  (void)target_name;
+// bool ManagedSecondary::sendFirmware(const std::string &data) {
```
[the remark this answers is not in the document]
mike-sul: will remove it.

src/virtual_secondary/managedsecondary.cc
```diff
@@ -139,44 +142,56 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
...
+    // injects an error on the middle of the control flow that would have happened if an installation error had happened
+    // in case of the virtual or the ip-secondary or any other secondary, e.g. add a mock secondary that returns an
+    // error to sendFirmware/install request we might consider passing the installation description message from
+    // Secondary, not just bool and/or data::ResultCode::Numeric
+    return data::ResultCode::Numeric::kInstallFailed;
+  }
+
+  image_reader(target_name)->writeToFile(sconfig.firmware_path);
+  Utils::writeFile(sconfig.target_name_path, expected_target_name);
+  sync();
```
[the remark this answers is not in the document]
mike-sul: fsync requires a file descriptor that is not available in the given context. Maybe just remove it?

src/virtual_secondary/managedsecondary.cc
```diff
@@ -139,44 +142,56 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
...
+//    }
+//  }
+//  detected_attack = "";
+//  const bool result = storeFirmware(expected_target_name, data);
+//  return result;
+//}
+
+data::ResultCode::Numeric ManagedSecondary::install(const Uptane::Target &target_name) {
+  if (fiu_fail((std::string("secondary_install_") + getSerial().ToString()).c_str()) != 0) {
+    // consider changing this approach of the fault injection, since the current approach impacts the non-test code flow
```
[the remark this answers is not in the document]
mike-sul: :) the comment was not introduced by this PR, I just had to move it from the virtual secondary (https://github.com/advancedtelematic/aktualizr/blob/a4d411d42e118282e23e0fa6313b32f1d760ab97/src/virtual_secondary/virtualsecondary.cc#L73) due to the API changes (removing sendFirmware from the interface). 
Anyway, I can remove that comment.

src/aktualizr_secondary/main.cc
```diff
@@ -85,10 +88,18 @@ int main(int argc, char *argv[]) {
...
-    auto secondary = AktualizrSecondaryFactory::create(config);
-    SecondaryTcpServer tcp_server(secondary->getDispatcher(), config.network.primary_ip, config.network.primary_port,
+    if (config.pacman.type != PACKAGE_MANAGER_OSTREE) {
+      secondary = std::make_shared<AktualizrSecondaryFile>(config);
+    }
+#ifdef BUILD_OSTREE
+    else {
+      secondary = std::make_shared<AktualizrSecondaryOstree>(config);
+    }
+#endif  // BUILD_OSTREE
```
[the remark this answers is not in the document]
mike-sul: > I think we should probably report at error if you try to use the OSTree package manager but didn't enable OSTree.

How exactly to report it and to what entity? There is no means to report errors from Secondary to BE at the moment. If you mean just outputting log message then I can added it but it will be output anyway in the exception catch statement.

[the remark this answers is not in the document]
patrickvacek: The log is good enough for now.

src/aktualizr_secondary/update_agent_file.cc
```diff
@@ -61,3 +67,57 @@ data::InstallationResult FileUpdateAgent::applyPendingInstall(const Uptane::Targ
...
+                            std::ofstream::out | std::ofstream::binary | std::ofstream::app);
+
+  if (!target_file.good()) {
+    LOG_ERROR << "Failed to open a new target image file";
+    return data::ResultCode::Numeric::kDownloadFailed;
+  }
+
+  auto current_new_image_size = target_file.tellp();
+  if (-1 == current_new_image_size) {
+    LOG_ERROR << "Failed to obtain a size of the new target image that is being uploaded";
```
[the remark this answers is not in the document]
mike-sul: Yes, Primary uploads a file to Secondary.

src/aktualizr_secondary/update_agent_file.cc
```diff
@@ -61,3 +67,57 @@ data::InstallationResult FileUpdateAgent::applyPendingInstall(const Uptane::Targ
...
+    target_file.close();
+    return data::ResultCode::Numeric::kDownloadFailed;
+  }
+
+  target_file.close();
+
+  LOG_INFO << "Received and stored data of a new target image;"
+              " received in this request: "
+           << size << " total received so far: " << (current_new_image_size + written_data_size)
+           << " expected total: " << target.length();
```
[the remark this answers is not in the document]
mike-sul: ok

src/aktualizr_secondary/update_agent_file.cc
```diff
@@ -61,3 +67,57 @@ data::InstallationResult FileUpdateAgent::applyPendingInstall(const Uptane::Targ
...
+
+  new_target_hasher_->update(data, size);
+
+  return data::ResultCode::Numeric::kOk;
+}
+
+Uptane::Hash FileUpdateAgent::getTargetHash(const Uptane::Target& target) {
+  // TODO check target.hashes() size
```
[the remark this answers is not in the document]
mike-sul: I suppose, you mean this ticket https://saeljira.it.here.com/browse/OTA-4831? Yes, this is exactly how it's done on Primary ATM.

src/aktualizr_secondary/update_agent.h
mike-sul: It can be just removed.

src/aktualizr_secondary/update_agent_file.cc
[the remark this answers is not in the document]
patrickvacek: Ah yes, thanks. Can you add the ticket number to this TODO so there's a more explicit link?

src/aktualizr_secondary/aktualizr_secondary.h
mike-sul: Well, I did it for purpose. I grouped protected methods into two groups, the first one are pure virtual protected methods (kind of protected interface that must be implemented by a child class) and the second group is methods that are defined in this class and can be used by children. That's the style I was taught, i.g. group methods and fields and explicitly specify an access modifier for each group.  
Anyway, it's not a big deal, I can remove it and just leave a comment and additional whitespace between the groups.

src/aktualizr_secondary/aktualizr_secondary.h
mike-sul: Not mandatory, just find it a bit cleaner way to access storage from child classes.

src/aktualizr_secondary/aktualizr_secondary_interface.h
mike-sul: yes, I thought, I removed it, it must be result of incorrect rebasing with master.

src/libaktualizr/crypto/crypto.h
```diff
@@ -58,10 +58,50 @@ class PublicKey {
...
+ * defeat infinite download attacks.
+ */
+class Hash {
+ public:
+  // order corresponds algorithm priority
+  enum class Type { kSha256, kSha512, kUnknownAlgorithm };
+
+  static Hash generate(Type type, const std::string &data);
+  Hash(const std::string &type, const std::string &hash);
+  Hash(Type type, const std::string &hash);
```
eu-smirnov, line 73: Does it make sense to delete the default constructor? Otherwise, it will initialize `type_` to `kSha256`, which is probably not what you want.
I'm also not sure what is the purpose of `kUnknownAlgorithm` objects.

src/libaktualizr/primary/aktualizr.h
```diff
@@ -215,6 +215,16 @@ class Aktualizr {
...
 
+  /**
+   * Return a CA certificate, a client certificates a private key along with a Treehub URL,
+   * so an user (e.g. some Secondary) can setup a TLS connection with Treehub and download
+   * ostree repository revision objects.
+   * The aforementioned artifacts are returned in the form of a std::string containing gzip archive
+   *
+   * @return a string containing an archive with TLS certificates/credentials for a connection to Treehub
+   */
+  std::string GetTreehubTlsCreds() const { return uptane_client_->secondaryTreehubCredentials(); }
```
eu-smirnov, line 226: It is a bit confusing why a generic primary's API call refers to "secondary" credentials internally. We assume that they are the same for primary and secondary, right?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1286,28 +1286,7 @@ std::future<data::ResultCode::Numeric> SotaUptaneClient::sendFirmwareAsync(Uptan
...
-      send_firmware_result = secondary.sendFirmware(data_to_send);
-    }
-
-    data::ResultCode::Numeric result =
-        send_firmware_result ? data::ResultCode::Numeric::kOk : data::ResultCode::Numeric::kInstallFailed;
-
-    if (send_firmware_result) {
-      result = secondary.install(target.filename());
-    }
+    data::ResultCode::Numeric result = secondary.install(target);
```
eu-smirnov, line 1289: Removing some code from `sotauptaneclient.cc` is nice!

src/libaktualizr/crypto/crypto.h
eu-smirnov: Sorry, I just noticed that you've actually simply moved this code.

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -548,8 +548,8 @@ class SecondaryInterfaceMock : public Uptane::SecondaryInterface {
   int32_t getRootVersion(bool director) const override { return getRootVersionMock(director); }
 
   bool putRoot(const std::string &, bool) override { return true; }
-  bool sendFirmware(const std::string &) override { return true; }
-  virtual data::ResultCode::Numeric install(const std::string &) override { return data::ResultCode::Numeric::kOk; }
+  // bool sendFirmware(const std::string &) override { return true; }
```
eu-smirnov, line 551: any reason to keep this?

src/aktualizr_secondary/aktualizr_secondary_test.cc
```diff
@@ -192,16 +272,16 @@ class SecondaryTestNegative : public ::testing::Test,
...
+//  EXPECT_CALL(_update_agent, download).Times(0);
+//  EXPECT_CALL(_update_agent, install).Times(0);
 
-  EXPECT_FALSE(_secondary->sendFirmware("firmware"));
+//  EXPECT_FALSE(secondary_->sendFirmware("firmware"));
 
-  EXPECT_NE(_secondary->install("target"), data::ResultCode::Numeric::kOk);
-}
+//  EXPECT_NE(secondary_->install("target"), data::ResultCode::Numeric::kOk);
+//}
```
patrickvacek, line 284: More commented out stuff... is this still WIP?

src/aktualizr_secondary/aktualizr_secondary_test.cc
```diff
@@ -216,92 +296,104 @@ INSTANTIATE_TEST_SUITE_P(SecondaryTestMalformedMetadata, SecondaryTestNegative,
...
 
-  EXPECT_TRUE(_secondary->putMetadata(_uptane_repo.getCurrentMetadata()));
-  auto image_data = getImageData();
-  image_data.append("\n");
-  EXPECT_FALSE(_secondary->sendFirmware(image_data));
-  EXPECT_NE(_secondary->install(_default_target), data::ResultCode::Numeric::kOk);
+  EXPECT_TRUE(secondary_->putMetadata(uptane_repo_.getCurrentMetadata()));
+
+  EXPECT_EQ(sendImageFile(bigger_target_), data::ResultCode::Numeric::kOk);
+  EXPECT_NE(secondary_->install(), data::ResultCode::Numeric::kOk);
```
patrickvacek, line 386: These tests are checking that the Secondary rejects an image that is smaller or larger than what's stated in the metadata, right?

src/aktualizr_secondary/secondary_rpc_test.cc
```diff
@@ -61,78 +211,190 @@ bool operator==(const Uptane::RawMetaPack& lhs, const Uptane::RawMetaPack& rhs)
...
-  secondary_server.stop();
-  secondary_server_thread.join();
+  EXPECT_EQ(installOstreeRev(), data::ResultCode::Numeric::kOk);
+  EXPECT_EQ(tls_creds_.hash(), secondary_.getReceivedTlsCredsHash());
 }
 
+INSTANTIATE_TEST_SUITE_P(
+    SecondaryRpcTestCases, SecondaryRpcTest,
+    ::testing::Values(std::make_pair(1024 * 3 + 1, true /* new API/messages */),
+                      std::make_pair(1024 * 3 + 1, false /* old messages/sendFirmware, fallback testing */),
```
patrickvacek, line 364: What do you mean by "fallback testing" here? I'm not quite sure what's being covered here. I think this file could use just a bit more documentation about what all is actually being tested here (and what isn't); it's a bit hard to follow.

src/aktualizr_secondary/update_agent_file.cc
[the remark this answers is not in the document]
mike-sul: done

src/aktualizr_secondary/update_agent.h
mike-sul: done

src/libaktualizr-isotp/isotpsecondary.cc
```diff
@@ -137,6 +137,12 @@ bool IsoTpSecondary::putMetadata(const RawMetaPack& meta_pack) {
   return conn.Send(out);
 }
 
+data::ResultCode::Numeric install(const Target& target) {
+  (void)target;
+  // TODO: implement an image sending data by using ImageReader and IsoTpSecondary::sendFirmware
+  return data::ResultCode::Numeric::kOk;
+}
+
```
patrickvacek, line 145: Historically this was just reusing the default string sending, right? How hard is it to just replicate that here? I know we don't have a good way to test that at the moment.

patrickvacek, line 145: Historically this was just reusing the default string sending, right? How hard is it to just replicate that here? I know we don't have a good way to test that at the moment.

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -120,34 +134,77 @@ bool IpUptaneSecondary::putMetadata(const RawMetaPack& meta_pack) {
...
-  LOG_INFO << "Invoking an installation of the target on the secondary: " << target_name;
+data::ResultCode::Numeric IpUptaneSecondary::install(const Uptane::Target& target) {
+  std::lock_guard<std::mutex> l(install_mutex);
+  auto install_result = install_v2(target);
+  if (install_result == data::ResultCode::Numeric::kUnknown) {
+    // fallback to the previous installation version
+    LOG_ERROR << "Failed to install " << target.filename() << " on Secondary " << getSerial()
+              << "\nTry to install by applying the previous version of the installation procedure";
+
+    install_result = install_v1(target);
```
patrickvacek, line 178: Can we find a better way to determine which installation method to use? I'd prefer if we could check once and then store the result (not in the database, just in the class). Otherwise this feels like a lot of wasted installation attempts. I know that will probably get covered by the MQTT work, so we might be able to punt on this for now.

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -163,37 +220,136 @@ data::ResultCode::Numeric IpUptaneSecondary::install(const std::string& target_n
   return static_cast<data::ResultCode::Numeric>(r->result);
 }
 
-Manifest IpUptaneSecondary::getManifest() const {
-  LOG_DEBUG << "Getting the manifest from secondary with serial " << getSerial();
+data::ResultCode::Numeric IpUptaneSecondary::install_v2(const Uptane::Target& target) {
+  data::ResultCode::Numeric data_delivery_result = data::ResultCode::Numeric::kDownloadFailed;
+
+  LOG_INFO << "Invoking an image delivery of the target  " << target.filename() << " to/on the Secondary "
```
patrickvacek, line 226: Maybe something like "Instructing Secondary <xyz> to receive target <abc>" would be clearer?

Also, don't these image receiving steps belong in the download phase, not installation? That might be out of scope here, but I'm trying to remember why we handle this here at all.

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -163,37 +220,136 @@ data::ResultCode::Numeric IpUptaneSecondary::install(const std::string& target_n
...
+    data_delivery_result = downloadOstreeRev(target);
+  } else {
+    data_delivery_result = uploadFirmware(target);
+  }
+
+  if (data::ResultCode::Numeric::kOk != data_delivery_result) {
+    return data_delivery_result;
+  }
+
+  LOG_INFO << "Invoking an installation of the target " << target.filename() << " on the Secondary " << getSerial();
```
patrickvacek, line 239: And similarly here "Instructing Secondary <xyz> to install target <abc>".

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -163,37 +220,136 @@ data::ResultCode::Numeric IpUptaneSecondary::install(const std::string& target_n
...
+    LOG_ERROR << "Failed to get response to sending firmware to Secondary";
+    return false;
+  }
+
+  auto r = resp->sendFirmwareResp();
+  return r->result == AKInstallationResult_success;
+}
+
+data::ResultCode::Numeric IpUptaneSecondary::downloadOstreeRev(const Uptane::Target& target) {
+  LOG_INFO << "Invoking donwload of the ostree target revision ( " << target.sha256Hash() << " ) on Secondary ( "
```
patrickvacek, line 263: Perhaps "Instructing Secondary <xyz> to download OSTree commit <hash>."

src/libaktualizr/crypto/crypto.h
patrickvacek: Why was it necessary to move this, though? It's weird to define `Uptane` namespace objects outside of the uptane subdirectory. I'd like to avoid that if possible. This might be a refactoring that we can pull out and merge immediately even if we can't merge the rest yet.

mike-sul:
> the main change in secondary iface is to remove sendFirmware step and to provide an image reader callback instead, that will be called internally in `install`?

Yes, the main change in the secondary interface is getting rid of sendFirmware() method.
The image reader callback is not exactly part of the interface (SecondaryInterface), this is rather an optional parameter of a specific secondary that might need it (e.g. the ostree secondary does not need it).

The whole point of this removal (sendFirmware) is to let a specific secondary decide what exactly to do in order to install a target and don't put any its specifics into a common interface. So, a concrete secondary needs to implement install(const Target&) method and how exactly it's done and what is an internal protocol/communication between primary and secondary ECUs is solely at its discretion. This is the next step after  "freeing" the IP Secondary from dependency on the secondary interface.

Before that the secondary interface reflected a specific use-case (file/binary update with full Uptnae verification) so any other implementation had a headache of adjusting its implementation into that interface and protocol.

The next logical step, IMHO, would be changing 'putMetadata(const RawMetaPack& meta_pack)' to 'putMetadata(const Target& target)'.

And the whole strategy here is to bring the update procedure on Primary and Secondary to a common denominator so the sotauptaneclient would perform the Uptane update on different entities (Primary ECU, different types of Virtual ECUs, different types of real ECUs) in a uniform way.

src/virtual_secondary/partialverificationsecondary.h
```diff
@@ -44,8 +44,8 @@ class PartialVerificationSecondary : public SecondaryInterface {
   int getRootVersion(bool director) const override;
   bool putRoot(const std::string& root, bool director) override;
 
-  bool sendFirmware(const std::string& data) override;
-  data::ResultCode::Numeric install(const std::string& target_name) override;
+  // bool sendFirmware(const std::string& data) override;
```
patrickvacek, line 47: ?

src/virtual_secondary/managedsecondary.h
```diff
@@ -57,8 +57,8 @@ class ManagedSecondary : public Uptane::SecondaryInterface {
   int getRootVersion(bool director) const override;
   bool putRoot(const std::string& root, bool director) override;
 
-  bool sendFirmware(const std::string& data) override;
-  data::ResultCode::Numeric install(const std::string& target_name) override;
+  // bool sendFirmware(const std::string& data) override;
```
patrickvacek, line 60: ?

src/aktualizr_primary/secondary.cc
```diff
@@ -66,11 +67,14 @@ void initSecondaries(Aktualizr& aktualizr, const boost::filesystem::path& config
...
  public:
-  SecondaryWaiter(uint16_t wait_port, int timeout_s, Secondaries& secondaries)
+  SecondaryWaiter(uint16_t wait_port, int timeout_s, Secondaries& secondaries,
+                  ImageReaderProvider image_reader_provider, TlsCredsProvider treehub_cred_provider)
       : endpoint_{boost::asio::ip::tcp::v4(), wait_port},
         timeout_{static_cast<boost::posix_time::seconds>(timeout_s)},
         timer_{io_context_},
-        connected_secondaries_(secondaries) {}
+        connected_secondaries_{secondaries},
+        image_reader__provider_{std::move(image_reader_provider)},
```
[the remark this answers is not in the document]
mike-sul: yep

src/virtual_secondary/managedsecondary.cc
```diff
@@ -139,44 +142,13 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
...
-          boost::algorithm::to_lower_copy(it->HashString())) {
-        detected_attack = "wrong_hash";
-        return false;
-      }
-    }
-  }
-  detected_attack = "";
-  const bool result = storeFirmware(expected_target_name, data);
-  return result;
-}
```
patrickvacek, line 176: Why get rid of the hash check?

src/libaktualizr/primary/aktualizr.h
mike-sul: Yes, the method name is misleading, I am renaming it.

src/libaktualizr/uptane/uptane_test.cc
mike-sul: no

src/aktualizr_secondary/aktualizr_secondary_test.cc
mike-sul: no, I am uncommenting it.

src/aktualizr_secondary/aktualizr_secondary_test.cc
mike-sul: yes

src/aktualizr_secondary/secondary_rpc_test.cc
mike-sul: The second instantiation (false) creates an instance of secondary mock that emulates the "old" Secondary, the one that can handle current RPC version between Primary and Secondary (specifically this sendFirmware request/response.) I'll add a comment into a test.

src/libaktualizr-isotp/isotpsecondary.cc
mike-sul: > Historically this was just reusing the default string sending, right? 
Yes, it's in IsoTpSecondary::sendFirmware. 

> How hard is it to just replicate that here?

It's easy. The problem is that nothing instantiates IsoTpSecondary nor tests it, the question is what the point in the functionality that is never triggered.

src/libaktualizr-posix/ipuptanesecondary.cc
mike-sul: OK, although IMHO, 'invoke' better reflects what actually happens than 'instruct' as the later implies that Secondary still has freedom not to execute the instruction :)

src/virtual_secondary/partialverificationsecondary.h
mike-sul: removed

src/virtual_secondary/managedsecondary.h
mike-sul: removed

src/virtual_secondary/managedsecondary.cc
mike-sul: I thought we discussed and agreed on it in the chat.

src/libaktualizr-posix/ipuptanesecondary.cc
patrickvacek: "Invoke" is a slightly more accurate verb but it makes the sentence structure somewhat more complicated, and I think the gain in simplicity is worth it.

src/virtual_secondary/partialverificationsecondary.h
mike-sul: > Why was it necessary to move this, though?

Because MultiPartHasher depends on Uptane::Hash now and I didn't want to make crypto.h depend on uptane/tuf.h since the later depends on crypto.h so we get a circle dependency here which can be overcomed. But, I think a notion of hash is not Uptane specific stuff, so I thought we can move Uptane::Hash just to Hash?

mike-sul:
> It was so that you would immediately know if you changed something that could break the compatibility between them.

First of all, that what tests are for, secondary the previous design could not help with detecting changes in the RPC messages itself. The bigger changes, e.g. RPC message format is changed   will be detected by the given PR's design. too as well as any changes in the secondary interface will require changes in the secondary implementation on Primary.

src/virtual_secondary/managedsecondary.cc
patrickvacek: Ah yes, true. Yeah, you're right, I think it's pretty redundant. I was trying to think if it added any benefit for testing purposes (which was the original goal of the VirtualSecondary), but now that we have more advanced Secondaries and Secondary test support, it's probably not necessary.

src/virtual_secondary/partialverificationsecondary.h
patrickvacek: > But, I think a notion of hash is not Uptane specific stuff, so I thought we can move Uptane::Hash just to Hash?

Agreed, the hashing logic should be removed to either the crypto section or its own new file.

src/libaktualizr-posix/ipuptanesecondary.cc
mike-sul: > Can we find a better way to determine which installation method to use?

Yes, as we discussed it, it requires the proper implementation of the discovery mechanism. Once Secondary is "discovered" it could inform Primary about an API version it supports. There are lots of small items here but in general, IMHO, it does not make sense to do it until we switch to the MQTT based-implementation. What probably can be done now is to add addition RPC request that Primary can perform at the secondary initialization phase and it this requests fails then Primary assumes that the given Secondary is the "old" one if it is successful than it gets the Secondary supported version from the response to that additional RPC request.

src/libaktualizr-posix/ipuptanesecondary.cc
mike-sul: ok

src/libaktualizr-posix/ipuptanesecondary.cc
mike-sul: ok

src/aktualizr_secondary/main.cc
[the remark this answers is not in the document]
mike-sul: done

src/libaktualizr/crypto/crypto.h
eu-smirnov: But there is also nothing Uptane specific here, it's just a generic crypto helper. I'd probably move it under the `Crypto` namespace, which can replace the `Crypto` class below.

src/virtual_secondary/partialverificationsecondary.h
mike-sul: So, can I get rid of this Uptane namespace from the Hash declaration? 

>  the hashing logic should be removed to either the crypto section or its own new file.
What is the crypto section?

patrickvacek: > So, can I get rid of this Uptane namespace from the Hash declaration?

Yeah, I don't think it needs to be in the Uptane namespace.

> What is the crypto section?

I just meant "crypto section of the code". I should've just said "file". :)

eu-smirnov:
I must admit I am not so sure about merging sendFirmware and install into one call.

Those are the separate steps in the Uptane spec:
5.4.2.7. Send images to Secondaries
and
5.4.3.x Installing images on Primary or Secondary ECUs

And I see at least one good reason for that: if primary doesn't have enough space to store the secondary image it can utilize secondary storage via the sendFirmware call, while still verifying the hash. If hash verification by the primary fails, it will not proceed to the installation step.

src/libaktualizr/crypto/crypto.h
mike-sul: So, I just move the Hash out of the Uptane namespace and leave it in crypto.h/cc?

src/libaktualizr-isotp/isotpsecondary.cc
mike-sul: just implemented it.

mike-sul: just implemented it.

src/virtual_secondary/partialverificationsecondary.h
mike-sul: So, I am moving Hash out of the Uptane namespace and leave its implementation in crypto.h/cc files?

mike-sul:
> I must admit I am not so sure about merging sendFirmware and install into one call.
> 
> Those are the separate steps in the Uptane spec:
> 5.4.2.7. Send images to Secondaries
> and
> 5.4.3.x Installing images on Primary or Secondary ECUs
> 
> And I see at least one good reason for that: if primary doesn't have enough space to store the secondary image it can utilize secondary storage via the sendFirmware call, while still verifying the hash. If hash verification by the primary fails, it will not proceed to the installation step.

That's not exactly what happens here,

src/libaktualizr/crypto/crypto.h
patrickvacek: > So, I just move the Hash out of the Uptane namespace and leave it in crypto.h/cc?

That sounds reasonable to me.

patrickvacek:
Two more questions:
1. How does this impact downloading large objects on the Secondary and storing them in memory? I know this gets rid of the std::string, though.
2. Have you tested manually upgrading a Secondary from the old version to the new version with qemu or whatever?

mike-sul:
> Two more questions:
> 
> 1. How does this impact downloading large objects on the Secondary and storing them in memory? I know this gets rid of the std::string, though.

It improves the uploading of large objects from Primary to Secondary as it doesn't store overall image data in RAM in general and doesn't pass them via function call stack (passing a string with >1GB size as a parameter is not exactly what we wanna do). Image data are transferred from Primary to Secondary by pieces.

> 2. Have you tested manually upgrading a Secondary from the old version to the new version with qemu or whatever?

Yes, details are in my first comment to this PR.

src/libaktualizr-posix/asn1/asn1_message.cc
```diff
@@ -77,6 +82,7 @@ Asn1Message::Ptr Asn1Rpc(const Asn1Message::Ptr& tx, int con_fd) {
   Asn1Message::Ptr msg = Asn1Message::FromRaw(&m);
 
   if (res.code != RC_OK) {
+    msg = Asn1Message::FromRaw(&m);
```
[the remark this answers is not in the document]
patrickvacek: I removed this; it seemed unnecessary.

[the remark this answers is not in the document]
mike-sul: Yes, that's correct, that's leftover from some previous version that wrongly assumed that Asn1Message::FromRaw() cannot handle `m` equal to null.

lbonn:
@patrickvacek Should we still test the migration scenario (on qemu for example) before we merge this?

patrickvacek:
> Should we still test the migration scenario (on qemu for example) before we merge this?

Yeah, I'd like to reproduce that (and see what happens if you do it the wrong way). We can discuss more at the meeting later today.

patrickvacek:
Sorry for all the rebasing; it's now based on https://github.com/advancedtelematic/aktualizr/pull/1650 and when that is merged this should be rebased again.

patrickvacek:
Well, it's only taken two months, but this PR is finally ready for review. I think all of it has already been reviewed at least once at some point, and I feel comfortable merging it after having done substantial testing on it. However, I'd like to provide one more chance for others to take a look and provide their input. @mike-sul I'm also curious if you see any causes for concern.

mike-sul:
> I'm also curious if you see any causes for concern

@patrickvacek  At first glance, I don't see any reasons for concern, will look into it closer later today.
In any case, I suppose, that it makes sense to communicate to the customers about the correct procedure of deployment/switching to the new version. (at first, deploy on Primary and then on Secondaries, iirc).

PR #1644, opened 2020-04-15
Add ability to provide custom hardware information which will be visible in UI

eu-smirnov:


eu-smirnov:
@kbushgit That's not a bad idea. But I've added this hwinfo file parameter mostly for testing purposes. Current users of this API will gather the needed information in the libaktualizr wrapper, so it should be easier for them to just pass the JSON object via this API.
I'm also not sure why do we even have putManifest call in the SendDeviceData, since we already do it in CheckUpdates and even have a separate API call for that.

kbushgit:
I mean that we also have an invocation of putManifestSimple, but you are right it can be done using sendManifest call, so, it is ok for now, I am just not sure that in the future, another custom information won't be needed during invocation of this method.

PR #1647, opened 2020-04-17
Updates to the versions of dependencies.

patrickvacek:
Mostly just bringing up to speed, although a few are still a bit vaguely defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,7 +86,7 @@ if(EXISTS /usr/local/opt/libarchive)
   list(APPEND CMAKE_PREFIX_PATH /usr/local/opt/libarchive)
 endif()
 
-find_package(Boost 1.57.0 COMPONENTS ${BOOST_COMPONENTS} REQUIRED)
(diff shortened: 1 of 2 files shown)


tkfu:
I'm a bit confused, but it's probably just that I don't understand cmake very well--why does the cmake `find_package` look for boost 1.58 while the boost version in `thirdparty.spdx` is 1.72?

patrickvacek:
> why does the cmake `find_package` look for boost 1.58 while the boost version in `thirdparty.spdx` is 1.72?

Yeah, I know, it's a bit silly. I'm trying to indicate that the minimum required version is 1.58, but the default in Yocto now is 1.72. I just realized that zeus still uses 1.71, though. The version specified in cmake is not a strict requirement, either; see https://cmake.org/cmake/help/v3.8/command/find_package.html. So I'm open to debate on what the right versions to specify are.

patrickvacek:
I decided it's best to be consistent. I've changed all the references to boost versions to 1.58 since that's what the REAMDE claimed was our minimum required version.

patrickvacek:
Finally passing CI now, but I blew away the approval by pushing.

PR #1651, opened 2020-04-22
Refactor/ota 4814/restore send firmware

patrickvacek:
Note that this is based on https://github.com/advancedtelematic/aktualizr/pull/1642. We are treating that as a feature branch for the moment and this should be merged into that. I originally tried to do this by skipping the original removal of sendFirmware and rebasing, but it was too much of a mess. This was much easier and ultimately rather straightforward.

### DIFF
--- a/src/libaktualizr/primary/sotauptaneclient.cc
+++ b/src/libaktualizr/primary/sotauptaneclient.cc
@@ -1292,7 +1292,11 @@ std::future<data::ResultCode::Numeric> SotaUptaneClient::sendFirmwareAsync(Uptan
     sendEvent<event::InstallStarted>(secondary.getSerial());
     report_queue->enqueue(std_::make_unique<EcuInstallationStartedReport>(secondary.getSerial(), correlation_id));
 
-    data::ResultCode::Numeric result = secondary.install(target);
+    data::ResultCode::Numeric result = secondary.sendFirmware(target);
+
+    if (result == data::ResultCode::Numeric::kOk) {
+      result = secondary.install(target);
+    }
 
     if (result == data::ResultCode::Numeric::kNeedCompletion) {
       report_queue->enqueue(std_::make_unique<EcuInstallationAppliedReport>(secondary.getSerial(), correlation_id));
--- a/src/virtual_secondary/managedsecondary.cc
+++ b/src/virtual_secondary/managedsecondary.cc
@@ -142,7 +142,12 @@ bool ManagedSecondary::putRoot(const std::string &root, const bool director) {
   return true;
 }
 
-data::ResultCode::Numeric ManagedSecondary::install(const Uptane::Target &target_name) {
+data::ResultCode::Numeric ManagedSecondary::sendFirmware(const Uptane::Target &target) {
+  (void)target;
+  return data::ResultCode::Numeric::kOk;
+}
+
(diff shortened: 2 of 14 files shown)


src/virtual_secondary/managedsecondary.cc
```diff
@@ -153,7 +158,7 @@ data::ResultCode::Numeric ManagedSecondary::install(const Uptane::Target &target
     return data::ResultCode::Numeric::kInstallFailed;
   }
 
-  image_reader(target_name)->writeToFile(sconfig.firmware_path);
+  image_reader(target)->writeToFile(sconfig.firmware_path);
   Utils::writeFile(sconfig.target_name_path, expected_target_name);
   sync();
```
mike-sul, line 163: I think I've forgotten to remove this `sync` as we agreed in the scope of my last PR discussion (or it was restored somehow).

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1292,7 +1292,11 @@ std::future<data::ResultCode::Numeric> SotaUptaneClient::sendFirmwareAsync(Uptan
     sendEvent<event::InstallStarted>(secondary.getSerial());
     report_queue->enqueue(std_::make_unique<EcuInstallationStartedReport>(secondary.getSerial(), correlation_id));
 
-    data::ResultCode::Numeric result = secondary.install(target);
+    data::ResultCode::Numeric result = secondary.sendFirmware(target);
```
mike-sul, line 1295: Not sure if I see any benefits in having two calls (sendFirmware and install) instead of just one `install` call in this context, although it doesn't cause any issue.

src/virtual_secondary/managedsecondary.cc
patrickvacek: Probably my bad restoring during multiple rounds of rebasing. It's gone now.

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: After some discussion, we decided it was best to split it, because we ultimately want to move the `sendFirmware` part to the download phase on the Primary. This is just the first step to enable that.

mike-sul: >  we ultimately want to move the sendFirmware part to the download phase on the Primary.

That's exactly what I'd considered too and that's why I suggested adding download() method to the secondary interface in that confluence document. In this case, a secondary implementer can implement streaming to Secondary ECU, provided that SotaUptaneClient calls ecu/secondary->download() at a download phase so the implementer just calls corresponding send data RPC call in its implementation, and sendFirmware is not required in the interface it is just needed at the RPC API level. But, it's just my IMHO :), and I don't see any issue with having sendFirmware in the interface neither.

patrickvacek: Thanks for your input! Sounds like your download concept is pretty similar to my intention for the sendFirmware once we move it. It's still up to the implementer to decide what to do in the download/sendFirmware call and what to do in the install call, and I imagine some will skip the sendFirmware part, especially if the Secondary doesn't have space to keep the update on disk. But I want to have the option for it if someone wants or needs it.

mike-sul: :+1:

PR #1653, opened 2020-04-22
Refuse to download OSTree targets with the fake/binary package manager.

patrickvacek:
Previously the download would succeed due to the empty length but the installation would fail. Better to fail as early as possible.

### DIFF
--- a/src/libaktualizr/package_manager/packagemanagerfake.cc
+++ b/src/libaktualizr/package_manager/packagemanagerfake.cc
@@ -110,5 +110,10 @@ bool PackageManagerFake::fetchTarget(const Uptane::Target& target, Uptane::Fetch
     return false;
   }
 
+  if (target.IsOstree()) {
+    LOG_ERROR << "Cannot download OSTree target " << target.filename() << " with the fake package manager!";
+    return false;
+  }
+
   return PackageManagerInterface::fetchTarget(target, fetcher, keys, progress_cb, token);
 }
--- a/src/libaktualizr/uptane/tuf.cc
+++ b/src/libaktualizr/uptane/tuf.cc
@@ -127,6 +127,7 @@ Target::Target(std::string filename, const Json::Value &content) : filename_(std
   std::sort(hashes_.begin(), hashes_.end(), [](const Hash &l, const Hash &r) { return l.type() < r.type(); });
 }
 
+// Internal use only.
 Target::Target(std::string filename, EcuMap ecus, std::vector<Hash> hashes, uint64_t length, std::string correlation_id)
     : filename_(std::move(filename)),
       ecus_(std::move(ecus)),
@@ -135,6 +136,7 @@ Target::Target(std::string filename, EcuMap ecus, std::vector<Hash> hashes, uint
       correlation_id_(std::move(correlation_id)) {
   // sort hashes so that higher priority hash algorithm goes first
   std::sort(hashes_.begin(), hashes_.end(), [](const Hash &l, const Hash &r) { return l.type() < r.type(); });
+  type_ = "UNKNOWN";
 }
 
 Target Target::Unknown() {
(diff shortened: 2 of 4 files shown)


src/libaktualizr/package_manager/packagemanagerfake.cc
```diff
@@ -110,5 +110,10 @@ bool PackageManagerFake::fetchTarget(const Uptane::Target& target, Uptane::Fetch
     return false;
   }
 
+  if (target.IsOstree()) {
```
kbushgit, line 113: Don't we do a similar check here https://github.com/advancedtelematic/aktualizr/blob/1d9509428441396674ac5b78ec2513de63f99edc/src/libaktualizr/primary/sotauptaneclient.cc#L651 ?

patrickvacek: Yes, but these two checks do different things. The first (the one you've pointed out) makes sure that we don't uselessly download OSTree commits for Secondaries since we can't do anything with them. The second (the one I've added) is specific to the fake/binary package manager and makes sure that we don't attempt to fetch OSTree commits (for the Primary) when we aren't using the OSTree package manager.

mike-sul: > makes sure that we don't attempt to fetch OSTree commits (for the Primary) when we aren't using the OSTree package manager.

I suppose the backend shouldn't allow such an update in the first place, i.e. it should not allow installation of targets type of which is not supported by the backend. 
Also, it could be more generic verification in SotaUptaneClient, something like if (!package_manager_->isTargetSupported()) {} or in the future when the logical ECU is in place then a container/array could expose an API allowing to search for an instance that supports a given target.

patrickvacek: > I suppose the backend shouldn't allow such an update in the first place, i.e. it should not allow installation of targets type of which is not supported by a device.

Yes, but I don't think the backend processes that information. But even if it did, we still have to check that on the device.

> Also, it could be more generic verification in SotaUptaneClient

Very true, but I don't think we're there yet. This is merely an incremental improvement.

eu-smirnov, line 113: I would probably prefer something like `if (target.type_ != this->type_)`, but that should be also good for now.

src/libaktualizr/uptane/tuf.cc
```diff
@@ -135,6 +135,7 @@ Target::Target(std::string filename, EcuMap ecus, std::vector<Hash> hashes, uint
       correlation_id_(std::move(correlation_id)) {
   // sort hashes so that higher priority hash algorithm goes first
   std::sort(hashes_.begin(), hashes_.end(), [](const Hash &l, const Hash &r) { return l.type() < r.type(); });
+  type_ = "binary";
```
eu-smirnov, line 138: Maybe, it is better to set to "unknown"? I also think that "BINARY" is capitalized, though you can never be sure. Should we use `.toLower()`?

patrickvacek: > Maybe, it is better to set to "unknown"?

Agreed, done.

> I also think that "BINARY" is capitalized, though you can never be sure. Should we use `.toLower()`?

We don't appear to use "BINARY"; the only comparison we use for Target types is with "OSTREE". I think for now the comparison is fine as is.

src/libaktualizr/package_manager/packagemanagerfake.cc
patrickvacek: Me too, but for now we don't expose package manager types... unless I've forgotten something?

eu-smirnov: No, we don't. I just was thinking that we might want to do something like this in the future,

mike-sul: PackageManagerInterface::name() is kind of a package manager type :)

eu-smirnov: @mike-sul True, but it still doesn't have 1:1 match with update type.

mike-sul: Yeah, I know, maybe adding something like package_manager->isTargetSupported(const Target&)  (or logicalEcu->isTargetSupported()) so the pack manager implementation decides whether it can handle the given target or not would be the best here.

PR #1655, opened 2020-04-23
build-agl: Bump AGL version to icefish 9.0.1.

patrickvacek:
This includes some fixes that mean we need fewer additional configuration steps.


tkfu:
Nice! Happy to see that got merged upstream.

PR #1658, opened 2020-04-24
Some exception refactoring

lbonn:
Tests do not pass yet and it will probably need some changes in tuf-test-vectors as well.

### DIFF
--- a/src/libaktualizr/primary/initializer.cc
+++ b/src/libaktualizr/primary/initializer.cc
@@ -73,8 +73,15 @@ bool Initializer::initEcuSerials() {
 
 void Initializer::resetEcuSerials() { storage_->clearEcuSerials(); }
 
-// Postcondition: (public, private) is in the storage. It should not be stored until Secondaries are provisioned
-bool Initializer::initPrimaryEcuKeys() { return keys_.generateUptaneKeyPair().size() != 0U; }
+// Postcondition: (public, private) is in the storage. It should not be stored until secondaries are provisioned
+bool Initializer::initPrimaryEcuKeys() {
+  try {
+    return keys_.generateUptaneKeyPair().size() != 0U;
+  } catch (const std::exception &e) {
+    LOG_ERROR << "Initializer error: " << e.what();
+  }
+  return false;
+}
 
 void Initializer::resetEcuKeys() { storage_->clearPrimaryKeys(); }
 
--- a/src/libaktualizr/primary/initializer.cc
+++ b/src/libaktualizr/primary/initializer.cc
@@ -77,7 +77,7 @@ void Initializer::resetEcuSerials() { storage_->clearEcuSerials(); }
 bool Initializer::initPrimaryEcuKeys() {
   try {
     return keys_.generateUptaneKeyPair().size() != 0U;
-  } catch (const std::exception &e) {
+  } catch (const std::exception& e) {
     LOG_ERROR << "Initializer error: " << e.what();
   }
   return false;
--- a/src/libaktualizr/primary/initializer.cc
+++ b/src/libaktualizr/primary/initializer.cc
@@ -11,11 +11,11 @@
 #include "storage/invstorage.h"
 
 // Postcondition: device_id is in the storage
-bool Initializer::initDeviceId() {
+void Initializer::initDeviceId() {
   // if device_id is already stored, just return
   std::string device_id;
   if (storage_->loadDeviceId(&device_id)) {
-    return true;
+    return;
   }
 
   // if device_id is specified in config, just use it, otherwise generate a  random one
@@ -26,25 +26,23 @@ bool Initializer::initDeviceId() {
     } else if (config_.mode == ProvisionMode::kDeviceCred) {
       device_id = keys_.getCN();
     } else {
-      LOG_ERROR << "Unknown provisioning method.";
-      return false;
+      throw Error("Unknown provisioning method");
     }
   }
 
   storage_->storeDeviceId(device_id);
-  return true;
 }
 
 void Initializer::resetDeviceId() { storage_->clearDeviceId(); }
 
 // Postcondition [(serial, hw_id)] is in the storage
-bool Initializer::initEcuSerials() {
+void Initializer::initEcuSerials() {
   EcuSerials ecu_serials;
 
   // TODO: the assumption now is that the set of connected ECUs doesn't change, but it might obviously
   //   not be the case. ECU discovery seems to be a big story and should be worked on accordingly.
   if (storage_->loadEcuSerials(&ecu_serials)) {
-    return true;
+    return;
   }
 
   std::string primary_ecu_serial_local = config_.primary_ecu_serial;
@@ -56,7 +54,7 @@ bool Initializer::initEcuSerials() {
   if (primary_ecu_hardware_id.empty()) {
     primary_ecu_hardware_id = Utils::getHostname();
     if (primary_ecu_hardware_id == "") {
-      return false;
+      throw Error("Could not get current host name, please configure an hardware ID explicitely");
     }
   }
 
@@ -68,19 +66,22 @@ bool Initializer::initEcuSerials() {
   }
 
   storage_->storeEcuSerials(ecu_serials);
-  return true;
 }
 
 void Initializer::resetEcuSerials() { storage_->clearEcuSerials(); }
 
 // Postcondition: (public, private) is in the storage. It should not be stored until secondaries are provisioned
-bool Initializer::initPrimaryEcuKeys() {
+void Initializer::initPrimaryEcuKeys() {
+  std::string key_pair;
   try {
-    return keys_.generateUptaneKeyPair().size() != 0U;
+    key_pair = keys_.generateUptaneKeyPair();
   } catch (const std::exception& e) {
-    LOG_ERROR << "Initializer error: " << e.what();
+    throw KeyGenerationError(e.what());
   }
-  return false;
+
+  if (key_pair.size() == 0U) {
+    throw KeyGenerationError("Unknow error");
+  }
 }
 
 void Initializer::resetEcuKeys() { storage_->clearPrimaryKeys(); }
@@ -91,14 +92,13 @@ bool Initializer::loadSetTlsCreds() {
 }
 
 // Postcondition: TLS credentials are in the storage
-InitRetCode Initializer::initTlsCreds() {
+void Initializer::initTlsCreds() {
   if (loadSetTlsCreds()) {
-    return InitRetCode::kOk;
+    return;
   }
 
   if (config_.mode != ProvisionMode::kSharedCred) {
-    LOG_ERROR << "Shared credentials expected but not found.";
-    return InitRetCode::kStorageFailure;
+    throw StorageError("Shared credentials expected but not found");
   }
 
   // Shared credential provision is required and possible => (automatically)
@@ -112,8 +112,7 @@ InitRetCode Initializer::initTlsCreds() {
   Json::Value data;
   std::string device_id;
   if (!storage_->loadDeviceId(&device_id)) {
-    LOG_ERROR << "Unable to load device_id during shared credential provisioning.";
-    return InitRetCode::kStorageFailure;
+    throw StorageError("Unable to load device_id during shared credential provisioning");
   }
   data["deviceId"] = device_id;
   data["ttl"] = config_.expiry_days;
@@ -122,10 +121,11 @@ InitRetCode Initializer::initTlsCreds() {
     Json::Value resp_code = response.getJson()["code"];
     if (resp_code.isString() && resp_code.asString() == "device_already_registered") {
       LOG_ERROR << "Device ID " << device_id << " is already registered.";
-      return InitRetCode::kOccupied;
+      throw ServerOccupied();
     }
-    LOG_ERROR << "Shared credential provisioning failed: " << response.http_status_code << " " << response.body;
-    return InitRetCode::kServerFailure;
+    const auto err = std::string("Shared credential provisioning failed: ") +
+                     std::to_string(response.http_status_code) + " " + response.body;
+    throw ServerError(err);
   }
 
   std::string pkey;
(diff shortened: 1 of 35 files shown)


lbonn:
@patrickvacek Just checking on uptane vectors

{director,image_repo}_root_rotation_version_{skip,unchanged} return "Uptane::SecurityException: Version in Root metadata doesn't match the expected value". I suppose it's better than unmet threshold? I'll probably change it to a more specific exception anyway.

patrickvacek:
> `{director,image_repo}_root_rotation_version_{skip,unchanged}` return "Uptane::SecurityException: Version in Root metadata doesn't match the expected value". I suppose it's better than unmet threshold? I'll probably change it to a more specific exception anyway.

"Version in Root metadata doesn't match the expected value" sounds correct to me. A lot of the errors expected by the vector tests are bogus because we were waiting on this refactoring before spending too much time getting everything just right. I think the "umet threshold" and a few others are common exceptions that really should be something more detailed.

patrickvacek:
> I'll probably change it to a more specific exception anyway.

Just saw the `RootVersionChainMismatch` that you added. That looks like an improvement!

lbonn:
> "Version in Root metadata doesn't match the expected value" sounds correct to me. A lot of the errors expected by the vector tests are bogus because we were waiting on this refactoring before spending too much time getting everything just right. I think the "umet threshold" and a few others are common exceptions that really should be something more detailed.

Thanks, that's what I thought. I've opened a PR with the changes, the text is now "One Root version is not a strict increment of the preceding." (to not re-use a plain SecurityException). I guess we can discuss names and descriptions there https://github.com/advancedtelematic/tuf-test-vectors/pull/71

lbonn:
I also realized that removing "plain throws" like I did here https://github.com/advancedtelematic/aktualizr/pull/1654 was a mistake. When done in a catch clause, it rethrows the caught exception.

patrickvacek:
> I also realized that removing "plain throws" like I did here #1654 was a mistake. When done in a catch clause, it rethrows the caught exception.

I don't think I understand the problem; that sounds like what I expected. What's wrong with rethrowing?

src/libaktualizr/uptane/imagerepository.cc
```diff
@@ -214,25 +209,24 @@ std::shared_ptr<Uptane::Targets> ImageRepository::verifyDelegation(const std::st
...
   return false;
 }
 
-bool ImageRepository::updateMeta(INvStorage& storage, const IMetadataFetcher& fetcher) {
+void ImageRepository::updateMeta(INvStorage& storage, const IMetadataFetcher& fetcher) {
   resetMeta();
 
   if (!updateRoot(storage, fetcher, RepositoryType::Image())) {
-    return false;
+    throw std::runtime_error("");
```
patrickvacek: Now I see. I think it makes sense to try to decide on some new meaningful exceptions to use here and in similar places. Did you have particular plans or ideas already?

lbonn:
> > I also realized that removing "plain throws" like I did here #1654 was a mistake. When done in a catch clause, it rethrows the caught exception.
> 
> I don't think I understand the problem; that sounds like what I expected. What's wrong with rethrowing?

Ah sorry it was unclear. Exactly, nothing wrong with rethrowing :).

```
try {
} catch (const std::exception& e) {
  throw;
}
```

is actually better than

```
try {
} catch (const std::exception& e) {
  throw e;
}
```

Because in the second case, we slice the original exception and loose the original context.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -796,12 +787,15 @@ result::UpdateCheck SotaUptaneClient::checkUpdates() {
...
-  if (!uptaneIteration(&updates, &ecus_count)) {
+  try {
+    uptaneIteration(&updates, &ecus_count);
+  } catch (const std::exception &e) {
     result = result::UpdateCheck({}, 0, result::UpdateStatus::kError, Json::nullValue, "Could not update metadata.");
     return result;
   }
 
   std::string director_targets;
+  // What if it fails?
```
patrickvacek, line 805: It should abort, right? What else is there to do?

src/libaktualizr/uptane/imagerepository.cc
```diff
@@ -267,9 +236,12 @@ bool ImageRepository::updateMeta(INvStorage& storage, const IMetadataFetcher& fe
     int local_version;
     std::string image_snapshot_stored;
     if (storage.loadNonRoot(&image_snapshot_stored, RepositoryType::Image(), Role::Snapshot())) {
-      if (verifySnapshot(image_snapshot_stored, true)) {
+      try {
+        verifySnapshot(image_snapshot_stored, true);
         fetch_snapshot = false;
         LOG_DEBUG << "Skipping Image repo Snapshot download; stored version is still current.";
+      } catch (const Uptane::Exception& e) {
+        LOG_ERROR << "Snapshot repo verification failed: " << e.what();
```
patrickvacek, line 244:
```suggestion
        LOG_ERROR << "Image repo Snapshot verification failed: " << e.what();
```

src/libaktualizr/uptane/imagerepository.cc
```diff
@@ -296,9 +264,12 @@ bool ImageRepository::updateMeta(INvStorage& storage, const IMetadataFetcher& fe
     int local_version = -1;
     std::string image_targets_stored;
     if (storage.loadNonRoot(&image_targets_stored, RepositoryType::Image(), Role::Targets())) {
-      if (verifyTargets(image_targets_stored, true)) {
+      try {
+        verifyTargets(image_targets_stored, true);
         fetch_targets = false;
         LOG_DEBUG << "Skipping Image repo Targets download; stored version is still current.";
+      } catch (const std::exception& e) {
+        LOG_ERROR << "Image repo verification failed: " << e.what();
```
patrickvacek, line 272:
```suggestion
        LOG_ERROR << "Image repo Root verification failed: " << e.what();
```

src/libaktualizr/uptane/directorrepository.cc
```diff
@@ -46,37 +40,33 @@ bool DirectorRepository::usePreviousTargets() const {
...
+void DirectorRepository::verifyTargets(const std::string& targets_raw) {
   try {
     // Verify the signature:
     latest_targets = Targets(RepositoryType::Director(), Role::Targets(), Utils::parseJSON(targets_raw),
                              std::make_shared<MetaWithKeys>(root));
     if (!usePreviousTargets()) {
       targets = latest_targets;
     }
-  } catch (const Uptane::Exception& e) {
+  } catch (const Uptane::Exception& e) {  // !
```
patrickvacek, line 51: Why the comment/concern?

src/libaktualizr/uptane/directorrepository.cc
```diff
@@ -32,12 +28,10 @@ bool DirectorRepository::targetsSanityCheck() {
       if (ecu_ids.find(ecu.first) == ecu_ids.end()) {
         ecu_ids.insert(ecu.first);
       } else {
-        last_exception = Uptane::InvalidMetadata(type.toString(), Role::Targets().ToString(), "Found repeated ECU ID.");
-        return false;
+        throw Uptane::InvalidMetadata(type.toString(), Role::TARGETS, "Found repeated ECU ID.");
```
patrickvacek, line 31: It would be ideal if we could list the repeated ID in question if that doesn't make the tuf-test-vector integration too challenging.

src/libaktualizr/uptane/tuf.h
```diff
@@ -23,14 +23,17 @@ class RepositoryType {
   enum class Type { kUnknown = -1, kImage = 0, kDirector = 1 };
 
  public:
+  static const std::string IMAGE;
+  static const std::string DIRECTOR;
+
```
patrickvacek, line 28: 👍👍👍👍👍👍 This should've been done long ago!

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -876,13 +883,14 @@ result::UpdateStatus SotaUptaneClient::checkUpdatesOffline(const std::vector<Upt
     const auto it = std::find_if(director_targets.cbegin(), director_targets.cend(), target_comp);
     if (it == director_targets.cend()) {
       LOG_ERROR << "No matching target in Director Targets metadata for " << target;
-      return result::UpdateStatus::kError;
+      throw Uptane::Exception(Uptane::RepositoryType::DIRECTOR, "No matching target in Director Targets metadata");
+      // return result::UpdateStatus::kError;
```
patrickvacek, line 887: ?

src/libaktualizr/uptane/imagerepository.cc
lbonn: It was changed in later commits. Maybe could have squashed some more stuff.

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: Yes, we can add some check then.

src/libaktualizr/uptane/directorrepository.cc
lbonn: This was just an help for me when I was going through the code. I thought I removed it, will do.

src/libaktualizr/uptane/directorrepository.cc
lbonn: Or in a LOG_ERROR?

lbonn:
> Already looks like a big improvement! What's the benefit of `last_exception` in SotaUptaneClient, though? And I'd like to make sure that @eu-smirnov is on board with this direction.

I tried to not change the external API that tried its best no to throw, so the last_exception is used to report the error reason as understood by tuf-test-vectors. There might be a better way to expose that through the Aktualizr API.

Basically my conclusion was that the way we constructed the API, asynchronous and sending events, using exceptions would be quite awkward and hard to handle (for example, we still want to report something to the backend even if an install fails because of a metadata check). For now I've chosen to reserve them for cases that are hard/impossible to recover from. Otherwise I'm afraid it would actually worsen the usability of the API.

src/libaktualizr/uptane/directorrepository.cc
patrickvacek: Yeah, that's probably good enough. :)

src/libaktualizr/uptane/imagerepository.cc
patrickvacek: No problem, that was my mistake then. I was trying to go through the commits one by one but crossreference the final version before leaving comments. I must've gotten confused somehow.

patrickvacek:
> I tried to not change the external API that tried its best no to throw, so the last_exception is used to report the error reason as understood by tuf-test-vectors. There might be a better way to expose that through the Aktualizr API.
> 
> Basically my conclusion was that the way we constructed the API, asynchronous and sending events, using exceptions would be quite awkward and hard to handle (for example, we still want to report something to the backend even if an install fails because of a metadata check). For now I've chosen to reserve them for cases that are hard/impossible to recover from. Otherwise I'm afraid it would actually worsen the usability of the API.
> 
> EDIT: so to summarize, the objective right now is to let roll exceptions as far as possible and catch them at a small number of points in API boundaries for error reporting.

Okay, that makes sense to me. Even if we want to throw exceptions through the API, we would need to do the work in this PR first to make sure we know what we're throwing. I'm open to discussion on what to do from here.

src/libaktualizr/primary/initializer.cc
```diff
@@ -73,8 +73,15 @@ bool Initializer::initEcuSerials() {
...
-bool Initializer::initPrimaryEcuKeys() { return keys_.generateUptaneKeyPair().size() != 0U; }
+// Postcondition: (public, private) is in the storage. It should not be stored until secondaries are provisioned
+bool Initializer::initPrimaryEcuKeys() {
+  try {
+    return keys_.generateUptaneKeyPair().size() != 0U;
+  } catch (const std::exception& e) {
+    LOG_ERROR << "Initializer error: " << e.what();
+  }
+  return false;
+}
```
eu-smirnov, line 84: I'm not sure that's the right place to catch it.
`initPrimaryEcuKeys` gets called only in Initializer ctor. I would definitely prefer if the Initializer ctor actually threw an exception if it failed to do something.

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -639,38 +619,45 @@ std::pair<bool, Uptane::Target> SotaUptaneClient::downloadImage(const Uptane::Ta
...
     }
-    if (!success) {
-      LOG_ERROR << "Download unsuccessful after " << tries << " attempts.";
-    }
-  } else {
-    // we emulate successfull download in case of the Secondary OSTree update
-    success = true;
+  } catch (const std::exception &e) {
+    LOG_ERROR << "Error downloading image: " << e.what();
+    last_exception = std::current_exception();
```
eu-smirnov, line 660: What prevents us from moving this try..catch block one level higher, to the `downloadImages` method and get rid of `success` return value?

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -683,16 +670,20 @@ std::pair<bool, Uptane::Target> SotaUptaneClient::downloadImage(const Uptane::Ta
...
-bool SotaUptaneClient::uptaneIteration(std::vector<Uptane::Target> *targets, unsigned int *ecus_count) {
-  if (!updateDirectorMeta()) {
-    LOG_ERROR << "Failed to update Director metadata: " << last_exception.what();
-    return false;
+void SotaUptaneClient::uptaneIteration(std::vector<Uptane::Target> *targets, unsigned int *ecus_count) {
+  try {
+    updateDirectorMeta();
+  } catch (const std::exception &e) {
+    LOG_ERROR << "Director metadata update failed: " << e.what();
+    throw;
```
eu-smirnov, line 678: I'm not sure moving `try..catch` from `updateDirectorMeta` and others to the upper-level function improves readability.

eu-smirnov:
Definitely looks like an improvement to me! I'm still unsure about a few places, where I left the comments, and about the idea of catching all exceptions before the API level. But we can discuss it later and it shouldn't prevent us from merging this PR.
It also would be cool if we could avoid
```
try {
  ...
} catch () {
  LOG_ERROR << "";
  throw;
}
```
blocks by having more information in the exception object, so we can move the prints higher in the call stack.

src/libaktualizr/primary/initializer.cc
lbonn: Good point, Initializer's methods are worth converting to throwing instead of returning booleans in general.

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: Because of the EcuDownloadCompleteReport and DownloadTarget event. I tried to move them out but then noticed that aktualizr-lite is calling `downloadImage()` directly, so it would miss these calls right now.

We'd have to make some changes there to fix this. I can add a comment.

eu-smirnov: Right, I missed them somehow.

src/libaktualizr/primary/initializer.cc
lbonn: Changed in latest version

src/libaktualizr/primary/sotauptaneclient.cc
lbonn: Changed in latest version

lbonn:
> It also would be cool if we could avoid
> 
> ```
> try {
>   ...
> } catch () {
>   LOG_ERROR << "";
>   throw;
> }
> ```
> 
> blocks by having more information in the exception object, so we can move the prints higher in the call stack.

That's true that they are a bit ugly but I thought it's often useful to know the high level operation that caused the error. Sometimes we only have one caller but otherwise, I don't know how to log that better.

Or we define a macro (:smiley:)

lbonn:
Should be ready to go

PR #1659, opened 2020-04-28
Add SecondaryProvider for getting data to Secondaries

patrickvacek:
There are several changes here grouped by commit. The biggest things are:
* using a new SecondaryProvider class to simply getting data into Secondary implementations on the Primary,
* changing to the IP/POSIX ASN.1 protocol to support more flexible metadata transmission, and
* creating a new MetaBundle type to replace RawMetaPack.

### DIFF
--- a/src/libaktualizr-posix/ipuptanesecondary.cc
+++ b/src/libaktualizr-posix/ipuptanesecondary.cc
@@ -108,6 +108,21 @@ bool IpUptaneSecondary::putMetadata(const Target& target) {
   }
 
   LOG_INFO << "Sending Uptane metadata to the Secondary";
+
+  bool put_result = putMetadata_v2(meta_pack);
+  if (!put_result) {
+    // Fall back to the previous metadata sending version.
+    // TODO(OTA-4793): refactor to negotiate versioning once during
+    // initialization.
+    LOG_ERROR << "Failed to send metadata for target " << target.filename() << " to Secondary " << getSerial()
+              << "\nFalling back to previous version of the metadata sending procedure.";
+
+    put_result = putMetadata_v1(meta_pack);
+  }
+  return put_result;
+}
+
+bool IpUptaneSecondary::putMetadata_v1(const Uptane::RawMetaPack& meta_pack) {
   Asn1Message::Ptr req(Asn1Message::Empty());
   req->present(AKIpUptaneMes_PR_putMetaReq);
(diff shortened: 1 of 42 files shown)


src/libaktualizr/primary/secondary_provider.h
```diff
@@ -0,0 +1,26 @@
+#ifndef UPTANE_SECONDARY_PROVIDER_H
+#define UPTANE_SECONDARY_PROVIDER_H
+
+#include <string>
+
+#include "config/config.h"
+#include "storage/invstorage.h"
+#include "uptane/tuf.h"
+
+class SecondaryProvider {
```
mike-sul, line 10: :+1: I had been considering adding such class too, just needed one more case for "justification" :). Not sure why an instance of this class cannot be passed to Secondary via ctor, though.

patrickvacek: > I had been considering adding such class too, just needed one more case for "justification"

Yeah, it's the only reasonable way I could find to also add the support for flexible metadata control.

> Not sure why an instance of this class cannot be passed to Secondary via ctor, though.

I was considering the same thing. I might still change it to work that way. I'd originally gone a different route here that more or less required the init() call and I hadn't gotten around to deciding for sure if I wanted to keep it.

src/libaktualizr-posix/ipuptanesecondary.cc
```diff
@@ -93,21 +87,26 @@ SecondaryInterface::Ptr IpUptaneSecondary::connectAndCheck(const std::string& ad
...
-      hw_id_{std::move(hw_id)},
-      pub_key_{std::move(pub_key)},
-      image_reader_provider_{std::move(image_reader_provider)},
-      treehub_cred_provider_{std::move(treehub_cred_provider)} {}
-
-bool IpUptaneSecondary::putMetadata(const RawMetaPack& meta_pack) {
+                                     HardwareIdentifier hw_id, PublicKey pub_key)
+    : addr_{address, port}, serial_{std::move(serial)}, hw_id_{std::move(hw_id)}, pub_key_{std::move(pub_key)} {}
+
+bool IpUptaneSecondary::putMetadata(const Target& target) {
```
mike-sul, line 97: :+1:

mike-sul:
@patrickvacek Great work!

PR #1662, opened 2020-04-30
Improve logging when stored metadata does not contain expected targets.

patrickvacek:


patrickvacek:
> were requested by whom?

Fair question. Requested by the caller to download/install. This is automatic when using aktualizr in the continuous mode, but if you use the commandline options or the libaktualizr API, you have more control over what Targets you can request to download/install.

> wordy (what's opposite of consciese?)

"Verbose". "Wordy" is obviously fine, too. Or "loquacious".

I'm not sure how to get more verbose than it already is, though.

PR #1665, opened 2020-05-04
dockerapp: avoid multiple current ostree installs

doanac:
Compare current against target OSTree hash and skip the install call if
the same.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

### DIFF
--- a/src/libaktualizr/package_manager/dockerapp_bundles.cc
+++ b/src/libaktualizr/package_manager/dockerapp_bundles.cc
@@ -131,7 +131,19 @@ bool DockerAppBundles::fetchTarget(const Uptane::Target &target, Uptane::Fetcher
 }
 
 data::InstallationResult DockerAppBundles::install(const Uptane::Target &target) const {
-  auto res = OstreeManager::install(target);
+  data::InstallationResult res;
+  Uptane::Target current = OstreeManager::getCurrent();
+  if (current.sha256Hash() != target.sha256Hash()) {
+    res = OstreeManager::install(target);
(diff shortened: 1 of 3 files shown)


patrickvacek:
This looks fine, but I'm a bit confused why this even happens. Shouldn't we catch this at an earlier stage during the check for updates? If a target is already installed, we shouldn't consider it a new update, and we should toss it out.

By coincidence, I was just looking at that code yesterday and will probably make a small PR to improve some things, but it won't change the basic idea here.

mike-sul:
> Shouldn't we catch this at an earlier stage during the check for updates?

I don't think that this stage present in aktualizr-lite if you mean SotaUptaneClient::checkUpdates() what eventually calls this line https://github.com/advancedtelematic/aktualizr/blob/ea1f81da5fb3cdc75a37754468604448bdb911b5/src/libaktualizr/primary/sotauptaneclient.cc#L490,  the lite doesn't use the uptane cycle.

doanac:
I didn't realize this was specific to aklite. I don't mind keeping this out of tree if you'd rather do that

patrickvacek:
> I didn't realize this was specific to aklite. I don't mind keeping this out of tree if you'd rather do that

Nah, sorry, didn't mean to imply that. I was just trying to understand why this was necessary, since we already do a similar check. I think Mike is right that this is because of the unusual nature of this package manager. It's fine for now, I think.

doanac:
> Nah, sorry, didn't mean to imply that.

You didn't imply anything. There's a cost to ever "if" in a code base, so its always worth thinking about.

PR #1666, opened 2020-05-05
Fix/ota 4941/installed version matching

patrickvacek:


src/libaktualizr/uptane/imagerepository.cc
```diff
@@ -115,7 +115,7 @@ void ImageRepository::fetchTargets(INvStorage& storage, const IMetadataFetcher&
   verifyTargets(image_targets, false);
 
   if (local_version > remote_version) {
-    throw Uptane::SecurityException(RepositoryType::IMAGE, "Mismatched target versions");
+    throw Uptane::SecurityException(RepositoryType::IMAGE, "Rollback attempt");
```
zabbal, line 118: That looks like smth which deserves big bold red error in the log output. Is this done where this exception is caught?

src/libaktualizr/package_manager/packagemanagerfake.cc
```diff
@@ -110,6 +110,7 @@ bool PackageManagerFake::fetchTarget(const Uptane::Target& target, Uptane::Fetch
     return false;
   }
 
+  // TODO(OTA-4939): Unify this and make it more generic.
```
zabbal, line 113: Perhaps explicitly saying unification with what exactly is desired would be helpful in future? Like function name or file for example.

src/libaktualizr/uptane/imagerepository.cc
patrickvacek: Yes, I believe so, although the real tragedy is we don't think we make this particularly obvious in the web UI.

src/libaktualizr/package_manager/packagemanagerfake.cc
patrickvacek: Fair point, done.

src/libaktualizr/uptane/imagerepository.cc
zabbal: Hmm... I though there's a ticket for that but can't find it now.

PR #1668, opened 2020-05-11
Remove accumulator.h

patrickvacek:
It was external code with someone else's copyright that we didn't even really need to be using anymore. It was a relic of the days when garage-push lived in a separate repo and didn't use the logger. Now that we just reuse aktualizr's logger, it's better to just standardize around that.


patrickvacek:
I love that I can remove code and the coverage goes down. Oh well.

PR #1671, opened 2020-05-15
Feat/ota 4843/secondary errors

patrickvacek:
~~FYI @eu-smirnov this is what I've got so far. We can cherry-pick to master if some parts are needed sooner rather than later.~~

~~Still to do:~~
* ~~upgrade IP Secondary error reporting protocol~~
* ~~better testing for sending error messages from IP Secondaries~~
* ~~improve fallback logic negotiation~~
* ~~write tests for protocol version negotiation~~
* ~~manually test upgrading again (desktop/binary and qemu/ostree)~~

### DIFF
--- a/src/libaktualizr-posix/ipuptanesecondary.h
+++ b/src/libaktualizr-posix/ipuptanesecondary.h
@@ -28,28 +28,31 @@ class IpUptaneSecondary : public SecondaryInterface {
   void init(std::shared_ptr<SecondaryProvider> secondary_provider_in) override {
     secondary_provider_ = std::move(secondary_provider_in);
   }
-  bool putMetadata(const Target& target) override;
+  data::InstallationResult putMetadata(const Target& target) override;
   int32_t getRootVersion(bool /* director */) const override { return 0; }
-  bool putRoot(const std::string& /* root */, bool /* director */) override { return true; }
+  // TODO(OTA-4552): Support multiple Root rotations.
+  data::InstallationResult putRoot(const std::string& /* root */, bool /* director */) override {
+    return data::InstallationResult(data::ResultCode::Numeric::kOk, "");
+  }
   Manifest getManifest() const override;
   bool ping() const override;
-  data::ResultCode::Numeric sendFirmware(const Uptane::Target& target) override;
-  data::ResultCode::Numeric install(const Uptane::Target& target) override;
+  data::InstallationResult sendFirmware(const Uptane::Target& target) override;
+  data::InstallationResult install(const Uptane::Target& target) override;
 
  private:
   const std::pair<std::string, uint16_t>& getAddr() const { return addr_; }
-  bool putMetadata_v1(const Uptane::MetaBundle& meta_bundle);
-  bool putMetadata_v2(const Uptane::MetaBundle& meta_bundle);
-  data::ResultCode::Numeric sendFirmware_v1(const Uptane::Target& target);
-  data::ResultCode::Numeric sendFirmware_v2(const Uptane::Target& target);
-  data::ResultCode::Numeric install_v1(const Uptane::Target& target);
-  data::ResultCode::Numeric install_v2(const Uptane::Target& target);
+  data::InstallationResult putMetadata_v1(const Uptane::MetaBundle& meta_bundle);
+  data::InstallationResult putMetadata_v2(const Uptane::MetaBundle& meta_bundle);
+  data::InstallationResult sendFirmware_v1(const Uptane::Target& target);
+  data::InstallationResult sendFirmware_v2(const Uptane::Target& target);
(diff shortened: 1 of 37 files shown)


lbonn:
Looking good so far!

patrickvacek:
Gotta say, the new clang-tidy-8 checks are impressively good at finding calls to virtual functions in constructors.

patrickvacek:
Sadly the version negotiation is still not quite right. It doesn't work if a reboot is required to complete the installation.

patrickvacek:
This is finally functionally complete, manually tested, and ready for review.

I'm not completely satisfied with the version negotiation logic, but it works, and we can try to improve it another time if possible. The problem is caused by Secondaries that require a reboot to complete installation. This means that at any arbitrary point between requests for the manifest from the Secondary, the protocol version could change. Since there was no way to ask the version before, we don't even have anything to fall back on. We have to ask for it every single time. This also means that an older (v1) Secondary will dump a lot of messages like `Unknown message received from Primary!` to the log. I don't really see a way around that, either.

patrickvacek:
@Raigi this will probably be interesting for you to test, even before we merge it to master if possible. I ran our full meta-updater tests on this and everything passed except for the one thing that is consistently failing on the master branch: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/pipelines/774964. However, when manually testing upgrades, downgrades, and especially flipping back and forth between commits, I encountered problems (which are now fixed).

Raigi:
> @Raigi this will probably be interesting for you to test, even before we merge it to master if possible. I ran our full meta-updater tests on this and everything passed except for the one thing that is consistently failing on the master branch: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/pipelines/774964. However, when manually testing upgrades, downgrades, and especially flipping back and forth between commits, I encountered problems (which are now fixed).

I can trigger otf for this change, plus add some more specific manual tests (after our test environments are usable again)

src/libaktualizr/primary/sotauptaneclient.cc
```diff
@@ -1290,40 +1301,50 @@ bool SotaUptaneClient::sendMetadataToEcus(const std::vector<Uptane::Target> &tar
...
+      result = rotateSecondaryRoot(Uptane::RepositoryType::Image(), *(sec->second));
+      if (!result.isSuccess()) {
+        final_result = result;
+        continue;
+      }
+      result = sec->second->putMetadata(target);
+      if (!result.isSuccess()) {
+        LOG_ERROR << "Sending metadata to " << sec->first << " failed: " << result.result_code << " "
+                  << result.description;
+        final_result = result;
```
eu-smirnov, line 1318: So if more than one ECU failed, we will report an error only for the last one?

eu-smirnov: Nevermind, didn't notice the TODO on top.

src/libaktualizr-posix/ipuptanesecondary.h
```diff
@@ -28,35 +26,39 @@ class IpUptaneSecondary : public SecondaryInterface {
...
 
  private:
   const std::pair<std::string, uint16_t>& getAddr() const { return addr_; }
-  bool putMetadata_v1(const Uptane::MetaBundle& meta_bundle);
-  bool putMetadata_v2(const Uptane::MetaBundle& meta_bundle);
-  data::ResultCode::Numeric sendFirmware_v1(const Uptane::Target& target);
-  data::ResultCode::Numeric sendFirmware_v2(const Uptane::Target& target);
-  data::ResultCode::Numeric install_v1(const Uptane::Target& target);
-  data::ResultCode::Numeric install_v2(const Uptane::Target& target);
+  void getSecondaryVersion() const;
```
eu-smirnov, line 42: What is the benefit of declaring this `const` and `protocol_version` `mutable`?

src/libaktualizr/primary/sotauptaneclient.cc
patrickvacek: Yeah, we should fix that, but since it doesn't affect the protocol or SecondaryInterface, we can follow up with a fix and it won't risk breaking anything.

src/libaktualizr-posix/ipuptanesecondary.h
patrickvacek: Yeah, it's annoying, I know. The problem is that `getManifest` is const, and that call this. The options are either this or changing `getManifest` in all Secondary implementations to non-const. What's your preference?

eu-smirnov: Ah, I see. I think it is fine to leave it like this.

PR #1672, opened 2020-05-18
Remove unnecessary call to saveSecondaryInfo

eu-smirnov:
I might be wrong, but to me call to `saveSecondaryInfo` in `addSecondary` seems unnecessary.
If a device is not yet registered, it will do nothing but producing "Can't save Secondary key: unknown error" message, as there is no serial yet for this secondary in DB. Otherwise the secondary info should  be already stored by `initEcuRegister()`.

### DIFF
--- a/src/libaktualizr/primary/aktualizr_test.cc
+++ b/src/libaktualizr/primary/aktualizr_test.cc
@@ -175,6 +175,7 @@ TEST(Aktualizr, DeviceInstallationResult) {
 
   EcuSerials serials{
       {Uptane::EcuSerial("primary"), Uptane::HardwareIdentifier("primary_hw")},
+      {Uptane::EcuSerial("secondary_ecu_serial"), Uptane::HardwareIdentifier("secondary_hw")},
       {Uptane::EcuSerial("ecuserial3"), Uptane::HardwareIdentifier("hw_id3")},
   };
   storage->storeEcuSerials(serials);
--- a/src/libaktualizr/primary/initializer.cc
+++ b/src/libaktualizr/primary/initializer.cc
@@ -8,7 +8,6 @@
 #include "bootstrap/bootstrap.h"
 #include "crypto/keymanager.h"
 #include "logging/logging.h"
-#include "storage/invstorage.h"
 
 // Postcondition: device_id is in the storage
 void Initializer::initDeviceId() {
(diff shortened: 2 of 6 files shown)


patrickvacek:
Does this break the logic with reloading Secondary configuration after reboot if the Secondary isn't immediately available? I think that was the problem that was being solved here. @lbonn can you confirm that this looks okay?

src/aktualizr_primary/primary_secondary_registration_test.cc
```diff
@@ -43,6 +43,7 @@ TEST(PrimarySecondaryReg, SecondariesMigration) {
     // prepare storage the "old" way:
     storage->storeDeviceId("device");
     storage->storeEcuSerials({{primary_serial, primary_hwid}, {secondary_serial, secondary_hwid}});
+    storage->saveSecondaryInfo(secondary_serial, "IP", PublicKey());
```
lbonn, line 46: You're cheating here, this part of the the database did not exist before this feature.
The code that's removed is supposed to handle this case.

Some explanations are in this commit message https://github.com/advancedtelematic/aktualizr/commit/72b47fb306b96571d67b2fc8af721d4d65279a1e but we should put them in a comment probably.

eu-smirnov: Ah, alright, I see. Yes, I've missed this case.
Why can't we still make it from one place though? If we call it from Initializer as a step separate from `initEcuRegister`, it would cover this case as well, wouldn't it?

lbonn: Yes, that sounds like a better plan. I think it could work.

eu-smirnov:
I've updated the PR according to the conversation with @lbonn 
One or two python tests were failing on Github Actions but it worked after a couple of retries :(

patrickvacek:
> I've updated the PR according to the conversation with @lbonn

LGTM! Have you manually tested the scenario in question? Or can we write a test for that?

> One or two python tests were failing on Github Actions but it worked after a couple of retries :(

Yeah, we have some flakiness emerging again, unfortunately.

eu-smirnov:
> LGTM! Have you manually tested the scenario in question? Or can we write a test for that?

Do you mean the scenario we discussed with @lbonn ? `primary_secondary_registration_test.cc `, which is unmodified now, tests exactly this.

PR #1673, opened 2020-05-19
Only send device data if it has changed

patrickvacek:
We now stored hashes of the hardware/system information, list of installed packages, network information, and configuration in the database. The hardware and network data were previously stored locally in SotaUptaneClient, but this persists over restarts, so it saves even more bandwidth.

Note that at least on my laptop, one field in the output of lshw changes every time I call it, so the hardware info gets sent every time. Thankfully, I could not reproduce that problem on qemu.

I also removed sending the manifest as part of device data. There's no reason to; it is always sent as part of checking for updates, and there is never a time that device data is sent that isn't immediately followed by checking for updates.

### DIFF
--- a/config/sql/migration/migrate.25.sql
+++ b/config/sql/migration/migrate.25.sql
@@ -1,0 +1,9 @@
+-- Don't modify this! Create a new migration instead--see docs/ota-client-guide/modules/ROOT/pages/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+CREATE TABLE device_data(data_type TEXT PRIMARY KEY, hash TEXT NOT NULL);
+
+DELETE FROM version;
+INSERT INTO version VALUES(25);
+
+RELEASE MIGRATION;
--- a/src/libaktualizr/storage/sqlstorage.cc
+++ b/src/libaktualizr/storage/sqlstorage.cc
@@ -1648,6 +1648,48 @@ void SQLStorage::clearInstallationResults() {
   db.commitTransaction();
 }
 
+void SQLStorage::storeDeviceDataHash(const std::string& data_type, const std::string& hash) {
+  SQLite3Guard db = dbConnection();
+
+  auto statement = db.prepareStatement<std::string, std::string>(
+      "INSERT OR REPLACE INTO device_data(data_type,hash) VALUES (?,?);", data_type, hash);
+  if (statement.step() != SQLITE_DONE) {
+    LOG_ERROR << "Can't set " << data_type << " hash: " << db.errmsg();
+    throw std::runtime_error("Can't set " + data_type + " hash: " + db.errmsg());
+  }
+}
+
+bool SQLStorage::loadDeviceDataHash(const std::string& data_type, std::string* hash) const {
+  SQLite3Guard db = dbConnection();
+
+  auto statement =
+      db.prepareStatement<std::string>("SELECT hash FROM device_data WHERE data_type = ? LIMIT 1;", data_type);
+
+  int result = statement.step();
+  if (result == SQLITE_DONE) {
+    LOG_TRACE << data_type << " hash not present in db";
+    return false;
+  } else if (result != SQLITE_ROW) {
+    LOG_ERROR << "Can't get " << data_type << " hash: " << db.errmsg();
+    return false;
+  }
+
+  if (hash != nullptr) {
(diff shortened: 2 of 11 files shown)


eu-smirnov:
QEMU doesn't have frequency scaling, most of the real hardware will behave as your laptop.

patrickvacek:
> I see that the `lshw` output is different only in section with `id: cpu` and the value `size`, is this value critical for us? Can we just ignore this field and remove it from the resulting json?

No, not critical, and yes, we could just ignore it, but that requires parsing something that we previously left untouched. We've since agreed to only send the device data once during initialization and only again if a user provides custom data through the new API function, so we don't have to worry about it anymore. I still need to update the PR, though.

patrickvacek:
I also don't understand why the tests are failing. I haven't been able to reproduce locally yet.

config/sql/migration/migrate.25.sql
```diff
@@ -0,0 +1,9 @@
+-- Don't modify this! Create a new migration instead--see docs/ota-client-guide/modules/ROOT/pages/schema-migrations.adoc
+SAVEPOINT MIGRATION;
+
+CREATE TABLE device_data(unique_mark INTEGER PRIMARY KEY CHECK (unique_mark = 0), hardware_info_hash TEXT NOT NULL DEFAULT "", installed_packages_hash TEXT NOT NULL DEFAULT "", network_info_hash TEXT NOT NULL DEFAULT "", configuration_hash TEXT NOT NULL DEFAULT "");
```
lbonn, line 4: I'm starting to dislike this kind of one-row tables.

Maybe we can actually simplify things here with something like a key/value:

```
CREATE TABLE device_data(data_type TEXT PRIMARY KEY, hash TEXT NOT NULL);
```

And then, we don't even have to duplicate the load/store methods completely, they can call these with data_type as a constant:

```
void storeDeviceDataHash(const std::string &data_type, const std::string &hash);
bool loadDeviceDataHash(const std::string &data_type, std::string *hash);
```

Plus, each of these can just use `INSERT OR REPLACE` instead of checking if the table is empty first.

patrickvacek: Yeah, I was debating whether I should add these new items to the existing `device_info` table. I decided against it just because it seemed like two separate concepts. This seems like it would be more flexible (as in, fewer migrations necessary to add new items). Should we migrate some of the other similar things to that model as well?

lbonn: That's a fair question. Perhaps I was too vague about what felt wrong but it's probably that these hashes are pretty independent from each others and read/written at different times. And they have the same format, so it felt more natural to have them in different rows.

As far as other examples we have, the primary_keys table looks similar but I think the public/private keys are not so independent and we store both at once. It also looks like the other "one-row" tables are updated all at once, with fields of various types.

Also (again, opinion :D), the pain threshold we should have before changing the table schema should be slightly higher than changing regular code as they are not completely free to migrate (resources as much as risk). So I don't think we have anything urgent to change there right now.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1648,6 +1648,48 @@ void SQLStorage::clearInstallationResults() {
...
 }
 
+void SQLStorage::storeDeviceDataHash(const std::string& data_type, const std::string& hash) {
+  SQLite3Guard db = dbConnection();
+
+  auto statement = db.prepareStatement<std::string, std::string>(
+      "INSERT OR REPLACE INTO device_data(data_type,hash) VALUES (?,?);", data_type, hash);
+  if (statement.step() != SQLITE_DONE) {
+    LOG_ERROR << "Can't set " << data_type << " hash: " << db.errmsg();
+    return;
```
eu-smirnov, line 1658: Should we throw an exception here?

eu-smirnov:
LGTM, except one comment

src/libaktualizr/storage/sqlstorage.cc
patrickvacek: Yes, probably.

patrickvacek: Although, to be fair, there are lots of other parts of the storage that don't throw given the same circumstances.

eu-smirnov: Yes, I know. Still, will be fewer places to fix in the future ;)

Issue #1675, opened 2020-05-25
clang-tidy-6.0 is not supported by debian:testing


mike-sul:
The test container ./docker/Dockerfile.debian.testing cannot be built anymore; it looks like clang-tidy v6  is not supported on debian:testing (it shows 'bullseye/sid' release for me).

`
mike@msul:~/work/here/projects/aktualizr$ docker build -t aklite-test -f ./docker/Dockerfile.debian.testing .
Sending build context to Docker daemon  160.1MB
Step 1/14 : FROM debian:testing
 ---> c5552d611861
Step 2/14 : LABEL Description="Aktualizr testing dockerfile for Debian Unstable + static checks"
 ---> Using cache
 ---> e9c5da304a85
Step 3/14 : ENV DEBIAN_FRONTEND noninteractive
 ---> Using cache
 ---> 46ef8e69a95d
Step 4/14 : RUN apt-get update && apt-get -y install --no-install-suggests --no-install-recommends   apt-transport-https   asn1c   autoconf   automake   bison   ccache   clang-6.0   clang-tidy-6.0   clang-tools-6.0   clang-format-6.0   cmake   curl   doxygen   e2fslibs-dev   g++   gcc   git   graphviz   jq   lcov   libarchive-dev   libboost-dev   libboost-log-dev   libboost-program-options-dev   libboost-system-dev   libboost-test-dev   libboost-thread-dev   libcurl4-openssl-dev   libengine-pkcs11-openssl   libexpat1-dev   libfuse-dev   libglib2.0-dev   libgpgme11-dev   libgtest-dev   liblzma-dev   libp11-dev   libsodium-dev   libsqlite3-dev   libssl-dev   libtool   lshw   make   ninja-build   net-tools   opensc   pkg-config   psmisc   python3-dev   python3-gi   python3-openssl   python3-venv   softhsm2   sqlite3   strace   valgrind   wget   xsltproc   zip
 ---> Running in c434a8a23e91
Get:1 http://deb.debian.org/debian testing InRelease [116 kB]
Get:2 http://deb.debian.org/debian testing/main amd64 Packages [7705 kB]
Fetched 7821 kB in 2s (3291 kB/s)
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package clang-tidy-6.0
E: Couldn't find any package by glob 'clang-tidy-6.0'
E: Couldn't find any package by regex 'clang-tidy-6.0'
`

[...]

PR #1677, opened 2020-05-27
Upgrade to clang-tidy and clang-format to version 8.

patrickvacek:
Debian testing no longer supports v6 so our hand has been forced. Fixes https://github.com/advancedtelematic/aktualizr/issues/1675.


lbonn:
You're confident the tests will pass? I was about to open https://github.com/advancedtelematic/aktualizr/pull/1678 to buy us some time.

patrickvacek:
> You're confident the tests will pass?

Nope, not at all. I was trying to test locally with docker but ran into some problems, so figured I'd enlist CI to speed up the process and share my progress. If this fails and https://github.com/advancedtelematic/aktualizr/pull/1678 succeeds I have no problem with merging that first just to get things running again.

src/libaktualizr/crypto/crypto.cc
```diff
@@ -95,6 +95,7 @@ Json::Value PublicKey::ToUptane() const {
 
 std::string PublicKey::KeyId() const {
   std::string key_content = value_;
+  // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
```
lbonn, line 98: nice!

patrickvacek:
I'm not so sure this works as I was hoping. This doesn't look good to me:
```
+ cmake --build . --target clang-tidy -- -j6 -k 0
ninja: no work to do.
```

lbonn:
> I'm not so sure this works as I was hoping. This doesn't look good to me:
> 
> ```
> + cmake --build . --target clang-tidy -- -j6 -k 0
> ninja: no work to do.
> ```

That's what I was afraid of, this file doesn't necessarily exist or is up to date at configure time.

Another hacky way which could work better is to add this check in the `COMMAND` argument of the `add_custom_target`, later down.

patrickvacek:
> Another hacky way which could work better is to add this check in the `COMMAND` argument of the `add_custom_target`, later down.

I tried that, but it's not so easy. If I do `grep ... || clang-tidy ...` then the command fails if grep fails. If I wrap that all in `|| true` but then it wouldn't fail if clang-tidy fails. I wish I could just make the target depend on `compile_commands.json` existing, but I'm not aware of an easy way to do that.

patrickvacek:
It appears to be working now, but for some reason it's dumping a bunch of output that wasn't there before, despite that the `-quiet` is still there. I can't figure that one out.

lbonn:
> > Another hacky way which could work better is to add this check in the `COMMAND` argument of the `add_custom_target`, later down.
> 
> I tried that, but it's not so easy. If I do `grep ... || clang-tidy ...` then the command fails if grep fails. If I wrap that all in `|| true` but then it wouldn't fail if clang-tidy fails. I wish I could just make the target depend on `compile_commands.json` existing, but I'm not aware of an easy way to do that.

What about?

```
! grep ... || clang-tidy ...
```

patrickvacek:
> What about?
> 
> ```
> ! grep ... || clang-tidy ...
> ```

Ah yes, that's what I was looking for. Well, the script is more powerful and easier to use anyway. And it has a check to make sure it's actually finding `compile_command.json` and not just skipping everything!

scripts/clang-tidy-wrapper.sh
```diff
@@ -0,0 +1,21 @@
...
+FILE="${4}"
+
+if [[ ! -e "${CMAKE_BINARY_DIR}/compile_commands.json" ]]; then
+  echo "compile_commands.json not found!"
+  exit 1
+fi
+
+# Filter out anything that we aren't trying to compile. Older clang-tidy
+# versions (<=6) skipped them automatically.
+if grep "${FILE}" "${CMAKE_BINARY_DIR}/compile_commands.json"; then
```
lbonn, line 17:
```suggestion
if grep -q "${FILE}" "${CMAKE_BINARY_DIR}/compile_commands.json"; then
```

lbonn: Oops 2 seconds too late :)

patrickvacek: Sorry! Now I see what's going on. Wanna make a PR for that or shall I?

lbonn: Ok I'll do it quick

PR #1678, opened 2020-05-27
Pin Debian docker file to a fixed debian version

lbonn:
debian:testing has switched to bullseye which misses some of our dependencies


lbonn:
doxygen is getting annoying...

patrickvacek:
> doxygen is getting annoying...

Is this just the thing where you need to put `/** \file */` at the top of the file for some stupid reason?

lbonn:
No, a bug. Probably this one https://sourceforge.net/p/doxygen/mailman/message/35481387/

PR #1679, opened 2020-05-27
Make path to downloaded binaries separately configurable

eu-smirnov:
Main changes:
- Move binary target handling logic from Storage class to package manager
- Add a configuration option for the path to downloaded files
- Use std::ifstream in public API instead of StorageRHandle

### DIFF
--- a/src/libaktualizr/package_manager/packagemanagerinterface.cc
+++ b/src/libaktualizr/package_manager/packagemanagerinterface.cc
@@ -1,5 +1,7 @@
 #include "packagemanagerinterface.h"
 
+#include <sys/statvfs.h>
+
 #include "http/httpclient.h"
 #include "logging/logging.h"
 
@@ -114,7 +116,7 @@ bool PackageManagerInterface::fetchTarget(const Uptane::Target& target, Uptane::
     }
 
     const uint64_t required_bytes = target.length() - ds->downloaded_length;
-    if (!storage_->checkAvailableDiskSpace(required_bytes)) {
+    if (checkAvailableDiskSpace(required_bytes)) {
       throw std::runtime_error("Insufficient disk space available to download target");
     }
 
@@ -191,4 +193,23 @@ TargetStatus PackageManagerInterface::verifyTarget(const Uptane::Target& target)
   }
 
   return TargetStatus::kGood;
+}
+
+bool PackageManagerInterface::checkAvailableDiskSpace(const uint64_t required_bytes) const {
+  struct statvfs stvfsbuf {};
+  const int stat_res = statvfs(config.images_path.c_str(), &stvfsbuf);
+  if (stat_res < 0) {
+    LOG_WARNING << "Unable to read filesystem statistics: error code " << stat_res;
+    return true;
+  }
+  const uint64_t available_bytes = (static_cast<uint64_t>(stvfsbuf.f_bsize) * stvfsbuf.f_bavail);
+  const uint64_t reserved_bytes = 1 << 20;
+
+  if (required_bytes + reserved_bytes < available_bytes) {
+    return true;
+  } else {
+    LOG_ERROR << "Insufficient disk space available to download target! Required: " << required_bytes
+              << ", available: " << available_bytes << ", reserved: " << reserved_bytes;
+    return false;
+  }
 }
--- a/src/libaktualizr/package_manager/packagemanagerinterface.cc
+++ b/src/libaktualizr/package_manager/packagemanagerinterface.cc
@@ -13,7 +13,7 @@ struct DownloadMetaStruct {
         progress_cb{std::move(progress_cb_in)} {}
   uintmax_t downloaded_length{0};
   unsigned int last_progress{0};
-  std::unique_ptr<StorageTargetWHandle> fhandle;
+  std::ofstream fhandle;
   const Hash::Type hash_type;
   MultiPartHasher& hasher() {
     switch (hash_type) {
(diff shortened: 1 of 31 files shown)


src/libaktualizr/primary/aktualizr_helpers.cc
```diff
@@ -17,7 +17,7 @@ void targets_autoclean_cb(Aktualizr &aktualizr, const std::shared_ptr<event::Bas
     auto start = entry.installs.size() >= 2 ? entry.installs.end() - 2 : entry.installs.begin();
     for (auto it = start; it != entry.installs.end(); it++) {
       auto fit = std::find_if(installed_targets.begin(), installed_targets.end(),
-                              [&it](const Uptane::Target &t2) { return it->sha256Hash() == t2.sha256Hash(); });
+                              [&it](const Uptane::Target &t2) { return it->filename() == t2.filename(); });
```
patrickvacek, line 20: Can we guarantee that filename will be unique? I'm wondering if we should match more than just that.

src/libaktualizr/primary/sotauptaneclient.h
```diff
@@ -75,6 +75,16 @@ class SotaUptaneClient {
   void completeInstall() const;
   Uptane::LazyTargetsList allTargets() const;
   Uptane::Target getCurrent() const { return package_manager_->getCurrent(); }
+  std::vector<Uptane::Target> getStoredTargets() const { return package_manager_->getTargetFiles(); }
+  void deleteStoredTarget(const Uptane::Target &target) { package_manager_->removeTargetFile(target); }
+  std::ifstream openStoredTarget(const Uptane::Target &target) { 
+    auto status = package_manager_->verifyTarget(target);
```
patrickvacek, line 81: +1

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1685,267 +1685,62 @@ void SQLStorage::clearDeviceData() {
...
-  boost::filesystem::path image_path_;
-  std::ifstream stream_;
-};
-
-std::unique_ptr<StorageTargetRHandle> SQLStorage::openTargetFile(const Uptane::Target& target) const {
-  return std_::make_unique<SQLTargetRHandle>(*this, target);
 }
 
-std::vector<Uptane::Target> SQLStorage::getTargetFiles() {
+std::vector<std::string> SQLStorage::getAllTargetnames() const {
```
patrickvacek, line 1716:
```suggestion
std::vector<std::string> SQLStorage::getAllTargetNames() const {
```

src/libaktualizr/storage/storage_common_test.cc
```diff
@@ -571,7 +571,7 @@ TEST(storage, list_remove_targets) {
   EXPECT_EQ(tfs.size(), 0);
   EXPECT_FALSE(boost::filesystem::exists(temp_dir.Path() / "images" / "HASH"));
 }
-
+#endif
```
patrickvacek, line 574: Can these tests get moved elsewhere and restored?

src/libaktualizr/uptane/uptane_test.cc
```diff
@@ -459,57 +461,58 @@ TEST(Uptane, InstallFakeBad) {
...
+  r.read(content, length);
+  EXPECT_EQ(r.gcount(), length);
+  r.close();
+  auto w = std::ofstream(image, std::ios::binary | std::ios::ate);
+  char content_bad[length + 1];
   memset(content_bad, 0, length + 1);
-  EXPECT_EQ(whandle->wfeed(content_bad, 3), 3);
-  whandle->wcommit();
+  w.write(content_bad, 3);
+  w.close();
```
patrickvacek, line 478: This test is such a hack but it does the job. :) It also conveniently tested a few of the old Storage APIs, though. Do we have good enough coverage of the new versions?

src/libaktualizr/primary/aktualizr_helpers.cc
eu-smirnov: That's the target name and I think there are a lot of places where we rely on it being unique.

src/libaktualizr/storage/storage_common_test.cc
eu-smirnov: I've added two tests to packagemanagerfake, which cover the same functionality.

src/libaktualizr/uptane/uptane_test.cc
eu-smirnov: It should be covered by new unit tests in packagemanagerfake and storage.

eu-smirnov:
I've added a few unit tests to cover refactored methods and hopefully solved the remaining CI issues.

src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -182,13 +180,97 @@ TargetStatus PackageManagerInterface::verifyTarget(const Uptane::Target& target)
...
+  if (!file) {
+    throw std::runtime_error("File doesn't exist for target " + target.filename());
+  }
+  boost::filesystem::remove(file->second);
+  storage_->deleteTargetInfo(target.filename());
+}
+
+std::vector<Uptane::Target> PackageManagerInterface::getTargetFiles() {
+  std::vector<Uptane::Target> v;
+  auto names = storage_->getAllTargetNames();
```
patrickvacek, line 270: Why return just the names as strings from `getAllTargetNames` if we just construct fake Targets here?

eu-smirnov: Do you mean why not construct fake targets in the storage method, or why construct them at all?

patrickvacek: Both, I guess. It looks like previously we were constructing Targets in SqlStorage::getTargetFiles() based on the hashes that were stored in the database. What's the benefit of losing that information?

eu-smirnov: There is no benefit in storing this information. Those Targets were still fake and used only for bookkeeping, as the comment in API header states:

>   Get list of targets currently in storage. This is intended to be used with
>   DeleteStoredTarget and targets are not guaranteed to be verified and
 >  up-to-date with current metadata.

For bookkeeping, it is sufficient to have only a filename.
I also think it is cleaner to keep in Storage only that logic that is directly related to reading/storing necessary information from db.

patrickvacek: Fair enough, thanks for clarifying!

PR #1686, opened 2020-06-08
Feat/ota 4984/secondary replacement

patrickvacek:
Allow ECU re-registration to support adding, removing, and replacing Secondaries. Tested manually and with unit tests for virtual and IP Secondaries.

### DIFF
--- a/src/aktualizr_primary/primary_secondary_registration_test.cc
+++ b/src/aktualizr_primary/primary_secondary_registration_test.cc
@@ -6,11 +6,11 @@
 #include "uptane_test_common.h"
 #include "utilities/utils.h"
 
-boost::filesystem::path uptane_repos_dir;
 boost::filesystem::path fake_meta_dir;
 
 /* This tests that a device that had an IP Secondary will still find it after
- * recent changes, even if it does not connect when the device starts. */
+ * recent changes, even if it does not connect when the device starts. Note that
+ * this is only supported for a single IP Secondary. */
 TEST(PrimarySecondaryReg, SecondariesMigration) {
   const Uptane::EcuSerial primary_serial{"p_serial"};
   const Uptane::EcuSerial secondary_serial{"s_serial"};
@@ -38,7 +38,7 @@ TEST(PrimarySecondaryReg, SecondariesMigration) {
   Json::Value sec_conf;
 
   {
-    // prepare storage the "old" way:
+    // Prepare storage the "old" way (without the secondary_ecus table):
     storage->storeDeviceId("device");
     storage->storeEcuSerials({{primary_serial, primary_hwid}, {secondary_serial, secondary_hwid}});
     storage->storeEcuRegistered();
@@ -51,9 +51,8 @@ TEST(PrimarySecondaryReg, SecondariesMigration) {
   }
 
   {
-    // verifies that aktualizr can still start if it can't connect to its
-    // secondary
-
+    // Verify that aktualizr can still start if it can't connect to its
+    // Secondary:
     UptaneTestCommon::TestAktualizr aktualizr(conf, storage, http);
     Primary::initSecondaries(aktualizr, sec_conf_path);
     aktualizr.Initialize();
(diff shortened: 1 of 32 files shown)


patrickvacek:
The LGTM failure looks pretty bogus to me; it failed because of a timeout fetching from the Ubuntu servers. I can't figure out how to retry it, though.

lbonn:
> The LGTM failure looks pretty bogus to me; it failed because of a timeout fetching from the Ubuntu servers. I can't figure out how to retry it, though.

If you log in, they have a "retry analysis" button. Just clicked it.

patrickvacek:
> If you log in, they have a "retry analysis" button. Just clicked it.

Thanks. I logged in and looked around but couldn't find it. Anyway, I just pushed another tiny commit, so it probably got retriggered by that, too.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1007,31 +1007,32 @@ bool SQLStorage::loadCachedEcuManifest(const Uptane::EcuSerial& ecu_serial, std:
...
-      return;
-    }
+  db.beginTransaction();
 
-    std::vector<MisconfiguredEcu>::const_iterator it;
-    for (it = ecus.begin(); it != ecus.end(); it++) {
-      auto statement = db.prepareStatement<std::string, std::string, int>(
-          "INSERT INTO misconfigured_ecus VALUES (?,?,?);", it->serial.ToString(), it->hardware_id.ToString(),
-          static_cast<int>(it->state));
+  auto statement = db.prepareStatement<std::string>("SELECT count(*) FROM misconfigured_ecus WHERE serial = ?;",
```
lbonn, line 1015: We could get rid of the count and use `INSERT OR REPLACE` here. `serial` would be better off being a primary key instead of just "unique" but it should also work.

patrickvacek: > We could get rid of the count and use `INSERT OR REPLACE` here.

Good point.

> `serial` would be better off being a primary key instead of just "unique" but it should also work.

True, but then I have to add a migration. Is that worth it?

lbonn: I think it should work without so probably not.

src/libaktualizr/primary/initializer.cc
```diff
@@ -53,23 +48,67 @@ void Initializer::initEcuSerials() {
...
+    // Check all configured Secondaries to see if any are new.
+    for (auto it = secondaries_.cbegin(); it != secondaries_.cend(); ++it) {
+      EcuCompare secondary_comp(std::make_pair(it->second->getSerial(), it->second->getHwId()));
+      store_it = std::find_if(stored_ecu_serials.cbegin(), stored_ecu_serials.cend(), secondary_comp);
+      if (store_it == stored_ecu_serials.cend()) {
+        LOG_INFO << "Configured Secondary ECU serial " << it->second->getSerial() << " with hardware ID "
+                 << it->second->getHwId() << " not found in storage.";
+        register_ecus = true;
+      } else {
+        found[static_cast<size_t>(std::distance(stored_ecu_serials.cbegin(), store_it))] = true;
```
lbonn, line 95:
```suggestion
        found[store_it - stored_ecu_serials.cbegin()] = true;
```

src/libaktualizr/primary/initializer.cc
```diff
@@ -53,23 +48,67 @@ void Initializer::initEcuSerials() {
...
+
+    EcuCompare primary_comp(new_ecu_serials[0]);
+    EcuSerials::const_iterator store_it;
+    store_it = std::find_if(stored_ecu_serials.cbegin(), stored_ecu_serials.cend(), primary_comp);
+    if (store_it == stored_ecu_serials.cend()) {
+      LOG_INFO << "Configured Primary ECU serial " << new_ecu_serials[0].first << " with hardware ID "
+               << new_ecu_serials[0].second << " not found in storage.";
+      register_ecus = true;
+    } else {
+      found[static_cast<size_t>(std::distance(stored_ecu_serials.cbegin(), store_it))] = true;
```
lbonn, line 83: See below

src/libaktualizr/primary/initializer.cc
lbonn: Ah this code was just moved...

patrickvacek: Indeed, but I think I also wrote it originally. Can't remember why exactly I used `std::distance` here. Do you see a good reason to prefer one over the other?

src/aktualizr_primary/primary_secondary_registration_test.cc
```diff
@@ -65,53 +64,11 @@ TEST(PrimarySecondaryReg, SecondariesMigration) {
     EXPECT_EQ(secs_info[0].type, "IP");
     EXPECT_EQ(secs_info[0].extra, R"({"ip":"127.0.0.1","port":9061})");
   }
-
```
lbonn, line 70: What's the reason to remove this part of the test?

patrickvacek: I knew you were going to ask about that! Well, I spent a lot of time trying to get it to work, and then I realized it was testing something that we never tried to fix: it's migrating two Secondaries from the old storage layout to the new storage. In the actual migration code for that (in `src/aktualizr_primary/secondary.cc`), it explicitly only does the migration for one Secondary. Previously, this test was doing something kind of strange. Now, this isn't allowed: we just reprovision the device with two Secondaries in this case. Migration doesn't really make sense here.

src/libaktualizr/primary/initializer.cc
lbonn: oh oops I thought I answered earlier, must have missed the button.

So after reading some docs, I can't really see the difference but the substraction is shorter and seems clearer to me.

src/aktualizr_primary/primary_secondary_registration_test.cc
lbonn: Thanks! That's a satisfying reason.

src/libaktualizr/primary/initializer.cc
patrickvacek: Fixed.

patrickvacek:
I've added a fix for a somewhat [obscure issue](https://saeljira.it.here.com/browse/OTA-5050) with trying to reregister Secondaries while an update is active. The backend will reject the registration, and the only resolution is to restore the previous Secodnary configuration, which was complicated by the fact that we'd already stored the new Secondary information. Now we wait to store it until registration is succesful.

patrickvacek:
@lbonn Sorry, just blasted your review because I realized I'd broken the networking test without realizing it. Thankfully the fix was easy. :)

patrickvacek:
Finally fixed all the tests.

src/libaktualizr/storage/sqlstorage.cc
```diff
@@ -1006,30 +1006,14 @@ bool SQLStorage::loadCachedEcuManifest(const Uptane::EcuSerial& ecu_serial, std:
...
-      }
-    }
+void SQLStorage::saveMisconfiguredEcu(const MisconfiguredEcu& ecu) {
+  SQLite3Guard db = dbConnection();
 
-    db.commitTransaction();
+  auto statement = db.prepareStatement<std::string, std::string, int>(
+      "INSERT OR REPLACE INTO misconfigured_ecus VALUES (?,?,?);", ecu.serial.ToString(), ecu.hardware_id.ToString(),
+      static_cast<int>(ecu.state));
+  if (statement.step() != SQLITE_DONE || sqlite3_changes(db.get()) != 1) {
```
eu-smirnov, line 1015: I'm not sure it's worth checking `sqlite3_changes` value here. My understanding was that it is only useful when you have `WHERE` in your statement or something similar.

patrickvacek: Oops, that was lazy copying on my part, didn't even think about it. I'll double check but will probably remove it.

patrickvacek: Fixed. Not is it not necessary, but according to https://www.sqlite.org/c3ref/changes.html it doesn't even do anything for `REPLACE`, so it would've caused a bug.

eu-smirnov: That was my first impression as well, but it looks like while `REPLACE` doesn't increment the counter, the subsequent `INSERT` does. As I understood, `REPLACE` does only the deleting part.

patrickvacek: Oh, I see. Well, I removed it anyway. :) Seems unnecessary.

PR #1687, opened 2020-06-09
Try to solve clone error for package tests

lbonn:
These errors are really getting annoying...


lbonn:
It still happened... :cry:

lbonn:
Let's merge anyway and see what's going on.

PR #1688, opened 2020-06-10
ref: little step forward to separate lib aktualizr headers

kbushgit:
refactoring regarding separating ```libaktualizr``` headers "internal/external"


patrickvacek:
Looks good so far, thanks!

patrickvacek:
It looks like the only problem here is with the formatting because of header include ordering issues. This is a bit unfortunate. I would need to look into this a bit to decide what the right way to do this is. @kbushgit or @eu-smirnov do either of you have strong opinions on how we should order these headers and if the stipulations of the Google style guide (and clang-tidy) are appropriate?

include/libaktualizr/aktualizr.h
```diff
@@ -6,12 +6,17 @@
...
+#include <libaktualizr/events.h>
+
+class SecondaryInterface;
+class SotaUptaneClient;
+class INvStorage;
+class SecondaryInfo;
+
+namespace api {
+  class CommandQueue;
+}
```
patrickvacek, line 19: I forget, does `class api::CommandQueue` work? I thought it did and it's a little cleaner.

include/libaktualizr/config.h
```diff
@@ -9,15 +9,16 @@
...
-#include "utilities/config_utils.h"
-#include "utilities/types.h"
+
+#include <libaktualizr/bootloader_config.h>
+#include <libaktualizr/keymanager_config.h>
+#include <libaktualizr/p11_config.h>
+#include <libaktualizr/logging_config.h>
+#include <libaktualizr/packagemanagerconfig.h>
+#include <libaktualizr/storage_config.h>
+#include <libaktualizr/telemetryconfig.h>
```
patrickvacek, line 19: If we end up having to move all these config files into a common directory, then it no longer makes sense for them to be separate files. We originally separated them out into their respective modules to keep them tied to their relevant code. But if that doesn't work, it's just unnecessary complexity to have them scattered across multiple tiny files.

src/libaktualizr/config/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-set(HEADERS config.h)
+#set(HEADERS config.h)
```
patrickvacek, line 1: This (and the other commented-out line below) should be removed if they really aren't needed anymore.

src/libaktualizr/primary/aktualizr.cc
```diff
@@ -28,7 +30,8 @@ Aktualizr::Aktualizr(Config config, std::shared_ptr<INvStorage> storage_in,
 
 void Aktualizr::Initialize() {
   uptane_client_->initialize();
-  api_queue_.run();
+  api_queue_ = std::make_shared<api::CommandQueue>();
```
patrickvacek, line 33: Can we move the object creation to the constructor?

kbushgit:
> Wow, that's a lot of boilerplate. But looks pretty good so far!
> 
> Why prefer including the public API headers with angle brackets instead of quotes? I know that might be preferable for other users of the library, Thanks for the review, This is a good point and it makes sense but only for the default paths, which are system paths plus current project 's directory. The compiler organizes include paths in the queue and then looking for include files from left to right in that queue, so if we explicitly specify the path using -I (include_directories) option this path will be appended to the left in that queue,  so the compiler will be treated user-added path first before look into the system path. In general, I agree that in our nonpublic headers it may be preferred to use quotes instead of angle brackets but not in the public headers. Since I do not see any disadvantages I would prefer to leave that as is and keep using the same approach for both public and private headers.but internally, shouldn't we be searching locally for the headers before searching system paths?

kbushgit:
> Wow, that's a lot of boilerplate. But looks pretty good so far!
> 
> Why prefer including the public API headers with angle brackets instead of quotes? I know that might be preferable for other users of the library, but internally, shouldn't we be searching locally for the headers before searching system paths?

This is a good point and it makes sense but only for the default paths, which are system paths plus current project 's directory. The compiler organizes include paths in the queue and then looking for include files from left to right in that queue, so if we explicitly specify the path using -I (include_directories) option this path will be appended to the left in that queue,  so the compiler will be treated user-added path first before look into the system path. In general, I agree that in our nonpublic headers it may be preferred to use quotes instead of angle brackets but not in the public headers. Since I do not see any disadvantages I would prefer to leave that as is and keep using the same approach for both public and private headers.

patrickvacek:
Did you mean to close this? :) I assume not.

> The compiler organizes include paths in the queue and then looking for include files from left to right in that queue, so if we explicitly specify the path using -I (include_directories) option this path will be appended to the left in that queue, so the compiler will be treated user-added path first before look into the system path. In general, I agree that in our nonpublic headers it may be preferred to use quotes instead of angle brackets but not in the public headers. Since I do not see any disadvantages I would prefer to leave that as is and keep using the same approach for both public and private headers.

I don't think this is completely accurate. At least [gcc](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html) is explicit that brackets should be used for system headers and quotes for "header files of your own program". My interpretation is that users of libaktualizr will want to include with brackets, but within the project, when referring to our own headers, we should use quotes.

include/libaktualizr/aktualizr.h
kbushgit: yes we use it, the disadvantage he is that we have to use smart ptr, the reason is that we cant simple uses forward declaration with member objects defined by the value https://github.com/advancedtelematic/aktualizr/blob/39f58c56010f99b278267b7d9ef188c8f97e8894/include/libaktualizr/aktualizr.h#L250

patrickvacek: Yeah, that's fine, I just meant if we could put this forward declaration on one line by using the scope resolution operator (`::`) instead of this three-line variant.

eu-smirnov:
While there is indeed no difference between `""` and `<>` if headers are located in a separate include directory, I agree with Patrick that we should use `""` for our headers within the project, as it makes it easier to distinguish them.

include/libaktualizr/aktualizr.h
eu-smirnov: Why not a unique_ptr, though?

kbushgit: > Yeah, that's fine, I just meant if we could put this forward declaration on one line by using the scope resolution operator (`::`) instead of this three-line variant.

Do you mean like this ```class api::CommandQueue;```

kbushgit: > Why not a unique_ptr, though?

It will cause a compilation error: 
```
Invalid application of ‘sizeof’ to incomplete type ‘api::CommandQueue’
  static_assert(sizeof(_Tp)>0,
```

It can be fixed by adding an empty dtor for Aktualizr. 

Here is an answer with a good explanation
https://stackoverflow.com/a/9954553

patrickvacek:
This was the github CI failure:
```
2020-07-01T16:30:54.1960145Z In file included from ../src/libaktualizr/bootloader/bootloader_config.cc:2:
2020-07-01T16:30:54.1960458Z In file included from ../src/libaktualizr/utilities/config_utils.h:9:
2020-07-01T16:30:54.1961475Z ../src/libaktualizr/logging/logging.h:8:1: error: class 'LoggerConfig' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
2020-07-01T16:30:54.1962516Z class LoggerConfig;
2020-07-01T16:30:54.1965238Z ^
2020-07-01T16:30:54.1967164Z ../include/libaktualizr/config.h:15:8: note: previous use is here
2020-07-01T16:30:54.1968343Z struct LoggerConfig {
2020-07-01T16:30:54.1969883Z        ^
2020-07-01T16:30:54.1971549Z ../src/libaktualizr/logging/logging.h:8:1: note: did you mean struct here?
2020-07-01T16:30:54.1973115Z class LoggerConfig;
2020-07-01T16:30:54.1974709Z ^~~~~
2020-07-01T16:30:54.1976155Z struct
2020-07-01T16:30:54.1977754Z 1 error generated.
```

patrickvacek:
I don't know if you're still unable to see the failures, but here's what I'm seeing:
```
 In file included from ../src/libaktualizr/primary/aktualizr.cc:5:
../include/libaktualizr/aktualizr.h:15:1: error: class 'SecondaryInfo' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
class SecondaryInfo;
^
../include/libaktualizr/types.h:443:8: note: previous use is here
struct SecondaryInfo {
       ^
../include/libaktualizr/aktualizr.h:15:1: note: did you mean struct here?
class SecondaryInfo;
^~~~~
struct
```
and
```
 /builds/aktualizr-debian-3833209/src/libaktualizr/primary/aktualizr.cc:20:12: error: use '= default' to define a trivial destructor [modernize-use-equals-default,-warnings-as-errors]
Aktualizr::~Aktualizr() {}
           ^            ~~
                        = default;
```
And it needs a rebase now. If you have clang-tidy-8 on your local machine, I suspect there are going to be more complaints from that, so I encourage running it locally.

src/libaktualizr-posix/asn1/asn1_message.cc
```diff
@@ -12,6 +12,8 @@
 #include <sys/socket.h>
 #include <sys/types.h>
 
+#include "utilities/utils.h"
```
patrickvacek, line 15: The ordering here is mixed up. Seems like it's split into two sections.

src/libaktualizr/config/CMakeLists.txt
```diff
@@ -13,4 +12,5 @@ add_aktualizr_test(NAME config SOURCES config_test.cc ARGS ${PROJECT_BINARY_DIR}
 # config file test for collisions between import and FS->SQL migration paths
 add_test(NAME config-import COMMAND ${PROJECT_SOURCE_DIR}/tests/run_import_clash_test.sh ${PROJECT_SOURCE_DIR}/config)
 
-aktualizr_source_file_checks(${SOURCES} ${HEADERS} config_test.cc)
+#aktualizr_source_file_checks(${SOURCES} ${HEADERS} config_test.cc)
```
patrickvacek, line 15: This commented-out line can be removed.

src/libaktualizr/package_manager/androidmanager.cc
```diff
@@ -1,17 +1,16 @@
 #include <boost/algorithm/string/find.hpp>
 #include <boost/filesystem.hpp>
 #include <forward_list>
-
-#include "androidmanager.h"
-#include "packagemanagerfactory.h"
-
-#include "utilities/utils.h"
-
+:
```
patrickvacek, line 4: ?

src/libaktualizr/utilities/CMakeLists.txt
```diff
@@ -8,14 +8,13 @@ set(SOURCES aktualizr_version.cc
...
-            config_utils.h
+	    config_utils.h
             dequeue_buffer.h
             exceptions.h
             fault_injection.h
             sig_handler.h
             timer.h
-            types.h
-            utils.h
+	    utils.h
```
patrickvacek, line 17: Indenting is off in these two changed lines.

src/virtual_secondary/partialverificationsecondary.h
```diff
@@ -4,12 +4,16 @@
 #include <string>
 #include <vector>
 
+#include <string>
+#include <vector>
```
patrickvacek, line 8: Duplicated.

include/libaktualizr/types.h
```diff
@@ -0,0 +1,461 @@
...
+enum class StorageType { kFileSystem = 0, kSqlite };
+
+/**
+ * @brief The BasedPath class
+ * Can represent an absolute or relative path, only readable through the BasePath::get() method.
+ *
+ * The intent is to avoid unintentional use of the "naked" relative path by
+ * mandating a base directory for each instantiation.
+ *
+ * @todo has to be moved into Utils namespace
```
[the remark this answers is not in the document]
kbushgit: this syntax is a ```doxygen``` compliant, I know we use ```TODO``` but what the reason to keep this approach in our public headers!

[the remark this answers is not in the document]
patrickvacek: But is that "todo" more meaningful for the API users reading the documentation or for us, the developers that maintain it and whose burden it would be to actually put it in the `Utils` namespace?

More importantly, is it actually important for `BasedPath` to live in the `Utils` namespace? I'm not convinced it "has to be moved".

kbushgit:
>  /builds/aktualizr-debian-3833209/src/libaktualizr/primary/aktualizr.cc:20:12: error: use '= default' to define a trivial destructor [modernize-use-equals-default,-warnings-as-errors]
> Aktualizr::~Aktualizr() {}
>            ^            ~~
>                         = default;
> ```
> 
> And it needs a rebase now. If you have clang-tidy-8 on your local machine, I suspect there are going to be more complaints from that, so I encourage running it locally.


on my machine installed clang-tidy-8 I don't see such error...

We cant use default destructor in the combination of forwarding declaration and unique_ptr because of this compilation error:
```
src/libaktualizr-c/CMakeFiles/aktualizr-c.dir/build.make:62: recipe for target 'src/libaktualizr-c/CMakeFiles/aktualizr-c.dir/libaktualizr-c.cc.o' failed
make[2]: *** [src/libaktualizr-c/CMakeFiles/aktualizr-c.dir/libaktualizr-c.cc.o] Error 1
CMakeFiles/Makefile2:15977: recipe for target 'src/libaktualizr-c/CMakeFiles/aktualizr-c.dir/all' failed
make[1]: *** [src/libaktualizr-c/CMakeFiles/aktualizr-c.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
In file included from /usr/include/c++/7/bits/locale_conv.h:41:0,
                 from /usr/include/c++/7/locale:43,
                 from /usr/include/boost/filesystem/path_traits.hpp:29,
                 from /usr/include/boost/filesystem/path.hpp:25,
                 from /usr/include/boost/filesystem.hpp:16,
                 from /home/bush/work/here/origin/master/aktualizr/src/aktualizr_primary/main.cc:3:
/usr/include/c++/7/bits/unique_ptr.h: In instantiation of ‘void std::default_delete<_Tp>::operator()(_Tp*) const [with _Tp = api::CommandQueue]’:
/usr/include/c++/7/bits/unique_ptr.h:263:17:   required from ‘std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = api::CommandQueue; _Dp = std::default_delete<api::CommandQueue>]’
/home/bush/work/here/origin/master/aktualizr/include/libaktualizr/aktualizr.h:25:7:   required from here
/usr/include/c++/7/bits/unique_ptr.h:76:22: error: invalid application of ‘sizeof’ to incomplete type ‘api::CommandQueue’
  static_assert(sizeof(_Tp)>0,
```

PR #1689, opened 2020-06-10
Fix a few tests that fail if /var/sota isn't writable.

patrickvacek:


patrickvacek:
> Oops, how did I miss those?

If you run the tests as non-root on your host machine, but `/var/sota` is owned by root, it will fail. If `/var/sota` is writable by the user running the test (which is apparently the case in CI), they succeed. This has come up a couple times in the past, but it's been a while. :) I once tried to think of ways to catch or prevent the issue, but I'd forgotten about it being an issue at all.

PR #1690, opened 2020-06-12
Delete obsolete pages

HalynaDumych:
Some of the pages don't appear in the navigation and are no longer valid.

Relates-to: OTA-4980

Signed-off-by: Halyna Dumych [ext-halyna.dumych@here.com](mailto:ext-halyna.dumych@here.com)


docs/ota-client-guide/modules/ROOT/pages/ecu_events.adoc
```diff
@@ -1,164 +0,0 @@
-= Client events reporting
-ifdef::env-github[]
-
-[NOTE]
-====
-We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}.html[view this article in our documentation portal]. Not all of our articles render correctly in GitHub.
-====
-endif::[]
-
-// AB: This page is currently out of date and will be temporarily hidden in the developer guide portal to be updated later and connected to the API documentation.
```
patrickvacek, line 10: This doc is still good information. It's purely internal, though. The compromise we made when we transitioned the documentation to this format was to include technical docs but just put them in the Reference or "For Contributors" sections. Could we instead just link to this from there? @lbonn how out of date is this doc really?

tkfu: I agree that we should keep this one around. If we could manage to get it checked/updated and added to some appropriate section, it would be great, but it's also ok to keep it as a hidden page.

HalynaDumych: The file is now back in the repo (but hidden).

PR #1695, opened 2020-06-22
Document OSTREE_DEVICETREE.

patrickvacek:
See https://github.com/advancedtelematic/meta-updater/pull/728.

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/build-configuration.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/build-configuration.adoc
@@ -21,7 +21,8 @@ Here, we provide reference documentation on how to configure link:https://github
 | `OSTREE_COMMIT_BODY`|Message attached to OSTree commit. Empty by default.
 | `OSTREE_COMMIT_SUBJECT`|Commit subject used by OSTree. Defaults to `Commit-id: $\{IMAGE_NAME}`
 | `OSTREE_UPDATE_SUMMARY`|Set this to '1' to update summary of OSTree repository on each commit. '0' by default.
-| `OSTREE_DEPLOY_DEVICETREE`|Set this to '1' to include devicetree(s) to boot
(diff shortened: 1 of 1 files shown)


docs/ota-client-guide/modules/ROOT/pages/build-configuration.adoc
```diff
@@ -21,7 +21,8 @@ Here, we provide reference documentation on how to configure link:https://github
 | `OSTREE_COMMIT_BODY`|Message attached to OSTree commit. Empty by default.
 | `OSTREE_COMMIT_SUBJECT`|Commit subject used by OSTree. Defaults to `Commit-id: $\{IMAGE_NAME}`
 | `OSTREE_UPDATE_SUMMARY`|Set this to '1' to update summary of OSTree repository on each commit. '0' by default.
-| `OSTREE_DEPLOY_DEVICETREE`|Set this to '1' to include devicetree(s) to boot
+| `OSTREE_DEPLOY_DEVICETREE`|Set this to '1' to include devicetree(s) to boot.
+| `OSTREE_DEVICETREE`|Overrides `KERNEL_DEVICETREE` when determining which device tree blob(s) to deploy with OSTree.
```
tkfu, line 25: This would be a filename, right? Or a list of files? This explanation wasn't enough for me to understand what the option does, so I read the related PR. But maybe this is a place where anyone reading it would have sufficient domain knowledge to understand. I'm gonna give a suggestion that I find to be a little bit clearer, but I may be getting some details wrong:

```suggestion
| `OSTREE_DEVICETREE`| Set the list device tree blob(s) to include in the OSTree image. Should be a strict subset of `KERNEL_DEVICETREE`. If not set, all devicetrees in `KERNEL_DEVICETREE` will be used.
```

patrickvacek: @tkfu I think you make a good point. I've taken the best of your suggestion, removed a not-strictly-accurate part, and made it conform closer to some of the other similar entries.

PR #1697, opened 2020-06-25
Remove shared cred p12 from the archive after provisioning

patrickvacek:
It should never be needed again and presents a potential security risk if left available for an attacker who gains access to the device.

Also remove `treehub.json`. It shouldn't be there in the first place, but in the distant past it was often accidentally left there.

There is now an explicit config option to prevent this behavior if you don't want it, which should only be used for tests. Set `provision.mode = "SharedCredReuse"` to use that.

### DIFF
--- a/src/libaktualizr/config/config.cc
+++ b/src/libaktualizr/config/config.cc
@@ -10,6 +10,43 @@
 #include "utilities/exceptions.h"
 #include "utilities/utils.h"
 
+std::ostream& operator<<(std::ostream& os, ProvisionMode mode) {
+  std::string mode_s;
+  switch (mode) {
+    case ProvisionMode::kSharedCred:
+      mode_s = "SharedCred";
+      break;
+    case ProvisionMode::kDeviceCred:
+      mode_s = "DeviceCred";
+      break;
+    case ProvisionMode::kSharedCredReuse:
+      mode_s = "SharedCredReuse";
+      break;
+    default:
+      mode_s = "Default";
+      break;
+  }
+  os << '"' << mode_s << '"';
+  return os;
+}
+
+template <>
+inline void CopyFromConfig(ProvisionMode& dest, const std::string& option_name, const boost::property_tree::ptree& pt) {
+  boost::optional<std::string> value = pt.get_optional<std::string>(option_name);
+  if (value.is_initialized()) {
+    std::string provision_mode{StripQuotesFromStrings(value.get())};
+    if (provision_mode == "SharedCred") {
+      dest = ProvisionMode::kSharedCred;
+    } else if (provision_mode == "DeviceCred") {
+      dest = ProvisionMode::kDeviceCred;
+    } else if (provision_mode == "SharedCredReuse") {
+      dest = ProvisionMode::kSharedCredReuse;
+    } else {
+      dest = ProvisionMode::kDefault;
+    }
+  }
+}
+
 void TlsConfig::updateFromPropertyTree(const boost::property_tree::ptree& pt) {
   CopyFromConfig(server, "server", pt);
   CopyFromConfig(server_url_path, "server_url_path", pt);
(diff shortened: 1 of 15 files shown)


src/libaktualizr/config/config.cc
```diff
@@ -122,7 +159,11 @@ KeyManagerConfig Config::keymanagerConfig() const {
 void Config::postUpdateValues() {
   logger_set_threshold(logger);
 
-  provision.mode = provision.provision_path.empty() ? ProvisionMode::kDeviceCred : ProvisionMode::kSharedCred;
+  if (provision.mode == ProvisionMode::kDefault) {
+    provision.mode = provision.provision_path.empty() ? ProvisionMode::kDeviceCred : ProvisionMode::kSharedCred;
+  } else if (provision.mode == ProvisionMode::kSharedCredReuse) {
+    LOG_INFO << "Provisioning mode is set to reuse shared credentials. This should only be used for testing!";
```
eu-smirnov, line 165: Would it make sense to fail early if `provision.provision_path.empty()`? Probably, doesn't matter, really

src/libaktualizr/utilities/utils.cc
```diff
@@ -589,6 +588,113 @@ void Utils::writeArchive(const std::map<std::string, std::string> &entries, std:
...
+  int r = archive_read_open(a_in, reinterpret_cast<void *>(state.get()), nullptr, read_cb, nullptr);
+  if (r != ARCHIVE_OK) {
+    LOG_ERROR << "archive error: " << archive_error_string(a_in);
+    archive_read_free(a_in);
+    throw std::runtime_error("archive error");
+  }
+
+  struct archive *a_out = archive_write_new();
+  if (a_out == nullptr) {
+    LOG_ERROR << "archive error: could not initialize archive object";
```
eu-smirnov, line 625: Don't we need `archive_read_free` here?

eu-smirnov:
Overall looks good!

src/libaktualizr/config/config.cc
patrickvacek: You mean if using shared creds but the path is empty? Maybe, but it will fail during initialization anyway. Also, nothing would stop you from still importing device creds even if you claim you want to use shared creds (I think). And if you've already provisioned and you get the URL from elsewhere, technically you don't need to keep the provisioning path configured; it'll never get used again.

src/libaktualizr/utilities/utils.cc
patrickvacek: Yep, good call. I did some RAII cleanup to this code and it should be better now. Also rebased.

eu-smirnov: Cool! I always forget that we have StructGuard for such cases.

src/libaktualizr/primary/initializer.cc
```diff
@@ -181,6 +181,16 @@ void Initializer::initTlsCreds() {
     throw Error("Failed to configure HTTP client with device credentials.");
   }
 
+  // Remove shared provisioning credentials from the archive; we have no more
+  // use for them.
+  Utils::removeFileFromArchive(config_.provision_path, "autoprov_credentials.p12");
```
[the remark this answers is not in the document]
patrickvacek: I don't think so; it should succeed. `treehub.json` is in a try-catch because normally it should fail; that file should not be there.

src/libaktualizr/utilities/utils.cc
```diff
@@ -547,45 +545,141 @@ static ssize_t write_cb(struct archive *a, void *client_data, const void *buffer
...
   if (r != ARCHIVE_OK) {
-    LOG_ERROR << "archive error: " << archive_error_string(a);
+    LOG_ERROR << "archive error: " << archive_error_string(a_out.get());
+  }
+
+  if (found) {
+    boost::filesystem::rename(outfile.Path(), archive_path);
+  } else {
+    boost::filesystem::remove(outfile.Path());
+    throw std::runtime_error("Requested file not found in archive!");
```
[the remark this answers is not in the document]
patrickvacek: To be honest, I just copied the style of the existing functions `readFileFromArchive` and `writeArchive`. I think consistency is good, and I don't see a compelling reason to prefer error codes over exceptions. Can you explain your preference?

patrickvacek:
I don't understand why the github CI job is failing; I fixed the problem and it works on my laptop. The line it is quoting as failing has been fixed; you can compare it with the actual code here.

PR #1701, opened 2020-06-29
Fix the leak test (so that it keeps leaking).

patrickvacek:
It appears that gcc 9 fixes the leak automatically, so valgrind doesn't detect it. Now we actually use the leaky variable, which seems to do the trick.

See the oe-selftest ptest results [here](https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/pipelines/825839) for proof that it works.


eu-smirnov:
CI failure seems unrelated...

PR #1711, opened 2020-07-07
Refactor/rm unused code

patrickvacek:


patrickvacek:
> could you also remove these unnecessary semicolons after CLEANUP_AND_RETURN_FAILED in the https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr-c/test/api-test.c
> 
> in the scope of this ref?

I guess, but that sounds like a pretty unrelated change. Are those semicolons hurting anything?

kbushgit:
> > could you also remove these unnecessary semicolons after CLEANUP_AND_RETURN_FAILED in the https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr-c/test/api-test.c
> > in the scope of this ref?
> 
> I guess, but that sounds like a pretty unrelated change. Are those semicolons hurting anything?

it related to ```-Wextar-semi-stmt``` warning but we do not use this flag, so not a problem to leave it as is

PR #1719, opened 2020-07-09
ref: aktualizr client + secondary interface

kbushgit:
Signed-off-by: Kostiantyn Bushko <kostiantyn.bushko@yahoo.com>


src/libaktualizr/primary/aktualizr.cc
```diff
@@ -1,10 +1,11 @@
-#include "aktualizr.h"
-
-#include <chrono>
-
 #include <sodium.h>
+#include <chrono>
```
patrickvacek, line 2: Not a big deal, but C++ headers should be listed before external libs.

docs/ota-client-guide/modules/ROOT/pages/libaktualizr-getstarted.adoc
```diff
@@ -131,9 +131,9 @@ Libaktualizr provides a C++ API for fetching information about available updates
 
 If you're yet not familiar with OTA Connect concepts such as "campaigns", have a look at the https://connect.ota.here.com/#/campaigns[OTA web app] first. You might need to register for a free developer account first.
 
-The main library header is https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr/primary/aktualizr.h[`primary/aktualizr.h`]. It also includes few other libaktualizr headers.
+The main library header is https://github.com/advancedtelematic/aktualizr/blob/master/include/libaktualizr/aktualizr.h[`libaktualizr/aktualizr.h`]. It also includes few other libaktualizr headers.
 
-To use the API, add the `src/libaktualizr` directory to your include path and add `# include "primary/aktualizr.h"` to your source file.
+To use the API, add the `aktualizr/include` directory to your include path and add `# include <libaktualizr/aktualizr.h>` to your source file.
```
patrickvacek, line 136: Not your change, but why is there a space between the `#` and `include`?

src/libaktualizr/primary/aktualizr.cc
kbushgit: It's not my fault,  ```make format``` makes this changed(:

include/libaktualizr/aktualizr.h
```diff
@@ -8,10 +8,15 @@
...
-#include "utilities/apiqueue.h"
+#include "libaktualizr/secondary/secondaryinterface.h"
+
+class SotaUptaneClient;
+class INvStorage;
+class SecondaryInfo;
+
+namespace api {
+class CommandQueue;
+}
```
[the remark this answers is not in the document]
kbushgit: I tried it but the compiler generate an error like:
```
forward declaration of class cannot have a nested name specifier
```

src/libaktualizr/primary/aktualizr.cc
patrickvacek: Even if you put empty lines between the C++ and external header sections?

include/libaktualizr/aktualizr.h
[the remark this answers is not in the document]
patrickvacek: Okay, no worries.

src/libaktualizr/primary/aktualizr.cc
kbushgit: thanks it works

patrickvacek:
I'm not so sure about putting some headers in a `secondary` subdirectory of the `include/libaktualizr` directory. Those headers are still specifically for the Primary to use, so to me it seems like a confusing separation. Or was there a specific goal you were trying to achieve with that?

kbushgit:
> I'm not so sure about putting some headers in a `secondary` subdirectory of the `include/libaktualizr` directory. Those headers are still specifically for the Primary to use, so to me it seems like a confusing separation. Or was there a specific goal you were trying to achieve with that?

No, there is no specific purpose, I'm just not sure if we have to mimic the structure of folders in the public headers as it is in the ```src```? 
Is the idea, to put all those headers in one folder more comfortable for you?

src/virtual_secondary/partialverificationsecondary.cc
```diff
@@ -37,14 +37,15 @@ data::InstallationResult PartialVerificationSecondary::putMetadata(const Target
   }
 
   // TODO(OTA-2484): check for expiration and version downgrade
-  root_ = Uptane::Root(RepositoryType::Director(), Utils::parseJSON(director_root), root_);
+  Uptane::Root root(Root::Policy::kAcceptAll);
+  root = Uptane::Root(RepositoryType::Director(), Utils::parseJSON(director_root), root);
```
patrickvacek, line 41: This is a downgrade in functionality by not storing the Root to check it the next time around. However, this whole class is woefully incomplete, so it doesn't really matter. It needs a lot of work and that isn't part of the scope here.

PR #1722, opened 2020-07-13
Change the guide version & modify 1 instruction

HalynaDumych:
* Change the Developer Guide version to 2020.8.
* Modify the instruction on how to create a custom device ID.

Relates-to: OTA-5118

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/use-your-own-deviceid.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/use-your-own-deviceid.adoc
@@ -8,32 +8,24 @@ We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}
 endif::[]
 
 
-In OTA Connect, a device has two types of identifier: an internal device UUID, and a standard device ID.
+In OTA Connect, a device has two types of identifiers: an internal device UUID and a standard device ID. By default, OTA Connect generates a random device ID for you, but you can also create your own device ID. You can later use this unique device ID to xref:ota-web::create-custom-device-fields.adoc[create custom device fields] and xref:ota-web::create-smart-group.adoc[group devices].
 
-// MC: use xref:otaconnect-identifiers.adoc[identifier]: when topic is finished
-By default, OTA Connect generates a random device ID for you, but you can override this behavior.
+You can create a custom device ID in one of the following ways:
 
-The standard way to specify device identity is in the device certificate -- specifically, in the Subject Distinguished Name (DN) field.
+* Specify your unique device ID in the aktualizr configuration. For more information, see the xref:aktualizr-config-options.adoc#_provision[reference documentation].
 
-This method requires that you use device-credential provisioning which tells OTA Connect to use your device certificates for authentication. This method is a little more complicated than the "shared-credential" provisioning method that you might have used in the "Get Started" guide.
-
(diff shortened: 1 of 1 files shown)


patrickvacek:
Something doesn't look right here based on all the commits being pulled in. Perhaps you should rebase on the intended base branch?

docs/ota-client-guide/modules/ROOT/pages/use-your-own-deviceid.adoc
```diff
@@ -8,32 +8,26 @@ We recommend that you link:https://docs.ota.here.com/ota-client/latest/{docname}
 endif::[]
 
 
-In OTA Connect, a device has two types of identifier: an internal device UUID, and a standard device ID.
+In OTA Connect, a device has two types of identifiers: an internal device UUID and a standard device ID. By default, OTA Connect generates a random device ID for you, but you can also create your own device ID. You can later use this unique device ID to xref:ota-web::create-custom-device-fields.adoc[create custom device fields] and group devices.
 
-// MC: use xref:otaconnect-identifiers.adoc[identifier]: when topic is finished
-By default, OTA Connect generates a random device ID for you, but you can override this behavior.
+If you want to specify a unique device ID, you will need to use the xref:client-provisioning-methods.adoc#_provisioning_with_device_credentials[device-credential provisioning] method rather than xref:client-provisioning-methods.adoc#_provisioning_with_shared_credentials[shared-credentials provisioning].
```
tkfu, line 13: This isn't actually true--you can specify a unique device ID in the aktualizr config: https://docs.ota.here.com/ota-client/latest/aktualizr-config-options.html#_provision

HalynaDumych: Please tell me if I got it right:
No matter if that's provisioning with shared credentials or device credentials, users can generate an ID for their device by updating the following command: `export device_id=${DEVICE_ID:-$\{DEVICE_UUID}}`.

Is there anything else users should specify? Maybe, the repo or device name? How can we combine it with the process of shared-credentials provisioning?

tkfu: This is a little bit more complex.

Each device running aktualizr needs to have a certificate, signed by the fleet root certificate authority. When you use device credential provisioning, you generate the certificate before you run aktualizr for the first time, and the device ID is taken from the common name indicator (CNI) of the device's certificate. When you use shared credential provisioning, the OTA Connect server generates and signs a certificate for you--either with a random CNI, or with something you specify.

The step where you do `export device_id=<whatever>` is part of the process of generating device certificates that we outline [here](https://docs.ota.here.com/ota-client/latest/generate-devicecert.html)--it's just setting an environment variable that is used later on in the openssl command that actually creates the certificate. As such, it's not at all applicable to the shared credential provisioning case.

HalynaDumych: Ok, so we can set a device ID using `export device_id=<customdeviceid>` only when we provision with device credentials.
How can we specify a device ID in the shared credential provisioning case? By modifying a configuration file, like this one: https://github.com/advancedtelematic/aktualizr/blob/master/tests/config/device_id.toml?

PR #1725, opened 2020-07-15
Fix/annoying logging

patrickvacek:
Mostly improving aktualizr-secondary logging, but also cleans up some other logging issues that have bothered me (and others).


mike-sul:
:+1:

PR #1727, opened 2020-07-15
Fix a typo in naming the Ubuntu Focal dockerfile.

patrickvacek:
Hopefully will fix https://github.com/advancedtelematic/aktualizr/runs/873458617.


patrickvacek:
> So it runs for master only.

Yes, this is just updating dockerfiles. The actual tests are run on PRs, and those are fine.

PR #1731, opened 2020-07-29
ref: add jsoncpp library to public API

kbushgit:
Since we use jsoncpp in our public API we have to add appropriate header/s. 
In this case, the user will not be able to use jsoncpp installed as a system package because of naming conflict.


include/libaktualizr/types.h
```diff
@@ -3,11 +3,11 @@
 /** \file */
 
 #include <algorithm>
+#include <boost/filesystem.hpp>
```
patrickvacek, line 6: External headers are supposed to come after C++ headers.

PR #1733, opened 2020-07-30
Add info on the aktualizr config options page

HalynaDumych:
Mention that users should not use `/` as a path in config.

Relates-to: OTA-5095

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/aktualizr-config-options.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/aktualizr-config-options.adoc
@@ -153,7 +153,10 @@ Options for how Aktualizr stores data locally.
 |==========================================================================================
 | Name                      | Default                   | Description
 | `type`                    | `"sqlite"`                | What type of storage driver to use. Options: `"sqlite"`. The former `"filesystem"` option is now disabled, existing devices will be migrated (see note below)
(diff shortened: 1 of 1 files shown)


docs/ota-client-guide/modules/ROOT/pages/aktualizr-config-options.adoc
```diff
@@ -153,7 +153,10 @@ Options for how Aktualizr stores data locally.
 |==========================================================================================
 | Name                      | Default                   | Description
 | `type`                    | `"sqlite"`                | What type of storage driver to use. Options: `"sqlite"`. The former `"filesystem"` option is now disabled, existing devices will be migrated (see note below)
-| `path`                    | `"/var/sota"`             | Directory for storage
+| `path`                    | `"/var/sota"`             | Directory for storage.
+
+Do not use `/` as a path to your storage directory because your folder permissions will be overwritten, and you will not be able to turn on your computer.
```
tkfu, line 158:
```suggestion
This should be a directory dedicated to aktualizr data. aktualizr will attempt to set permissions on this directory, so this option should not be set to anything that is used for another purpose. In particular, do not set it to `/` or to your home directory, as this may render your system unusable.
```

HalynaDumych: Changed

PR #1734, opened 2020-08-03
Fix/ota 4544/secondary libs

patrickvacek:
@kbushgit can you confirm that my separation of a few functions into base_config.cc still works for external users of libaktualizr?


kbushgit:
> @kbushgit can you confirm that my separation of a few functions into base_config.cc still works for external users of libaktualizr?

yes, I've checked ```Config``` ctor with the vector of paths, this ctor invokes ``BaseConfig::updateFromDirs```, works as expected, tested it with the installed ```libaktualizr```

PR #1737, opened 2020-08-06
Document exceptions thrown through the API.

patrickvacek:
This list is not exhaustive, but should give a high-level view of what to expect.

### DIFF
--- a/include/libaktualizr/aktualizr.h
+++ b/include/libaktualizr/aktualizr.h
@@ -24,8 +24,15 @@ class CommandQueue;
 class Aktualizr {
  public:
   /** Aktualizr requires a configuration object. Examples can be found in the
-   *  config directory. */
+   *  config directory.
+   *
+   * @throw SQLException
+   * @throw boost::filesystem::filesystem_error
+   * @throw std::bad_alloc
+   * @throw std::runtime_error (filesystem and json parsing failures; libsodium initialization failure)
+   */
   explicit Aktualizr(const Config& config);
+
   Aktualizr(const Aktualizr&) = delete;
   Aktualizr& operator=(const Aktualizr&) = delete;
   ~Aktualizr();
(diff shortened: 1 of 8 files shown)


include/libaktualizr/aktualizr.h
```diff
@@ -34,18 +41,40 @@ class Aktualizr {
    * Initialize aktualizr. Any Secondaries should be added before making this
    * call. This will provision with the server if required. This must be called
    * before using any other aktualizr functions except AddSecondary.
+   *
+   * @throw Initializer::Error and subclasses
+   * @throw Uptane::Exception and subclasses
```
kbushgit, line 46: Looking into ```Aktualizr/SotaUptaneClient``` classes I can't find the place where the ```Uptane::Exception``` is not handled by the ```SotaUptaneClient```. Is it necessary right now to expose ```Uptane::Exception```, until it is not necessary?

patrickvacek: `SotaUptaneClient::finalizeAfterReboot()` -> `director_repo.dropTargets()`. This is an annoying hole that I'd like to patch up, actually. I too was surprised by this.

xaviergully:
LGTM

include/libaktualizr/aktualizr.h
kbushgit: Sorry, could you point me where actually we can get ```Uptane::Exception``` in the ```DirectoRepo::dropTargets``` ,  instead I see that this call can throw ```SQLException```, If I missed nothing...

patrickvacek: `DirectorRepository::dropTargets()` -> `DirectorRepository::resetMeta()` -> `RepositoryCommon::resetRoot()` -> `Root::Root()` -> `Uptane::Root::UnpackSignedObject()`. There's already one Uptane exception there and more in `MetaWithKeys::UnpackSignedObject()`.

This is admittedly something of a stretch, but it's theoretically possible.

PR #1744, opened 2020-08-13
Add missing configuration values to simulate a device.

houcros:
Found a couple of missing configuration parameters that are needed after upgrading aktualizr to 2020.8 on a Mac.


patrickvacek:
Looks good, but docs changes should actually be pointed at the 2020.8-docs branch so that they'll get reflected in the portal faster.

houcros:
Ah I see, updated.

PR #1745, opened 2020-08-14
Get rid of veracode helpers

zabbal:
The corresponding functionality will be provided by SAST integration in
follow-up patches.


patrickvacek:
Should we leave this until the new thing is ready? Or is it useless or otherwise problematic at present?

zabbal:
It's useless at present - last time I've checked veracode didn't support necessary version of gcc. It was still left in here to facilitate testing for the time when they'll add it.

PR #1749, opened 2020-08-19
Enable additional security and compliance checks

zabbal:


zabbal:
Note to self: this can be greatly optimized by explicitly stating job dependencies. Will try that in next PR.

patrickvacek:
Can you give me a bit more information about what these are actually doing, or how to tell that they are doing meaningful reviews? In particular, the license scanner says "No active and installed package managers found for project." (https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/jobs/4695378)

zabbal:
Sure, here's brief overview:
* SAST - static code analysis for potential vulnerabilities (integer, buffer overflow etc)
* Secret-Detection - check for high-entropy strings which could be secrets (tokens, access keys etc)
* Dependency-Scanning - check dependencies versions for know vulnerabilities (CVE)
* License-Scanning - check for licences (code and dependencies)

All those are far from perfect and give false positives or simply fail. They are still under active development. The only mandatory right now is SAST checker while others are expected to become part of recommended battery of tests as they mature. I've just thought it's easier to add them all in one go.

The failing part of the license checker is an attempt to looks up dependencies licenses. The check for our own code license should still work.

patrickvacek:
> All those are far from perfect and give false positives or simply fail. They are still under active development.

Okay, thanks for the info. I think this is probably fine, but if they do start yielding lots of unhelpful output, can we switch them to be optional? Or can we put them into a separate phase of the pipeline? I think at the moment it's a bit confusing to have them mixed in with the regular test suite.

zabbal:
They all are already optional. I can also move them to separate stage or remove completely.

patrickvacek:
> They all are already optional.

Oh nice, great! Sorry, I didn't realize that.

> I can also move them to separate stage or remove completely.

I think I would prefer if they were bundled as a separate stage from the unit tests.

zabbal:
I see multiple "dependencies" entries which looks like a misnomer caused by confusing gitlab language. Are those what humans think (this job depends on that job) or it's what gitlab thinks (take artifacts from those jobs only while depending on all the previously defined jobs)? If it's the former than I can greatly simplify and speedup the pipeline by using proper syntax.

patrickvacek:
> I see multiple "dependencies" entries which looks like a misnomer caused by confusing gitlab language. Are those what humans think (this job depends on that job) or it's what gitlab thinks (take artifacts from those jobs only while depending on all the previously defined jobs)? If it's the former than I can greatly simplify and speedup the pipeline by using proper syntax.

Oh, interesting! Yes, I agree, it's the former meaning.

zabbal:
Shall I add commits to this PR or make another one?

patrickvacek:
> Shall I add commits to this PR or make another one?

Let's merge this one since it looks good now and then you can make another.

patrickvacek (approved):

PR #1750, opened 2020-08-20
Use proper syntax for gitlab job dependencies

zabbal:
The 'dependencies' actually means the list of artifacts, while the dependencies
are expressed with 'needs' keyword which by default implies artifact list as
well.

### DIFF
--- a/ci/gitlab/.gitlab-ci.yml
+++ b/ci/gitlab/.gitlab-ci.yml
@@ -271,7 +271,7 @@ OE Docker setup:
 OE Checkout:
   extends: .bb_checkout
   stage: oe-checkout
-  dependencies: []
+  needs: []
   variables:
       GIT_CHECKOUT: 'false'
   only:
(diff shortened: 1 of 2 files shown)


patrickvacek:
I see some sort of error I haven't entirely understood yet: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/aktualizr/-/pipelines/1021912

zabbal:
Doh! I really wish gitlab's yaml was designed from scratch by someboby who knows what they are doing. Alas, we have to deal with haphazard collection of features randomly bolted on each other. The fix is waiting for repo sync.

ci/gitlab/.gitlab-ci.yml
```diff
@@ -293,6 +293,7 @@ Ptest qemux86_64:
       GIT_CHECKOUT: false
   dependencies:
     - OE Checkout
+  needs: ["OE Checkout"]
```
patrickvacek, line 296: Why does this one need both keywords specified?

zabbal: Pre-lunch commits are rarely good idea :-/

zabbal:
I've split changes which could affect release process into separate commit to make it easier to rollback.

patrickvacek:
The test flakiness is getting really annoying, but it's not new. It's possible that this PR has somehow exacerbated the problems, but I don't really understand how. For now, let's go ahead and merge and see how it goes. We have a ticket already for looking into these tests (https://saeljira.it.here.com/browse/OTA-5194) which can be prioritized if necessary.

patrickvacek (approved):

PR #1751, opened 2020-08-26
Use default ostree package in Ubuntu Focal dockerfile.

patrickvacek:
It's much newer than the version we compile manually.


patrickvacek:
Looks like this works!

aodukha (approved):

PR #1756, opened 2020-09-08
OTA-4864 - improve binary file download progress logging

aodukha:
Log into output download progress each 15sec for big binary files

### DIFF
--- a/src/libaktualizr/package_manager/packagemanagerinterface.cc
+++ b/src/libaktualizr/package_manager/packagemanagerinterface.cc
@@ -77,8 +77,7 @@ static int ProgressHandler(void* clientp, curl_off_t dltotal, curl_off_t dlnow, 
       auto now = std::chrono::steady_clock::now();
       auto milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(now - ds->time_lastreport);
       if(milliseconds.count() > LogProgressInterval || progress == 100) {
-        auto clocktime = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
-        LOG_INFO << std::put_time(std::localtime(&clocktime), "%T") <<" Downloading "<< ds->target.filename() <<" progress " << progress << std::endl;
+        LOG_INFO << "Download progress for file " << ds->target.filename() <<": " << progress << std::endl;
         ds->time_lastreport = now;
       }
     }
--- a/src/libaktualizr/package_manager/packagemanagerinterface.cc
+++ b/src/libaktualizr/package_manager/packagemanagerinterface.cc
@@ -35,8 +35,9 @@ struct DownloadMetaStruct {
   Uptane::Target target;
   const api::FlowControlToken* token;
   FetcherProgressCb progress_cb;
-  std::chrono::time_point<std::chrono::steady_clock> time_start; //download start time
-  std::chrono::time_point<std::chrono::steady_clock> time_lastreport; // each XXX sec report dowload progress to user to do not stuck with bit files
+  std::chrono::time_point<std::chrono::steady_clock> time_start;  // download start time
+  std::chrono::time_point<std::chrono::steady_clock>
+      time_lastreport;  // each XXX sec report dowload progress to user to do not stuck with bit files
 
  private:
   MultiPartSHA256Hasher sha256_hasher;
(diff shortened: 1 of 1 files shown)


src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -98,6 +112,7 @@ bool PackageManagerInterface::fetchTarget(const Uptane::Target& target, Uptane::
       return true;
     }
     std::unique_ptr<DownloadMetaStruct> ds = std_::make_unique<DownloadMetaStruct>(target, progress_cb, token);
+    ds->time_start = ds->time_lastreport = std::chrono::steady_clock::now(); //use this time to report download progress it download take long time
```
patrickvacek, line 115: Is `time_start` used for anything? Seems unnecessary.

src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -67,6 +72,15 @@ static int ProgressHandler(void* clientp, curl_off_t dltotal, curl_off_t dlnow,
   if (ds->progress_cb && progress > ds->last_progress) {
     ds->last_progress = progress;
     ds->progress_cb(ds->target, "Downloading", progress);
+    //OTA-4864:Improve binary file download progress logging. Report each XX sec report event that notify user
+    if (!ds->target.IsOstree()) {
```
patrickvacek, line 76: This handler isn't used for OSTree, so you shouldn't need to use this check.

src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -67,6 +72,15 @@ static int ProgressHandler(void* clientp, curl_off_t dltotal, curl_off_t dlnow,
   if (ds->progress_cb && progress > ds->last_progress) {
     ds->last_progress = progress;
     ds->progress_cb(ds->target, "Downloading", progress);
+    //OTA-4864:Improve binary file download progress logging. Report each XX sec report event that notify user
+    if (!ds->target.IsOstree()) {      
+      auto now = std::chrono::steady_clock::now();
+      auto milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(now - ds->time_lastreport);
+      if(milliseconds.count() > LogProgressInterval || progress == 100) {
```
patrickvacek, line 79: I don't think you need to do the `progress == 100` check, because once the download is complete, a separate event is sent to make that clear, so that's already covered.

src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -67,6 +72,15 @@ static int ProgressHandler(void* clientp, curl_off_t dltotal, curl_off_t dlnow,
...
+        LOG_INFO << "Download progress for file " << ds->target.filename() <<": " << progress << std::endl;
```
patrickvacek, line 80: No need for `std::endl`; the logger already handles newlines automatically.

patrickvacek (commented):
Cool, thanks! Normally we'd require a test to prove this works, but I don't think it's necessary here. I think if you could do a quick demo, that would suffice.

src/libaktualizr/package_manager/packagemanagerinterface.cc
aodukha: removed. it was artifact from older and a bit more complicated version

src/libaktualizr/package_manager/packagemanagerinterface.cc
aodukha: done

src/libaktualizr/package_manager/packagemanagerinterface.cc
aodukha: done

src/libaktualizr/package_manager/packagemanagerinterface.cc
aodukha: removed

src/libaktualizr/package_manager/packagemanagerinterface.cc
```diff
@@ -98,6 +110,8 @@ bool PackageManagerInterface::fetchTarget(const Uptane::Target& target, Uptane::
       return true;
     }
     std::unique_ptr<DownloadMetaStruct> ds = std_::make_unique<DownloadMetaStruct>(target, progress_cb, token);
+    ds->time_lastreport =
+        std::chrono::steady_clock::now();  // use this time to report download progress it download take long time
```
patrickvacek, line 114: Two minor issues:
1. I think this initialization should either be in the constructor of the `DownloadMetaStruct` or right before the download is initiated (before the loop with `http_->download` below).
2. If a comment is so long as to cause the code to require a linebreak, I think it would make sense to put it on the preceding line instead.

Otherwise looks good!

aodukha: done

patrickvacek (approved):
LGTM, thanks! I tested locally and it worked as expected. Normally I would suggest squashing the fixup commits, but in this case I can do that when I merge.

PR #1761, opened 2020-09-24
packagemanage: Fix output of pacman config

doanac:
When the config is dumped out for the docker-app manager, the custom
config key/vals are printed val=key. eg:

 "/tmp/sota/docker-apps" = docker_apps_root

This fixes the issue.


patrickvacek:
This is perhaps counterintuitive, and I'm surprised no one noticed until now. :) I think this might deserve a small unit test to prevent future breakage.

doanac:
>  I think this might deserve a small unit test to prevent future breakage.

Great idea. I force pushed a unit test for this.

patrickvacek (approved):
Great, thanks!

PR #1767, opened 2020-10-15
support new backend authentication for garage-push and garage-deploy

aodukha:
it is  OTA-5285
Before back-end start support cognito authentication we do in code something like:

curl -X POST <auth-plus-url>/token \
      -d "grant_type=client_credentials" \
      -u <client-id>:<client-secret> 

now we have to keep backward compatabitly with old credetials.zip and for new use :
curl -X POST <auth-server-url> \
      -d "grant_type=client_credentials" \
      -d "scope=<scope>" \
      -u <client-id>:<client-secret> 

difference bewteen new and old credentials.zip are in treehub.json. New have "scope" and auth-server-url ends with /token


.vscode/launch.json
```diff
@@ -21,5 +21,29 @@
...
+            "externalConsole": false,
+            "MIMode": "gdb",
+            "setupCommands": [
+                {
+                    "description": "Enable pretty-printing for gdb",
+                    "text": "-enable-pretty-printing",
+                    "ignoreFailures": true
+                }
+            ]
+        },
```
patrickvacek, line 47: I was never particularly fond of this file being part of the repo in the first place, but if it's going to be there, it really shouldn't have absolute paths that won't work for anyone else. I'd prefer if this file were removed and put it .gitignore, but I'd accept it if the paths were relative.

src/sota_tools/oauth2.cc
```diff
@@ -24,8 +24,29 @@ size_t curl_handle_write_sstream(void *buffer, size_t size, size_t nmemb, void *
...
-  curlEasySetoptWrapper(curl_handle.get(), CURLOPT_URL, (server_ + "/token").c_str());
+  // We need this check for backwards-compatibility with previous versions of treehub.json.
+  // Previous versions have the server URL *without* the token path, so it needs to be hardcoded.
+  // The new version has the full URL with the `/oauth2/token` path at the end, so nothing needs
+  // to be appended. Also we can't send the `scope` to Auth+, it confuses it.
+  // This check can be removed after we finish the migration to Cognito. At that point there's
+  // no need to attempt to be backwards-compatible anymore, since old credentials will have a
+  // client_id that will have been removed from user-profile, and a server URL to Auth+, which
+  // will be in computer heaven.
+  // check similar implementation in Scala for garage-sigh here:
```
patrickvacek, line 43:
```suggestion
  // check similar implementation in Scala for garage-sign here:
```
Sounds good in general, though!

patrickvacek (commented):
Looks fine, but this needs to be covered by unit tests. All the existing authentication methods are currently covered in `authenticate_test.cc` and we have to do something similar for this.

I was also wondering if there are any documentation changes necessary, but it looks like https://docs.ota.here.com/ota-client/latest/provisioning-methods-and-credentialszip.html doesn't going into details about the contents of `treehub.json`. I don't know if anything else does, either. Your thorough comment is probably good enough.

.vscode/launch.json
aodukha: yep. make sense remove the file. anyway configuration for aktualizr does not work

src/sota_tools/authenticate_test.cc
```diff
@@ -98,12 +99,22 @@ TEST(authenticate, no_json_zip) {
...
   boost::filesystem::path filepath = "tests/sota_tools/auth_test_good.json";
   TreehubServer treehub;
   int r = authenticate("", ServerCredentials(filepath), treehub);
   EXPECT_EQ(0, r);
 }
 
+TEST(authenticate, good_json_v2) {
+  // Authenticates with new backend.
+  // Note: update auth_test_goog_v2.json after deploy on prod. current file uses HAT
+  boost::filesystem::path filepath = "tests/sota_tools/auth_test_goog_v2.json";
```
patrickvacek, line 112: Is it intentional that the comment and path here are `goog`, but the json is `goox` and the zip is `good`?

I suggest not supporting json directly but only the zip file. I also suggest creating the zip file with a script at build or test time instead of hardcoding a binary file into the repo. We try to avoid binaries as much as possible, and it makes things much easier to read, debug, and maintain to do it with a script. There are some examples of this already in the repo and it was one of our goals to do this everywhere.

ci/gitlab/.gitlab-ci.yml
```diff
@@ -1,7 +1,7 @@
 stages:
   - docker
   - test
-  - static scans
+  - "static scans"
```
zabbal: Why?

zabbal (dismissed):
The space in the stage name is a perfectly valid construct according to gilab docs.

patrickvacek (commented):
I recommend rebasing on master and squashing a few of those tiny fixup commits.

src/sota_tools/authenticate_test.cc
```diff
@@ -98,12 +99,22 @@ TEST(authenticate, no_json_zip) {
 /* Extract credentials from a provided JSON file. */
 TEST(authenticate, good_json) {
   // Authenticates with the ATS portal to the SaaS instance.
+  // Outdated. Keep backward compatibility
```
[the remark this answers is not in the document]
aodukha: https://saeljira.it.here.com/browse/OTA-5341 to keep it and update test data after backend release

src/sota_tools/authenticate_test.cc
aodukha: typo fixed

patrickvacek (changes requested):
The code looks fine. However, there are still some lingering issues here:

* Now up to nine commits including merges. Please rebase and squash some of the smaller fixup commits.
* The zip file is currently unused. Since the json auth method is deprecated anyway, I would prefer if the zip file were used instead and the new json file excluded.
* The zip file should be constructed dynamically at build or test time via a script. Binaries should only be added to the repo if absolutely necessary.

aodukha:
done. also  "create zit at run-time" added to OTA-5341 task

patrickvacek (approved):
Okay, that'll work. Thanks for the cleanup effort!

PR #1771, opened 2020-11-09
OTA-5170 Update max allowed binary file size (from 3GB to 15GB)

molitskyi:


HalynaDumych (approved):
LGTM :)
Please, also update the following page: https://docs.ota.here.com/ota-web/dev/upload-software-ui.html

molitskyi:
> LGTM :)
> Please, also update the following page: https://docs.ota.here.com/ota-web/dev/upload-software-ui.html

Updated
Thanks)

PR #1772, opened 2020-11-19
fixed checks for cert re-import

aodukha:
do not throw exception if device id and CN in cert are different. E.g. if happens if we provides device_id in config. Added test to cover this case

### DIFF
--- a/src/libaktualizr/storage/storage_common_test.cc
+++ b/src/libaktualizr/storage/storage_common_test.cc
@@ -603,9 +603,8 @@ TEST(StorageImport, ImportData) {
   Utils::writeFile(import_config.tls_clientcert_path.get(import_config.base_path).string(), tls_cert_in2);
   Utils::writeFile(import_config.tls_pkey_path.get(import_config.base_path).string(), tls_pkey_in2);
 
-  // Attempt to re-import, but expect failure because the TLS cert's device ID
-  // changed.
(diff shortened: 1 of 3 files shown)


patrickvacek:
I think you need to rework the test I'd originally written that expects an exception: https://github.com/advancedtelematic/aktualizr/pull/1743/files#diff-e8048692df9ddf9062c09436bbed8460e081dd6c6bfbf0250fbb586753940477R608

src/libaktualizr/storage/storage_common_test.cc
```diff
@@ -641,7 +639,7 @@ TEST(StorageImport, ImportData) {
   EXPECT_TRUE(storage->loadTlsCert(&tls_cert));
   EXPECT_TRUE(storage->loadTlsPkey(&tls_pkey));
 
-  // All TLS objects should be updated.
+  // All TLS objects should be updated exept primary keys.
```
patrickvacek, line 642: BTW the keys are for Uptane, not TLS, which is why the previous comment was accurate.

patrickvacek (approved):
For future reference, as stated in the [contribution guidelines](https://github.com/advancedtelematic/aktualizr/blob/master/CONTRIBUTING.md), "[t]he short description (first line of the commit text) should not exceed 72 chars. The rest of the text (if applicable) should be separated by an empty line."

PR #1773, opened 2020-11-25
fixed outdated credentials used for ctest

aodukha:
backend does not work with old credentials.zip and json after moving to Amazon Cognito.  Also removed one test that duplicate with new cred is full copy of "good_json" test


Raigi (approved):
Looks good. Just one question, the tests will be running on dev environment now, is this desired?

aodukha:
> Looks good. Just one question, the tests will be running on dev environment now, is this desired?

yes. really we need only auth here.   Dev is stable when working with it.

PR #1774, opened 2020-12-01
Use CI credentials for auth tests

aodukha:
remove stored credentials stored real credentials (auth_test_good.zip and treehub json "auth_test_good.json " have client secret inside) and use CI SOTA_PACKED_CREDENTIALS instead. Two auth tests are moved from authenticate_test.cc into /test directory because needs if(SOTA_PACKED_CREDENTIALS). It is safe. System check all during coverage stage 


tests/authenticate_cred_test.cc
```diff
@@ -0,0 +1,63 @@
...
+  }
+  good_test_cred = argv[1];
+
+  // prepare auth_test_good.json, use treehub.json from SOTA_PACKED_CREDENTIALS
+  // extract treehub.json from and save it as auth_test_good.json.
+  TemporaryDirectory tmp_data_dir;
+  good_auth_json = tmp_data_dir / "auth_test_good.json";
+
+  int res =
+      system((std::string("unzip -p ") + argv[1] + std::string(" treehub.json >") + good_auth_json.string()).c_str());
```
patrickvacek, line 54: We have `Utils::shell` as a convenience function for system calls.

patrickvacek (commented):
Seems like an upgrade, so that's cool! I think the only downside is that more tests become dependent on providing credentials, but that seems like an acceptable and minor loss.

I'm not sure I understand why some of the tests were moved to a separate file in a different directory. I think the `sota_tools` separation is a bit arbitrary; it's definitely a historical artifact that doesn't really matter.

I'd still like to get rid of the other zip files and use scripts instead (like `tests/sota_tools/authentication/generate-certs.sh`) but that is a bigger task and isn't urgent, at least until some of the certs in them start expiring. Actually, it's possible that they already have. I think that happened once before and we didn't notice because the tests were still failing. Hmmm.

I also think we can probably safely remove the json-only auth from sota_tools, but I understand that it's easier to leave it as is than remove it.

tests/authenticate_cred_test.cc
aodukha: >>I'm not sure I understand why some of the tests were moved to a separate file in a different directory.
there is additional check in cmakelists.txt for test files and suggestion use "test" for everything that require conditional compilation. so there are builds without credentials at all and builds with fake value (dummy_credentials) instead of real file. So real tests happens on gitlab and for other builds there are 2 ways to exclude some test - test dir + if()-endif() and label "credential" on test
https://github.com/advancedtelematic/aktualizr/blob/d5bc8bad0327f3d9bfa474c9a9636f457caed4a3/CMakeLists.txt#L399

patrickvacek: The correct way to prevent tests that require credentials from running when credentials are not available is to use `if(SOTA_PACKED_CREDENTIALS)` in the CMake code, which is already done for some tests in `src/sota_tools/CMakeLists.txt`.

Using the labels like `credentials` is a good way to control which tests are run if what circumstances, but I think that should be secondary to the if clauses. There is no point in even trying to run tests that require credentials if credentials weren't provided to CMake.

patrickvacek (dismissed):
Looks good, but let's try to figure out what's up with CI...

patrickvacek:
Strange that the gitlab CI is missing from the checks. Do you have a link to a successful run with this branch?

PR #1775, opened 2020-12-11
Dockerapp pac man deletion and a test emulating aktualizr-lite

mike-sul:


pattivacek (commented):
Looks pretty good to me! Good idea, too. I'd still love to formalize the API at some point, but this is a nice interim solution.

mike-sul:
> Looks pretty good to me! Good idea, too. I'd still love to formalize the API at some point, but this is a nice interim solution.

Thanks, Patrick! In addition to removing the dockerapp package manager and adding the test along with mocks, I also made a few tiny changes to other parts of libaktualizr (separate commits). If they are not acceptable I will adjust the test & mocks accordingly.

pattivacek (approved):
All looks good to me. Thanks for this!

PR #1823, opened 2022-09-15
OTA-5352 add page about how to add/remove key, set threshold

wtmount:


docs/ota-client-guide/modules/ROOT/pages/add-remove-targets-key.adoc
```diff
@@ -0,0 +1,88 @@
+= Manage targets keys and set thresholds
+
+You can <<add, add>> and <<remove, remove>> target keys or <<set, change>> the number of keys needed to sign for targets.json.
```
HalynaDumych, line 3:
```suggestion
You can <<add, add>> and <<remove, remove>> target keys or <<set, change>> the number of keys needed to sign targets.json.
```

HalynaDumych (changes requested):

HalynaDumych (approved):

PR #1828, opened 2022-10-25
OTA-6069 document targets.json size limit

wtmount:
Signed-off-by: Oleksii Hubrii <ext-oleksii.hubrii@here.com>

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/upload-large-binary.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/upload-large-binary.adoc
@@ -82,6 +82,8 @@ garage-sign targets sign \
 // tag::gs-push-targets[]
 . Push the new `targets.json` file to OTA Connect.
 +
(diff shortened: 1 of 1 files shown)


docs/ota-client-guide/modules/ROOT/pages/upload-large-binary.adoc
```diff
@@ -82,6 +82,8 @@ garage-sign targets sign \
 // tag::gs-push-targets[]
 . Push the new `targets.json` file to OTA Connect.
 +
+NOTE: Pay attention that the size of targets.json file cannot be larger than 8,286,208 bytes (approximately 8MB). When targets.json reaches 75% of its size limit, a warning message will be shown after each push to the server. Upon reaching the limit, you won't be able to upload new software versions. In this case, you'll first need to delete some unnecessary versions.
```
HalynaDumych, line 85:
```suggestion
NOTE: The size of the targets.json file cannot be larger than 8,286,208 bytes (approximately 8 MB). When targets.json reaches 75% of its size limit, you get a warning message after each push to the server. When the limit is reached, new software versions cannot be uploaded. In this case, delete unnecessary versions.
```

HalynaDumych (changes requested):

HalynaDumych (approved):

PR #1829, opened 2022-11-17
OTA-5642 update targets upload and add-uploaded options

wtmount:
Signed-off-by: Oleksii Hubrii <ext-oleksii.hubrii@here.com>

### DIFF
--- a/docs/ota-client-guide/modules/ROOT/pages/upload-large-binary.adoc
+++ b/docs/ota-client-guide/modules/ROOT/pages/upload-large-binary.adoc
@@ -51,6 +51,7 @@ garage-sign targets upload \
   --name <name-of-target> \
   --version <target-version>
 ----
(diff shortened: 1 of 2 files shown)


docs/ota-client-guide/modules/ROOT/pages/upload-large-binary.adoc
```diff
@@ -51,6 +51,7 @@ garage-sign targets upload \
   --name <name-of-target> \
   --version <target-version>
 ----
+NOTE: To skip checking whether the file has already been added to the targets, run the command with `--force` option.
```
HalynaDumych, line 54:
```suggestion
NOTE: Not to check if the file is in targets, run the command with the `--force` option.
```

HalynaDumych (changes requested):

HalynaDumych (approved):

